PR #41, opened 2023-01-29
prompt to install plugin on `rtx x nodejs@14 -- node -v`

### LINKED ISSUE #40: rtx exec not fetching new plugins
I'm trying to strike a balance with prompting at the right times and not at the wrong times. However look at this command and imagine I'm on a brand new box (but one that has node on it):

```sh-session
$ rtx exec nodejs@14 -- node -v
```

I would think this would fetch node 14 and run it, but what actually happens is it runs the system node since the plugin isn't installed. I think ignoring runtimes when the plugin isn't installed makes sense for `.tool-versions`. Otherwise it's too complex.

However in this case, this should prompt to install nodejs plugin and node. Or maybe just if you specify a version with `@`, but I think if any argument is specified it should be treated similar to `rtx install` (which will prompt to install plugins).

Unfortunately this is kind of tricky to fix. I have a PR up but it's not a complete fix. It needs to call `.fetch_latest()` on the version because otherwise it tried to literally install "14" but it needs to install the latest version of 14. The code isn't well suited for this, unfortunately.

BTW if anyone is reading this and wondering about Rust: this is exactly what I hate about it. Everything basically has to be immutable pretty much so I can't really make this change without a huge design change. If I was writing this in Node or Go or most other languages I could just change the value of the runtime and it would be fine, but that's a lot harder in Rust. I don't dislike Rust, but it certainly is taking me a lot longer to do relatively simple things for sure.

jdxcode:
Fixes #40

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -133,7 +133,9 @@ impl Config {
 
     pub fn ensure_installed(&self) -> Result<()> {
         for rtv in self.ts.list_current_versions() {
-            if rtv.plugin.is_installed() {
+            eprintln!("Ensuring {} is installed {}", rtv.plugin.name , rtv.version);
+            let source = self.ts.get_source_for_plugin(&rtv.plugin.name);
+            if rtv.plugin.is_installed() || matches!(source, Some(PluginSource::Argument(_))) {
                 rtv.ensure_installed(self)?;
             }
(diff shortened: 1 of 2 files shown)


jdxcode:
this was added in a different PR

PR #91, opened 2023-02-03
[feat] add Xonsh shell support

eugenesvk:
Seems to be working (it compiles, and since it's Rust, should be fine :)), installed a tool and check entering/exiting a dir with a `.tool-versions` file, did all the contributing list of checks
(though hate that the all the linters hate properly aligned code and turn ↓ tabular beauty into a ziz-zagged mess
```rs
Some(escaped_char) => escaped_string.push_str(escaped_char),
None               => escaped_string.push(ch), // ← cleaner to see the diff vs ↓
```
```rs
Some(escaped_char) => escaped_string.push_str(escaped_char),
None => escaped_string.push(ch),
```
)

Meanwhile, a few questions/comments:
  - var name: I've changed the escape rules for variable values (xonsh uses Python escape rules, not unix shell's), but __NOT__ for variable names — escapes in `$KEY_VAR_with_weird_chars` won't help, it will still result in an error as it's considered an identifier. Don't think anyone would use weird names in variables, but even if they would I think erroring out would be correct as I'm not sure the plugin would expect a silent rename? For now, don't plan to add any fix (and anyway Python has a bunch of identifier rules defined as unicode char groups with some legacy exceptions etc, so didn't want to recreate them from scratch either, do you maybe know a crate that does that?)
  - perf: There is a chance to eek out a ~15% performance on execution when `cd`-ing into a new dir with a `.tool-versions` file (maybe a bit more on new empty dir, but the base is much smaller there), but then you lose autocolor detection (user's would need to use `$FORCE_COLOR` env var to get it back). Worth it?
  - path: I see that you replace the whole `$PATH` every time. This might have a slight negative affect on user configs who have shell `$PATH` different from OS environment `PATH` for one reason or another since both are currently overwritten with values from rtx (shell PATH needs to be overwritten to find `rtx` and other executables, OS `PATH` needs to be overwritte since that's where `rtx` reads info from). Meanwhile, xonsh is capable of adding/removing individual items to/from a `$PATH`. Is this something that `rtx` could pass instead of the wholesale replacement of PATH (afaik you're moving to more granular tracking approach anyway due to a conflict with direnv)?

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,6 +243,25 @@ $ echo 'eval "$(rtx activate bash)"' >> ~/.bashrc
 $ echo 'rtx activate fish | source' >> ~/.config/fish/config.fish
 ```
 
+### Xonsh
+
+Since `.xsh` files are [not compiled](https://github.com/xonsh/xonsh/issues/3953) you may shave a bit off startup time by using a pure Python import: add the code below to, for example, `~/.config/xonsh/rtx.py` config file and `import rtx` it in `~/.config/xonsh/rc.xsh`:
+```xsh
+from pathlib        	import Path
+from xonsh.built_ins	import XSH
+
+ctx = XSH.ctx
+rtx_init = subprocess.run([Path('~/bin/rtx').expanduser(),'activate','xonsh'],capture_output=True,encoding="UTF-8").stdout
+XSH.builtins.execx(rtx_init,'exec',ctx,filename='rtx')
+```
+
+Or continue to use `rc.xsh`/`.xonshrc`:
+```xsh
+echo 'execx($(~/bin/rtx activate xonsh))' >> ~/.config/xonsh/rc.xsh # or ~/.xonshrc
+```
+
+Given that `rtx` replaces both shell env `$PATH` and OS environ `PATH`, watch out that your configs don't have these two set differently (might throw `os.environ['PATH'] = xonsh.built_ins.XSH.env.get_detyped('PATH')` at the end of a config to make sure they match)
+
 ### Something else?
 
 Adding a new shell is not hard at all since very little shell code is
@@ -490,7 +509,7 @@ Arguments:
   [SHELL_TYPE]
           Shell type to generate the script for
           
-          [possible values: bash, fish, zsh]
+          [possible values: bash, fish, xonsh, zsh]
 
 Options:
   -q, --quiet
@@ -504,6 +523,7 @@ Examples:
     $ eval "$(rtx activate bash)"
     $ eval "$(rtx activate zsh)"
     $ rtx activate fish | source
+    $ execx($(rtx activate xonsh))
 
 ```
 ### `rtx alias ls`
@@ -599,7 +619,7 @@ Arguments:
   [SHELL_TYPE]
           shell type to generate the script for
           
-          [possible values: bash, fish, zsh]
+          [possible values: bash, fish, xonsh, zsh]
 
 Options:
   -h, --help
@@ -610,6 +630,7 @@ Examples:
     $ eval "$(rtx deactivate bash)"
     $ eval "$(rtx deactivate zsh)"
     $ rtx deactivate fish | source
+    $ execx($(rtx deactivate xonsh))
 
 ```
 ### `rtx direnv activate`
@@ -674,7 +695,7 @@ Options:
   -s, --shell <SHELL>
           Shell type to generate environment variables for
           
-          [possible values: bash, fish, zsh]
+          [possible values: bash, fish, xonsh, zsh]
 
   -h, --help
           Print help (see a summary with '-h')
@@ -684,6 +705,7 @@ Examples:
   $ eval "$(rtx env -s bash)"
   $ eval "$(rtx env -s zsh)"
   $ rtx env -s fish | source
+  $ execx($(rtx env -s xonsh))
 
 ```
 ### `rtx exec`
--- a/src/cli/hook_env.rs
+++ b/src/cli/hook_env.rs
@@ -19,8 +19,6 @@ use crate::{dirs, env, hook_env};
 #[clap(hide = true)]
 pub struct HookEnv {
     /// Shell type to generate script for
-    ///
-    /// e.g.: bash, zsh, fish
     #[clap(long, short)]
     shell: Option<ShellType>,
 }
--- a/src/shell/mod.rs
+++ b/src/shell/mod.rs
@@ -5,12 +5,14 @@ use crate::env;
 
 mod bash;
 mod fish;
+mod xonsh;
 mod zsh;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, clap::ValueEnum)]
 pub enum ShellType {
     Bash,
     Fish,
+    Xonsh,
     Zsh,
 }
 
@@ -21,6 +23,8 @@ impl ShellType {
             Some(ShellType::Bash)
         } else if shell.ends_with("fish") {
             Some(ShellType::Fish)
+        } else if shell.ends_with("xonsh") {
+            Some(ShellType::Xonsh)
         } else if shell.ends_with("zsh") {
             Some(ShellType::Zsh)
         } else {
@@ -34,6 +38,7 @@ impl Display for ShellType {
         match self {
             Self::Bash => write!(f, "bash"),
             Self::Fish => write!(f, "fish"),
+            Self::Xonsh => write!(f, "xonsh"),
             Self::Zsh => write!(f, "zsh"),
         }
     }
@@ -49,8 +54,9 @@ pub trait Shell {
 pub fn get_shell(shell: Option<ShellType>) -> Box<dyn Shell> {
     match shell.or_else(ShellType::load) {
         Some(ShellType::Bash) => Box::<bash::Bash>::default(),
-        Some(ShellType::Zsh) => Box::<zsh::Zsh>::default(),
         Some(ShellType::Fish) => Box::<fish::Fish>::default(),
+        Some(ShellType::Xonsh) => Box::<xonsh::Xonsh>::default(),
+        Some(ShellType::Zsh) => Box::<zsh::Zsh>::default(),
         _ => panic!("no shell provided, use `--shell=zsh`"),
     }
 }
--- a/src/shell/xonsh.rs
+++ b/src/shell/xonsh.rs
@@ -1,0 +1,142 @@
+use std::path::Path;
+
+use indoc::formatdoc;
+
+use crate::shell::{is_dir_in_path, Shell};
+
+#[derive(Default)]
+pub struct Xonsh {}
+
+use std::borrow::Cow;
+fn xonsh_escape_sq(input: &str) -> Cow<str> {
+    for (i, ch) in input.chars().enumerate() {
+        if xonsh_escape_char(ch).is_some() {
+            let mut escaped_string = String::with_capacity(input.len());
+
+            escaped_string.push_str(&input[..i]);
+            for ch in input[i..].chars() {
+                match xonsh_escape_char(ch) {
+                    Some(escaped_char) => escaped_string.push_str(escaped_char),
+                    None => escaped_string.push(ch),
+                };
+            }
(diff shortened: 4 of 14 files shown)


src/shell/mod.rs
```diff
@@ -6,12 +6,14 @@ use crate::env;
...
 mod fish;
 mod zsh;
+mod xonsh;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, clap::ValueEnum)]
 pub enum ShellType {
     Bash,
     Fish,
     Zsh,
+    Xonsh,
```
jdx, line 15: please alphabetize this list

src/shell/xonsh.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_hook_init() {
+        insta::assert_snapshot!(Xonsh::default().activate(Path::new("/some/dir/rtx")));
```
jdx, line 130: you will need to update the snapshots with `cargo insta test --accept`

README.md
```diff
@@ -25,6 +25,8 @@ Hook rtx into to your shell. This will automatically add `~/bin` to `PATH` if it
 $ echo 'eval "$(~/bin/rtx activate bash)"' >> ~/.bashrc
 $ echo 'eval "$(~/bin/rtx activate zsh)"' >> ~/.zshrc
 $ echo '~/bin/rtx activate fish | source' >> ~/.config/fish/config.fish
+$ echo 'execx($(~/bin/rtx activate xonsh))' >> ~/.config/xonsh/rc.xsh
```
jdx: let's leave this out of the quickstart, I think we can just leave the most common 3 shells in this section

.gitignore
```diff
@@ -26,3 +26,7 @@
 **/snapshots/*.snap.new
 
 package-lock.json
+
+# project's built/temp docs
+Changelog.html
+ReadMe.html
```
jdx: can you explain why this is necessary?

jdxcode:
> path: should the path to rtx at activation be the first one like it is for other shells (PATH={}:$PATH)? Didn't understand the rationale behind it, so just use a regular adding function for xonsh, which by default doesn't put it at the front to avoid messing up with user paths as much as possible

That's not really what that code is doing. It's just adding the directory that rtx is in to PATH so that `rtx` functions correctly. asdf does the same thing. That way you can't get into a situation where you activated rtx but it isn't actually on PATH. If the directory is already in PATH it doesn't modify it. It's a minor optimization I don't think is crucial to add for xonsh.

> path: I see that you replace the whole $PATH every time. This might have a slight negative affect on user configs who have shell $PATH different from OS environment PATH for one reason or another since both are currently overwritten with values from rtx (shell PATH needs to be overwritten to find rtx and other executables, OS PATH needs to be overwritte since that's where rtx reads info from). Meanwhile, xonsh is capable of adding/removing individual items to/from a $PATH. Is this something that rtx could pass instead of the wholesale replacement of PATH (afaik you're moving to more granular tracking approach anyway due to a conflict with direnv)?

I'm not entirely sure what you mean by shell path vs os path. In bash/zsh there isn't a distinction that I'm aware of. In fish there kind of is with fish_user_paths, but rtx doesn't make use of that at all. For now I would try to stay close to bash without thinking about how rtx works under the hood.

> env unset: Am I correct that unset_env is NOT used for PATH? In xonsh adding/removing PATHs uses a different syntax, which I've added only for set_env, but not for unset_env since I haven't noticed any code where you use unset for PATH
var name: I've changed the escape rules for variable values (xonsh uses Python escape rules, not unix shell's), but NOT for variable names — escapes in $KEY_VAR_with_weird_chars won't help, it will still result in an error as it's considered an identifier. Don't think anyone would use weird names in variables, but even if they would I think erroring out would be correct as I'm not sure the plugin would expect a silent rename? For now, don't plan to add any fix (and anyway Python has a bunch of identifier rules defined as unicode char groups with some legacy exceptions etc, so didn't want to recreate them from scratch either, do you maybe know a crate that does that?)

Are you asking if we ever run the equivalent of `unset PATH`? I can't think of a good reason for us to ever want to do that.

> perf: There is a chance to eek out a ~15% performance on execution when cd-ing into a new dir with a .tool-versions file (maybe a bit more on new empty dir, but the base is much smaller there), but then you lose autocolor detection (user's would need to use $FORCE_COLOR env var to get it back). Worth it?

Where is this performance coming from? If it's rtx, are you benchmarking in debug mode? If so you'll want to test against a release built for a proper benchmark.

> complete: Am I correct that you use https://github.com/nihaals/clap-complete-command to autogenerate shell completions?

I wouldn't worry about xonsh completions for now.

.gitignore
eugenesvk: to exclude locally built md→html docs

jdx: would you mind removing it? I want to keep the ignore file free from dev tool config. You can put this in your `.git/info/exclude` if you want it ignored from your personal clone

eugenesvk: k, though that's not a portable solution across clones

src/cli/hook_env.rs
```diff
@@ -20,7 +20,7 @@ use crate::{dirs, env, hook_env};
 pub struct HookEnv {
     /// Shell type to generate script for
     ///
-    /// e.g.: bash, zsh, fish
+    /// e.g.: bash, zsh, fish, xonsh
```
jdx: sorry I just noticed but this should be removed, this is the output currently:

```
Options:
  -s, --shell <SHELL>
          Shell type to generate script for

          e.g.: bash, zsh, fish

          [possible values: bash, fish, zsh]
```

this is my fault of course but this is a good time to fix it

jdx: if we get that I think this is good to go

jdxcode:
# [Codecov](https://codecov.io/gh/jdxcode/rtx/pull/91?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Jeff+Dickey) Report
Base: **81.47**% // Head: **81.14**% // Decreases project coverage by **`-0.33%`** :warning:
> Coverage data is based on head [(`c15bad5`)](https://codecov.io/gh/jdxcode/rtx/pull/91?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Jeff+Dickey) compared to base [(`6798ab1`)](https://codecov.io/gh/jdxcode/rtx/commit/6798ab1afc4d4d8b530ff0ffc7532b3deb6b8175?el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Jeff+Dickey).
> Patch coverage: 64.44% of modified lines in pull request are covered.

<details><summary>Additional details and impacted files</summary>


```diff
@@            Coverage Diff             @@
##             main      #91      +/-   ##
==========================================
- Coverage   81.47%   81.14%   -0.33%     
==========================================
  Files          78       79       +1     
  Lines        2375     2419      +44     
==========================================
+ Hits         1935     1963      +28     
- Misses        440      456      +16     
```


| [Impacted Files](https://codecov.io/gh/jdxcode/rtx/pull/91?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Jeff+Dickey) | Coverage Δ | |
|---|---|---|
| [src/cli/activate.rs](https://codecov.io/gh/jdxcode/rtx/pull/91?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Jeff+Dickey#diff-c3JjL2NsaS9hY3RpdmF0ZS5ycw==) | `87.50% <ø> (ø)` | |
| [src/cli/deactivate.rs](https://codecov.io/gh/jdxcode/rtx/pull/91?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Jeff+Dickey#diff-c3JjL2NsaS9kZWFjdGl2YXRlLnJz) | `100.00% <ø> (ø)` | |
| [src/cli/env.rs](https://codecov.io/gh/jdxcode/rtx/pull/91?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Jeff+Dickey#diff-c3JjL2NsaS9lbnYucnM=) | `100.00% <ø> (ø)` | |
| [src/cli/hook\_env.rs](https://codecov.io/gh/jdxcode/rtx/pull/91?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Jeff+Dickey#diff-c3JjL2NsaS9ob29rX2Vudi5ycw==) | `74.19% <ø> (ø)` | |
| [src/cli/render\_help.rs](https://codecov.io/gh/jdxcode/rtx/pull/91?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Jeff+Dickey#diff-c3JjL2NsaS9yZW5kZXJfaGVscC5ycw==) | `100.00% <ø> (ø)` | |
| [src/shell/mod.rs](https://codecov.io/gh/jdxcode/rtx/pull/91?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Jeff+Dickey#diff-c3JjL3NoZWxsL21vZC5ycw==) | `32.14% <20.00%> (-5.36%)` | :arrow_down: |
| [src/shell/xonsh.rs](https://codecov.io/gh/jdxcode/rtx/pull/91?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Jeff+Dickey#diff-c3JjL3NoZWxsL3hvbnNoLnJz) | `70.00% <70.00%> (ø)` | |

Help us with your feedback. Take ten seconds to tell us [how you rate us](https://about.codecov.io/nps?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Jeff+Dickey). Have a feature suggestion? [Share it here.](https://app.codecov.io/gh/feedback/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Jeff+Dickey)

</details>

[:umbrella: View full report at Codecov](https://codecov.io/gh/jdxcode/rtx/pull/91?src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Jeff+Dickey).   
:loudspeaker: Do you have feedback about the report comment? [Let us know in this issue](https://about.codecov.io/codecov-pr-comment-feedback/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Jeff+Dickey).

src/cli/hook_env.rs
eugenesvk: Removed
So it should be like this?
```
Options:
  -s, --shell <SHELL>
          Shell type to generate script for

          e.g.: bash, zsh, fish

          [possible values: bash, fish, xonsh, zsh]
```

jdx: remove lines 22 and 23 entirely so that `.e.g.:` line does not display

eugenesvk: yeah, sry, was wondering why remove xonsh from the e.g. only

eugenesvk:
Just FYI what the difference is in executing a function that executes a loop of 2 `cd`s (in/out of a folder with a tool file) a bunch of times as reported by the shell itself (changed rtx to hook into `cd` instead of prompt for easier time tracking)

|Mode          	|event   	| time|Δ to py |
|-             	|-       	| -| -|
|no rtx        	|on_chdir	| 0.507 | |
|rtx in python 	|on_chdir	| 4.292 | |
|rtx in subproc	|on_chdir	| 7.106 |+65%|

eugenesvk:
and for completions opened a feature request upstream https://github.com/clap-rs/clap/issues/4693, so whenever it's added should be auto-picked up by your build scripts

Issue #74, opened 2023-02-01
add version update check on `rtx version`


jdxcode:
`rtx doctor` and `rtx version` should pull https://rtx.jdxcode.com/VERSION to compare and see if a new version is out. If so it should show a warning.

This should have a short timeout and fail gracefully (show nothing) if it can't reach it for any reason.

PR #142, opened 2023-02-17
add version update check on rtx version #74

roele:
Thought i give this a try as my very first attempt at working with Rust, so bare with me if the code is far away from best practices.

I wonder if the warning should be written as a log message with log level warn or to the standard out. I am also struggling to add tests for this since i could not figure out how to test log output for simplelog or mock the minreq request. Both topics i need to read up on.


src/cli/version.rs
```diff
@@ -47,6 +50,7 @@ pub fn print_version_if_requested(args: &[String], out: &mut Output) {
     if args.len() == 2 {
         let cmd = &args[1].to_lowercase();
         if cmd == "version" || cmd == "-v" || cmd == "--version" {
+            show_latest();
```
jdx: this should go after `show_version()`. I know that might seem a bit weird from a UX perspective, but it makes it so the user sees the version immediately regardless of anything that might happen with the remote fetching

src/cli/version.rs
```diff
@@ -57,6 +61,36 @@ fn show_version(out: &mut Output) {
...
 
+fn show_latest() {
+    if let Some(latest) = check_for_new_version() {
+        warn!("rtx version {} available", latest)
+    }
+}
+
+pub fn check_for_new_version() -> Option<String> {
+    let current = env!("CARGO_PKG_VERSION").to_string();
+    let current = Versioning::new(current.as_str()).unwrap();
```
jdx:
```suggestion
    let current = Versioning::new(env!("CARGO_PKG_VERSION")).unwrap();
```

what you're doing here is taking `env!("CARGO_PKG_VERSION")`, converting it from a `&str` into a String, then calling `as_str()` which converts it back again. There's no need to do that conversion.

src/cli/version.rs
```diff
@@ -57,6 +61,36 @@ fn show_version(out: &mut Output) {
...
+
+    if current < latest {
+        return Some(latest.to_string());
+    } else {
+        return None;
+    }
+}
+
+fn get_latest_version() -> Result<String, Error> {
+    let response = minreq::get("https://rtx.jdxcode.com/VERSION")
```
jdx: let's go with reqwest since it's the more popular client. By default it would use async which would require us adding tokio which would require a fair bit of changes, but you can use the blocking version of reqwest that won't require tokio.

Follow these instructions and it should be about the same as this: https://github.com/seanmonstar/reqwest#blocking-client

Another thing this must have is a short timeout. We don't want to block on this for any significant amount of time. I think to do that you'll need to construct a RequestBuilder with reqwest: https://dtantsur.github.io/rust-openstack/reqwest/struct.ClientBuilder.html

Make the timeout 1 second.

src/cli/version.rs
```diff
@@ -57,6 +61,36 @@ fn show_version(out: &mut Output) {
...
+        .with_timeout(1)
+        .send()?;
+    match response.status_code {
```
jdx: with reqwest, you can use `error_for_status` to do this instead: https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.error_for_status

if there is an error, record it with `debug!()` and return `None`. The output of this function should be either `None` if there was an error with http parsing. (This means the return type should not be `Result<>` but `Option<Version>`)

I would put the version parsing in here as well, but instead of calling `.unwrap()`, call [`.ok()`](https://doc.rust-lang.org/std/result/enum.Result.html#method.ok). What that does is make it so if version parsing fails it won't return an error, it'll return None.

The point of all of this is that this function wouldn't be able to return any error, only None or Some(Version). If it's none, we can compare the value, otherwise we don't check the version.

src/cli/version.rs
```diff
@@ -36,6 +37,8 @@ pub static VERSION: Lazy<String> = Lazy::new(|| {
     )
 });
 
+const LATEST_VERSION_UNKNOWN: &str = "0.0.0";
```
jdx: let's get rid of this, as explained below if the version isn't what we expect or anything goes wrong we simply just shouldn't attempt to show the warning

src/cli/doctor.rs
```diff
@@ -24,6 +25,10 @@ impl Command for Doctor {
             }
         }
 
+        if let Some(latest) = cli::version::check_for_new_version() {
+            warn!("rtx version {} available", latest)
```
jdx: this is a good addition, but let's change the messaging a bit
```suggestion
            warn!("new rtx version {} available, currently on {}", latest, env!("CARGO_PKG_VERSION"))
```

jdxcode (changes requested):
this is a great start, apologies that my feedback is so nitpicky here but I definitely want to make sure we do this correctly since I don't want to mess up `rtx version` in any situation

jdxcode:
let me know if you struggle with any parts here, join the discord if you want to chat about it

jdxcode:
don't worry about any new automated tests for this. I'm not concerned about validating this and it would be hard to do.

src/cli/version.rs
jdx: this should also resolve the openssl build failures we're getting in CI

jankatins:
This is BTW a nice crate for self updates from e.g. github releases, if you want that as well: https://github.com/jaemk/self_update

roele:
@jdxcode Thanks a lot for the detailed comments, this helps a lot and is much appreciated.

jdxcode:
run `just lint-fix` to fix the linter issues

Cargo.toml
```diff
@@ -47,6 +47,7 @@ once_cell = "1.17.0"
 owo-colors = { version = "3.5.0", features = ["supports-colors"] }
 rayon = "1.6.1"
 regex = "1.7.1"
+reqwest = { version = "0.11.14", features = ["blocking"] }
```
jdx, line 53:
```suggestion
reqwest = { version = "0.11.14", features = ["blocking"] }
openssl = { version = "0.10", features = ["vendored"] }
```

I think this will fix the build

arguably we could probably get away with disabling https since there isn't any reason http wouldn't be fine for this use-case, but I have plans for using https for downloading things in other places in rtx so I think we'll need it anyways

src/cli/version.rs
```diff
@@ -48,6 +49,7 @@ pub fn print_version_if_requested(args: &[String], out: &mut Output) {
         let cmd = &args[1].to_lowercase();
         if cmd == "version" || cmd == "-v" || cmd == "--version" {
             show_version(out);
+            show_latest();
```
jdx: do you think we could move this to line 59 instead? I think if we did that the tests would run it (of course this wouldn't validate it's actually working, but we can at least verify nothing is panicking)

Cargo.toml
roele: agree on keeping it since https is pretty much standard these days

jdxcode:
I believe that E2E test is failing in main for some reason since I had it on #145

roele:
e2e seems to fail because of direnv which also fails on my local machine (unit, e2e) but i don't have it intalled. also `test-coverage` won't run on my machine because tarpaulin cannot be found and adding it via cargo causes dependency issues with indexmap.

jdxcode:
yeah don't worry about that E2E failure, I'm fixing it in a different PR. This one looks ready, do you want to mark it not as draft?

jdxcode (approved):

jdxcode:
thanks for this!

Issue #98, opened 2023-02-04
tool-versions ref: and path:


jdxcode:
in asdf you can have tool-versions files with [ref: and path: tags](https://asdf-vm.com/manage/configuration.html). rtx doesn't currently support these, but it should.

jdxcode:
I've started to lay some groundwork in this in #145.

PR #145, opened 2023-02-18
feat: parallel install

### LINKED ISSUE #72: `rtx install --parallel`
I think this speaks for itself, we should be able to install runtimes (maybe plugins too, but that's less important since they're not that slow).

I'm not sure how well this would work, or how much it would improve performance since most builds will be parallel anyways, but I think we should offer this as an option. The tricky part here is that we'll probably need to not show spinners on each runtime install, but one spinner for all of them. Ideally it should get updated when each runtime is done installing.

### LINKED ISSUE #80: `rtx exec --plugin`
there should be a way to run `rtx exec` that will autoinstall the plugin

### LINKED ISSUE #133: show progress while installing runtimes
we can do a few things here:
- [x] show git url and sha after installing plugin
- [x] split progress text into download/install steps
- [x] show last line of output from install script

### LINKED ISSUE #134: `rtx install --all` should be the default
After demoing `rtx install --all` and using it for a while I'm a big fan of it. I think we should just make it the default behavior.

jdxcode:
Most of this is a huge refactor that simplifies the Toolset struct massively. Before it was heavily coupled to the Config struct. Now it just manages runtime versions. There was code strewn throughout the codebase to fetch versions and manually install, this consolidates a ton of that logic.

This made it possible to add several UX improvements:

* Parallel plugin and runtime installation (Fixes #72)
* Show last line of install output in UI (Fixes #133)
* Defaults to installing plugins alongside runtimes (Fixes #134 and Fixes #80)

It also lays some groundwork for fixing #98.

Originally I thought this would require a MV bump, but I think this is actually pretty safe to do even though it's a massive change under the hood. I don't think parallel install should mess up workflows (it can be disabled with `--jobs 1` or with equivalent settings).

This also changes the behavior to make `rtx install --all` the default behavior, so it will automatically install plugins. While this is a significant change, I don't think it would be a breaking change for anyone. The flag is still there it just is hidden and won't do anything.

The downside of this is we can't just output things to stderr since we have a huge progress bar on top of everything, so we have to be careful to suspend it if we want to output things.


jdxcode:
looks like I need to add openssl to the build image

jdxcode:
this is turning into a big change, but it's got a lot of great UX improvements and should vastly simplify the codebase. Here's a vide of where it is at so far.

https://user-images.githubusercontent.com/216188/219876716-4efba553-2741-47a5-90ff-5cf5f053c170.mov

jdxcode:
I think this is finally ready

PR #155, opened 2023-02-20
feat: added `rtx activate --status`

### LINKED ISSUE #154: Configuration option to hide a prompt?
When I switch between projects I get the output that looks like this:

```
rtx node@16.19.0 ruby@3.2.1
```

I don't really need this information all the time, is there an option to hide these messages?

jdxcode:
This changes the behavior of activate to default to not displaying the status line.

Fixes #154

### DIFF
--- a/README.md
+++ b/README.md
@@ -547,8 +547,8 @@ Arguments:
           [possible values: bash, fish, xonsh, zsh]
 
 Options:
-  -q, --quiet
-          Hide the "rtx: <PLUGIN>@<VERSION>" message when changing directories
+      --status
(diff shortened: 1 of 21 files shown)


tycooon:
What if some people still like the message? Not sure how to turn it on now 🤔

jdxcode:
huh? it's in the title of this PR: `rtx activate --status`

tycooon:
But I never use the `rtx activate` command 🤔

tycooon:
Actually I do that in my rc-file, sorry :)

Issue #25, opened 2023-01-28
alias management commands


jdxcode:
I've implemented commands for CRUD operations on settings as `rtx settings *`, we need the same for aliases:

```sh-session
$ rtx aliases
$ rtx aliases set "nodejs.lts/hydrogen" 18.0.0
$ rtx aliases get "nodejs.lts/hydrogen"
```

This is a great issue for someone that wants to contribute to the project since it'll be fairly straightforward how to do this. You can copy most of `src/cli/settings/*` to get started.

PR #159, opened 2023-02-21
feat: alias management commands (#25)

roele:
This is a draft on the `alias` management commands.

Couple of caveats
* not sure if 2-3 arguments should be used rather than the proposed dot notation `"<PLUGIN>.<ALIAS>"` (or even `"<PLUGIN@ALIAS>"`?)
* `ls` command uses option `-p` to specify a plugin but since this is for filtering this might not be an inconsistency
* adding an alias adds an empty `[alias]` section in the `config.toml`
* found a discrepancy with `aliases` (singular/plural) in `rtxrc.rs`


src/cli/alias/get.rs
```diff
@@ -0,0 +1,71 @@
...
+    }
+}
+
+static COLOR: Lazy<Color> = Lazy::new(|| Color::new(Stream::Stdout));
+static AFTER_LONG_HELP: Lazy<String> = Lazy::new(|| {
+    formatdoc! {r#"
+    {}
+      $ rtx alias get nodejs lts/hydrogen
+      18.0.0
+    "#, COLOR.header("Examples:")}
```
jdx: update this to be like other commands (I just changed this)

src/cli/alias/mod.rs
```diff
@@ -22,12 +25,18 @@ pub struct Alias {
 #[derive(Debug, Subcommand)]
 enum Commands {
     Ls(ls::AliasLs),
+    Get(get::AliasGet),
```
jdx, line 27: please alphabetize this (it's the order that will be shown to the user, so it matters)

src/cli/alias/set.rs
```diff
@@ -0,0 +1,61 @@
...
+use crate::ui::color::Color;
+
+/// Add/update an alias for a plugin
+///
+/// This modifies the contents of ~/.config/rtx/config.toml
+#[derive(Debug, clap::Args)]
+#[clap(visible_aliases = ["add", "create"], after_long_help = AFTER_LONG_HELP.as_str(), verbatim_doc_comment)]
+pub struct AliasSet {
+    /// The plugin to set the alias for
+    pub plugin: String,
```
jdx, line 18: these shouldn't use `pub` (unless I'm missing something)

jdxcode (changes requested):

jdxcode:
> not sure if 2-3 arguments should be used rather than the proposed dot notation "<PLUGIN>.<ALIAS>" (or even "<PLUGIN@ALIAS>"?)

I think I like multiple arguments the most. For runtimes I've settled on using `plugin@version` but that's so we can use multiple, e.g.: `rtx i nodejs@10 nodejs@12`. That seems less useful here.

> ls command uses option -p to specify a plugin but since this is for filtering this might not be an inconsistency

This is because it's a root command. It has to use an option because if you ran `rtx alias get` it wouldn't be possible to know if "get" was the name of a plugin or the name of a subcommand.

> adding an alias adds an empty [alias] section in the config.toml
worth noting but I think we can live with this

> found a discrepancy with aliases (singular/plural) in rtxrc.rs
yep, good catch

src/cli/alias/set.rs
roele: i was wondering about this as well but it seems other commands also define those as public

src/config/config_file/rtxrc.rs
```diff
@@ -309,6 +309,32 @@ impl RTXFile {
...
+                plugin_aliases.remove(from);
+                if plugin_aliases.is_empty() {
+                    aliases.remove(plugin);
+                }
+            }
+            if aliases.is_empty() {
+                doc.as_table_mut().remove("alias");
+            }
+        }
+    }
```
jdx: This can be cleaned up (possibly even more than this, but this was the best I could do). I tested this so I'm sure it's good.

```suggestion
    pub fn remove_alias(&mut self, plugin: &str, from: &str) {
        let doc = self.get_or_create_edit();
        if let Some(aliases) = doc.get_mut("alias").and_then(|v| v.as_table_mut()) {
            if let Some(plugin_aliases) = aliases.get_mut(plugin).and_then(|v| v.as_table_mut()) {
                plugin_aliases.remove(from);
                if plugin_aliases.is_empty() {
                    aliases.remove(plugin);
                }
            }
            if aliases.is_empty() {
                doc.as_table_mut().remove("alias");
            }
        }
    }
```

src/cli/alias/get.rs
```diff
@@ -0,0 +1,69 @@
...
+use indoc::formatdoc;
+use once_cell::sync::Lazy;
+
+use crate::cli::command::Command;
+use crate::config::Config;
+use crate::output::Output;
+
+/// Show an alias for a plugin
+///
+/// This is the contents of an alias.<PLUGIN> entry in ~/.config/rtx/config.toml
```
jdx, line 12: this is probably out of scope, but I wonder if we could generate this path using `env::CONFIG_FILE_PATH` or whatever that is.

Don't worry about this one unless you're particularly curious.

jdxcode (changes requested):

jdxcode (approved):

jdxcode:
I'll go ahead and merge this once the tests pass

jdxcode:
thanks again for the help!

src/config/config_file/rtxrc.rs
jdx: and btw I realize you probably just copy+pasted similar code that I had written in this file. This file is some of the oldest code in the codebase and has some of the worst code since I'm still new at rust.

PR #171, opened 2023-02-21
Allow global tool-versions file to be in config dir from a setting in the config.toml file

triarius:
In this PR, I introduce a configuration value that changes the location of the global tool-versions file to be a sibling of the `config.toml` file. By default, the location is `~/.config/rtx/tool-versions`. The config value is `false` by default to maintain backwards compatibility.

There are a few reasons one would want to do this. One is to eliminate yet another dotfile cluttering the `$HOME` directory. Motivated by this, I have expressed a more principled reason below:

The `~/.tool-verisons` file is unlike the `.tool-versions` files in other directories in that it controls the versions of rtx managed executables in directories that are not descendants in its directory hierarchy. For example, if you're on a linux system and `cd` to the directory `/tmp`, rtx will set the path to match what is in `$HOME/.tool-versions`, even though neither directory is a descendant of the other. This is an exception to the usual rule that rtx will search all ancestors of the current directory and set the path according to the `.tool-versions` it finds.

Of course, this behaviour is intentional because `~/.tool-versions` is the global tool versions file and such global configuration of tool versions is a feature that many users would want. However, I think it is important to separate the concepts because there is a collision between the global `.tool-versions` file and the local `.tool-versions` file for the `$HOME` directory. In my opinion, this collision warrants a different location for the global tool versions file. Furthermore, the global tool-versions file belongs in the same directory as the other global configuration in the `config.toml` file, so I have decided not to make the location configurable.

Something similar used to be achievable in asdf by setting the environment variable `ASDF_DEFAULT_TOOL_VERSIONS_FILENAME`, but it was changed in asdf to only affect the filename and not the path. I suspect rtx inherited this change as the equivalent `RTX_DEFAULT_TOOL_VERSIONS_FILENAME` behaves similarly. In any case, there are a few comments and issues floating around the asdf issue tracker by people that used to use that env variable to change the location of the global tool-versions file, for example: https://github.com/asdf-vm/asdf/issues/1365.

PS: I added and ran unit tests, but the running the e2e tests locally did not work for me, even on the main branch.


jdxcode:
I realize this isn't clear or working yet, but my intention is that `config.toml` could store what's in `.tool-versions` as the global configuration so that you wouldn't need a global `.tool-versions` file.

PR #192, opened 2023-02-24
feat: added `rtx plugins test`

jdxcode:


jdxcode:
I'm going to close as I'm not sure this is necessary right now, it's just syntax sugar on top of one of these:

```
rtx i nodejs@latest
rtx x nodejs@latest -- ...
```

or 

```
RTX_MISSING_RUNTIME_BEHAVIOR=autoinstall rtx x nodejs@latest -- ...
```

the latter I'm thinking could benefit from a flag like `--autoinstall`

PR #204, opened 2023-02-25
bug: immediately call hook-env after activating

### LINKED ISSUE #199: `rtx activate bash` and PATH inconsistencies
I use bash as my login shell, and as such have placed `eval "$(rtx activate bash)"` in my `.bashrc`.
The unexpected thing is that doing so adds *each* runtime's binary path to the shell, as shown

```
/home/karan/.local/share/rtx/installs/nodejs/19.3.0/bin:/home/karan/.local/share/rtx/installs/poetry/1.3.1/bin:/home/karan/.local/share/rtx/installs/rust/stable/bin:/home/karan/.local/share/rtx/installs/leiningen/2.10.0/bin:/home/karan/.local/share/rtx/installs/php/8.2.3/bin:/home/karan/.local/share/rtx/installs/php/8.2.3/sbin:/home/karan/.local/share/rtx/installs/php/8.2.3/.composer/vendor/bin:/home/karan/.local/share/rtx/installs/tokei/12.1.2/bin:/home/karan/.local/share/rtx/installs/zoxide/0.9.0/bin:/home/karan/.local/share/rtx/installs/yarn/1.22.19/bin:/home/karan/.local/bin:/home/karan/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/sbin:/sbin:/var/lib/snapd/snap/bin:/home/karan/.yarn/bin:/home/karan/.yarn/bin
```

I don't believe this is the intended behavior? I can say that that specific line produces this output becomes removing it and logging back in again does not produce this PATH.

Not sure what's going on here, but any help is appreciated. Please do let me know if any other information is required.

Info:
rtx ver: 1.16.1
shell: bash 5.2.15
OS: linux-x64 (Fedora 37)

jdxcode:
Fixes #199

If you use activate in an rc file, this will make it so the binaries are available immediately, e.g.:

```
eval "$(rtx activate bash)"
jq # before this change this wouldn't work if it came from rtx
```

### DIFF
--- a/e2e/cd/test_bash
+++ b/e2e/cd/test_bash
@@ -6,7 +6,6 @@ orig_path="$PATH"
 
 # shellcheck disable=SC1090
 eval "$(rtx activate bash --status)"
-_rtx_hook
 
 assert_path() {
   local expected="${1//$HOME/\~}:"
--- a/e2e/cd/test_fish
+++ b/e2e/cd/test_fish
@@ -4,7 +4,7 @@
 #set -l fish_trace 1
 rtx install nodejs@18.0.0 nodejs@16.0.0; or exit
 
-rtx activate fish | source && __rtx_env_eval
+rtx activate fish | source
 #rtx i
 
 test (node -v) = "v18.0.0"; or exit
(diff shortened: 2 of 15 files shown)


jdxcode:
this appears to have broke usage with direnv so I'm going to need to revert it

PR #242, opened 2023-03-01
chore: use GITHUB_TOKEN for E2E tests

jdxcode:
This secret should be available on forks so external
contributors should now have passing E2E tests

### DIFF
--- a/.github/workflows/rtx.yml
+++ b/.github/workflows/rtx.yml
@@ -30,7 +30,6 @@ jobs:
       - name: Run cargo nextest
         run: cargo nextest run --features clap_mangen
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/rtx.yml
```diff
@@ -30,7 +30,6 @@ jobs:
       - name: Run cargo nextest
         run: cargo nextest run --features clap_mangen
         env:
-          GITHUB_API_TOKEN: ${{ secrets.RTX_GITHUB_BOT_TOKEN }}
```
jdx, line 33: this shouldn't be necessary for unit tests anymore

PR #258, opened 2023-03-04
feat: dockerize dev setup

roele:
Gave this a try but this obviously still needs work. 

```bash
cd .devcontainer
docker build -t local/rtxdevcontainer .
docker run --rm -it -v "$(pwd)"/../:/workspaces/cached local/rtxdevcontainer
```

Mounting the project caused some trouble regarding the CARGO_TARGRET_DIR that clashes with the host which might be of another architecture and therefore causes linking issues etc.

Running in VS Code Dev Containers or plain docker gives mixed results. VS Code adds some environment variables of its own which troubles some tests. Other tests are failing because of different plugin installation paths. Also E2E fail because they can't find the generated `rtx` binary (seems it does not use CARGO_TARGET_DIR/debug/rtx).

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,26 @@
+FROM mcr.microsoft.com/devcontainers/rust:0-1-bullseye
+
+WORKDIR /workspaces/cached
+
+# Use another target directory to avoid conflicts with the host target directory
+RUN mkdir /workspaces/target
+ENV CARGO_TARGET_DIR=/workspaces/target
+
+# Install rust tools
+RUN rustup component add clippy llvm-tools rustfmt
+RUN cargo install cargo-insta cargo-llvm-cov
+
+# Install dependencies
+RUN export DEBIAN_FRONTEND=noninteractive \
+    && curl -fsSL https://deb.nodesource.com/setup_18.x | bash - \
+    && apt-get update \
+    && apt-get -y install --no-install-recommends \
+    # shells, direnv, shellcheck
+    bash fish zsh direnv nodejs shellcheck \
+    && apt-get clean \
+    # shfmt
+    && curl -sS https://webi.sh/shfmt | sh \
+    # just
+    && curl --proto '=https' --tlsv1.2 -sSf https://just.systems/install.sh | bash -s -- --to /usr/bin
+
+ENTRYPOINT [ "/bin/bash" ]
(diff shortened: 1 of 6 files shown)


jdxcode:
thanks for taking a look at this!

roele:
The image now runs most `just` targets in usual Docker mode. E2E test run but require manual cleanup of the directories first since it otherwise clashes with host architecture binaries. Target that still causes permissions issues is `render-help`.

jdxcode:
@roele is this ready for me to test?

roele:
@jdxcode Generally seems to work but tests still pose issues. Test `bin_paths` sometimes fails due to different paths, `e2e` fail intermittently and also `envrc` fails when run via VS Code Devcontainers due to an additional environment variables. Is there a way to see why e2e fails i just get an exit code 1.

jdxcode:
Add set -x to the script

roele:
The E2E that currently fail are `test_{bash,fish,zsh}` because of the following command which does not seem to work as expected in the container.

```
rtx shell nodejs@16.0.0 && _rtx_hook
test "$(node -v)" = "v16.0.0"
```

roele:
@jdxcode Feel free to play around. I would say it's 90% working (exept the 3 failing e2e tests and failing render-help). Ensure you run `just clean` first when you switch between local,  plain container or VS Code Devcontainer.

src/cli/direnv/envrc.rs
```diff
@@ -74,6 +74,11 @@ mod tests {
         let stdout = assert_cli!("direnv", "envrc");
         let envrc = fs::read_to_string(stdout.trim()).unwrap();
         let envrc = envrc.replace(dirs::HOME.to_string_lossy().as_ref(), "~");
+        let envrc = envrc
```
roele, line 77: @jdxcode I find that a bit ugly to have to deal with .devcontainer specifics like an additional env var in the test but could not think of anything better.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+RUN mkdir /workspaces/target
+ENV CARGO_TARGET_DIR=/workspaces/target
+
+# Install rust tools
+RUN rustup component add clippy llvm-tools rustfmt
+RUN cargo install cargo-insta cargo-llvm-cov
+
+# Install dependencies
+RUN export DEBIAN_FRONTEND=noninteractive \
+    && curl -fsSL https://deb.nodesource.com/setup_18.x | bash - \
```
roele, line 15: @jdxcode Not quite sure what system version is expected here but i guess 18.x should be fine?

jdx: I think there is a test somewhere that checks that it uses system node, I think it doesn't care which version of node is used

roele:
@jdxcode  Finally found some time to hunt down the e2e issues and finish this up. All just targets now run through in both variants (plain, VS Code). Also added a short description in the CONTRIBUTING.md

PR #268, opened 2023-03-05
Fix zsh deactivation

jaybosamiya:
Turns out the existing `rtx deactivate` doesn't actually work as expected. In particular, instead of removing `_rtx_hook`, it was actually removing everything _except_ `_rtx_hook` from the `precmd_functions` and `chpwd_functions`.

This commit fixes this.

Relevant documentation from zsh (https://zsh.sourceforge.io/Guide/zshguide05.html)
> The `(r)` flag takes a pattern and substitutes the first element of the array matched [...]

and

From https://zsh.sourceforge.io/Doc/Release/Expansion.html#Parameter-Expansion:
> `${name:#pattern}`
> [...] If `name` is an array the matching array elements are removed

### DIFF
--- a/src/cli/snapshots/rtx__cli__deactivate__tests__deactivate-2.snap
+++ b/src/cli/snapshots/rtx__cli__deactivate__tests__deactivate-2.snap
@@ -3,8 +3,8 @@ source: src/cli/deactivate.rs
 expression: output
 ---
 export PATH='$PATH'
-precmd_functions=( ${precmd_functions[(r)_rtx_hook]} )
-chpwd_functions=( ${chpwd_functions[(r)_rtx_hook]} )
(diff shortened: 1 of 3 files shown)


jdxcode:
hahaha, copilot wrote that and I really should've tested it

jdxcode:
I'll fix the snapshot tests for you

jdxcode:
tests are failing because github is rate limiting me

jaybosamiya:
Thanks for updating the snapshot tests; I didn't realize that I needed to update them.

With this fixed, I have now finally completely removed pyenv and anyenv and such from my system and switched over fully to rtx :D

jdxcode:
another happy customer! glad to hear it!

PR #284, opened 2023-03-07
Nix macos

jonathanmorley:


default.nix
```diff
@@ -9,7 +9,7 @@ rustPlatform.buildRustPackage {
     lockFile = ./Cargo.lock;
   };
 
-  buildInputs = with pkgs; [ coreutils bash direnv gnused git gawk ];
+  buildInputs = with pkgs; [ coreutils bash direnv gnused git gawk darwin.apple_sdk.frameworks.Security ];
```
jdx: What happens if this is run on Linux?

jonathanmorley: I dont have a linux nix host to test with unfortunately

jdx: I suspect it will fail, which means we can't merge this. cc @chadac

chadac: I can check this out, but generally platform-specific packages should be added via a conditional check on `system` rather than a catch-all. Why does the security package need to be added?

default.nix
```diff
@@ -27,7 +27,8 @@ rustPlatform.buildRustPackage {
   # is excluded by default from Nix.
   checkPhase = ''
     RUST_BACKTRACE=full cargo test --features clap_mangen -- \
-      --skip cli::plugins::ls::tests::test_plugin_list_urls
+      --skip cli::plugins::ls::tests::test_plugin_list_urls \
+      --skip cli::prune::tests::test_prune
```
chadac: Interesting that this fails during build on MacOS. What is the error specifically?

default.nix
jonathanmorley: Building with `jdxcode/rtx/main` causes:
> error: linking with `/nix/store/cqag9fv2gia03nzcsaygan8fw1ggdf4g-clang-wrapper-11.1.0/bin/cc` failed: exit status: 1
> ...
>   = note: ld: framework not found Security
>           clang-11: error: linker command failed with exit code 1 (use -v to see invocation)
>
>
> error: could not compile `rtx-cli` due to previous error

default.nix
jonathanmorley: building with `jonathanmorley/rtx/0f26a7fef9cb4ff917bea934cfb1f3c0c43d3fac` causes:

> failures:
>
> ---- cli::prune::tests::test_prune stdout ----
> thread 'cli::prune::tests::test_prune' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/cli/prune.rs:92:50

default.nix
chadac: In that case, could you append this package with a conditional check on macos systems like https://stackoverflow.com/a/67457556/4246250 ?

default.nix
chadac: I'm surprised this succeeded on my system. I think we need to include another substitution for `cmd!("git"` to be `cmd!("${git}/bin/git"`.

jdx: I believe there was a bug around this that was recently resolved, maybe try building from latest

chadac (approved):
Nice! Works for me, LGTM

Issue #288, opened 2023-03-08
env variables diff issue with shells other than bash


issmo:
The env diff is broken with some plugins exporting environment variables with bash
some environment variables are rendered differently by bash and their bash version gets included by rtx as patched variables while they should not.

I have provided a pull request which runs the plugin bash scripts with a new minimal environment to avoid this behaviour.

jdxcode:
can you give me more specifics on exactly what you're seeing? I'm not sure what we want is for this script to run in a completely empty environment.

issmo:
I am having issue with some environment variables being modified as a result of the execution of plugins' bash scripts.
I am using fish shell on macOS with tide plugin as prompt. setting a java version in .tools-version triggers the execution of ~/.local/share/rtx/plugins/java/bin/exec-env which sets the JAVA_HOME environment variable.
rtx executes this script and reads back the environment variables from bash. Some of the environment variables are detected as modified while they are not (for ex, in my case, the _tide_location_color containing escape character and KITTY_PUBLIC_KEY). I have noticed that bash returns slightly different values (uppercase escape char) for these user env variables which sets back these bash modified values in fish and breaking some functionalities (prompt, etc..).
The rationale of the suggested change is to build a new environment with only the necessary ASDF environment variables needed.
And you are completely right, the need is not to have a completely blank slate for these scripts but rather and "ASDF" compatible environment without user defined env variables. I am still working on completing the pull request

jdxcode:
Your solution is not going to work. Many plugins need access to variables that are not just ASDF_.

For one example, see asdf-golang which only sets GOPATH if not already set.

jdxcode:
it sounds like what we need to do is fix the escaping/unescaping logic to be compatible with these cases so they don't appear as a change

issmo:
This looks a safer way. I'll cancel the pull request which is way too specific to my use case.

jdxcode:
I hate that a lot of that escaping/unescaping is manual, however I couldn't figure out a way to do it that's compatible with bashv3.

If you can find a way to do more of this in bash and less of it that's manually escaping everything I'm all ears.

issmo:
Do you think it would be safer to have a dedicated rtx command/utility (for example: rtx env) that is responsible to output the environment variables instead of the command "export -p"? The variables would be collected through rust vars()
Would this be a viable option?

jdxcode:
[...]

PR #289, opened 2023-03-08
fix: run plugin scripts with a clean bash environment (#288)

issmo:


src/env_diff.rs
```diff
@@ -66,7 +66,11 @@ impl EnvDiff {
         let env: HashMap<String, String> =
             env.into_iter().map(|(k, v)| (k.into(), v.into())).collect();
         let out = cmd!(
+            "env", 
+            "-i",
             "bash",
+            "--norc",
+            "--noprofile",
```
jdx, line 73: I think these are doing nothing. These would only take effect if `-i` or `-l` were also passed.

issmo: True. I was testing in an interactive shell.
These are not needed.

issmo:
too specific and does not cover all cases

PR #295, opened 2023-03-11
bug: fix environment variables escaping (#288)

issmo:
Added a function to escape environment variables (see normalize_escape_sequences). It is mainly based on [ansi-c escape sequences](https://en.wikipedia.org/wiki/Escape_sequences_in_C#Table_of_escape_sequences) plus some extra chars ('`', '$'), these  were escaped by bash in my current environment.

I also rewrote the bash script output parsing to escape the values after parsing (mainly  for multiline variables).

**Unit tests**
The unit test were failing in my environment due to the original escaping method. They all now succeed.
I have modified one value in 'test_from_bash_script' which looks that it was incorrectly escaped.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,6 +1,0 @@
-coverage:
-  status:
-    project:
-      default:
-        target: 80%
-        informational: true
--- a/src/env_diff.rs
+++ b/src/env_diff.rs
@@ -78,44 +78,93 @@ impl EnvDiff {
         .read()?;
 
         let mut additions = HashMap::new();
-        let mut cur_key = None;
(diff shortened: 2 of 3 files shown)


jdxcode:
@issmo it wasn't clear to me what the changes to from_bash_script were intended to do. The tests didn't behave any differently with those changes. If I'm missing something from that let me know what test case I could add.

issmo:
it works fine with the initial parsing then looping through the variables and clean the additions hashmap. I just tried to do everything in one pass.

I have tested the update and so far everything works fine for me.
Thanks a lot.

PR #296, opened 2023-03-11
bug: fix environment variables escaping

### LINKED ISSUE #288: env variables diff issue with shells other than bash

jdxcode:
Fixes #288 

This is an extension of #295 which I didn't have write permissions to make extra commits to.

### DIFF
--- a/src/env_diff.rs
+++ b/src/env_diff.rs
@@ -78,44 +78,93 @@ impl EnvDiff {
         .read()?;
 
         let mut additions = HashMap::new();
-        let mut cur_key = None;
-        for line in out.lines() {
-            match line.strip_prefix("declare -x ") {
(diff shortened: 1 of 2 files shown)


jdxcode:
@issmo it wasn't clear to me what the changes to `from_bash_script` were intended to do. The tests didn't behave any differently with those changes. If I'm missing something from that let me know what test case I could add.

jdxcode:
huh seems I actually do have write permissions to #295, closing this in favor of reusing the existing PR

PR #317, opened 2023-03-13
feat: core plugins

jdxcode:


jdxcode:
I decided to implement this differently and have core plugin be written in rust instead of bash

PR #348, opened 2023-03-18
feat: plugin autoupdating

jdxcode:


jdxcode:
closing since this will need to change quite a bit

PR #354, opened 2023-03-18
WIP: Nushell support

toadslop:
Hello @jdxcode . I have a draft of support for nushell. It's still a bit rough and doesn't completely handle the deactivate command and I also didn't confirm that the tests work, but I was hoping you could take a look and let me know if you have any suggestions or noticed anything that I missed other than the two things I mentioned above.


src/shell/nushell.rs
```diff
@@ -0,0 +1,123 @@
...
+                    |before, after| _rtx_hook
+                  }}]
+              }}
+          }})
+            "#});
+
+        out
+    }
+
+    // TODO: properly handle deactivate
```
jdx: can you echo out this message for now?

```
echo "TODO: properly deactivate nushell"
```

we'll fix that later

toadslop: Sure.

toadslop: Do you mean that I should print it to the console when the command runs just to inform users that deactivate won't work properly?

jdx: Yes

toadslop: I added the line, but when I tested it for some reason RTX doesn't call Nushell's deactivate. It looks like it's calling zsh deactivate instead :/

toadslop: Ah, found the issue. Accidentally deleted my RTX_SHELL environment variable.

toadslop: Got deactivate working.

src/shell/mod.rs
```diff
@@ -14,6 +15,7 @@ pub enum ShellType {
     Fish,
     Xonsh,
     Zsh,
+    Nu,
```
jdx: Please alphabetize here and everywhere else

.tool-versions
```diff
@@ -1,4 +0,0 @@
-shellcheck 0.9.0
```
jdx: ?

README.md
```diff
@@ -433,6 +433,12 @@ echo 'eval "$(rtx activate bash)"' >> ~/.bashrc
 echo 'rtx activate fish | source' >> ~/.config/fish/config.fish
 ```
 
+### Nushell
+
+```sh-session
+$ rtx activate nu | save $nu.config-path --append
```
jdx: I recently changed this in most other places, but remove the $ here

toadslop: Ok

toadslop:
@jdxcode Added some changes to make the nushell activate script more robust. Please merge if there are no issues.

PR #368, opened 2023-03-22
added "trust" support to make rtx a bit more secure

jdxcode:
config files that do "dangerous" things will now need to be marked
as "trusted"


dhanak:
Cool and very much needed feature! One comment, though. The way `direnv` works is that it hashes the contents of the `.envrc` file, not just its path, and remembers that on its allow list. This way, when the file changes, it requires an explicit reauthorization. `rtx trust` remembers the path only, so it would be possible to sneak some malicious code into an already allowed `.rtx.toml` file. Please consider adjusting the trust system according to that. Thanks!

jdxcode:
I think that’s a mistake by direnv. It’s incredibly annoying to deal with and offers only a tiny advantage.

jdxcode:
I do have plans for a "paranoid" mode which file hashing could fall under

PR #410, opened 2023-03-26
refactor: complete command

### LINKED ISSUE #409: Consider renaming `complete` command to `completion` and use positional args for shell
It is extremely common for modern CLI's to use the convention `[cli] completion <shell>` for outputting shell completions. In fact it's so common I created a shell script to quickly install and refresh zsh completions using this convention.

I might suggest renaming `complete` as `completion` and using a position argument for the `shell` name rather then a flag to be more consistent with other CLIs. Or at the very least aliasing it and supporting both.

andrewthauer:
Fixes #409 

I'm very much a rust newbie. Any tips or suggestions would be appreciated.

- Renames `complete` to `completion` throughout
- Adds support for `shell_type` as a positional arg and the preferred method
- Hidden alias for `complete` and support for the `--shell` flag for backwards compatibility

One thing that currently isn't working and could use some help, is making the `[SHELL_TYPE]` required. My knowledge of rust pattern matching, etc. isn't great, so wasn't able to figure out the best way to do this.


src/cli/completion.rs
```diff
@@ -0,0 +1,67 @@
...
+    shell_type: Option<clap_complete::Shell>,
+
+    /// Shell type to generate completions for
+    #[clap(long, short)]
+    shell: Option<clap_complete::Shell>,
+}
+
+impl Command for Completion {
+    fn run(self, _config: Config, out: &mut Output) -> Result<()> {
+        let shell_type = self.shell_type.or(self.shell);
```
andrewthauer: This needs improvement. Was hoping to do something like https://github.com/jdxcode/rtx/blob/main/src/cli/activate.rs#L42 & https://github.com/jdxcode/rtx/blob/main/src/shell/mod.rs#L59.

src/cli/completion.rs
```diff
@@ -0,0 +1,67 @@
...
+        let shell_type = match shell_type {
+            Some(clap_complete::Shell::Bash) => clap_complete::Shell::Bash,
+            Some(clap_complete::Shell::Zsh) => clap_complete::Shell::Zsh,
+            Some(clap_complete::Shell::Fish) => clap_complete::Shell::Fish,
+            _ => clap_complete::Shell::Zsh,
+        };
```
jdx: This should be all you need I think
```suggestion
        let shell_type = self.shell_type.or(self.shell).expect("shell must be specified");
```

src/cli/completion.rs
```diff
@@ -0,0 +1,67 @@
...
+/// Generate shell completions
+#[derive(Debug, clap::Args)]
+#[clap(alias = "complete", verbatim_doc_comment, after_long_help = AFTER_LONG_HELP.as_str())]
+pub struct Completion {
+    /// Shell type to generate completions for
+    #[clap()]
+    shell_type: Option<clap_complete::Shell>,
+
+    /// Shell type to generate completions for
+    #[clap(long, short)]
```
jdxcode, line 23: add `hide = true` to this

jdx:
```suggestion
    #[clap(long, short, hide = true)]
```

jdx: also switch the names around so this hidden one is the one titled `shell_type`

jdxcode:
don't worry about that test failure, I'll fix that for you so you don't need to setup the snapshot stuff.

src/cli/completion.rs
jdx: btw `.expect()` will turn the type from `Option<clap_complete::Shell>` to `clap_complete::Shell`. That is also done in your pattern match but without needing to have the `_ => ` at the bottom.

You could also do it like this:

```rust
        let shell_type = self.shell_type.or(self.shell);
        let shell_type = match shell_type {
            Some(clap_complete::Shell::Bash) => clap_complete::Shell::Bash,
            Some(clap_complete::Shell::Zsh) => clap_complete::Shell::Zsh,
            Some(clap_complete::Shell::Fish) => clap_complete::Shell::Fish,
            None => panic!("no shell provided"),
        };
```

or this:

```rust
        let shell_type = match self.shell_type.or(self.shell) {
            Some(shell) => shell,
            None => panic!("no shell provided"),
        };
```

but I still think `.expect()` is the cleanest method

andrewthauer: Thanks!

src/cli/completion.rs
andrewthauer: Done

PR #424, opened 2023-03-29
docs: post-install configuration docs no longer requires rtx to be in ~/bin

decafdev:
The doc as is, are a little confusing if you used an install method that does not place rtx in `~/bin` here is a suggested update.


jdxcode:
this is intentional. It's not necessary for it to be on PATH: https://github.com/jdxcode/rtx#how-it-works

PR #436, opened 2023-04-04
docs: README: don't untar Github release

tombh:
README previously stated that the plain binary Github release needed untarring. Whereas it can in fact just be installed directly without untarring.


jdxcode:
thanks for the correction!

PR #455, opened 2023-04-09
core plugin: python

jdxcode:
adds python as a core plugin built in rust


jdxcode:
Will create a new issue for remaining work as this logic is behind an env var.

PR #477, opened 2023-04-12
fixes bug/typo

luccahuguet:
simple fix that addresses #470.


jdxcode:
you need to update the snapshots with `cargo insta test  --review`. I tried to do it for you but I don't have write permission to the PR.

luccahuguet:
thank you @jdxcode, I have since synced my forked with your commit on both branches and I expect it to work now. 

If it does not work, I'll updated the snapshots the way you described

PR #501, opened 2023-04-22
added npm shim

jdxcode:
this will make `npm i -g` automatically reshim

https://github.com/rtx-plugins/rtx-nodejs/issues/6

### DIFF
--- a/src/plugins/core/assets/nodejs_npm_shim
+++ b/src/plugins/core/assets/nodejs_npm_shim
@@ -1,0 +1,82 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# This script wraps npm so to run `rtx reshim` after global installs and uninstalls
+# Any other cases are passed-through to npm
+
+this_dir=$(dirname "${BASH_SOURCE[0]}")
+plugin_name=$(basename "$(dirname "$this_dir")")
+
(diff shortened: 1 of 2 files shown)


jdxcode:
@pdecat I'll cut a release with this by eod, once that happens would you mind trying the core plugin to see if it works for you? Just uninstall node/nodejs, enable experimental mode, then run `rtx i node@latest`. You can tell if it is using the core plugin by reading the output of `rtx doctor`.

pdecat:
It's working great, thanks @jdxcode!

pdecat:
FWIW, the difference between node and nodejs core plugins was unclear to me.

From https://github.com/jdxcode/rtx/blob/07160fa309b824c1437dca07d4c53722286e3a39/src/plugins/core/mod.rs#L19-L20 and https://github.com/jdxcode/rtx/blob/07160fa309b824c1437dca07d4c53722286e3a39/src/plugins/core/node.rs#L192

I understand it only concerns the [support for `.node-version`](https://github.com/jdxcode/rtx/blame/v1.29.1/README.md#L549).

Is that right?

jdxcode:
nodejs is hidden to retain compatibility with asdf's name for the plugin

PR #502, opened 2023-04-22
shims GA

### LINKED ISSUE #478: rtx reshim should NOT delete everything in the target folder
**Describe the bug**
`rtx reshim` deletes the `shims_dir` folder content

**To Reproduce**
- set in your config
```toml
experimental	= true              	# |false|	enable experimental features such as shims
shims_dir   	= '~/.local/bin/rtx'	# ||     	directory where shims are stored
```
- create a file in `~/.local/bin/rtx`
- run `rtx reshim`
- your file is gone

**Expected behavior**
Rtx should __only__ touch the files that __it__ __generates__

<details>
  <summary>rtx doctor output</summary>

```
~ ❯ rtx doctor
rtx version:
  1.27.12 macos-x64 (cc5b77f 2023-04-05)

build:
  Target: x86_64-apple-darwin
  Features: BREW
  Built: Wed, 05 Apr 2023 20:19:35 +0000
  Rust Version: rustc 1.68.0 (2c8cc3432 2023-03-06)
  Profile: release

shell:
  /opt/local/bin/fish
  fish, version 3.6.1

rtx environment variables:
  (none)

settings:
  {"experimental": "true", "missing_runtime_behavior": "prompt", "always_keep_download": "false", "legacy_version_file": "true", "plugin_autoupdate_last_check_duration": "10080", "trusted_config_paths": "[]", "verbose": "false", "asdf_compat": "false", "jobs": "4", "disable_default_shorthands": "false", "log_level": "INFO", "shims_dir": "~/.local/bin/rtx/shim", "raw": "false"}

config files:
  ~/.config/rtx/config.toml
  ~/.tool-versions

plugins:
  direnv   https://github.com/asdf-community/asdf-direnv.git#e3622e6
  nodejs   https://github.com/asdf-vm/asdf-nodejs.git#644ada3

toolset:
  nodejs@18

No problems found
```
</details>

**Additional context**
I had a few other rtx-related binaries in that folder that I thought would be nice to colocate and have a single path entry, but now they're poof

jdxcode:
This makes shims no longer only enabled in experimental mode. It also removes the ability to set a custom shims_dir.

Fixes #478

Apologies if this is disruptive (specifically removing the shims_dir config), but this was marked experimental so I could change the behavior.


Aeron:
It was unexpected to get the `Unknown config setting: settings.shims_dir` error message after an update, even considering a clear experimental label. But I fully understand where it came from, even the `~/.local/share` directory use.

Yet I suppose there are people who treat RTX (and ASDF before) similarly to Homebrew and have it installed in a directory like the `/opt/rtx`. So, the shims path would be the `/opt/rtx/shims` that causes no problems. Just like ASDF has it, along with the `installs` and `downloads` directories.

For those who used it that way, the only solution is to sym-link the existing `/opt/rtx/shims` (or whatever they have) to the implied `~/.local/share/rtx/shims` before the next re-shimming. A bit messy to my taste, but it’s an ok workaround.

PR #529, opened 2023-05-06
ci: Add AUR rtx-bin package

tombh:
See https://github.com/jdxcode/rtx/discussions/440

This won't run until ownership of the [rtx-bin AUR package](https://aur.archlinux.org/packages/rtx-bin) has been changed to jdxcode

TODO:
  * Would be good to refactor the code shared with the original `rtx` package scripts.

### DIFF
--- a/.github/workflows/rtx.yml
+++ b/.github/workflows/rtx.yml
@@ -272,6 +272,9 @@ jobs:
       - name: Release to aur
         run: scripts/release-aur.sh
         working-directory: rtx
+      - name: Release aur-bin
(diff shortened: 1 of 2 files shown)


jdxcode:
lgtm, I think you can just make me a co-maintainer. If we had that seems like we can merge this. I'm fine refactoring it in a later follow-up.

jdxcode (approved):

tombh:
Ah yes, co-maintainer, good thinking. Added you ✅

PR #534, opened 2023-05-07
added rtx outdated and upgrade command

### LINKED ISSUE #140: `rtx install node@16.1` installs node@1.1* not node@1.1.* as expected
this is a hard problem since the fix might break the behavior in unexpected ways. However because we're doing dumb prefix matching, that means `node@1.1` matches things like `1.10.0` when that's almost certainly not what the user wants.

This can be worked around by doing `node@1.1.` but ideally that shouldn't be necessary.

jdxcode:
Fixes #140
Fixes https://github.com/jdxcode/rtx/discussions/413
Fixes https://github.com/jdxcode/rtx/discussions/525


jdxcode:
Fixes #413 
Fixes #525

PR #581, opened 2023-05-27
Completions for aur/aur-bin and man pages for aur releases.

Sacro:
Nice that you generate them as files, nicer still if we can get them added to the package!


tombh:
I think this PR is causing the following error in `rtx-bin`:

```
==> Entering fakeroot environment...
==> Starting package()...
install: cannot stat 'rtx/completions/rtx.bash': No such file or directory
==> ERROR: A failure occurred in package().
    Aborting...
error: failed to build 'rtx-bin-1.30.6-1':
error: packages failed to build: rtx-bin-1.30.6-1
```

Which is because https://github.com/jdxcode/rtx/releases/download/v1.30.6/rtx-v1.30.6-linux-x64.tar.gz doesn't contain the shell completion files. So is it only the source file releases that contain the completion files?

jdxcode:
ok I'll revert this for now until I can test this properly

Sacro:
Give me a few minutes and I'll have a look. Sorry.

PR #592, opened 2023-06-01
use shasum instead of sha256sum

### LINKED ISSUE #590: install.sh uses sha256sum which is not always available
macos does not include sha256sum, we should switch to shasum for install.sh I think.

jdxcode:
Fixes #590


jdxcode:
WOMM

PR #601, opened 2023-06-06
Include man and bash completion in nix build

szaffarano:


jdxcode:
I don't know much about nix, but can we just get rid of this in favor of: https://github.com/NixOS/nixpkgs/pull/218450

PR #642, opened 2023-07-05
Update Nix install instructions

booniepepper:
This addresses https://github.com/jdxcode/rtx/issues/635

But!

It should only be merged _after_ `rtx` is available in some stable Nix release. See this backport PR for Nix 23.05: https://github.com/NixOS/nixpkgs/pull/241594

### DIFF
--- a/README.md
+++ b/README.md
@@ -389,37 +389,10 @@ makepkg -si
 
 #### nix
 
-For NixOS or those using the Nix package manager:
+For the Nix package manager, at release 23.05 or later:
 
-```nix
-{
-  inputs = {
-    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
-    flake-utils.url = "github:numtide/flake-utils";
-    rtx-flake = {
-      url = "github:jdxcode/rtx";
-      inputs.nixpkgs.follows = "nixpkgs";
-      inputs.flake-utils.follows = "flake-utils";
-    };
-  };
-
-  outputs = { self, nixpkgs, flake-utils, rtx-flake }:
-    flake-utils.lib.eachDefaultSystem(system:
-      let
-        pkgs = import nixpkgs {
-          inherit system;
-          overlays = [ rtx-flake.overlay ];
-        };
-      in {
-        devShells.default = pkgs.mkShell {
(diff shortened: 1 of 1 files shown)


jdxcode:
I can't imagine it would work on NixOS but it should work for Nix on other OSs

booniepepper:
I could remove the NixOS bit.

Where does rtx store its installations? A directory relative to the exe, or in some OS-specific place like `$XDG_DATA_DIRS`? I'm seeing `~/.local/share/...` on my machine with a non-Nix installation.

If the assets are stored in an area considered mutable, I thought they should work with NixOS just fine.

jdxcode:
rtx uses either prebuilt binaries which would probably not run on NixOS or compiles the tools from scratch. I don't know how well compiling something on NixOS would work since rtx relies on some tools/libraries on the host machine.

jdxcode:
It uses directories in your HOME directory so that shouldn't be a problem

booniepepper:
> ...but the operating system doesn't recognize them.

I'm going to assume that you mean that the executables are installed, but not on PATH in a shell. (As opposed to other contexts like running an IDE launched outside a user login shell context.)

This I'm not 100% sure how to setup that for Nix, but I think this should be handled in the Nix recipe rather than in `rtx` itself.

booniepepper:
> rtx uses either prebuilt binaries which would probably not run on NixOS or compiles the tools from scratch. I don't know how well compiling something on NixOS would work since rtx relies on some tools/libraries on the host machine.

I think this is the same with any Linux distro.

There's more ceremony around either setting up a nix recipe, or setting up something like LD_LIBRARY_PATH from (for example) `nix eval --raw nixpkgs.zlib` and so on

---

As a whole other idea, maybe just rip out all the installation instructions?

For what it's worth, I've generally avoided putting per-OS installation instructions in projects I've created. Usually I'll put [a repology badge](https://repology.org/project/rtx/badges) in the README and expect that people can figure out how to install with their package manager.

A wiki can be a different self-service way for people to manage installation instructions without as much ceremony.

But... I also don't have any CLIs with nearly as many stars yet, so I'd trust your judgement here. :)

PR #645, opened 2023-07-06
Port the "--refs" option of "plugin list" from asdf

knu:
This option is not documented in the reference but has been in `asdf plugin list` since [0.7.6](https://github.com/asdf-vm/asdf/blob/master/CHANGELOG.md#076).
https://github.com/asdf-vm/asdf/blob/d1a563dcc0107d5c631f73b114044898b5cadcf9/lib/functions/plugins.bash#L14

I was in the process of migrating my setup from asdf to rtx and found my local asdf plugin manager relied on this feature, so I thought I could add it to rtx myself.  This is my first time submitting a piece of Rust code, so please guide me if I'm doing anything wrong.


jdxcode:
thanks!

PR #664, opened 2023-07-09
alias nodejs -> node and golang -> go

### LINKED ISSUE #560: plugin aliases (node/nodejs)
node and nodejs are essentially copies of each other which do the same thing. IMO this is better than nothing and needing to remember if rtx uses node or nodejs, but it creates problems of its own. Projects could use one or the other which might make it so you need to install runtimes for both names.

We should fix this somehow. The solution might be as easy (though still not that easy) as always redirecting directories like `~/.local/share/rtx/installs/nodejs` to `~/.local/share/rtx/installs/node` if using the core plugin. A workaround might simply be adding a symlink to do this but I think we could do it internally.

Node isn't the only language we should do this for, go/golang is another. I don't think this necessarily has to be functionality that is accessible by plugins, we can just hardcode these 2 languages for now.

Note that when this is fixed I plan to remove the ability to have *separate* node|nodejs and go|golang plugins entirely. I do not think that's a use-case anyone will need, but if you do please let me know.

jdxcode:
Fixes #560

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,7 @@ v20.0.0
 - [Core Plugins](#core-plugins)
 - [FAQs](#faqs)
   - [I don't want to put a `.tool-versions` file into my project since git shows it as an untracked file.](#i-dont-want-to-put-a-tool-versions-file-into-my-project-since-git-shows-it-as-an-untracked-file)
+  - [What is the difference between "nodejs" and "node" (or "golang" and "go")?](#what-is-the-difference-between-nodejs-and-node-or-golang-and-go)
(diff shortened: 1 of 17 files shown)


maciej-szlosarczyk:
Hey @jdxcode!

I think this breaks compatibility with asdf. If I have the following .tool-versions file:

```
python 3.10.2
nodejs 18.14.0
```

And then when I run `rtx local python@3.11.2` it becomes:

```
python 3.11.2
node 18.14.0
```

At which point node no longer has a valid version definition in asdf.

jdxcode:
makes sense, I can probably make this a special case and keep the existing names

PR #679, opened 2023-07-14
feat: add envs for rtx go core plugin

dfang:
https://github.com/jdxcode/rtx/discussions/677

for env names: `RTX_GO_REPO` `RTX_GO_REPOSITORY` `RTX_GO_DOWNLOAD_MIRROR` `RTX_GOLANG_DOWNLOAD_MIRROR`, since there is `RTX_GO_SKIP_CHECKSUM` already in [src/env.rs](https://github.com/jdxcode/rtx/blob/main/src/env.rs#L168C12-L168C32), I choose short version too.

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -166,6 +166,8 @@ pub static RTX_GO_DEFAULT_PACKAGES_FILE: Lazy<PathBuf> = Lazy::new(|| {
     var_path("RTX_GO_DEFAULT_PACKAGES_FILE").unwrap_or_else(|| HOME.join(".default-go-packages"))
 });
 pub static RTX_GO_SKIP_CHECKSUM: Lazy<bool> = Lazy::new(|| var_is_true("RTX_GO_SKIP_CHECKSUM"));
+pub static RTX_GO_REPO: Lazy<String> = Lazy::new(|| "https://github.com/golang/go");
+pub static RTX_GO_DOWNLOAD_MIRROR: Lazy<String> = Lazy::new(|| "https://dl.google.com/go");
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum Confirm {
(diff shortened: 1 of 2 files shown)


src/env.rs
```diff
@@ -166,6 +166,8 @@ pub static RTX_GO_DEFAULT_PACKAGES_FILE: Lazy<PathBuf> = Lazy::new(|| {
     var_path("RTX_GO_DEFAULT_PACKAGES_FILE").unwrap_or_else(|| HOME.join(".default-go-packages"))
 });
 pub static RTX_GO_SKIP_CHECKSUM: Lazy<bool> = Lazy::new(|| var_is_true("RTX_GO_SKIP_CHECKSUM"));
+pub static RTX_GO_REPO: Lazy<String> = Lazy::new(|| "https://github.com/golang/go");
```
jdx: see https://github.com/jdxcode/rtx/pull/679/files#diff-4da8e8ee5c65a27bd048a2f2e55de0a9c75e87618fc02e17b235cc5413e625a0R14 for how to do this properly

dfang: thanks for your review, never write rust code before :)

src/plugins/core/go.rs
```diff
@@ -33,7 +33,7 @@ impl GoPlugin {
                 "git",
                 "ls-remote",
                 "--tags",
-                "https://github.com/golang/go",
+                env::RTX_GO_REPO,
```
jdx:
```suggestion
                &*env::RTX_GO_REPO,
```

src/plugins/core/go.rs
```diff
@@ -97,7 +97,7 @@ impl GoPlugin {
     fn download(&self, tv: &ToolVersion, pr: &ProgressReport) -> Result<PathBuf> {
         let http = http::Client::new()?;
         let filename = format!("go{}.{}-{}.tar.gz", tv.version, platform(), arch());
-        let tarball_url = format!("https://dl.google.com/go/{}", &filename);
+        let tarball_url = format!("{}/{}", env::RTX_GO_DOWNLOAD_MIRROR, &filename);
```
jdx:
```suggestion
        let tarball_url = format!("{}/{}", &*env::RTX_GO_DOWNLOAD_MIRROR, &filename);
```

jdxcode (changes requested):

src/plugins/core/go.rs
jdx: this must be dereferenced to get the string out of the wrapped Lazy<String>, then you have to reference that string since it can't be moved

jdxcode:
I've gone ahead and run `cargo fmt` for you to fix the lint issues in this PR but please run that yourself on your other PRs

dfang:
> I've gone ahead and run `cargo fmt` for you to fix the lint issues in this PR but please run that yourself on your other PRs

ok, I will update code on local not GitHub web editor and run `cargo fmt` before commit.

PR #680, opened 2023-07-14
add envs for rtx node core plugin

dfang:
https://github.com/jdxcode/rtx/discussions/677


dfang:
@jdxcode the unit test failed, I can't fix this, can you help on this

jdxcode:
I don't think it's related to your change. Let's see if it passes after we update the branch.

PR #685, opened 2023-07-14
added merge_group

jdxcode:


jdxcode:
looks like this is only available to organizations right now

PR #689, opened 2023-07-14
refactor version installs

jdxcode:
this is mostly a refactor but it happens to fix #688

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -4,7 +4,7 @@ use std::path::PathBuf;
 
 use indoc::indoc;
 
-use crate::{assert_cli, env};
+use crate::env;
 
 #[ctor::ctor]
 fn init() {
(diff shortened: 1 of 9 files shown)


src/test.rs
```diff
@@ -27,9 +27,6 @@ fn init() {
     env::set_var("RTX_MISSING_RUNTIME_BEHAVIOR", "autoinstall");
     //env::set_var("TERM", "dumb");
     reset_config();
-    assert_cli!("trust");
```
jdx, line 30: for some reason this causes the following now:

```
❯ cargo test
   Compiling linked-hash-map v0.5.6
   Compiling yansi v0.5.1
   Compiling similar v2.2.1
   Compiling diff v0.1.13
   Compiling yaml-rust v0.4.5
   Compiling pretty_assertions v1.4.0
   Compiling insta v1.31.0
   Compiling rtx-cli v1.34.2 (/Users/jdx/src/rtx)
    Finished test [unoptimized + debuginfo] target(s) in 9.77s
     Running unittests src/main.rs (target/debug/deps/rtx-43fac8e16d9010f0)
fatal runtime error: assertion failed: thread_info.is_none()
error: test failed, to rerun pass `--bin rtx`

Caused by:
  process didn't exit successfully: `/Users/jdx/src/rtx/target/debug/deps/rtx-43fac8e16d9010f0` (signal: 6, SIGABRT: process abort signal)
```

seems threads don't work right inside of ctor or something

Issue #711, opened 2023-07-25
different behaviors for rtx global local shell 


dfang:
**Describe the bug**

if I'm not mistaken, `global`, `local`, `shell` are compatible with `asdf`.  but there are some insignificant issues for now.

1. `global` and `local` are hidden from `rtx -h`, `shell` not
2. `global` and `local` got short alias `g` and `l`,  `shell` not
3.  `rtx global -h` and `rtx local -h` print help message, but `rtx shell` raise errors
4.  `rtx global` prints global versions seted in ~/.tool-versions, `rtx local` just cat the content `.rtx.toml`, `rtx shell` prints nothing

<img width="458" alt="Screen Shot 2023-07-25 at 9 00 31 AM" src="https://github.com/jdxcode/rtx/assets/1277047/2aba085e-d750-45f8-aeaa-0fd1cd09c0ba">

**To Reproduce**

```
λ rtx version
1.34.1 macos-x64 (76f8de3 2023-07-14)
```

**Expected behavior**

`global`, `local`, `shell` keep the same pattern of behaviors, I mean show all or hide all, all short aliased or not, etc.

**`rtx doctor` output**

```
λ rtx doctor
[WARN] rtx::toolset::tool_version_list: failed to resolve tool version: Failed listing remote versions for plugin bun: Failed to run /Users/kb24/.local/share/rtx/plugins/bun/bin/list-all: timed out
rtx version:
  1.34.1 macos-x64 (76f8de3 2023-07-14)

build:
  Target: x86_64-apple-darwin
  Features: SELF_UPDATE
  Built: Fri, 14 Jul 2023 17:59:57 +0000
  Rust Version: rustc 1.70.0 (90c541806 2023-05-31)
  Profile: release

shell:
  /bin/bash
  GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin20)
  Copyright (C) 2007 Free Software Foundation, Inc.

rtx environment variables:
  RTX_SHELL=bash
  RTX_GO_REPO=https://ghproxy.com/https://github.com/golang/go
  RTX_RUBY_BUILD_REPO=https://ghproxy.com/https://github.com/rbenv/ruby-build.git
  RTX_ALWAYS_KEEP_INSTALL=1
  RTX_PYENV_REPO=https://ghproxy.com/https://github.com/pyenv/pyenv.git
  RTX_EXPERIMENTAL=1
[...]

PR #713, opened 2023-07-25
rtx global shell local

dfang:
#711


dfang:
@jdxcode  for 3, I never used `nushell` and `xnosh`, can you help?
and also 4 is a little tough hard for me.

src/shell/bash.rs
```diff
@@ -29,7 +29,11 @@ impl Shell for Bash {
 
               case "$command" in
               deactivate|shell)
-                eval "$(command rtx "$command" "$@")"
+                if [ "$1" = "-h" ]; then
```
jdx: This is missing some cases like:
* `rtx shell --help`
* `rtx shell node@18 -h`

dfang: cases:
- `rtx shell -h`
- `rtx shell --help`
- `rtx shell node@18 -h`


bash and zsh

```
STR1="rtx shell --help"
STR2="rtx shell node@18 -h"
STR3="rtx shell -h"
SUB='-h'
if [[ "$STR1" == *"$SUB"* ]]; then
  echo "yes"
fi
if [[ "$STR2" == *"$SUB"* ]]; then
  echo "yes"
fi
if [[ "$STR3" == *"$SUB"* ]]; then
  echo "yes"
fi
```

fish: 
```
set STR1 "rtx shell --help"
set STR2 "rtx shell node@18 -h"
set STR3 "rtx shell -h"
set substring_to_find "*-h*"
if string match -q $substring_to_find $STR1
    echo "yes"
end
if string match -q $substring_to_find $STR2
    echo "yes"
end
if string match -q $substring_to_find $STR3
    echo "yes"
end
```

hi @jdxcode, I've added some sample script here, PTAL.

notpeelz:
This fails with `rtx shell -vhy` and other niche scenarios.

Try this instead (zsh/bash):
```diff
   case "$command" in
   deactivate|shell)
+    for arg in "$@"; do
+      if [[ "$arg" == "--help" || "$arg" =~ ^-[a-zA-Z]*h.*$ ]]; then
+        command rtx "$command" "$@"
+        return
+      fi
+    done
     eval "$(command rtx "$command" "$@")"
     ;;
```

I've never used fish so I don't know what the equivalent syntax would be.

jdx:
superseded by https://github.com/jdx/rtx/pull/976

PR #744, opened 2023-07-30
add openssl to cross image

jdxcode:


jdxcode:
going to try going with #745 instead in case openssl isn't on the client machine

PR #755, opened 2023-08-01
Improve bash _rtx_hook PROMPT_COMMAND insertion

scop:
Remove unnecessary use of a regex, match stricter.


jdxcode:
change is fine but you need to update the test snapshots

scop:
Done & rebased.

PR #783, opened 2023-08-09
feat: add --all to rm

dfang:
add --all to rm,  `rtx rm --all xxx` is available when this is merged


jdxcode (changes requested):
please add a test for this

also, this will not work correctly for something like `rtx uninstall --all node@18` which should remove all node versions with "18" as the prefix

jdx:
closing in favor of #989

PR #786, opened 2023-08-09
print load and unload envs like direnv

### LINKED ISSUE #733: Show output when loading env vars
With direnv, when I'd `cd` into a dir with a `.envrc` file, I would get an output in my terminal showing which vars were loaded.

Similarly when I exited the folder there was an "Unloaded" message.

Generally I liked this/didn't consider it noise as it's good to know when your shell env vars are changing.

Having replaced direnv with rtx and switched to `.rtx.toml` to define the env vars, while the env vars get loaded/unloaded correctly I get no output in my terminal, even with `settings.verbose` enabled.

dfang:
close #733

@jdxcode  maybe add an ENV to toggle this, like `RTX_ENV_LOAD_MESSAGE` or else?


src/hook_env.rs
```diff
@@ -158,9 +158,15 @@ pub fn build_env_commands(shell: &dyn Shell, patches: &EnvDiffPatches) -> String
     for patch in patches.iter() {
         match patch {
             EnvDiffOperation::Add(k, v) | EnvDiffOperation::Change(k, v) => {
+                if !k.starts_with("__RTX") && k != "PATH" {
+                    info!("load env {}={}", k, v);
```
jdx: This is not the right place for this code

jdxcode (changes requested):

src/hook_env.rs
jdx: this function is just a utility function to build the change list, it shouldn't be displaying any UI, that should be in the command

src/cli/hook_env.rs
```diff
@@ -54,6 +54,21 @@ impl Command for HookEnv {
         patches.push(self.build_diff_operation(&diff)?);
         patches.push(self.build_watch_operation(&config)?);
 
+        // output load and unload env like direnv
+        for patch in patches.iter() {
```
jdx, line 75: This should go inside of display_status() I think

src/cli/hook_env.rs
```diff
@@ -65,7 +55,47 @@ impl Command for HookEnv {
...
+                EnvDiffOperation::Remove(k) if !k.starts_with("__RTX") && k != "PATH" => {
+                    info!("unload env {}", k);
+                }
+                _ => {}
+            }
+        }
+    }
+
     fn display_status(&self, config: &Config, ts: &Toolset, out: &mut Output) {
+        self.display_env_load_unload_status(config, ts);
```
jdx: can we pass `patches` through so we don't need to regenerate it?

jdx: keep the extraction that added `get_patches()` though, that's a useful refactor

src/cli/hook_env.rs
```diff
@@ -65,7 +55,47 @@ impl Command for HookEnv {
...
+        patches.push(self.build_watch_operation(config)?);
+
+        Ok(patches)
+    }
+
+    fn display_env_load_unload_status(&self, config: &Config, ts: &Toolset) {
+        let patches = match self.get_patches(config, ts) {
+            Ok(patches) => patches,
+            Err(error) => {
+                eprintln!("Error: {}", error);
```
jdx: don't ever say "error" unless something actually failed, this should say "warn" and use `warn!()`

src/cli/hook_env.rs
```diff
@@ -65,7 +55,47 @@ impl Command for HookEnv {
...
+                return;
+            }
+        };
+
+        for patch in patches {
+            match patch {
+                EnvDiffOperation::Add(k, v) | EnvDiffOperation::Change(k, v) if !k.starts_with("__RTX") && k != "PATH" => {
+                    info!("load env {}={}", k, v);
```
jdx: don't display the value, the value could be massive. I think we should use direnv's syntax here for `+ADDED` `-REMOVED` `~CHANGED` https://github.com/direnv/direnv/blob/517e9f5b4159db2ce63695e4f51ca7b38b867876/internal/cmd/cmd_export.go#L102

jdx: also show everything on one line

PR #805, opened 2023-08-16
test out mold linker

jdx:


jdxcode:
looks to be maybe 10% faster, not sure it's worth it

PR #822, opened 2023-08-18
Adding check for `CI`

yhakbar:
Per the convo [here](https://github.com/jdxcode/rtx-action/pull/123), you wanted a `CI` check before error-ing on lack of prompt.

It's possible you also want this check for marking `.rtx.toml` files as trusted implicitly, but I think tackling that might be more involved.


jdxcode:
I beat you to it https://github.com/jdxcode/rtx/commit/c670aba007c578490479bdf499bafb3804572cdb

jdxcode:
I actually think there is already a check for trusted configs somewhere, though I forget where

PR #827, opened 2023-08-21
skip redownload if file exists

dfang:
skip redownload if file exists for go, bun, deno, java
for node and python, they are controlled by NODE_BUILD_BUILD_PATH and PYTHON_BUILD_BUILD_PATH


src/plugins/core/java.rs
```diff
@@ -110,6 +110,15 @@ impl JavaPlugin {
         let filename = m.url.split('/').last().unwrap();
         let tarball_path = tv.download_path().join(filename);
 
+        if tarball_path.exists() {
+            pr.set_message(format!(
+                "file {} exists, skipping download",
+                tarball_path.to_string_lossy()
+            ));
+            hash::ensure_checksum_sha256(&tarball_path, &m.sha256)?;
```
jdx: we need to still check the checksum to see if it is valid, if not we should redownload it. For languages that do not support checksums we should always redownload.

jdxcode (changes requested):

src/plugins/core/java.rs
dfang: currently go and java plugin have checksum support, deno and bun are downloading from GitHub,  bun has SHASUMS256.txt in release assets, but deno doesn't

dfang: If the network is not smooth, re-downloading might be very slow. I will try to add checksum support or env like RTX_GO_SKIP_CHECKSUM

src/tool.rs
```diff
@@ -333,7 +333,7 @@ impl Tool {
 
     fn create_install_dirs(&self, tv: &ToolVersion) -> Result<()> {
         let _ = remove_all_with_warning(tv.install_path());
-        let _ = remove_all_with_warning(tv.download_path());
+        // let _ = remove_all_with_warning(tv.download_path());
```
dfang: this applies to all plugins? not only core plugins?

PR #830, opened 2023-08-22
stop overwriting pre_prompt and PWD in nushell.rs

tsukimizake:
I tried to use rtx along with direnv on nushell, and found `source "~/Library/Application Support/nushell/rtx.nu"` overwrites hook from direnv.
Editing my rtx.nu like this fixes the problem on my environment.


tsukimizake:
Seems like tests failed 😕

jdxcode:
you just need to update the snapshots

tsukimizake:
Passed. Thanks!

PR #833, opened 2023-08-22
fix race condition with cache file writing

jdxcode:


jdxcode:
multiple processes were writing to the `.part` file which could only be moved by the first process. This makes each filename unique.

src/cache.rs
```diff
@@ -88,7 +89,9 @@ where
         if let Some(parent) = self.cache_file_path.parent() {
             fs::create_dir_all(parent)?;
         }
-        let partial_path = self.cache_file_path.with_extension("part");
+        let partial_path = self
+            .cache_file_path
+            .with_extension(format!("part-{}", random_string(8)));
```
booniepepper, line 94: I'm assuming whatever happened at this location before was overwriting previous work. If so, that means there was sort of some control over the total disk space that it would use. This is a recipe for many files (so might use up disk space or inodes depending on what goes here)

booniepepper (approved):

PR #858, opened 2023-08-28
add support for macos integration

roele:
This is first rough draft. Which works with with `sudo`, what i currently don't like yet is that using `sudo` installs the JDK with `root` as the owner which currently fails with an unspecific error (since installation is removed before the symlink remove is attempted). Having root as owner potentially leads to other issues as well.

```
$ sudo env RTX_JAVA_MACOS_INTEGRATION=true .bin/rtx install java@corretto-20.0.2.10.1
$ /usr/libexec/java_home -V
$ sudo env .bin/rtx uninstall java@corretto-20.0.2.10.1
$ /usr/libexec/java_home -V
```

### DIFF
--- a/docs/java.md
+++ b/docs/java.md
@@ -15,9 +15,9 @@ The code for this is inside the rtx repository at
 The following installs the latest version of java-openjdk-17.x (if some version of openjdk-17.x is 
 not already installed) and makes it the global default:
 
-```sh
-rtx use -g java@openjdk-17
-rtx use -g java@17         # alternate shorthands for openjdk-only
+```sh-session
+$ rtx use -g java@openjdk-17
+$ rtx use -g java@17         # alternate shorthands for openjdk-only
 ```
 
 See available versions with `rtx ls-remote java`.
@@ -28,9 +28,9 @@ Some applications in macOS rely on `/usr/libexec/java_home` to find installed Ja
 
 To integrate an installed Java runtime with macOS run the following commands for the proper version (e.g. openjdk-20).
 
-```sh
-sudo mkdir /Library/Java/JavaVirtualMachines/openjdk-20
-sudo ln -s ~/.local/share/rtx/installs/java/openjdk-20/Contents /Library/Java/JavaVirtualMachines/openjdk-20/Contents
+```sh-session
+$ sudo mkdir /Library/Java/JavaVirtualMachines/openjdk-20
+$ sudo ln -s ~/.local/share/rtx/installs/java/openjdk-20/Contents /Library/Java/JavaVirtualMachines/openjdk-20/Contents
 ```
 
 The distribution from  Azul Systems does support the integration but the symlink target location will differ from the example above (e.g `~/.local/share/rtx/installs/java/zulu-11.64.190/zulu-11.jdk/Contents`).
--- a/docs/java.md
+++ b/docs/java.md
@@ -22,23 +22,15 @@ rtx use -g java@17         # alternate shorthands for openjdk-only
 
 See available versions with `rtx ls-remote java`.
 
-## Configuration
-
-- `RTX_JAVA_MACOS_INTEGRATION` [bool]: enables macOS JAVA_HOME integration, defaults to false
-
 ## macOS JAVA_HOME Integration
 
 Some applications in macOS rely on `/usr/libexec/java_home` to find installed Java runtimes.
 
-The environment variable `RTX_JAVA_MACOS_INTEGRATION` enables the integration during the installation. On removal
-the integration will be cleaned up regardless of the value of the environment variable.
-
-Since the integration relies on symlinks within `/Library/Java/JavaVirtualMachines`, administrative privileges are
-required when using this feature.
+To integrate an installed Java runtime with macOS run the following commands for the proper version (e.g. openjdk-20).
 
 ```sh
-sudo RTX_JAVA_MACOS_INTEGRATION=true rtx install java@openjdk-20
+sudo mkdir /Library/Java/JavaVirtualMachines/openjdk-20
+ln -s ~/.local/share/rtx/installs/java/openjdk-20/Contents /Library/Java/JavaVirtualMachines/openjdk-20/Contents
 ```
 
-> Note: Not all distributions of Java SDK support this integration (e.g liberica). The integration will be simply
-> ignored in such cases even if the environment variable is set to true.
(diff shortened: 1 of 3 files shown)


docs/java.md
```diff
@@ -15,9 +15,30 @@ The code for this is inside the rtx repository at
...
+Some applications in macOS rely on `/usr/libexec/java_home` to find installed Java runtimes.
+
+The environment variable `RTX_JAVA_MACOS_INTEGRATION` enables the integration during the installation. On removal
+the integration will be cleaned up regardless of the value of the environment variable.
+
+Since the integration relies on symlinks within `/Library/Java/JavaVirtualMachines`, administrative privileges are
+required when using this feature.
+
+```sh
+sudo RTX_JAVA_MACOS_INTEGRATION=true rtx install java@openjdk-20
```
jdx: I don't think we can do this. It's going to create a bunch of root-owned files in your home directory that are going to cause problems later.

jdx (changes requested):

docs/java.md
jdx: what we could do is create the files with `chown $SUDO_USER:$SUDO_GROUP` if it's running under sudo. However I think rtx just creates too many files in too many places for us to be able to do that reliably.

jdx: I think the safest thing to do would be something like this instead:

```
warn!("to integrate java_home with this install, run `sudo ln -s ~/.local/share/rtx/... /Library/Java/JavaVirtualMachines`");
```

just print out a command with the ln instruction they need to run under sudo if it doesn't already exist

roele: I agree, having root owned files is definitely a no-go. Maybe your proposal to show an info message after the installation with the command to create the symlink manually (if the vendor supports it) is a better approach.

In this case we also might not need the environment variable to trigger the feature.

Will adapt the PR accordingly later today...

src/plugins/core/java.rs
```diff
@@ -255,6 +296,28 @@ impl Plugin for JavaPlugin {
...
+
+    if os() == "macosx" && is_enabled && !is_unsupported {
+        return true;
+    }
+    false
+}
+
+fn java_home_path(tv: &ToolVersion) -> PathBuf {
+    if os() == "macosx" {
+        if tv.install_path().join("Contents/Home").exists() {
```
roele: @jdx since we are removing the folding i assume this is required to ensure that already installed tools would continue to work. is this correct?

jdx: this looks correct, though by default rtx will only add `~/.local/share/installs/java/*/bin` to PATH by default. You will need to modify that to `~/.local/share/installs/ava/*/Contents/Home/bin` by implementing `list_bin_paths()` (or a function named something almost exactly that). Similar to what you did for exec_env.

jdx: I have some concerns here that people may be direct-linking their JDK in other places to `~/.local/share/rtx/installs/java/*` and we're moving that to a different place.

When I was thinking about this I was considering doing something like this instead:

```
ln -s ~/.local/share/rtx/installs/java/./Content/Home/* ~/.local/share/rtx/installs/java/./
```

that way we would merge the files and it would have the same structure as it does now but would also contain the files needed for macos integration

jdx:
thanks for taking this on, I have been looking at this over the past couple of weeks but have't found the time to work on it

docs/java.md
```diff
@@ -15,9 +15,22 @@ The code for this is inside the rtx repository at
...
+
+## macOS JAVA_HOME Integration
+
+Some applications in macOS rely on `/usr/libexec/java_home` to find installed Java runtimes.
+
+To integrate an installed Java runtime with macOS run the following commands for the proper version (e.g. openjdk-20).
+
+```sh
+sudo mkdir /Library/Java/JavaVirtualMachines/openjdk-20
+ln -s ~/.local/share/rtx/installs/java/openjdk-20/Contents /Library/Java/JavaVirtualMachines/openjdk-20/Contents
```
jdx: needs sudo as well

jdx (commented):
will try to test this myself probably tomorrow

roele:
Great thanks @jdx. 

Ran some manual tests with almost all distributions. Iberica does not ship with the required files and Zulu has the files but they're located in a different sub-folder (e.g. zulu-11.jdk). Added an addtional comment in the docs about that.

src/plugins/core/java.rs
```diff
@@ -153,6 +156,34 @@ impl JavaPlugin {
...
+                let dest = tv.install_path().join("Contents").join(entry.file_name());
+                trace!("moving {:?} to {:?}", entry.path(), &dest);
+                file::rename(entry.path(), dest)?;
+            }
+            file::make_symlink(
+                tv.install_path().as_path(),
+                &tv.install_path().join("Contents").join("Home"),
+            )?;
+            info!(
+                "To enable macOS integration, run the following commands: \
```
jdx: remove the backslashes so it prints on multiple lines, this is hard to read
CleanShot 2023-08-29 at 14 04 21@2x

also, at least on my machine it looks like the convention is to add ".jdk" to the end of the directory name

roele: agree, the output is not optimal and mixes even different output types (progress message `java -version`).
you're probably right, package installation folders usually end in `.jdk` even though it also seems to work fine without.

src/plugins/core/java.rs
```diff
@@ -153,6 +156,34 @@ impl JavaPlugin {
...
+                let dest = tv.install_path().join("Contents").join(entry.file_name());
+                trace!("moving {:?} to {:?}", entry.path(), &dest);
+                file::rename(entry.path(), dest)?;
+            }
+            file::make_symlink(
+                tv.install_path().as_path(),
+                &tv.install_path().join("Contents").join("Home"),
+            )?;
+            info!(
+                "To enable macOS integration, run the following commands:\n\
```
jdx: rust supports linebreaks inside of strings so you can just remove the end of these lines entirely

roele: tried that, it interpreted the spaces in front of the commands

jdx: oh! use [indoc](https://docs.rs/indoc/latest/indoc/) then, e.g.: https://github.com/jdx/rtx/blob/10647a97d7e05867549d77eb178aedadd128caa7/src/fake_asdf.rs#L22-L26

PR #896, opened 2023-09-14
fix: Always return a list on 'rtx ls <plugin> --json'

### LINKED ISSUE #882: Let `rtx ls --json` always return an array
A tool is installed, it returns an array:
```
# rtx ls vault --current --json
[
  {
    "version": "1.14.2",
    "requested_version": "1.14.2",
    "install_path": "/home/patrick/.local/share/rtx/installs/vault/1.14.2",
    "source": {
      "type": ".tool-versions",
      "path": "/home/patrick/.tool-versions"
    }
  }
]
```

When a tool isn't, it returns an object:
```
# rtx ls tflint --current --json
{}
```

That makes the output json hard to process with other tools like jq.
Always returning an array would help with that.

booniepepper:
closes #882

### DIFF
--- a/src/cli/ls.rs
+++ b/src/cli/ls.rs
@@ -126,25 +126,23 @@ impl Ls {
     }
 
     fn display_json(&self, runtimes: Vec<RuntimeRow>, out: &mut Output) -> Result<()> {
+        if let Some(plugin) = &self.plugin {
+            // only runtimes for 1 plugin
+            let runtimes: Vec<JSONToolVersion> = runtimes
+                .into_iter()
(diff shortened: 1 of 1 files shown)


src/cli/ls.rs
```diff
@@ -273,6 +271,19 @@ impl Ls {
 
 type RuntimeRow = (Arc<Tool>, ToolVersion, Option<ToolSource>);
 
+impl From<RuntimeRow> for JSONToolVersion {
```
jdx, line 274: so clean!

PR #949, opened 2023-10-18
fix deactivate with existing PROMPT_COMMAND

jdx:
Fixes https://github.com/jdx/rtx/issues/920

### DIFF
--- a/src/shell/bash.rs
+++ b/src/shell/bash.rs
@@ -52,6 +52,7 @@ impl Shell for Bash {
 
     fn deactivate(&self) -> String {
         formatdoc! {r#"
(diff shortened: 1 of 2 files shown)


src/shell/bash.rs
```diff
@@ -52,6 +52,7 @@ impl Shell for Bash {
 
     fn deactivate(&self) -> String {
         formatdoc! {r#"
+            PROMPT_COMMAND="${{PROMPT_COMMAND//_rtx_hook;/}}"
```
jdx, line 55: this maybe isn't the cleanest but I couldn't figure out a simpler way off the top of my head

PR #955, opened 2023-10-23
Updating README initial install instructions to make it easier on folks

yhakbar:
This updates the README.md initial install instructions so that the default location the binary is installed into is a location that's in most macOS/Linux users' PATHs, and so that the shell hooks don't assume an install location, and rely on the location detected based on PATH.

I think this might make the instructions easier to digest for folks that aren't very familiar with this kind of thing, as they should work OOB on a brand new macOS ARM machine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,19 +36,30 @@ Note that calling `which node` gives us a real path to node, not a shim.
 Install rtx on macOS (other methods [here](#installation)):
 
 ```sh-session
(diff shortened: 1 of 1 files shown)


jdx:
it works just fine if it's not on PATH

yhakbar:
> it works just fine if it's not on PATH
Is it not expected that the average user is going to install it in their PATH?

jdx:
no, rtx will add itself it PATH if it is not already on it

jdx:
closing since I don't want to have to deal with sudo

PR #994, opened 2023-11-13
fix: xonsh hook

CharString:
It was raising a TypeError:

```
xonsh: For full traceback set: $XONSH_SHOW_TRACEBACK = True
TypeError: 'method' object is not subscriptable
Exception raised in event handler; ignored.
```

I was able to build, but running the tests failed with `fatal runtime error: failed to initiate panic, error 5`. I have been using the debug build and it solves the problems.


jdx:
thanks!

CharString:
Thank you!

PR #998, opened 2023-11-15
Fix quoting in zsh configuration instructions

benweint:


benweint:
Ah, actually I don't think this will work because of the fact that the installation script seems to be rendered via `envsubst` somewhere not on the user's machine. Let me think on it more.

benweint:
(Updated the description here with more detail after finding https://github.com/jdx/rtx/pull/963, which I think introduced this issue!)

jdx:
Thanks!

PR #1013, opened 2023-11-23
Update README.md: add hint for section asdf to rtx

muescha:
add hint for section asdf to rtx

Since rtx main claim is to replace asdf, there should be more visible the steps a user need to change, if they have previous installed asdf.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 
 ## Features
 
-- **asdf-compatible** - rtx is compatible with asdf plugins and `.tool-versions` files. It can be used as a drop-in replacement.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -64,6 +64,8 @@ $ node -v
 v20.0.0
 ```
 
+**Note:** If you're moving from asdf to rtx, please review [How do I migrate from asdf?](#how-do-i-migrate-from-asdf) for guidance.
```
jdx: this doesn't feel like the right place, what do you think about instead modifying the first bullet point in the features: https://github.com/jdx/rtx#features

to something like this:

* asdf-compatible - rtx is compatible with asdf plugins and .tool-versions files. It can be used as a drop-in replacement. [See below for migration instructions](#how-do-i-migrate-from-asdf)

muescha: great idea - that is a much better place

PR #1047, opened 2023-12-02
load installed/current plugins when installing plugins

jdx:
Possibly a cheap workaround for https://github.com/jdx/rtx/issues/393. It should make the following
possible:

    rtx install python
    rtx install poetry # poetry depends on python

If you were to just run `rtx install` and try to install python and
poetry at the same time it would not work, however.


jdx:
Then it will use system python

PR #1054, opened 2023-12-02
fix deadlock

jdx:
This was introduced in #1047


cluther:
Thanks for fixing this so quickly! It took me a bit to figure out why my prompts stopped appearing without a CTRL-C, but once I did, a quick apt update and upgrade to rtx 2023.12.13 fixed it.

In case it matters, I'm using the rtx.pub/deb repository on Ubuntu 22.04.3.

PR #1111, opened 2023-12-07
[ISSUE-1110] handle null in java plugin

hkoosha:
Resolves https://github.com/jdx/rtx/issues/1110

### DIFF
--- a/src/plugins/core/java.rs
+++ b/src/plugins/core/java.rs
@@ -83,7 +83,10 @@ impl JavaPlugin {
             .sorted_by_cached_key(|(v, m)| {
                 let is_shorthand = regex!(r"^\d").is_match(v);
                 let vendor = &m.vendor;
-                let is_jdk = m.image_type == "jdk";
+                let is_jdk = m
+                    .image_type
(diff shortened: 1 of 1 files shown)


src/plugins/core/java.rs
```diff
@@ -239,7 +239,9 @@ impl JavaPlugin {
             .http
             .json::<Vec<JavaMetadata>, _>(url)?
             .into_iter()
-            .filter(|m| JAVA_FILE_TYPES.contains(&m.file_type))
+            .filter(|m| {
+                m.file_type.is_some() && JAVA_FILE_TYPES.contains(&m.file_type.clone().unwrap())
```
jdx: use [is_some_and](https://doc.rust-lang.org/std/option/enum.Option.html#method.is_some_and) for this and the next change

hkoosha: Even in `Display`? (it will allocate).

hkoosha: nvm, got it working with `as_ref`.

PR #1118, opened 2023-12-08
(nodejs plugin) add support for automatically shimming package managers with `corepack enable`

jasisk:
This is more to convey an idea than anything else—I'm not a rusty kind of person.

I use [corepack](https://github.com/nodejs/corepack)—which has shipped with node since v14 and v16—since I bounce between a bunch of different nodejs codebases that use different package managers. Love `rtx` but I have to remember to reenable corepack every time I install a new version.

So this is my attempt at running `corepack enable` automatically after installing a new node version.

What do you think? Figured this might convey what I was thinking better than opening a discussion not necessarily because I think this code should be merged. 😄


src/env.rs
```diff
@@ -167,6 +167,8 @@ pub static RTX_NODE_DEFAULT_PACKAGES_FILE: Lazy<PathBuf> = Lazy::new(|| {
         HOME.join(".default-npm-packages")
     })
 });
+pub static RTX_NODE_COREPACK_ENABLE: Lazy<bool> =
```
jdx: let's just make it `RTX_NODE_COREPACK=1`

src/plugins/core/node.rs
```diff
@@ -295,6 +325,11 @@ impl Plugin for NodePlugin {
         self.install_npm_shim(&ctx.tv)?;
         self.test_npm(ctx.config, &ctx.tv, &ctx.pr)?;
         self.install_default_packages(ctx.config, &ctx.tv, &ctx.pr)?;
+        if *env::RTX_NODE_COREPACK_ENABLE && self.corepack_path(&ctx.tv).exists() {
```
jdx: I think we should check the version and if it's before corepack was released (not sure when that was) we should not try to enable it even if they set this to true

src/plugins/core/node.rs
```diff
@@ -295,6 +325,11 @@ impl Plugin for NodePlugin {
...
+            self.test_corepack(ctx.config, &ctx.tv, &ctx.pr)?;
```
jdx: this step isn't really necessary since we call it in the next step

jdx:
in general looks good

src/plugins/core/node.rs
jasisk: Sounds good. It was introduced in two minors—v14.19.0 and v16.9.0—so I'll add a method to compare with versions that ship with corepack. Recommend I leave the `exists()` check in place even with the version check?

jdx: oh sorry I didn't notice that, I think that check is sufficient

PR #1185, opened 2023-12-14
README: Extra warning for Fish activation

Aeron:
Previously it was ok to have the `rtx activate fish | source` inside of an `if status --is-interactive` block, but not anymore. Most likely, it happened because of the auto-activation release. 

Now, it’ll give you the following error every time Fish initiates itself:

```text
fish: Unknown command: rtx
/opt/homebrew/share/fish/vendor_conf.d/rtx-activate.fish (line 2): 
  rtx activate fish | source
  ^~^
from sourcing file /opt/homebrew/share/fish/vendor_conf.d/rtx-activate.fish
	called on line 248 of file /opt/homebrew/Cellar/fish/3.6.4/share/fish/config.fish
from sourcing file /opt/homebrew/Cellar/fish/3.6.4/share/fish/config.fish
	called during startup
```

And I don’t think it’s a bug, but users must know it’s no longer a viable option. At least until the `RTX_FISH_AUTO_ACTIVATE` is not set to 0.

Also, sprinkled it with [GitHub alert formatting syntax][gh-md-alerts] a little.

[gh-md-alerts]: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts

### DIFF
--- a/README.md
+++ b/README.md
@@ -471,12 +471,15 @@ echo 'eval "$(rtx activate zsh)"' >> "${ZDOTDIR-$HOME}/.zshrc"
 echo 'rtx activate fish | source' >> ~/.config/fish/config.fish
 ```
 
-> **Note:**
->
+> [!NOTE]
 > For homebrew and possibly other installs rtx is automatically activated so
 > this is not necessary.
 >
 > See [`RTX_FISH_AUTO_ACTIVATE=1`](#rtx_fish_auto_activate1) for more information.
+
+> [!WARNING]
+> Do not place the activation code inside the `if status --is-interactive` block
+> until you explicitly set the `RTX_FISH_AUTO_ACTIVATE` to zero.
 
 #### Nushell
 


jdx:
this was resolved by using a full path in the activate script, though I'm wondering if `rtx activate` in the auto activation script should be in this conditional regardless. I use fish but personally never in a non-interactive situation. Is there a reason you have it in the block? My understanding is that while I don't think `rtx activate` should be used in a non-interactive session—I also don't think it would actually do anything either.

Well there would be a tiny benefit, `rtx shell` would work I guess.

Aeron:
I don’t think it’s a super-crucial matter, especially with 3rd-party shells like Fish. People are using it mostly interactively indeed.

But I like to separate things a bit. I use this block to isolate various initializers like Zoxide, Starship, etc. Also, my short aliases (sometimes a replacement and not always backward-compatible) and even environment variables. So that, in any possible case of shell invoked non-interactively, there’ll be no overheads and no risk to break or leak things. It’s just a bit of extra safety.

Otherwise, some applications can invoke it and mess things up a lot. Probably, it’ll be `sh` or `bash` anyway, but some apps can decide to use the default user shell. Or it can be a user automation script or something that uses Fish explicitly out of convenience. It’s all about corner cases, but it could be too late when it happens.

PR #1207, opened 2023-12-18
plugins: add rust core plugin

justinholt:
Added an experimental core plugin for rust.

It's essentially a native implementation of the [asdf-rust](https://github.com/asdf-community/asdf-rust) plugin.

The only configuration currently is `RTX_RUST_WITHOUT` which is a carbon copy of the `RUST_WITHOUT` variable from asdf-rust.

This project is incredible! Thanks for all the hard work!

### DIFF
--- a/docs/rust.md
+++ b/docs/rust.md
@@ -1,0 +1,23 @@
+# Rust in rtx
+
+The following are instructions for using the rust rtx core plugin. This is used when there isn't a
+git plugin installed named "rust".
+
+If you want to use [asdf-rust](https://github.com/asdf-community/asdf-rust)
+then use `rtx plugins install rust GIT_URL`.
+
+The code for this is inside the rtx repository at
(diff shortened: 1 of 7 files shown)


jdx:
sorry for the delay on this. I'm not too familiar with this plugin but I want to try using it a bit. The thing is I've always found rustup so good that rust is the one language I've never wanted to use asdf/rtx for.

I need to consider what the "official" recommendation is for rust development with rtx. I want to make sure that we can provide an experience that's better than rustup—otherwise I don't think we should even offer a core plugin.

justinholt:
@jdx Ah, I was actually unfamiliar with rustup (I've always used asdf-rust). It definitely seems superior, so this tool does seem unnecessary as a core plugin. I'll close for now.

jdx:
yeah frankly I would just use rustup and forget about asdf/rtx for rust. In particular you can use `rustup override set beta` to pick specific channels for different projects. If you ever do cross-compiling it's capable of selecting different toolchains as well—something I have a hard time imagining how rtx would be able to support.

PR #1287, opened 2023-12-28
docs: fix allowed values for `RTX_TASK_OUTPUT`

jasisk:
`rtx` panics if `RTX_TASK_OUTPUT` is not set to `Prefix` or `Interleave` ([here](https://github.com/jdx/rtx/blob/460e54772b2eda6bf75358c50ac6a694394c051d/src/cli/run.rs#L300-L301)).

Correcting the values in the readme as a stopgap solution.


jdx:
thanks for pointing this out!

Issue #1329, opened 2024-01-02
Error installing java zulu-21.30.15


chojnac:
**Describe the bug**
I'm unable to install java version `zulu-21.30.15`

**To Reproduce**
when running: `rtx install java@zulu-21.30.15` 
I'm getting 
```
rtx failed to execute command: ~/.local/share/rtx/installs/java/zulu-21.30.15/bin/java -version
rtx No such file or directory (os error 2)
rtx Run with --verbose or RTX_VERBOSE=1 for more information
```

**Expected behavior**
`zulu` is among versions supported by rtx (`rtx ls-remote java`), so expected behavior is that we should be able to install it. 

**`rtx doctor` output**

```text
rtx version:
  2024.0.0 macos-arm64 (2023-12-31)

build:
  Target: aarch64-apple-darwin
  Features: DEFAULT, NATIVE_TLS
  Built: Sun, 31 Dec 2023 22:05:35 +0000
  Rust Version: rustc 1.74.0 (79e9716c9 2023-11-13) (Homebrew)
  Profile: release

shell:
  /bin/zsh
  zsh 5.9 (x86_64-apple-darwin22.0)

rtx data directory:
  /Users/someuser/.local/share/rtx

rtx environment variables:
  RTX_SHELL=zsh

settings:
  {
    "experimental": false,
    "color": true,
    "always_keep_download": false,
    "always_keep_install": false,
    "legacy_version_file": true,
    "legacy_version_file_disable_tools": [],
    "plugin_autoupdate_last_check_duration": "7d",
    "trusted_config_paths": [],
    "log_level": "info",
    "trace": false,
    "debug": false,
    "verbose": false,
    "quiet": false,
    "asdf_compat": false,
    "jobs": 4,
    "shorthands_file": null,
    "disable_default_shorthands": false,
    "disable_tools": [],
    "raw": false,
    "yes": false,
    "task_output": null,
    "not_found_auto_install": true,
    "ci": false,
    "cd": null
  }

config files:
  /Users/someuser/.config/rtx/config.toml

plugins:
  bun      (core)
  deno     (core)
  go       (core)
  java     (core)
  node     (core)
  python   (core)
  ruby     (core)
  tuist    https://github.com/cprecioso/asdf-tuist.git#3ecc489

toolset:
  ruby@3.2.2

No problems found
```

**Additional context**
rtx installed using homebrew

gutaker:
[...]

PR #1343, opened 2024-01-03
fix(java): use tar.gz archives to enable symlink support

roele:
IMO easiest fix for #1329 would be to only support `tar.gz` for now. It seems all vendors ship archives in this format which also seems to support symlinks properly as opposed to ZIP (rs-zip).


jdx:
thanks!

PR #1345, opened 2024-01-03
aur: add "replaces" field

### LINKED ISSUE #1344: arch package rename
**Describe the bug**

Since `rtx` got renamed to `mise` there are two packages in the AUR which should probably be deprecated properly.

- https://aur.archlinux.org/packages/rtx
- https://aur.archlinux.org/packages/rtx-bin

Looking at the structure for `PKGBUILD` it looks like the new scripts should include a replaces section.

See: https://wiki.archlinux.org/title/PKGBUILD#replaces

jdx:
Fixes #1344
cc @inverse


inverse (approved):
LGTM - not an arch expert but this looks like it should do the job :)

PR #1346, opened 2024-01-03
Add additional conflicts

inverse:
Fixes for #1344 

when testing the previous change I was expecting to to prompt me to remove `rtx` with the following changes it did so.

I had `rtx-bin` installed and build the `miso-bin` package locally using `DRY_RUN=1 ./scripts/scripts/release-aur-bin.sh` and then running `makepkg` installing the artifact using `yay`.

```
malachi@pulsar aur-bin $ yay -U mise-bin-2024.1.3-1-x86_64.pkg.tar.zst 
loading packages...
resolving dependencies...
looking for conflicting packages...
:: mise-bin and rtx-bin are in conflict. Remove rtx-bin? [y/N] y

Package (2)  Old Version  New Version  Net Change

rtx-bin      2024.0.0-1                -15.84 MiB
mise-bin                  2024.1.3-1    15.83 MiB

Total Installed Size:  15.83 MiB
Net Upgrade Size:      -0.01 MiB

:: Proceed with installation? [Y/n] 
```

Feels a bit convuluted mentioning all varations of what it conlifcts with but I guess it's complete :)


inverse:
unless we really do want to allow them to be installed side by side? :thinking: 

in which case we can close this :)

PR #1381, opened 2024-01-06
fix(java): enable macOS integration hint for Zulu distribution

roele:
Currently the macOS integration hint is not shown for Zulu distributions since the archive does not contain the `/Contents` folder in the root directory but within a `zulu-{major}.jdk` directory.

We can simply symlink that directory and then check for a `/Contents` directory in the installation root to show the hint for all vendors that support the integration.

### DIFF
--- a/src/plugins/core/java.rs
+++ b/src/plugins/core/java.rs
@@ -174,8 +174,21 @@ impl JavaPlugin {
             file::rename(entry.path(), dest)?;
         }
 
+        if os() == "macosx" {
+            self.handle_macos_integration(&contents_dir, tv, m)?;
+        }
+
+        Ok(())
+    }
+
+    fn handle_macos_integration(
+        &self,
+        contents_dir: &Path,
+        tv: &ToolVersion,
+        m: &JavaMetadata,
+    ) -> Result<()> {
         // move Contents dir to install path for macOS, if it exists
-        if os() == "macosx" && contents_dir.exists() {
+        if contents_dir.exists() {
             file::create_dir_all(tv.install_path().join("Contents"))?;
             for entry in fs::read_dir(contents_dir)? {
                 let entry = entry?;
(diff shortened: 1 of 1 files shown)


src/plugins/core/java.rs
```diff
@@ -174,8 +174,21 @@ impl JavaPlugin {
             file::rename(entry.path(), dest)?;
         }
 
+        if os() == "macosx" {
```
jdx: this is nitpicky and arguably personal preference but I'd prefer this
```suggestion
        if cfg!(target_os = "macos") {
```

jdx: I'm not sure but I think there are some (certainly meaningless in this case) performance benefits to this. In theory the compiler should be able to just remove the conditional entirely

src/plugins/core/java.rs
```diff
@@ -191,6 +204,21 @@ impl JavaPlugin {
                 tv.install_path().as_path(),
                 &tv.install_path().join("Contents").join("Home"),
             )?;
+        }
+
+        // if vendor is Zulu, symlink zulu-{major_version}.jdk/Contents to install path for macOS
+        if m.vendor.as_str() == "zulu" {
+            let major_version = m.version.split('.').next().unwrap();
```
jdx: I think this would be a bit better:

```suggestion
            let (major_version, _) = m.version.split_once('.').unwrap_or_default(&m.version);
```

assuming it compiles that is

jdx: I know they should always have a dot but unwraps make me uncomfortable and I've found if nothing else they're actually very noisy. I've recently started to attempt to remove them whenever I can. If I'm diagnosing them I'm always drawn to `.unwrap()` even if it should always work

roele:
Thanks for your comments @jdx, they have been addressed.

jdx:
thanks!

henricook:
@roele I'm sorry to trouble you, would you mind providing a bit of background on why the whole macOS integration step is necessary, is it a problem with all java runtimes? Is it something I might be able to contribute to to make more automatic in mise (any guidance welcome if so).

PR #1395, opened 2024-01-07
Correct PATH for python venvs

### LINKED ISSUE #1394: Running python tasks is not hitting the venv binaries
**Describe the bug**
When using python in tasks, the used binary is the on on `mise` installed folders and not the `venv` one. Typically tasks to `pip install` are not targeting the venv

**To Reproduce**
```toml
[tools]
python = { version = "3.12", virtualenv = "{{config_root}}/.venv" }

[tasks.deps]
description = "install deps"
run = [
  "pip install -U pip",
  "pip install -U -r requirements.txt",
  "pip install -U -r requirements.dev.txt",
]
```
**Expected behavior**
Expecting the venv `python` to be called.

alikefia:
Try to fix #1394

### DIFF
--- a/src/toolset/mod.rs
+++ b/src/toolset/mod.rs
@@ -281,14 +281,14 @@ impl Toolset {
     }
     pub fn env_with_path(&self, config: &Config) -> BTreeMap<String, String> {
         let mut path_env = PathEnv::from_iter(env::PATH.clone());
+        for p in config.path_dirs.clone() {
+            path_env.add(p);
+        }
         let mut env = self.env(config);
         if let Some(path) = env.get("PATH") {
             path_env.add(PathBuf::from(path));
         }
         for p in self.list_paths() {
-            path_env.add(p);
-        }
-        for p in config.path_dirs.clone() {
             path_env.add(p);
         }
         env.insert("PATH".to_string(), path_env.to_string());
--- a/src/toolset/mod.rs
+++ b/src/toolset/mod.rs
@@ -281,15 +281,15 @@ impl Toolset {
     }
     pub fn env_with_path(&self, config: &Config) -> BTreeMap<String, String> {
         let mut path_env = PathEnv::from_iter(env::PATH.clone());
-        for p in config.path_dirs.clone() {
-            path_env.add(p);
+        let mut env = self.env(config);
+        if let Some(path) = env.get("PATH") {
+            path_env.add(PathBuf::from(path));
         }
         for p in self.list_paths() {
             path_env.add(p);
         }
-        let mut env = self.env(config);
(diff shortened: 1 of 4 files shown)


jdx:
I have not had time to dig into this but if I have an opportunity I will soon. At first glance this appears wrong though and the test cases failing seem to confirm that. I think this order should remain the same. The issue must be somewhere else.

alikefia:
This is what I am not sure about : the logic / intent of PathEnv (used only in Toolset.env_with_path)
- My first propal : fix the callee (change the logic)
- I have pushed another fix fixing the caller (adding paths in the rev order)

Up to you to decide on the logic you want to expose via PathEnv
- As-is (PR now) 
```
p.add("p1");
p.add("p2");
# p1 is fetched first which is less instinctive IMO but compliant with PATH logic (locations fetched in order)
```
- Proposal (Failed PR)
```
p.add("p1");
p.add("p2");
# p2 is fetched first : you add paths to override / precise
```

src/toolset/mod.rs
```diff
@@ -281,15 +281,15 @@ impl Toolset {
...
+        if let Some(path) = env.get("PATH") {
+            path_env.add(PathBuf::from(path));
         }
         for p in self.list_paths() {
             path_env.add(p);
         }
-        let mut env = self.env(config);
-        if let Some(path) = env.get("PATH") {
-            path_env.add(PathBuf::from(path));
+        for p in config.path_dirs.clone() {
```
jdx: ~~why is this on the bottom now? this is for users that manually set PATH in their `.mise.toml` files. That config should override any tool.~~

EDIT: oh I see what you're saying—that when you call `.add()` it actually works like a stack and the last thing becomes first. Let me look at this a bit more.

jdx: ok I see what we need to do here. Thanks for helping diagnose this most of the way but the solution is slightly different than what you have here. I'll publish the changes here shortly.

jdx:
the solution was slightly different than what you had. The issue was that "bin paths"—meaning ~/.local/share/installs/python/*/bin—was placed before the "env" paths, which are paths added in the "exec-env" step of plugins. That was incorrect, it should be "env" paths before "bin paths".

I'm not really sure when this broke. I don't think it was related to the name change. It's broken for everything aside from `hook-env` (`mise activate`).

jdx:
It looks like it broke in this commit: https://github.com/jdx/mise/commit/8920713ec71779ceffb4fe75eec6f5c0fb535fda

Fortunately it shouldn't happen again with the e2e test in place.

alikefia:
Thank you very much for support !

Issue #1380, opened 2024-01-06
PATH updating seems a bit dodgy


defhacks:
**Describe the bug**
relative or unqualified paths to mise alter $PATH unexpectedly

**To Reproduce**

Confirmed behaviour in both bash/zsh

relative paths yield an interesting result:
```
% ./mise activate zsh| grep PATH
export PATH=".:$PATH"
export __MISE_ORIG_PATH="$PATH"
```

but possibly more subtle and concerning is this case:
```
% mise activate zsh | grep PATH
export PATH=":$PATH"
export __MISE_ORIG_PATH="$PATH"
```

I was surprised to that both bash/zsh seem to interpret the empty string entry being pre-pended to $PATH here as adding the current working directory to $PATH.

I would propose patches to fix, but when I modified my local copy to prevent this behavior, I saw a comment about direnv, which I don't really use so I thought I'd just kick off the discussion rather than propose something that causes grief to direnv users.

jdx:
do you have experimental enabled? that actually goes down a completely different code path (see #863 if you're curious what)

PR #1400, opened 2024-01-08
prevent adding relative/empty paths during activation

defhacks:
attempting to address #1380

After some initial reading/testing, this seems to fix the problematic cases I found, and hopefully the test is helpful.

Also, because the behavior was a bit non-obvious to me, a link to some is_relative() examples

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e806075e086f5d669ae4dfc0f0b40aa9

### DIFF
--- a/e2e/test_path_safety
+++ b/e2e/test_path_safety
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+set -euo pipefail
+# shellcheck source-path=SCRIPTDIR
+source "$(dirname "$0")/assert.sh"
+
+export MISE_EXPERIMENTAL=1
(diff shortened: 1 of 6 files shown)


jdx:
oh I see, I hadn't considered people might be activating mise in this way but it makes sense now

defhacks:
`eval "$(mise activate zsh)"` is directly from the documentation I believe, and ends up being affected by this.
In testing, it seems that <empty string> being added to path is equivalent to adding "." in most shells (zsh/bash/nushell). It also seems that fish itself might have prevented the addition to PATH.

Issue #1272, opened 2023-12-24
visualize task dependency tree


jdx:
I want a way to see a graph of the dependency tree. I'm not at all sure what UI we could/should display here. If anyone knows of a CLI that is able to print decent graphs let me know (doesn't need to be rust)

mkos11:
@jdx Use Graphviz for a CLI solution. Install it, create a DOT file specifying dependencies, then run `dot -Tpng` to visualize.

roele:
Could start out as simple as

```
> brew deps --installed --tree
bat
├── libgit2
│   ├── libssh2
│   │   └── openssl@3
│   │       └── ca-certificates
│   └── openssl@3
│       └── ca-certificates
└── oniguruma
```

or

```
> mvn dependency:tree
[INFO] com.example:playground:jar:0.0.1-SNAPSHOT
[INFO] +- org.projectlombok:lombok:jar:1.18.24:provided
[INFO] +- com.google.code.gson:gson:jar:2.10.1:compile
[INFO] +- org.junit.jupiter:junit-jupiter:jar:5.8.2:test
[INFO] |  +- org.junit.jupiter:junit-jupiter-api:jar:5.8.2:test
[INFO] |  |  +- org.opentest4j:opentest4j:jar:1.2.0:test
[INFO] |  |  +- org.junit.platform:junit-platform-commons:jar:1.8.2:test
[INFO] |  |  \- org.apiguardian:apiguardian-api:jar:1.1.2:test
[INFO] |  +- org.junit.jupiter:junit-jupiter-params:jar:5.8.2:test
[INFO] |  \- org.junit.jupiter:junit-jupiter-engine:jar:5.8.2:test
[INFO] |     \- org.junit.platform:junit-platform-engine:jar:1.8.2:test
[INFO] +- org.mockito:mockito-core:jar:4.6.0:test
[INFO] |  +- net.bytebuddy:byte-buddy:jar:1.12.10:test
[INFO] |  +- net.bytebuddy:byte-buddy-agent:jar:1.12.10:test
[INFO] |  \- org.objenesis:objenesis:jar:3.2:test
[INFO] +- org.mockito:mockito-inline:jar:4.6.0:test
[INFO] \- org.mockito:mockito-junit-jupiter:jar:4.6.0:test
```

jdx:
gah I can't find it now but I did find a task runner or something similar that had a nice command for this.

[...]

PR #1440, opened 2024-01-12
feat: add command to print task dependency tree

roele:
First rough draft to visualize task depencies (#1272).

_Changes_
- moved `Deps` to `tasks/deps.rs` and made crate/struct public
- added new `task` command `deps [--dot] [TASKS]...`
- uses crate `ptree` to render  a tree directly from a `petgraph::Graph`
- uses crate `petgraph` to render a DOT representation which can be piped via `| dot -Tpng > ./graph.png`

_Open issues_
- [tree] ptree does not seem to allow customization of output and simply uses `fmt::Display` which currently has the format `[TASK] CMD...`
- [tree] sub-tasks seem to be in wrong order
- [dot] only renders nodes with incoming edges

Looking forward to some comments, as a Rust newby i feel there is a lot of room for improvement in that first rough draft.

### DIFF
--- a/src/cli/task/deps.rs
+++ b/src/cli/task/deps.rs
@@ -146,8 +146,8 @@ static AFTER_LONG_HELP: &str = color_print::cstr!(
   $ <bold>mise task deps</bold>
   Shows dependencies for all tasks
 
-  $ <bold>mise task deps task1</bold>
-  Shows dependencies for task1
+  $ <bold>mise task deps lint test check</bold>
+  Shows dependencies for the "lint", "test" and "check" tasks
 
   $ <bold>mise task deps --dot</bold>
   Shows dependencies in DOT format
--- a/src/cli/task/deps.rs
+++ b/src/cli/task/deps.rs
@@ -1,7 +1,7 @@
 use crate::{
     config::{Config, Settings},
-    task::{Deps, Task, TreeItem, TreeItemIndent, TREE_ITEM_CHARS},
-    ui::style,
+    task::{Deps, Task},
+    ui::{style, tree::print_tree},
 };
 use console::style;
 use itertools::Itertools;
@@ -94,43 +94,8 @@ impl TaskDeps {
         });
         // iterate over selected graph nodes and print tree
         for idx in start_indexes {
-            let _ = self.print_tree(&(&deps.graph, idx));
+            let _ = print_tree(&(&deps.graph, idx));
         }
-        Ok(())
-    }
-
-    pub fn print_tree<T: TreeItem>(&self, item: &T) -> Result<()> {
-        let indent = TreeItemIndent::new(4, 1, &TREE_ITEM_CHARS);
-        self.print_tree_item(item, String::from(""), String::from(""), &indent, 0)
-    }
-
-    fn print_tree_item<T: TreeItem>(
-        &self,
-        item: &T,
-        prefix: String,
-        child_prefix: String,
-        indent: &TreeItemIndent,
-        level: u32,
-    ) -> Result<()> {
-        print!("{}", prefix);
-        item.write_self()?;
-        println!("");
-
-        let children = item.children();
-        if let Some((last_child, children)) = children.split_last() {
-            let rp = child_prefix.clone() + &indent.regular_prefix;
-            let cp = child_prefix.clone() + &indent.child_prefix;
-
-            for c in children {
-                self.print_tree_item(c, rp.clone(), cp.clone(), indent, level + 1)?;
-            }
-
-            let rp = child_prefix.clone() + &indent.last_regular_prefix;
-            let cp = child_prefix.clone() + &indent.last_child_prefix;
-
-            self.print_tree_item(last_child, rp, cp, indent, level + 1)?;
-        }
-
         Ok(())
     }
 
--- a/src/cli/task/deps.rs
+++ b/src/cli/task/deps.rs
@@ -1,0 +1,265 @@
+use std::{
+    collections::{HashMap, HashSet},
+    sync::mpsc,
+};
+
+use crate::{
+    config::{Config, Settings},
+    task::Task,
+    ui::style,
+};
+use console::style;
+use itertools::Itertools;
+use miette::Result;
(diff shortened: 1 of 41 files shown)


src/cli/task/deps.rs
```diff
@@ -0,0 +1,265 @@
...
+  $ <bold>mise task deps task1</bold>
+  Shows dependencies for task1
+
+  $ <bold>mise task deps --dot</bold>
+  Shows dependencies in DOT format
+"#
+);
+
+#[derive(Debug)]
+pub struct Deps {
```
jdx: this should go into `src/tasks.rs` I think

jdx:
> moved Deps to tasks/deps.rs and made crate/struct public

I wonder if we should reconsider the term `depends` used on task definitions to also be "depends".

> [tree] ptree does not seem to allow customization of output and simply uses fmt::Display which currently has the format [TASK] CMD...

we could fork it (no update in 2 years), or change the way we use fmt::Display in other places, or you could use a different data structure maybe

I think it's definitely on the right track though, mostly the fmt::Display issue makes it not look great but I think when that is fixed it will look incredible. DIdn't test the dot stuff, I'll wait a bit till you get the last thing fixed.

Definitely a good start though, especially as a rust newbie. That petgraph stuff definitely isn't easy to work with.

src/cli/task/deps.rs
roele: agree, that looks like a better place for this

roele:
- moved `Deps` to `task.rs`
- removed crate `ptree` and re-implemented rendering of the `Graph` as tree

_TODO_
- move generic tree rendition structs and code to `ui/tree.rs`
- clean up and simplify code where possible
- revisit DOT format because of missing tasks without edges

src/cli/task/deps.rs
```diff
@@ -0,0 +1,155 @@
...
+  $ <bold>mise task deps</bold>
+  Shows dependencies for all tasks
+
+  $ <bold>mise task deps task1</bold>
+  Shows dependencies for task1
+
+  $ <bold>mise task deps --dot</bold>
+  Shows dependencies in DOT format
+"#
+);
```
jdx, line 154:
```suggestion
);

#[cfg(test)]
mod tests {
    #[test]
    fn test_tasks_deps() {
        assert_cli_snapshot!("tasks", "deps", @"");
    }
}
```

then just run `mise run snapshots` and it will update the string to have the correct output

src/cli/task/deps.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+}
+
+static AFTER_LONG_HELP: &str = color_print::cstr!(
+    r#"<bold><underline>Examples:</underline></bold>
+  $ <bold>mise task deps</bold>
+  Shows dependencies for all tasks
+
+  $ <bold>mise task deps task1</bold>
+  Shows dependencies for task1
```
jdx: would be good to show a concise output here

src/cli/task/deps.rs
```diff
@@ -0,0 +1,155 @@
...
+                &|_, _| String::new(),
+                &|_, nr| format!("label = \"{}\"", nr.1.name),
+            ),
+        );
+        Ok(())
+    }
+
+    fn err_no_task(&self, config: &Config, t: &str) -> miette::Report {
+        let tasks = config.tasks();
+        let task_names = tasks.keys().sorted().map(style::ecyan).join(", ");
```
jdx, line 137: I think you said you're new at rust but based on this line I think that's clearly not the case

src/cli/task/deps.rs
```diff
@@ -0,0 +1,155 @@
...
+        });
+        match tasks {
+            Some(Ok(tasks)) => Ok(tasks),
+            Some(Err(e)) => Err(e),
+            None => Ok(vec![]),
+        }
+    }
+
+    ///
+    /// Print dependencies as a tree
```
jdx, line 76: nice comments, I'm not as good as I should be about this sort of thing

src/cli/task/deps.rs
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        Ok(())
+    }
+
+    fn get_all_tasks(&self, config: &Config) -> Result<Vec<Task>> {
+        config
+            .tasks()
+            .iter()
+            .filter(|(n, t)| *n == &t.name) // filter out aliases
```
jdx: maybe the aliases shouldn't even be in this? and we should have `config.tasks_with_aliases()` instead?

I said "maybe" but I actually think "probably"

src/cli/task/deps.rs
```diff
@@ -0,0 +1,155 @@
...
+        let tasks = if self.tasks.is_none() {
+            self.get_all_tasks(&config)?
+        } else {
+            self.get_task_lists(&config)?
+        };
+
+        if self.dot {
+            let _ = self.print_deps_dot(&config, tasks);
+        } else {
+            let _ = self.print_deps_tree(&config, tasks);
```
jdx: not clear to me why you're swallowing the error here, if you can't figure something out with rust lmk

roele: good catch

src/cli/task/deps.rs
roele: what do you mean by that exactly?

src/cli/task/deps.rs
roele: that is used somewhere else already so i took the liberty of copying it

roele: ah i assume concise with the `run` command docs...

src/cli/task/deps.rs
roele: ah i assume concise with the `run` command docs

roele:
I revisited the DOT format export but it seems there is no way to show tasks which do not have any edges (dependencies). Also the export has almost none configuration options.

Also `EndBug/add-and-commit@v9` fails the unit test step currently and might require a smarter condition to prevent trying to update my repository to which it has no access to.

jdx:
nice work!

PR #1452, opened 2024-01-13
added support for installing directly from npm

jdx:
e.g.: `mise use -g npm-prettier@3.2.0`


jdx:
cc @pepicrft, not actually very much code at all

Issue #1383, opened 2024-01-06
simplify `mise settings` commands


jdx:
here's what I would like to see:

```sh
$ mise settings # list all
a = true
b = true
$ mise settings a # get a value
true
$ mise settings a=false # set a value
```

I believe `mise env-vars` already works this way so we should be able to copy that logic.

4ydx:
FYI

```
HAL:~/test$ mise version
2024.1.14 linux-x64 (187cc24 2024-01-09)
HAL:~/test$ mise env-vars a=b
HAL:~/test$ mise env-vars
a=b ~/test/.mise.toml
HAL:~/test$ mise env-vars a
error: invalid value 'a' for '[ENV_VARS]...'

For more information, try '--help'.
```

Are you wanting env-vars to also have "get a value" functionality?

jdx:
definitely

PR #1454, opened 2024-01-13
change settings interface

JosiasAurel:
addresses #1383 

```shell
$ mise settings
  List all settings

$ mise settings a
  Get the value of setting with name 'a'

$ mise settings a=b
  Set setting with name 'a' to 'b'
```


src/cli/settings/set.rs
```diff
@@ -1,133 +0,0 @@
-use eyre::Result;
```
jdx, line 1: users may still be using the old syntax so they need to be able to still use these old commands

camsteffen:
I think it would be simpler if the setting value is a separate argument instead of using `=`? This is how `git config` works. - `mise settings set x value`

jdx:
@camsteffen I want to support both styles, but this matches how `mise set` works for consistency

camsteffen:
Sorry I meant `mise settings x value`. Would that work?

jdx:
ah yeah, that could be done I think

jdx:

Issue #1274, opened 2023-12-26
`rtx run` to show interactive selector


jdx:
I think if there is no "default" task, then `rtx run` should show a list of available tasks allowing you to select one to execute

PR #1459, opened 2024-01-14
feat: add interactive selection for tasks if task was not found

roele:
Adds an interactive task selector as mentioned in #1274.

- adds cask `inquire`
- replaces `task not found` error with interactive selection of available tasks

Did not find an easy way to test this. Not sure if theres an easy way send user input to the console.

### DIFF
--- a/src/cli/run.rs
+++ b/src/cli/run.rs
@@ -10,8 +10,9 @@ use std::time::SystemTime;
 
 use clap::ValueHint;
 use console::{style, Color};
+use demand::{DemandOption, Select};
 use duct::IntoExecutablePath;
-use eyre::Result;
+use eyre::{OptionExt, Result};
 use globwalk::GlobWalkerBuilder;
 use itertools::Itertools;
 use once_cell::sync::Lazy;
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -63,6 +63,7 @@ indenter = "0.3"
 indexmap = { version = "2", features = ["serde"] }
 indicatif = { version = "0.17", features = ["default", "improved_unicode"] }
 indoc = "<3"
+inquire = { version = "0.6.2", default-features = false, features = ["console"] }
```
roele: Not sure if this is the right choice or if `console-rs/dialoguer` would be a better one

jdx: use demand, you can see an example with `mise up --interactive`

roele: thought i've seen something interactive somewhere but couldn't find it anymore. thanks for the hint!

jdx:
wow does it work? it's crazy how little code that was

jdx:
nice, it works great!
Screenshot 2024-01-14 at 07 48 54

I saw you filed a bug on demand which isn't too surprising, it hasn't been used that much yet but I think it's got a lot fo promise. I wasn't very happy with inquirer

roele:
Yeah was surpised too how easy this was. The issue in `demand` is probably an edge-case which i ran into because i used the terminal in my IDE with aprox. 5 rows height only.

PR #1460, opened 2024-01-14
fix: bail out of task suggestion if there are no tasks

roele:
we should probably `bail` if the user is in a directory with no taks defined


src/cli/run.rs
```diff
@@ -332,6 +332,9 @@ impl Run {
 
     fn prompt_for_task(&self, config: &Config, t: &str) -> Result<Task> {
         let tasks = config.tasks();
+        if tasks.is_empty() {
+            bail!("no tasks defined in .mise.toml or .mise/tasks/")
```
jdx: throw a link to the docs https://mise.jdx.dev/tasks/

PR #1462, opened 2024-01-15
fix: no panic on missing current dir

tamasfe:
The shell is spammed with a panic message after each command if the current directory is not available, with this PR simply mise ignores/logs if that is the case instead of a hard crash.

I noticed the issue while browsing a remote snapshot directory tree that was FUSE mounted by [kopia](https://github.com/kopia/kopia).


tamasfe:
I've just realized that global/system config could still be read, considering this is such a niche case I assume it doesn't matter, still would it be worth it to open a separate PR for that?

jdx:
we should probably support that. Is there a user that is going to have problems with it? I would expect probably not for a long time, but I also think it _will_ happen one day and if we don't fix this now I think it would be very hard to diagnose.

PR #1470, opened 2024-01-16
add support for installing directly with go modules

endigma:
This mimics the NPM/Cargo functionality for Go install.

Things you should look at:
1. is that the best way to get Go into the build/test process?
2. should I be only installing the binary into the rtx path or moving the whole gopath like I am currently? The whole GOPATH benefits sandboxing I guess but would almost definitely result in duplication of libraries across similar project installs. Could alternatively set GOBIN and do an mkdir of the /bin inside the version folder.

Things I noticed while contributing:
- docs on *running* mise or tests without the devcontainer are sparse at best, couldn't figure out how to deactivate mise enough to satisfy the tests
- perhaps an option to run tests in a container would be good? Maybe stepchowfun/toast or earthly or similar?
- It is also pretty hard to wipe a messed up package install, or at least undocumented.
- you mention something about a shim in CONTRIBUTING.md, but I couldn't figure out how this was meant to work it seems like not all the info required to set it up
- pre commit hooks don't work without some md-magic thing I don't have installed

### DIFF
--- a/src/forge/go.rs
+++ b/src/forge/go.rs
@@ -16,7 +16,7 @@ pub struct GoForge {
 
 impl Forge for GoForge {
     fn get_type(&self) -> ForgeType {
-        ForgeType::Npm
+        ForgeType::Go
     }
 
     fn fa(&self) -> &ForgeArg {
--- a/src/forge/go.rs
+++ b/src/forge/go.rs
@@ -1,0 +1,68 @@
+use std::fmt::Debug;
+
+use crate::cache::CacheManager;
+use crate::cli::args::ForgeArg;
+use crate::cmd::CmdLineRunner;
+use crate::config::{Config, Settings};
+
+use crate::forge::{Forge, ForgeType};
+use crate::install_context::InstallContext;
+
+#[derive(Debug)]
+pub struct GoForge {
+    fa: ForgeArg,
+    remote_version_cache: CacheManager<Vec<String>>,
+}
+
+impl Forge for GoForge {
+    fn get_type(&self) -> ForgeType {
+        ForgeType::Npm
+    }
+
+    fn fa(&self) -> &ForgeArg {
+        &self.fa
+    }
+
+    fn list_remote_versions(&self) -> eyre::Result<Vec<String>> {
+        self.remote_version_cache
+            .get_or_try_init(|| {
+                let raw = cmd!("go", "list", "-m", "-versions", "-json", self.name()).read()?;
+                let mod_info: GoModInfo = serde_json::from_str(&raw)?;
+                Ok(mod_info.versions)
+            })
+            .cloned()
+    }
+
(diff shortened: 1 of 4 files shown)


src/forge/mod.rs
```diff
@@ -39,6 +40,7 @@ pub enum ForgeType {
     Asdf,
     Cargo,
     Npm,
+    Go,
```
jdx, line 42: please alphabetize

endigma:
Failing dogfooding, maybe just nix? Nix works, even if its a little complex for a new user, its documented and reliable.

src/forge/go.rs
```diff
@@ -0,0 +1,69 @@
...
+
+#[derive(Debug)]
+pub struct GoForge {
+    fa: ForgeArg,
+    remote_version_cache: CacheManager<Vec<String>>,
+}
+
+impl Forge for GoForge {
+    fn get_type(&self) -> ForgeType {
+        ForgeType::Npm
```
jdx:
```suggestion
        ForgeType::Go
```

endigma:
I've updated it to function as per your suggestion, but I'm still not sure this is totally correct. Go already does build caching in `~/.cache/go-build`, the stuff in $GOPATH which is now in `~/.cache/mise` is literally the downloaded source for the version, which normally goes in your system GOPATH to avoid dupes. 

Intuition and experience tells me this should function fine if I removed the GOPATH bit entirely, as that's how Go install normally works, but I'll defer to whatever you want for the project here.

Could also update GOCACHE if you wanted to make it totally portable, I suppose.

jdx:
> Go already does build caching in ~/.cache/go-build

Yeah I think that's fine to keep there, I don't think everything needs to be centralized.

> the stuff in $GOPATH which is now in ~/.cache/mise is literally the downloaded source for the version, which normally goes in your system GOPATH to avoid dupes

Yes this is what I think is best. I wouldn't want my project sources to be commingled with stuff from CLIs I'm installing. The cache I'll probably never look at, but the source adds a lot of noise that I will see.

jdx:
> I've updated it to function as per your suggestion, but I'm still not sure this is totally correct. Go already does build caching in ~/.cache/go-build, the stuff in $GOPATH which is now in ~/.cache/mise is literally the downloaded source for the version, which normally goes in your system GOPATH to avoid dupes.
> Intuition and experience tells me this should function fine if I removed the GOPATH bit entirely, as that's how Go install normally works, but I'll defer to whatever you want for the project here.

wanted to let you know that I think I was wrong and you were correct here. I did a lot of research into this in https://github.com/jdx/mise/discussions/1638 and came to the same conclusion. I'm going to remove GOPATH from the forge and just use whatever the system is using.

My concern is just that $HOME/go/src would be full of a bunch of junk, but I think that's actually kind of expected from go developers.

Issue #1447, opened 2024-01-13
`env._.source`


jdx:
Support sourcing bash files in `[env]`:

```toml
[env]
_.source = "./script.sh"
```

The variables exported within the script should be imported. This will require running scripts in bash. I'll make a separate ticket for other types of scripts/binaries.

hahuang65:
Is the aim of this to replace direnv, essentially?

I've just installed `mise` and was under the impression that I could get rid of `asdf` and `direnv`.

However, it doesn't look like `mise` is currently compatible with `.envrc` files. Rather, `mise` provides it's own mechanism for env variables.

Our `.envrc` files have scripts to run inside them, which this seems to suggest supporting. I'd like to use `mise` as a replacement for `asdf` and `direnv` without forcing my co-workers to alter their development environments yet.

jdx:
there needs to be a lot more infra in place before mise could natively support envrc files. I'm also not sure that's the direction I would want to go. I generally like `.tool-versions` files but after being a direnv user for about a decade I really don't like the netrc format very much. I prefer something more declarative.

That said, if you really want envrc support a plugin would be capable of providing that with a `bin/exec-env` script.

hahuang65:
Curious. I've seen `exec-env` mentioned in the codebase a few times, but I haven't seen it in the documentation. Is there a link to it?

I'm really just after a way to re-use our `.envrc` files without having to port them over to `.mise.toml` since that would involve syncing changes between the two files, unless I convince my coworkers to use `mise` as well.

jdx:
> Curious. I've seen exec-env mentioned in the codebase a few times, but I haven't seen it in the documentation. Is there a link to it?

read the asdf plugin docs

> I get an error:

this ticket is open.

Issue #1303, opened 2023-12-31
new venv syntax


jdx:
I think I want to move venv configuration for python/poetry into the `[env]` section rather than putting it on individual tools:

```toml
[env]
_.python.venv = "~/.venv/myvenv"
_.poetry.pyproject = "pyproject.toml"
```

this would make it still driven by the plugin (allowing other plugins to do similar things), but also make it possible to use with the system python resolving #1248.

More importantly though, this just feels like the more correct place to put it since it's configuring the "environment"—not the python installation. That said it will depend on that so a bit of care around making sure if there is a python version set it needs to be installed _before_ setting the venv.

rsyring:
I don't follow mise development close enough to know the genesis of the `_` environment variables.  That being said, I find it odd the settings would be there and not a part of the tool.

camsteffen:
I think it's good that `[env]` has the single purpose of setting environment variables. It's analogous to a `.env` file.

Issue #1261, opened 2023-12-24
Cache env vars with slow exec function calls in templating


pdecat:
Setting environment variables from commands that are slow affects all prompt executions, e.g. setting:
```sh
mise env-vars TEST_MISE="{{exec(command='sleep 1')}}"
```
which translates into:
```toml
[env]
TEST_MISE = "{{exec(command='sleep 1')}}"
```
will add a 1 second latency at each prompt.

It would be awesome to allow caching such commands.

Note: that's just an example, my actual use case is invoking gopass to set an environment variable with GPG operations depending on my yubikey.

jdx:
I think me putting caching around templates is too difficult. I have no idea what sort of invalidation logic would work since I have no idea what people might be putting in their files.

Perhaps what we could do though is expose caching inside the templates so you could read/write to the cache instead of performing expensive computations.

hinricht:
Caching would be great, please have in mind that exec templates can be used to retrieve secrets so env vars shouldn't be cached/stored on disk unencrypted.

i.e. 
```
[env]
TOKEN = "{{exec(command='gopass show --password token')}}"
```

hinricht:
Oh just realized this use-case was the original motivation for this FR :)

pdecat:
I believe the need for caching templates used in environment variables could be eliminated by leveraging the `env._.source` feature which should only re-execute the configured script if something has changed.

However, there seems to be an issue with the short circuiting logic as mentioned in https://github.com/jdx/mise/issues/1617#issuecomment-1939772306

Doing a bit of investigation, I've added the following traces in the code:
```rust
diff --git a/src/config/env_directive.rs b/src/config/env_directive.rs
index dc995816..9c398a5c 100644
--- a/src/config/env_directive.rs
+++ b/src/config/env_directive.rs
@@ -155,11 +155,11 @@ impl EnvResults {
         };
         let mut paths: Vec<(PathEntry, PathBuf)> = Vec::new();
         for (directive, source) in input.clone() {
-            // trace!(
-            //     "resolve: directive: {:?}, source: {:?}",
-            //     &directive,
-            //     &source
-            // );
+            trace!(
+                "resolve: directive: {:?}, source: {:?}",
+                &directive,
+                &source
+            );
             let config_root = source
                 .parent()
                 .map(Path::to_path_buf)
diff --git a/src/hook_env.rs b/src/hook_env.rs
index 24ccba1e..c5830ad1 100644
--- a/src/hook_env.rs
[...]

PR #1519, opened 2024-01-24
env: resolve env vars in order

### LINKED ISSUE #1262: Allow env variables to reference previously defined env vars in same configuration file
I often find myself having to reference environment variables from others, e.g.:
```sh
rtx env-vars TEST_RTX1="1"
rtx env-vars TEST_RTX2="{{env.TEST_RTX1}}/bin"
rtx error parsing config file: ~/.rtx.toml
rtx failed to parse template: env.TEST_RTX2='{{env.TEST_RTX1}}/bin'
rtx Failed to render '__tera_one_off'
rtx Variable `env.TEST_RTX1` not found in context while rendering '__tera_one_off'
rtx Run with --verbose or RTX_VERBOSE=1 for more information
```

```toml
[env]
TEST_RTX1 = "1"
TEST_RTX2 = "{{env.TEST_RTX1}}"
```

Would it possible to add support for this?

jdx:
This is a major refactor of how env vars are handled. Previously, they were resolved as part of parsing but they really need to be done one at a time so that the results of each var can be used for the next one. This should also make it possible to greatly simplify how env vars are handled in tools by using the same logic.

Fixes #1262 
Prereq for #1447
Prereq for #1303
Prereq for #1261


jdx:
@Ajpantuso wanted to draw your attention to this PR just so you're aware of it since I know you worked on some of the env stuff a bit. I'd hold off on any env-related things for a couple of days until I can get this out since all of that logic is going to change.

PR #1520, opened 2024-01-25
feat(tasks): infer bash task topics from folder structure

### LINKED ISSUE #1269: task topics
we should make it possible to namespace topics with bash tasks.

e.g.: if I create a task at `.rtx/tasks/foo/bar/baz`, I should be able to run it with something like `rtx run foo:bar:baz`. Currently this directory only supports a single level.

Ajpantuso:
### Summary

Closes #1269

These changes make it possible to define bash tasks in a nested folder structure under the predefined task directories. Tasks defined in directories below the parent `tasks` directory will have implied scope in their names by joining their relative path with `:`s.

For example a task defined at `{{config_root}}/.mise/tasks/a/b/task` will have task name `a:b:task`.

### Additional Notes

- While `walkdir` was already a transitive dependency it is now a direct dependency to provide recursive walks through the filesystem.
- If a bash script task contains the `:` character it will be replaced by `_` to prevent name collisions.
- Simple enough to test locally so I did, but some integration/e2e tests wouldn't hurt here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1367,6 +1367,7 @@ dependencies = [
  "toml_edit",
  "url",
  "versions",
+ "walkdir",
  "which",
  "zip",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,6 +100,7 @@ toml = { version = "0.8.8", features = ["parse"] }
 toml_edit = { version = "0.21.0", features = ["parse"] }
 url = "2.5.0"
 versions = { version = "6.1.0" , features=["serde"]}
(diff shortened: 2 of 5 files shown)


jdx:
incredible! Your rust is excellent btw, have you been writing it long?

Ajpantuso:
> incredible! Your rust is excellent btw, have you been writing it long?

Thanks, only about a year now, but I spent a good amount of time with Haskell in college so iterators/function composition/Algebraic Data Types are a little more natural for me now.

PR #1524, opened 2024-01-25
feat(doctor): identify missing/extra shims

### LINKED ISSUE #1511: doctor: warn if shims are missing
mise should warn if `mise reshim` needs to be run

Ajpantuso:
### Summary

Closes #1511 

These changes enable `mise doctor` to identify missing or extra shims and will request users to run `mise reshim` if needed.

### Additional Notes
- Quite a bit of refactoring here really just to make the changes more obvious.
- Uses `checks` rather than a warning, but can be trivially updated to warnings if desired

### DIFF
--- a/src/cli/doctor.rs
+++ b/src/cli/doctor.rs
@@ -13,27 +13,29 @@ use crate::file::display_path;
 use crate::git::Git;
 use crate::plugins::PluginType;
 use crate::shell::ShellType;
+use crate::toolset::Toolset;
 use crate::toolset::ToolsetBuilder;
+use crate::ui::style;
 use crate::{cli, cmd, dirs, forge};
 use crate::{duration, env};
+use crate::{file, shims};
 
 /// Check mise installation for possible problems.
 #[derive(Debug, clap::Args)]
 #[clap(verbatim_doc_comment, after_long_help = AFTER_LONG_HELP)]
-pub struct Doctor {}
+pub struct Doctor {
+    #[clap(skip)]
+    checks: Vec<String>,
+}
 
 impl Doctor {
-    pub fn run(self) -> Result<()> {
-        let mut checks = Vec::new();
-        if let Err(err) = Config::try_get() {
-            checks.push(format!("failed to load config: {}", err));
-        }
-
+    pub fn run(mut self) -> Result<()> {
         miseprintln!("{}", mise_version());
         miseprintln!("{}", build_info());
         miseprintln!("{}", shell());
         miseprintln!("{}", mise_data_dir());
         miseprintln!("{}", mise_env_vars());
+
         match Settings::try_get() {
             Ok(settings) => {
                 miseprintln!(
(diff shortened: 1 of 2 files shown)


src/cli/doctor.rs
```diff
@@ -44,58 +45,94 @@ impl Doctor {
...
-                    Ok(ts) => {
-                        miseprintln!("{}\n{}\n", style("toolset:").bold(), indent(ts.to_string()))
-                    }
-                    Err(err) => warn!("failed to load toolset: {}", err),
-                }
+            Ok(config) => self.analyze_config(config)?,
+            Err(err) => {
+                warn!("failed to load config: {}", err);
+
+                self.checks.push(format!("failed to load config: {}", err));
```
jdx, line 53: nice addition, but I think we can get rid of the warning in this case (just keep the check)

Ajpantuso: Done, was trying to keep parity with the original [code](https://github.com/jdx/mise/blob/main/src/cli/doctor.rs#L75), but yeah that was definitely redundant.

src/cli/doctor.rs
```diff
@@ -13,27 +13,28 @@ use crate::file::display_path;
...
 use crate::{duration, env};
+use crate::{file, shims};
 
 /// Check mise installation for possible problems.
 #[derive(Debug, clap::Args)]
 #[clap(verbatim_doc_comment, after_long_help = AFTER_LONG_HELP)]
-pub struct Doctor {}
+pub struct Doctor {
+    #[clap(skip)]
+    checks: Vec<String>,
```
jdx, line 28: good call

src/cli/doctor.rs
```diff
@@ -44,58 +45,92 @@ impl Doctor {
...
+                self.checks
+                    .push(format!("plugin {} is not installed", &plugin.id()));
+                continue;
+            }
+        }
+
+        if !config.is_activated() && !shims_on_path() {
+            let cmd = style("mise help activate").yellow().for_stderr();
+            let url = style("https://mise.jdx.dev").underlined().for_stderr();
+            let shims = style(dirs::SHIMS.display()).cyan().for_stderr();
```
jdx: this is nitpicky, but I've been trying to move towards [these helpers](https://github.com/jdx/mise/blob/main/src/ui/style.rs) to simplify the common styling stuff

src/cli/doctor.rs
```diff
@@ -44,58 +45,92 @@ impl Doctor {
...
+            ));
+        }
+
+        match ToolsetBuilder::new().build(config) {
+            Ok(ts) => {
+                self.analyze_shims(&ts);
+
+                miseprintln!("{}\n{}\n", style("toolset:").bold(), indent(ts.to_string()));
+            }
+            Err(err) => warn!("failed to load toolset: {}", err),
```
jdx: I think while we're at it this should be a check also

src/cli/doctor.rs
```diff
@@ -44,58 +45,92 @@ impl Doctor {
...
+
+    fn analyze_shims(&mut self, toolset: &Toolset) {
+        let mise_bin = file::which("mise").unwrap_or(env::MISE_BIN.clone());
+
+        if let Ok((missing, extra)) = shims::get_shim_diffs(mise_bin, toolset) {
+            let cmd = style("mise reshim").yellow().for_stderr();
+
+            if !missing.is_empty() {
+                self.checks
+                    .push(formatdoc!("shims are missing, run {cmd} to create them"));
```
jdx: I would just use `format!()` here. I haven't been too happy with indoc personally just because it doesn't work well in my IDE. In this case though since it's a single line `formatdoc` isn't doing anything special

src/shims.rs
```diff
@@ -86,29 +86,7 @@ pub fn reshim(ts: &Toolset) -> Result<()> {
...
-            list_tool_bins(t.clone(), &tv).unwrap_or_else(|e| {
-                warn!("Error listing bin paths for {}: {:#}", tv, e);
-                Vec::new()
-            })
-        })
-        .collect();
-
-    let shims_to_add = shims.difference(&existing_shims);
-    let shims_to_remove = existing_shims.difference(&shims);
+    let (shims_to_add, shims_to_remove) = get_shim_diffs(&mise_bin, ts)?;
```
jdx, line 89: I was sure this would need to be refactored, but it looks like it wasn't too bad. I actually refactored this at one point to use the set difference and I think that paid off. Before it just did the calculation and symlink edits at the same time.

src/shims.rs
```diff
@@ -171,6 +166,32 @@ fn list_executables_in_dir(dir: &Path) -> Result<HashSet<String>> {
...
+            })
+        })
+        .collect())
+}
+
+// lists all the paths to bins in a tv that shims will be needed for
+fn list_tool_bins(t: Arc<dyn Forge>, tv: &ToolVersion) -> Result<Vec<String>> {
+    Ok(t.list_bin_paths(tv)?
+        .into_iter()
+        .par_bridge()
```
jdx, line 186: I do wonder if using parallelization is actually worth it for stuff like this. I think maybe if someone had like 100 tools (and I know some people do), it might help.

That said, when I have tried to benchmark stuff like this in the past with rust it's basically impossible to tell the difference at least with hyperfine. I think to really find out we'd have to have isolated benchmarks.

jdx:
also, keep in the back of your mind other potential checks we might be able to add here

PR #1528, opened 2024-01-25
feat(run): match tasks to run with glob patterns

### LINKED ISSUE #1271: `rtx run lint:*`
wildcards are supported as task dependencies but they should be allowed in `rtx run` CLI commands as well.

Ajpantuso:
### Summary

Closes #1271 

This change make it possible to select multiple tasks to run using glob style matching.
Task topics, separated by `:`, are treated like directories in shell matching such that patterns like `a:**:d` will match tasks named `a:d`, `a:b:d` or `a:b:c:d`.

### Additional Notes
- This brings `globset` in as a direct dependency, but it was transitive already.
- Used an Extension Trait since `tasks_with_aliases` is used in multiple places making a more local change preferable though it may make sense to move this trait or change the return type of `tasks_with_aliases` in the future.


src/cli/run.rs
```diff
@@ -521,6 +531,26 @@ fn get_color() -> Color {
     static COLOR_IDX: AtomicUsize = AtomicUsize::new(0);
     COLORS[COLOR_IDX.fetch_add(1, Ordering::Relaxed) % COLORS.len()]
 }
+trait GetMatchingExt<T> {
+    fn get_matching(&self, pat: &str) -> Result<Vec<&T>>;
+}
+
+impl<T> GetMatchingExt<T> for std::collections::HashMap<String, T> {
```
jdx, line 539: can we go with BTreeMap? I think the output will look a little cleaner being not randomized by HashMap

IndexMap might also be a good idea but I'm not sure what order the tasks are in

Ajpantuso: I guess that decision would really have to be made [here](https://github.com/jdx/mise/blob/59c721690fc7fe5ae5f54e3a530f1523114bdba0/src/config/mod.rs#L186), but I can definitely add sorting to guarantee ordering.

src/cli/run.rs
```diff
@@ -134,11 +135,20 @@ impl Run {
...
+                        Ok(vec![self.prompt_for_task(config)?])
+                    }
+                    (_, tasks) if tasks.is_empty() => bail!("no task {} found", style::ered(t)),
+                    (_, tasks) => Ok(tasks
+                        .iter()
+                        .cloned()
+                        .map(|t| t.clone().with_args(args.to_vec()))
+                        .collect()),
+                },
+            )
```
jdx: I think the pattern matching is getting a bit crazy here, what do you think of this?

```suggestion
            .map(|(t, args)| {
                let tasks = config.tasks_with_aliases().get_matching(&t)?;
                if tasks.is_empty() {
                    ensure!(t == "default", "no task {} found", style::ered(t));
                    Ok(vec![self.prompt_for_task(config)?])
                } else {
                    Ok(tasks
                        .iter()
                        .cloned()
                        .map(|t| t.clone().with_args(args.to_vec()))
                        .collect())
                }
            })
```

probably not shorter but I think it's easier to grok

src/cli/run.rs
```diff
@@ -134,11 +135,20 @@ impl Run {
...
+            .flatten_ok()
```
jdx, line 152: wow I did not know about this, definitely need to be using this in a lot of places

jdx:
clever solution using the glob here. I was just going to have a naive wildcard match

src/cli/run.rs
Ajpantuso: Honestly just found it myself. Pretty nifty.

src/cli/run.rs
jdx: up to you, could be in a different PR if you want. I'm not sure how many places that would need to change off the top of my head

jdx:
this could happen later, but this code should probably be updated to use glob logic as well:

https://github.com/jdx/mise/blob/59c721690fc7fe5ae5f54e3a530f1523114bdba0/src/task.rs#L137

src/cli/run.rs
Ajpantuso: Lol, played around with this a bit only to realize parallelization ends up randomizing the output ordering anyway. So I think we can defer the decision for now since ordering is only relevant for single threaded runs (which I imagine is the less likely use case).

Ajpantuso:
> this could happen later, but this code should probably be updated to use glob logic as well:
> 
> https://github.com/jdx/mise/blob/59c721690fc7fe5ae5f54e3a530f1523114bdba0/src/task.rs#L137

I can pick that up in a follow up.

PR #1536, opened 2024-01-26
feat(tasks): add --timings option for run command

### LINKED ISSUE #1265: tasks: show elapsed time of tasks
optionally of course

Ajpantuso:
### Summary

Closed #1265 

Adds a `--show-timings` option to the `mise run` command to display elapsed time after each task.

### Open Questions
- Should timings be shown even after failure?
- Should total time also be displayed after all tasks have run?
- What style/format would look best for the new output?

### DIFF
--- a/src/cli/run.rs
+++ b/src/cli/run.rs
@@ -107,6 +107,10 @@ pub struct Run {
     #[clap(long, short, verbatim_doc_comment)]
     pub raw: bool,
 
+    /// Shows elapsed time after each task
+    #[clap(long, alias = "timing", verbatim_doc_comment)]
+    pub timings: bool,
+
     #[clap(skip)]
     pub is_linear: bool,
 }
@@ -166,12 +170,16 @@ impl Run {
             env.insert("FORCE_COLOR".into(), "1".into());
         }
 
-        let tasks = Mutex::new(Deps::new(config, tasks)?);
-        self.is_linear = tasks.lock().unwrap().is_linear();
-
-        for task in tasks.lock().unwrap().all() {
+        let tasks = Deps::new(config, tasks)?;
+        for task in tasks.all() {
             self.validate_task(task)?;
         }
+
+        let num_tasks = tasks.all().count();
+        self.is_linear = tasks.is_linear();
+
+        let tasks = Mutex::new(tasks);
+        let timer = std::time::Instant::now();
 
         let pool = rayon::ThreadPoolBuilder::new()
             .num_threads(self.jobs() + 1)
@@ -195,6 +203,12 @@ impl Run {
                 run(&task);
             }
         });
+
+        if self.timings && num_tasks > 1 {
+            let msg = format!("finished in {}", format_duration(timer.elapsed()));
+            info!("{}", style::edim(msg));
+        };
+
         Ok(())
     }
 
@@ -211,6 +225,8 @@ impl Run {
             .map(|(k, v)| (k.clone(), v.clone()))
             .collect();
 
+        let timer = std::time::Instant::now();
+
         if let Some(file) = &task.file {
             self.exec_file(file, task, &env, &prefix)?;
         } else {
@@ -221,6 +237,14 @@ impl Run {
                 };
                 self.exec_script(cmd, &args, task, &env, &prefix)?;
             }
+        }
+
+        if self.timings {
+            miseprintln!(
+                "{} finished in {}",
+                prefix,
+                format_duration(timer.elapsed())
+            );
         }
 
         self.save_checksum(task)?;
@@ -531,6 +555,15 @@ fn get_color() -> Color {
     static COLOR_IDX: AtomicUsize = AtomicUsize::new(0);
     COLORS[COLOR_IDX.fetch_add(1, Ordering::Relaxed) % COLORS.len()]
 }
+
+fn format_duration(dur: std::time::Duration) -> String {
+    if dur < std::time::Duration::from_secs(1) {
+        format!("{:.0?}", dur)
+    } else {
+        format!("{:.2?}", dur)
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use crate::file;


jdx:
let's just call it `--timings`—with a hidden alias `--timing`

src/cmd.rs
```diff
@@ -394,10 +396,91 @@ impl Display for CmdLineRunner<'_> {
 enum ChildProcessOutput {
     Stdout(String),
     Stderr(String),
-    ExitStatus(ExitStatus),
+    ExitStatus(TimedProcessExitStatus),
     Signal(i32),
 }
 
+struct TimedProcess {
```
jdx: couldn't this just be a part of every ChildProcess? we're just recording a timestamp so I feel like we could do it every time even though we'll only read the result in rare situations

Ajpantuso: Yeah, this approach is really just faux-inheritance so that the `std::process::Child` API can be accessed as is while still providing the timing functionality which I think is a little more dev friendly then fully encapsulating the `Child` struct.

Of course `TimedProcess` is very specific and could be made a generic `Process` struct which would be more extensible long term. In which case a dedicated `process` module would make sense. The primary entrypoint for that API would probably then just be a public `spawn(cmd: &mut Command)` function.

With a re-mod `ExecStatus` would still stay in this module to be the public return where we could control what information gets propagated back to callers. I think it's best to minimize side-effects in `execute` and send any stats back to the caller for formatting/processing since that will allow for aggregated results and tabular or structured output.

Long term if users are encourage to run `mise` tasks in CI having a JSON report or something machine-parsable would make sense.

Ajpantuso: Added a refactor commit to demonstrate what I'm talking about.

roele:
Since tasks might run in parallel the task prefix should probably be added to be able to tell which timing belongs to what task. 
The current output does not make that clear.
Screenshot 2024-01-29 at 10 45 12

Ajpantuso:
@roele yeah I figured that would happen since we are streaming stdout/stderr; was just getting lucky on my test runs.

How would you feel about just adding a summary report of timings after all parallel execution?
I already added a metrics recorder for this purpose so there could eventually be structured output options. I guess prefixed output would be okay as well just think it may be hard to read with many tasks or lots of output.

roele:
Yes prefixed will help a little but with many tasks could still be hard to read. An overall summary at the end might be better. Maybe we can find some good examples from other tools.

Screenshot 2024-01-29 at 11 10 57

Ajpantuso:
@roele cool, added prefixes for now just so folks can review what it would look like. In parallel I'll play around with summary formats.

src/cli/run.rs
```diff
@@ -195,6 +209,24 @@ impl Run {
...
+
+        let elapsed = match self.recorder.lock() {
+            Ok(inner) => inner,
+            Err(poison) => poison.into_inner(),
+        }
+        .elapsed()
+        .unwrap();
+
+        if self.timings {
+            miseprintln!("total time: {:?}", style::nbold(elapsed))
```
jdx: can we reduce the precision on this?

Ajpantuso: Done, set to `2` decimal places.

Ajpantuso:
@roele I have a quick summary table implementation [here](https://github.com/Ajpantuso/mise/commit/7d80fccf1205b94a5a5ad07fff6a7fccde04701d) if you wanted to review/refine it.

After looking at it I feel like hacking a table on to the default output doesn't look right. I think a `--summarize` or `--summary` option in the future would feel more natural. The default summary could just be a user-consumable table with no stdout/stderr from tasks while additional output options like `---output={json,yaml,...}` could dump stdout/stderr from tasks into a structured format for processing by CI runners.

src/cli/run.rs
```diff
@@ -299,18 +331,37 @@ impl Run {
...
-                    error!("{prefix} exited with code {code}");
-                    exit(code);
-                } else if let Some(signal) = status.signal() {
-                    error!("{prefix} killed by signal {signal}");
-                    exit(1);
+        match cmd.execute() {
+            Ok(stat) => {
+                if self.timings {
+                    miseprintln!(
+                        "{prefix} elapsed time: {:.2?}",
```
jdx: how does this work? does it always show ms or does it switch to seconds if it's large?

At least for ms I think we shouldn't show any decimal places. I think anything more precise isn't going to be useful.

src/cli/run.rs
```diff
@@ -173,6 +181,12 @@ impl Run {
             self.validate_task(task)?;
         }
 
+        match self.recorder.lock() {
```
jdx: a couple things: first, I generally just do `.unwrap()` on mutexes since we lock them so little the idea of a deadlock is pretty much impossible unless there is a pretty serious issue—in which case panicking is probably fine.

also, I don't like exposing the mutex into the CLI command. That should be pushed into the Recorder<T> I think. You may need an inner struct or possibly a [tuple struct](https://doc.rust-lang.org/reference/items/structs.html) to make that more concise.

src/cli/run.rs
```diff
@@ -107,8 +108,15 @@ pub struct Run {
     #[clap(long, short, verbatim_doc_comment)]
     pub raw: bool,
 
+    /// Shows elapsed time after each task
+    #[clap(long, alias = "timing", verbatim_doc_comment)]
+    pub timings: bool,
```
jdx, line 112: after looking at this I am considering if maybe we just want to always display the timings. Not sure though.

jdx:
I might be not understanding this so much but I think this could be simplified by doing something like this:

```rs
fn run_all() {
  let start = time::Instant::now()
  for task in tasks {
    task.run();
  }
  println!("elapsed all: ...")
}

fn run_task() {
  let start = time::Instant::now()
  run()
  println!("elapsed: ...")
}
```

src/cli/run.rs
Ajpantuso: It uses the largest unit possible where the adjusted duration is `>=1` so yes it will bump up to seconds after 999.99... milliseconds and minutes after 59.99... seconds and so on.

Dropping decimals for exclusively milliseconds would be a little annoying just cause you'd have to conditionally change the format string based on the `Duration` value, but it's doable.

src/cli/run.rs
roele: i feel in either case it should interfere less with the actual command output by maybe showing the entire line (after prefix) dimmed?

src/cli/run.rs
Ajpantuso: I'm fine using `.unwrap()` since like you said in this case we have strong assumptions about what's happening during the lock period, but I prefer to capture this stuff in the user facing bits (i.e. cli commands) to make it easier to handle panics and provide meaningful output back to the console.

Regarding hiding the `Mutex` in the `Recorder` I don't like the idea since it specializes the `Recorder` as a thread-safe, i.e. less performant, implementation which is something the caller should really decide. However the `Recorder` certainly doesn't need to be a member of the command struct. And with additional use of channels we really wouldn't need a thread-safe implementation at all, but I'll hit that in the next comment.

Ajpantuso:
> then we wouldn't need to modify anything outside of the cli command or use any mutexes or anything

Correct, the only reason to have `Mutexes` or `Channels` would be to capture summaries. If there is no appetite for having some kind of summary report I can drop this to the simplest implementation. Really just playing around with possibilities to scope this out.

roele:
> After looking at it I feel like hacking a table on to the default output doesn't look right.

@Ajpantuso you're right it looks a bit odd. So i guess we should start small by refining the current inline approach.

Ajpantuso:
@jdx @roele I've reset this back to barebones incorporating some suggestions.

Making it so timings are always displayed would have some implications for testing so leaving the flag in for now.

src/cli/run.rs
```diff
@@ -531,6 +552,15 @@ fn get_color() -> Color {
     static COLOR_IDX: AtomicUsize = AtomicUsize::new(0);
     COLORS[COLOR_IDX.fetch_add(1, Ordering::Relaxed) % COLORS.len()]
 }
+
+fn format_duration(dur: std::time::Duration) -> String {
+    if dur < std::time::Duration::new(1, 0) {
```
jdx:
```suggestion
    if dur < std::time::Duration::from_secs(1) {
```

src/cli/run.rs
```diff
@@ -195,6 +201,11 @@ impl Run {
                 run(&task);
             }
         });
+
+        if self.timings {
+            miseprintln!("finished in {}", format_duration(timer.elapsed()));
```
jdx: let's wrap this in `style::edim()` also we should only display it if there are >1 tasks. Lastly, let's put it on stderr. I could make a case for stderr or stdout but we're already using stderr for the task start message so I think stderr makes sense for this to match.

~~`info_unprefix_trunc` is what you'll want to use for this since that will respect the log level and hide the output under `--quiet`~~ EDIT: we probably only want to do that if we make it non-optional

src/cli/run.rs
jdx: I'm still undecided on this. I think I'm leaning towards always displaying though curious if you all have any thoughts.

jdx: hmm maybe not. Let's keep it optional for now, I think we should err on the side of less noise.

src/cli/run.rs
Ajpantuso: Done, used `info!` so the dim `mise` prefix is added, but I think that makes sense.

src/cli/run.rs
Ajpantuso: Yeah, the other choice is optional, but default to enabled. Seems reasonable however to just take user feedback and make a final decision before tasks are a GA feature.

PR #1562, opened 2024-01-31
Additional arch install

tlockney:
Took a stab at supporting the additional architectures in the generated install script. I admit I haven't thoroughly tested this by any means and the shell scripting probably leads a lot to be desired, but perhaps it can be a jumping off point, at least.


jdx:
there is a [script](https://github.com/jdx/mise/blob/main/scripts/test-standalone.sh) you can use to test this

packaging/standalone/install.envsubst
```diff
@@ -114,9 +165,15 @@ install_mise() {
 	version="$MISE_VERSION"
 	os="$(get_os)"
 	arch="$(get_arch)"
+	musl="$(get_musl)"
 	install_path="${MISE_INSTALL_PATH:-$HOME/.local/bin/mise}"
 	install_dir="$(dirname "$install_path")"
-	tarball_url="https://github.com/jdx/mise/releases/download/${version}/mise-${version}-${os}-${arch}.tar.gz"
+	if [ -z "$musl" ]; then
+		tarball_url="https://github.com/jdx/mise/releases/download/${version}/mise-${version}-${os}-${arch}-${musl}.tar.gz"
```
jdx: I think if you just made musl part of $arch it'll clean up a lot of this logic

packaging/standalone/install.envsubst
```diff
@@ -28,6 +28,19 @@ error() {
 }
 #endregion
 
+get_musl() {
+	if type ldd >/dev/null 2>/dev/null; then
+		libc=$(ldd /bin/ls | grep 'musl' | head -1 | cut -d ' ' -f1)
+		if [ -z "$libc" ]; then
+			echo "musl"
+		else
+			echo ""
```
jdx: you can just get rid of these else conditions and do nothing

packaging/standalone/install.envsubst
```diff
@@ -46,6 +59,8 @@ get_arch() {
 		echo "x64"
 	elif [ "$arch" = aarch64 ] || [ "$arch" = arm64 ]; then
 		echo "arm64"
+	elif [ "$arch" = armv7l ]; then
```
jdx, line 58: go ahead and add "armv6l" idk if it will work, but it's not like not trying to handle it would do anything useful anyways

of course if you want to find a way to test it that's great but I don't really mind

packaging/standalone/install.envsubst
tlockney: Got rid of this entirely. Moving the check to `get_arch()` cleaned things up a bit.

packaging/standalone/install.envsubst
tlockney: Done. Also made the code to check for `musl` slightly cleaner.

packaging/standalone/install.envsubst
tlockney: Ha, I actually thought I had added that earlier, but it's there now.

packaging/standalone/install.envsubst
```diff
@@ -66,21 +77,48 @@ get_checksum() {
...
+			echo "$checksum_macos_x86_64"
+		elif [ "$arch" = "arm64" ]; then
+			echo "$checksum_macos_arm64"
+		else
+			warn "no checksum for $os-$arch"
+		fi
 	else
 		warn "no checksum for $os-$arch"
 	fi
+
```
jdx:
```suggestion
```

PR #1571, opened 2024-01-31
feat(tasks): make script task dirs configurable

### LINKED ISSUE #1270: change task directory
right now the directories `.config/rtx/tasks` and `.rtx/tasks` are hardcoded and cannot be changed. This should be configurable.

Ajpantuso:
### Summary

Closes #1270 

Adds a top-level config Table, `task_config`, which currently has one property `includes`. `includes` is an array of paths which can be used to alter the locations from which file tasks are loaded from.

These paths can be absolute, relative, inside/outside the project, and accepts `~` as a substitution for `$HOME`.

### Additional Notes
- `includes` is represented as a `BTreeSet` to preserve ordering, allows users to influence load ordering, and to ensure uniqueness.

### DIFF
--- a/src/config/config_file/mise_toml.rs
+++ b/src/config/config_file/mise_toml.rs
@@ -14,7 +14,7 @@ use toml_edit::{table, value, Array, Document, Item, Value};
 use versions::Versioning;
 
 use crate::cli::args::ForgeArg;
-use crate::config::config_file::{trust_check, ConfigFile, ConfigFileType};
+use crate::config::config_file::{trust_check, ConfigFile, ConfigFileType, TaskConfig};
 use crate::config::env_directive::EnvDirective;
 use crate::config::AliasMap;
 use crate::file::{create_dir_all, display_path};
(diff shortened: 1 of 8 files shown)


src/config/config_file/mise_toml.rs
```diff
@@ -50,6 +50,8 @@ pub struct MiseToml {
     #[serde(default)]
     plugins: HashMap<String, String>,
     #[serde(skip)]
+    pub task_config: TaskConfig,
```
jdx, line 53: can you change this to parse via derive? I'm trying to get rid of the manual parsing

Ajpantuso: Yes, that's what I actually started with, but didn't quite understand how templates are being rendered to the deserialized values.

Ajpantuso: nvm, I see now. Was looking in the wrong place.
https://github.com/jdx/mise/blob/3070bf8f1ddb517d7d0f329b0d242d7433a1c487/src/config/env_directive.rs#L151-L165

Ajpantuso: Just dropping support for templating in `includes` for now. It just adds too much noise for the moment and will be simpler to incorporate after custom parsing is gone.

PR #1588, opened 2024-02-02
add service to task

sirenkovladd:
Motivation
https://github.com/google/wireit?tab=readme-ov-file#services

### DIFF
--- a/schema/mise.json
+++ b/schema/mise.json
@@ -227,6 +227,10 @@
                 "description": "glob pattern or path to files created by this task",
                 "type": "string"
               }
+            },
+            "service": {
+              "description": "If true or an object, treat this script as a long-running process.\nServices are automatically brought up and down as they are depended upon by other scripts.",
+              "type": "boolean"
             }
           },
           "additionalProperties": false
--- a/src/cli/run.rs
+++ b/src/cli/run.rs
@@ -6,6 +6,7 @@ use std::path::{Path, PathBuf};
 use std::process::{exit, Stdio};
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Mutex;
+use std::thread::sleep;
 use std::time::SystemTime;
 
 use clap::ValueHint;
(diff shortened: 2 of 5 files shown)


src/cli/run.rs
```diff
@@ -303,38 +304,49 @@ impl Run {
...
+            }
+            if !task.service || i == 0 {
+                break;
+            }
+            if i > 0 {
+                i -= 1;
             }
-            error!("{err}");
-            exit(1);
+            sleep(std::time::Duration::from_millis(50));
```
jdx: is the goal here to restart it if it exits? I don't think we should do that, just leave it stopped

sirenkovladd: Indeed, service should only restart if it has finished with an error

fixed

schema/mise.json
```diff
@@ -227,6 +227,10 @@
                 "description": "glob pattern or path to files created by this task",
                 "type": "string"
               }
+            },
+            "service": {
+              "description": "If true or an object, treat this script as a long-running process.\nServices are automatically brought up and down as they are depended upon by other scripts.",
```
jdx, line 232: why "an object"? doesn't this only accept a bool?

src/cli/run.rs
```diff
@@ -303,38 +304,50 @@ impl Run {
...
-        cmd.with_pass_signals();
-        match &self.output(task)? {
-            TaskOutput::Prefix => cmd = cmd.prefix(format!("{prefix} ")),
-            TaskOutput::Interleave => {
-                cmd = cmd
-                    .stdin(Stdio::inherit())
-                    .stdout(Stdio::inherit())
-                    .stderr(Stdio::inherit())
+        // TODO: add attribute to task
+        let mut i = -1;
```
jdx, line 308: is this being incremented anywhere? I'm not sure if I just keep missing it but I'm not seeing that

sirenkovladd: Not used yet
that's why I added a TODO to add the maximum number of restarts possible in the future

jdx: let's just skip the stubbed restarting logic for now

src/cli/watch.rs
```diff
@@ -92,6 +92,10 @@ impl Watch {
             })
             .collect::<Result<Vec<_>>>()?;
         let mut args = vec![];
+        // check if any tasks has a service = true
+        if let Some(_service) = tasks.iter().find_map(|t| Some(t.service)) {
+            args.push("-r".to_string());
```
jdx, line 97: I think I would prefer users just manually add this argument. Doing it for them means they can't disable it.

schema/mise.json
jdx: also it should say "treat this task" not script

jdx:

PR #1589, opened 2024-02-02
disable prefix for task

sirenkovladd:
Disable prefix for task

### DIFF
--- a/schema/mise.json
+++ b/schema/mise.json
@@ -227,6 +227,10 @@
                 "description": "glob pattern or path to files created by this task",
                 "type": "string"
               }
+            },
+            "prefix": {
+              "description": "prefix output with task name",
+              "type": "boolean"
             }
           },
           "additionalProperties": false
--- a/src/config/config_file/mise_toml.rs
+++ b/src/config/config_file/mise_toml.rs
@@ -183,6 +183,7 @@ impl MiseToml {
                         "raw" => task.raw = self.parse_bool(&key, v)?,
                         // "script" => task.script = Some(self.parse_string_tmpl(&key, v)?),
                         "sources" => task.sources = self.parse_string_array(&key, v)?,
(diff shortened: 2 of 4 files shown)


schema/mise.json
```diff
@@ -227,6 +227,10 @@
                 "description": "glob pattern or path to files created by this task",
                 "type": "string"
               }
+            },
+            "prefix": {
+              "description": "prefix output with task name",
+              "type": "boolean"
```
jdx, line 233: I think a string might be better

src/task.rs
```diff
@@ -91,6 +93,7 @@ impl Task {
             dir: p.parse_str("dir")?,
             env: p.parse_hashmap("env")?.unwrap_or_default(),
             file: Some(path.to_path_buf()),
+            prefix: p.parse_bool("prefix").unwrap_or(true),
```
jdx, line 96: I realize that hide has the same problem but this should be p.parse_bool("prefix")? otherwise parsing errors won't be displayed

schema/mise.json
sirenkovladd: To clarify, if an empty string is passed, then in this case not to display the prefix?

jdx: yeah, that way people can customize the output to whatever they want and aren't limited to our output

src/config/config_file/mise_toml.rs
```diff
@@ -183,6 +183,7 @@ impl MiseToml {
                         "raw" => task.raw = self.parse_bool(&key, v)?,
                         // "script" => task.script = Some(self.parse_string_tmpl(&key, v)?),
                         "sources" => task.sources = self.parse_string_array(&key, v)?,
+                        "prefix" => task.prefix = self.parse_bool(&key, v)?,
```
jdx, line 186: this will need to be refactored as the parsing has been moved to serde_derive

jdx:

PR #1600, opened 2024-02-03
fix(tasks): ensure task precedence is preserved during loading

### LINKED ISSUE #1593: Tasks are overwritten by inherited configurations
**Describe the bug**
Tasks with the same name, overwrite local tasks with the highest config in the directory hierarchy.

**To Reproduce**
With given directory structure:

- root
  - .mise.toml
  - project
    .mise.toml
    
Assuming both configs define a task called `test`, and running `mise task` inside the "project" directory, it will only show the task from the `root` .mise.toml. So there is no way for me to execute the `test` task in the "project" directory.

**Expected behavior**
I would expect the task from the config closest to the current directory to be available.

Ajpantuso:
### Summary

Fixes #1593

[This](https://github.com/jdx/mise/blob/main/src/config/mod.rs#L246) initial call to `rev()` ensures that lower precedence config files are processed first and consequently their tasks and aliases are inserted into the resulting Map first and overridden by high precedence tasks.

The additional call to `rev()` removed here inverted this behavior.


jdx:
given the test failure I think you removed the wrong one. That logic is so complicated I'm not sure though.

Ajpantuso:
Lolol I see what the problem is:

Created an e2e script with just the following:
```
MISE_EXPERIMENTAL=1 mise config
MISE_EXPERIMENTAL=1 mise tasks
```

This is the output:
```
./e2e/run_test ./e2e/test_debug
::group::E2E ./e2e/test_debug
Path                                             Plugins          
~/dev/github.com/jdx/mise/e2e/.e2e.mise.toml     tiny             
~/dev/github.com/jdx/mise/e2e/.e2e-tool-versions shellcheck, shfmt
~/dev/github.com/jdx/mise/e2e/.node-version      node             
~/dev/github.com/jdx/mise/.node-version          node             
Name        Description                 Source                                         
build                                   ~/dev/github.com/jdx/mise/e2e/.mise/tasks/build
configtask                              ~/dev/github.com/jdx/mise/e2e/.e2e.mise.toml   
filetask    This is a test build script ~/dev/github.com/jdx/mise/.mise/tasks/filetask 
lint                                    ~/dev/github.com/jdx/mise/e2e/.e2e.mise.toml   
test                                    ~/dev/github.com/jdx/mise/e2e/.e2e.mise.toml
```

And after removing `~/dev/github.com/jdx/mise/.node-version`:
```
./e2e/run_test ./e2e/test_bang
::group::E2E ./e2e/test_bang
Path                                             Plugins          
~/dev/github.com/jdx/mise/e2e/.e2e.mise.toml     tiny             
~/dev/github.com/jdx/mise/e2e/.e2e-tool-versions shellcheck, shfmt
~/dev/github.com/jdx/mise/e2e/.node-version      node             
Name        Description                 Source                                            
build                                   ~/dev/github.com/jdx/mise/e2e/.mise/tasks/build   
configtask                              ~/dev/github.com/jdx/mise/e2e/.e2e.mise.toml      
filetask    This is a test build script ~/dev/github.com/jdx/mise/e2e/.mise/tasks/filetask
lint                                    ~/dev/github.com/jdx/mise/e2e/.e2e.mise.toml      
test                                    ~/dev/github.com/jdx/mise/e2e/.e2e.mise.toml      
./e2e/test_bang: 0s
```

So I'll need to get e2e to ignore the legacy `.node-version` in the project root since that is generating the default task includes directories which rightfully should have higher precedence.

Ajpantuso:
Soft-blocking this on #1612 since difficult to identify if current e2e suite is correctly failing or not.

jdx:
> So I'll need to get e2e to ignore the legacy .node-version in the project root since that is generating the default task includes directories which rightfully should have higher precedence.

Related to this, I'm unsure if my strategy of using the fact there is a config file is even a good idea. I could see a user wanting to have tasks without a config file at all and the current solution doesn't allow for that.

jdx:
> So I'll need to get e2e to ignore the legacy .node-version in the project root since that is generating the default task includes directories which rightfully should have higher precedence.

looking at this a bit deeper, I don't understand. If both ~/src/mise and ~/src/mise/e2e define a task, then the one in ~/src/mise/e2e should override what is in ~/src/mise.

Ajpantuso:
Closing in favor of #1625

jdx:
I did some work in #1625 to refactor and improve this. A test is failing I need to sort out though.

PR #1612, opened 2024-02-03
tests(integration): introduce rust based integration suite

Ajpantuso:
### Summary

Introduces an integration suite in `rust` to complement the `e2e` suite. Some `e2e` tests can be converted to simpler integration tests which is partially demonstrated here, but others which rely on shell integration for completeness should not be.

### Additional Notes
- For the `e2e` tests being converted it will be important to scrutinize what global setup from the e2e suite must be included for valid testing
- Test helpers are being built as I go so are highly subject to change


Ajpantuso:
@jdx PTAL. Can keep this minimal in scope if you want to just help with #1600, but I'm happy to flesh this out in to a more complete refactor of e2e tests, of course leaving the true e2e tests alone. Obviously if you don't like the style of integration I can just hack around the current `test_run` file instead.

To get these to run from this commit you just need to uncomment [these lines](https://github.com/jdx/mise/pull/1612/files#diff-eb771f54a1fdc1c330a82981d7d27d1202a5cbe3a59581ff67af5e3361f5f90eR1-R2)

jdx:
I'm down to try it

Cargo.toml
```diff
@@ -112,9 +112,13 @@ exec = "0.3.1"
 built = { version = "0.7.1", features = ["chrono", "git2"] }
 
 [dev-dependencies]
+assert_cmd = "2.0.13"
 ctor = "0.2.6"
 insta = { version = "1.34.0", features = ["filters", "json"] }
+predicates = "3.1.0"
 pretty_assertions = "1.4.0"
+tempfile = "3.9.0"
```
jdx: we already have this

Ajpantuso:
@jdx alright I think this is good for a review. Took too long and bigger than I wanted, but I like the idea of having a DSL. Should hopefully make it easier for contributors to add integration tests. Noted in `tests/cli/mod.rs` the e2e tests which have not been converted. Also did not update the test targets for the time being, but should be trivial to add the filters to exclude these from `test:units` when desired.

PR #1661, opened 2024-02-12
fix(go): reflect on proper path for `GOROOT`

wheinze:
The last [patches](https://github.com/jdx/mise/commit/786220c6178625980bdcc61403c32db19d51360f) to `GOROOT` did miss to take over the `/go` suffix in the path, which ends up in various arbitrary errors when running go CLI commands.

The suffix has been added properly before: https://github.com/jdx/mise/commit/786220c6178625980bdcc61403c32db19d51360f#diff-352f25ac643733cdd2fb54a78ced508b3658cab947230ab610b968f154840413L54

### DIFF
--- a/src/plugins/core/go.rs
+++ b/src/plugins/core/go.rs
@@ -157,7 +157,7 @@ impl GoPlugin {
             set("GOBIN", tv.install_path().join("bin"));
         }
         if settings.go_set_goroot {
-            set("GOROOT", tv.install_path());
+            set("GOROOT", tv.install_path().join("go"));
         }
         if settings.go_set_gopath {
             set("GOPATH", self.gopath(tv));
--- a/src/plugins/core/go.rs
+++ b/src/plugins/core/go.rs
@@ -62,7 +62,11 @@ impl GoPlugin {
 
     // Represents GOROOT environment variable
     fn goroot(&self, tv: &ToolVersion) -> PathBuf {
-        tv.install_path().join("go")
+        let old_path = tv.install_path().join("go");
+        if old_path.exists() {
+            return old_path;
+        }
+        tv.install_path()
     }
 
     // Represents GOBIN environment variable
--- a/src/plugins/core/go.rs
+++ b/src/plugins/core/go.rs
@@ -50,11 +50,24 @@ impl GoPlugin {
         })
     }
 
+    // Represents go binary path
     fn go_bin(&self, tv: &ToolVersion) -> PathBuf {
         tv.install_path().join("bin/go")
     }
+
+    // Represents GOPATH environment variable
     fn gopath(&self, tv: &ToolVersion) -> PathBuf {
         tv.install_path().join("packages")
+    }
+
+    // Represents GOROOT environment variable
+    fn goroot(&self, tv: &ToolVersion) -> PathBuf {
+        tv.install_path().join("go")
+    }
+
+    // Represents GOBIN environment variable
+    fn gobin(&self, tv: &ToolVersion) -> PathBuf {
+        tv.install_path().join("bin")
     }
 
     fn install_default_packages(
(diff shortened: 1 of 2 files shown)


src/plugins/core/go.rs
```diff
@@ -154,10 +167,10 @@ impl GoPlugin {
         let gobin = settings.go_set_gobin;
         let gobin_env_is_set = env::PRISTINE_ENV.contains_key("GOBIN");
         if gobin == Some(true) || (gobin.is_none() && !gobin_env_is_set) {
-            set("GOBIN", tv.install_path().join("bin"));
+            set("GOBIN", self.gobin(tv));
         }
         if settings.go_set_goroot {
-            set("GOROOT", tv.install_path());
+            set("GOROOT", self.goroot(tv));
```
jdx, line 177: you need to also extract the tarball to this directory

wheinze: oh i see, so this is much more of a migration issue. when reinstalling the go versions with mise 2024.2.14, it's fine, as the directory layout doesn't differ from the expectations. 

so I'm wondering what your actual suggestion might be? is there a better way to transition already installed versions of Go into the new layout, or would you revert to the old directory layout (by extracting the tarball etc.)?

jdx: what's the error you're seeing exactly?

wheinze: > what's the error you're seeing exactly?

once (re-)installed with a recent version of mise (2024.2.14), the issues do not occur anymore. 

unfortunately, i did a reinstall of all my Go versions by running `mise uninstall go --all` before, so i could just give a recap of my journey, but cannot grep any errors anymore without doing a repro of the old setup:

Having a Go version installed with an older version of mise before upgrading to one of the recent versions (I would have to figure out which version exactly did cause this issue, but would assume it to be 2024.2.8, as this is the version including [these changes)](https://github.com/jdx/mise/commit/786220c6178625980bdcc61403c32db19d51360f)), results in issue related to GOROOT being set to `...mise/installs/go/<version>`, while still being installed to `...mise/installs/go/<version>/go`, with a directory layout like this in `...mise/installs/go/<version>`:
```
├── go
├── ...
└── packages
```

In concrete this means, I ran `go get -u ./...` in a project with a Go 1.22 installed via mise which returned: 
`go: GOPROXY list is not the empty string, but contains no entries`. 

Having a look into my configs, it turned out that I did not set `GOPROXY` anywhere, nor did mise. I tried the same Go command with a system-wide version of Go, which did succeed. So the issues must be related to mise. 
Then I did fire an strace to check, what paths are seeked to find the `go.env` file, that is defining the defaults for `GOPROXY`.


The strace:
```
LC_ALL=C strace -fe open,openat go get -u ./...
```
did show, that a `go.env` was tried to be found here: `...mise/installs/go/<version>/go.env`, but it was located here: `...mise/installs/go/<version>/go/go.env`.

Out of curiosity I just copied `go.env` up a directory, so this actual error was gone, but other ones popped up unsurprisingly:

Trying to install dependencies in a Go project resulted in errors that the std library wasn't present like `sync: package sync is not in std...`.

This led me to add the proper path back to mise via this PR, as `GOROOT´ wasn't aligned with my mise-issued installations of Go. But now knowing this has also changed in the installation process made up my question on how to properly solve this, i would call it, migration issue.

Probably the easiest way to reproduce would be:
1. install mise <2024.2.8
2. install Go 1.22
3. upgrade to latest mise
4. try to use Go

wheinze: tl;dr I wouldn't propose to merge this PR anymore, as I think other solutions would be more valuable here. 

@jdx what do you suggest on how to proceed here? should I close this PR and we just keep this in mind, whenever other users might encounter some issues related to this changes?

jdx: I haven't heard any other reports around this. We could just modify GOROOT if we detect the "go" directory is in the install dir

wheinze: > I haven't heard any other reports around this. We could just modify GOROOT if we detect the "go" directory is in the install dir

yep, sounds reasonable, i added a check: https://github.com/jdx/mise/pull/1661/commits/54eb4634f11944f3514f9368555f6cfa577ed5a0

jdx (approved):

PR #1688, opened 2024-02-20
fix(bun): install bunx symlink

### LINKED ISSUE #1680: command not found: bunx
**Describe the bug**
When install bun following the official installer `curl -fsSL https://bun.sh/install | bash`, `bunx` cli is also installed.
However when installing bun via mise, `bunx` is missing.

**To Reproduce**
```
$ mise use -g bun@latest
$ bunx
zsh: command not found: bunx
```

**Expected behavior**
`bunx` command executable.

**`mise doctor` output**

```text
mise version:
  2024.2.16 macos-arm64 (2024-02-15)

build:
  Target: aarch64-apple-darwin
  Features: DEFAULT, NATIVE_TLS
  Built: Thu, 15 Feb 2024 02:24:55 +0000
  Rust Version: rustc 1.75.0 (82e1608df 2023-12-21) (Homebrew)
  Profile: release

shell:
  /bin/zsh
  zsh 5.9 (x86_64-apple-darwin23.0)

mise dirs:
  data: /Users/user/.local/share/mise
  config: /Users/user/.config/mise
  cache: /Users/user/Library/Caches/mise
  state: /Users/user/.local/state/mise
  shims: /Users/user/.local/share/mise/shims

mise environment variables:
  MISE_SHELL=zsh

settings:
  activate_aggressive = false
  all_compile = false
  always_keep_download = false
  always_keep_install = false
  asdf_compat = false
  cargo_binstall = true
  color = true
  disable_tools = []
  experimental = false
  go_set_gopath = false
  go_set_goroot = true
  go_skip_checksum = false
  jobs = 4
  legacy_version_file = true
  node_compile = false
  not_found_auto_install = true
  paranoid = false
  python_compile = false
  python_default_packages_file = "/Users/user/.default-python-packages"
  raw = false
  trusted_config_paths = []
  quiet = false
  verbose = false
  yes = false
  ci = false
  debug = false
  trace = false
  log_level = "info"

  [status]
  show_env = false
  show_tools = false


activated: yes
shims_on_path: no
config files:
  ~/.tool-versions
  ~/.config/mise/config.toml

plugins:
  bun      (core)
  deno     (core)
  go       (core)
  java     (core)
  node     (core)
  python   (core)
  ruby     (core)

toolset:
  python@3.11.7, bun@latest, node@20.10.0, deno@1.39.1

No problems found
```

**Additional context**

booniepepper:
Closes #1680

### DIFF
--- a/src/plugins/core/bun.rs
+++ b/src/plugins/core/bun.rs
@@ -87,6 +87,7 @@ impl BunPlugin {
             self.bun_bin(&ctx.tv),
         )?;
         file::make_executable(&self.bun_bin(&ctx.tv))?;
+        file::make_symlink(Path::new("./bun"), &ctx.tv.install_path().join("bin/bunx"))?;
         Ok(())
     }
 
--- a/e2e/test_bun
+++ b/e2e/test_bun
@@ -8,6 +8,8 @@ export MISE_EXPERIMENTAL=1
 cat <<EOF >.bun-version
 1.0.17
 EOF
+
 mise i bun
 assert_contains "mise x bun -- bun -v" "1.0.17"
+assert_contains "mise x bun -- bunx oclif --version" "oclif"
 rm .bun-version


booniepepper:
Oops, should have started with a download of the distribution zip instead of reading docs

The zip files from bun contain only the `bun` executable, and `bunx` is very literally an alias for `bun x`

booniepepper:
More specifically, if I install with the script at `https://bun.sh/install` then I get `bunx` as a symlink to `bun`. Sounds like the bun binary checks its argv[0] process name to decide behavior, [I've done that](https://github.com/so-dang-cool/dt/blob/8da0c6b96f8aad69f15144321aa0f358de3aba55/src/main.zig#L32-L36) before too for shebang support, it's a fun trick

src/plugins/core/bun.rs
```diff
@@ -87,6 +87,7 @@ impl BunPlugin {
             self.bun_bin(&ctx.tv),
         )?;
         file::make_executable(&self.bun_bin(&ctx.tv))?;
+        file::make_symlink(&self.bun_bin(&ctx.tv), &ctx.tv.install_path().join("bin/bunx"))?;
```
jdx: you should use a relative symlink because the way this is written it would fail if any of the parent directories are ever moved

jdx:
it'd be good to add a line to e2e/test_bun for a simple test

src/plugins/core/bun.rs
booniepepper: Oh, good call

src/plugins/core/bun.rs
```diff
@@ -87,6 +87,7 @@ impl BunPlugin {
             self.bun_bin(&ctx.tv),
         )?;
         file::make_executable(&self.bun_bin(&ctx.tv))?;
+        file::make_relative_symlink(Path::new("bun"), &ctx.tv.install_path().join("bin/bunx"))?;
```
jdx: can you not just use the former `make_symlink` call with `Path::new("./bun")`?

booniepepper: Let me test it. I was assuming that would create a path relative to where the process is executing, rather than relative to the target/destination

booniepepper: Ok, it seems like it's already running in the right context, so yeah the new function is redundant

PR #1690, opened 2024-02-21
Retry with https if request fails

grant0417:
Port 80 is blocked CI for more security so I am getting the following error:

```
Error: 
   0: error sending request for url (http://mise-versions.jdx.dev/python-precompiled): error trying to connect: operation timed out
   1: error trying to connect: operation timed out
   2: operation timed out

Location:
   src/http.rs:56
```

I added the same logic for retying with https as below which would only occur before if the server responded with an error status. Now it should retry even if the net level request fails.

### DIFF
--- a/src/http.rs
+++ b/src/http.rs
@@ -45,7 +45,7 @@ impl Client {
     }
 
     pub fn get<U: IntoUrl>(&self, url: U) -> Result<Response> {
-        let url = url.into_url().unwrap();
+        let mut url = url.into_url().unwrap();
         debug!("GET {}", url);
         let mut req = self.reqwest.get(url.clone());
         if url.host_str() == Some("api.github.com") {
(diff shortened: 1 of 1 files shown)


src/http.rs
```diff
@@ -45,33 +45,44 @@ impl Client {
...
         let mut req = self.reqwest.get(url.clone());
         if url.host_str() == Some("api.github.com") {
             if let Some(token) = &*env::GITHUB_API_TOKEN {
                 req = req.header("authorization", format!("token {}", token));
             }
         }
-        let resp = req.send()?;
+
+        let res = req.send();
+        let resp = match res {
```
jdx:
```suggestion
        let resp = match req.send() {
```

grant0417: Fixed

PR #1693, opened 2024-02-21
build: use vendored openssl crate feature to mitigate missing headers

wheinze:
Hopefully fixes these kinda errors: https://github.com/jdx/mise/actions/runs/7996715229/job/21839765074

Not quite sure yet, but found [hints](https://github.com/sfackler/rust-openssl/issues/1021#issuecomment-605602642) that openssl headers might be missing. Using the vendored feature of openssl crate, these should be used from openssl-src.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1564,6 +1564,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ff011a302c396a5197692431fc1948019154afc178baf7d8e37367442a4601cf"
 
 [[package]]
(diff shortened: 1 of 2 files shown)


wheinze:
> we're already using this I think:
> 
> https://github.com/jdx/mise/blob/4adac60c41767bb18b479ce2532324bf33d1c946/.github/workflows/release.yml#L86
> 
> not sure what's going on here but it looks like maybe a bug around the cc crate

oh, i c - kk, then i'll close this ;)

PR #1697, opened 2024-02-23
feat: added "en" command

jdx:
@booniepepper this isn't working right but I'm curious if you have any ideas.

The problem is that it will still load your rc files, so if you do something like `export PATH="/opt/homebrew/bin:$PATH" in your rc file that will make the mise tools lower priority than homebrew so it won't use mise tools if brew provides any.

One idea I had was running `mise env` when the shell starts but I'm not sure if it's possible to have zsh (for example) load a script like that.

Technically it does _work_ though.

### DIFF
--- a/docs/.vitepress/cli_commands.ts
+++ b/docs/.vitepress/cli_commands.ts
@@ -96,6 +96,9 @@ export const commands: { [key: string]: Command } = {
   doctor: {
     hide: false,
   },
+  en: {
+    hide: false,
+  },
   env: {
     hide: false,
   },
--- a/docs/cli/en.md
+++ b/docs/cli/en.md
@@ -9,12 +9,6 @@ This is an alternative to `mise activate` that allows you to explicitly start a 
 It will have the tools and environment variables in the configs loaded.
 Note that changing directories will not update the mise environment.
 
-It's a lightweight alternative to `mise activate` if you don't want to put that into your shell rc but
-still don't want to deal with shims. It probably only makes sense for interactive use-cases.
-
-It's sort of akin to manually running `source .venv/bin/activate` for Python virtualenvs if you're
-familiar with that workflow.
-
 ## Arguments
 
 ### `[DIR]`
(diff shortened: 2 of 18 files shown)


jdx:
I think I figured out a decent way forward. I just put some docs on how to disable the rc files with bash or zsh. I think that's better than trying to automatically do anything.

docs/cli-reference.md
```diff
@@ -319,6 +319,42 @@ Examples:
     [WARN] plugin node is not installed
 ```
 
+## `mise en [OPTIONS] [DIR]`
+
+```text
+[experimental] starts a new shell with the mise environment built from the current configuration
```
booniepepper: I think `$SHELL` should be the default command launched

For the extra configuration like `--shell SHELL`... should this just support any command and be more like `mise x -- <etc>` but with a working directory set?

(The description of `--cd` uses "command" instead of "shell")

Asking because even with a shell, someone may want to be able to pass additional arguments like `-r` for restricted shell, or `-l` for login session, etc

jdx: > should this just support any command and be more like mise x -- <etc> but with a working directory set?

well this is a very good point. I am going to need to ponder on this one a bit

jdx: `$SHELL` is the default, but it has edge cases

jdx:
I think my issue is actually limited to fish (what I use) and how `fish_user_paths` work. It's a universal variable that always gets loaded. Skipping rc files isn't really a solution because that means you get no prompt or anything. I handled fish uniquely by running it with `fish -C "fish_add_path -gm ..."` which causes the mise paths to get loaded after fish gets loaded. That should fix the issue I had. I don't think this would be a problem with bash or zsh since in those shells users should be putting that stuff into ~/.bash_profile and ~/.zprofile I think.

Well, it would be an issue if it was in ~/.zshenv or whatever the bash equivalent to that is but I think in that case I can just tell users not to do that since that shouldn't be there anyways.

Anyhow, this might be just a fish thing mostly but I fixed fish.

PR #1699, opened 2024-02-23
Update mise.json - fix missing_tools type

fxsalazar:


schema/mise.json
```diff
@@ -343,8 +343,9 @@
           "properties": {
             "missing_tools": {
               "description": "display warning when a tool is not installed",
-              "type": "boolean",
-              "default": true
+              "type": "string"
```
jdx:
```suggestion
              "type": "string",
```

PR #1708, opened 2024-02-24
Add Swift plugin

pepicrft:
I'm implementing a core plugin to manage Swift versions. As a follow-up work, I'll implement a backend to install community packages representing CLIs.

The current [Swift asdf-plugin](https://github.com/fcrespo82/asdf-swift) turns out to be using [swiftenv](https://github.com/kylef/swiftenv) under the hood installing it if absent in the environment using Homebrew (too much indirection).


src/forge/swift.rs
```diff
@@ -0,0 +1,59 @@
...
+impl Forge for SwiftForge {
+    fn get_type(&self) -> ForgeType {
+        ForgeType::Npm
+    }
+
+    fn fa(&self) -> &ForgeArg {
+        &self.fa
+    }
+
+    fn list_remote_versions(&self) -> eyre::Result<Vec<String>> {
```
pepicrft: @jdx, the Swift Package Manager, does decentralized dependency resolution, so packages are scattered across Git repositories with references between them. There was a community effort to index the packages on a website, but that website still needs to provide an API. Luckily, they have a [repo](https://raw.githubusercontent.com/SwiftPackageIndex/PackageList/main/packages.json) that contains a list of all the packages. Since they are repositories, what do you think about using the `{owner}/{repo}` convention for identifying the packages?

Does the logic described in the comments make sense? If so, do you mind pointing me to a place in the codebase with some network code?

src/forge/swift.rs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    fn install_version_impl(&self, ctx: &InstallContext) -> eyre::Result<()> {
+        // TODO
+        // 1. Check that the Swift Package Manager is installed.
+        // 2. Clone the package into a temporary directory.
+        // 3. Use `swift build` there to compile the tool.
+        // 4. Copy the right artifacts to the install path.
+        //      Note that as part of this, we need to ensure we copy any dynamic frameworks that are required for the CLI to function.
+        //      We can start simple, and copy all the dynamic frameworks and libraries from the build directory.
```
pepicrft: Does this make sense @jdx? If so, do I get a temporary directory through the `remote_version_cache` property? I'll need it to clone the repository into it so that I can compile it using the Swift Package Manager.

src/forge/swift.rs
```diff
@@ -0,0 +1,59 @@
...
+
+        Ok(())
+    }
+}
+
+impl SwiftForge {
+    pub fn new(fa: ForgeArg) -> Self {
+        Self {
+            remote_version_cache: CacheManager::new(
+                fa.cache_path.join("remote_versions.msgpack.z"),
```
pepicrft: What does `fa` stand for? Why is it `remote_versions.msgpack.z`?

src/forge/swift.rs
jdx: the go backend might be helpful since that is similar to this—though the go toolchain does offer a CLI command to fetch versions.

Here's example http/json code https://github.com/jdx/mise/blob/4893e09ffe77a11dfd671cdfe0ce2591eebf5d83/src/plugins/core/bun.rs#L37

src/forge/swift.rs
jdx: you can use `ctx.tv.cache_path()` or `ctx.tv.download_path()`. You don't want to use `remote_version_cache`, that's for storing the version strings.

src/forge/swift.rs
jdx: "ForgeArg". And because it's messagepack compressed with zlib.

src/forge/swift.rs
pepicrft: Sweet! I'll use that one as a reference. Thanks @jdx

src/forge/swift.rs
pepicrft: Cool! I'll stick to them then.

src/plugins/core/swift.rs
```diff
@@ -0,0 +1,215 @@
...
+        if os() == "osx" {
+            CmdLineRunner::new("/usr/sbin/installer")
+            .with_pr(ctx.pr.as_ref())
+            .arg("-pkg")
+            .arg(download_path)
+            .arg("-target")
+            .arg(&bin_directory)
+            .execute()?;
+        } else {
+            file::unzip(download_path, &ctx.tv.download_path())?;
```
pepicrft, line 124: @jdx I noticed the tars contain other artifacts beyond the binary. I assume we have to respect the directory structure here. Should I untar everything into `bin_directory` and use [`make_executable`](https://github.com/jdx/mise/blob/16e7da08fc135166e0f44e64d44fb3b3325943aa/src/plugins/core/bun.rs#L89)?

src/plugins/core/swift.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+    if distribution == "amzn" {
+        return format!("amazonlinux{}", version);
+    } else {
+        format!("{}{}", distribution, version)
+    }
+}
+
+fn os() -> String {
+    // ASK: Is it ok if unlike other plugins, we return a String here instead of a static str?
```
pepicrft, line 201: Other modules return a `&str` with a static lifetime. However, since I need to get the Linux distribution at runtime, I needed to change the return type to `String`. Do you think it's fine?

src/plugins/core/swift.rs
```diff
@@ -0,0 +1,215 @@
...
+
+    fn install(&self, ctx: &InstallContext, download_path: &Path) -> Result<()> {
+        let filename = download_path.file_name().unwrap().to_string_lossy();
+        ctx.pr.set_message(format!("installing {filename}"));
+        file::remove_all(ctx.tv.install_path())?;
+        let bin_directory = ctx.tv.install_path().join("bin");
+        file::create_dir_all(&bin_directory)?;
+
+        if os() == "osx" {
+            CmdLineRunner::new("/usr/sbin/installer")
```
pepicrft, line 116: When using `installer`, the two values that `-target` can take are `CurrentUserHomeDirectory` or `LocalSystem`. In the case of the former, the Swift toolchain ends up being installed at `~/Library/Developer/Toolchains/swift-5.9.2-RELEASE.xctoolchain/`. In the case of the later, the directory is `/Library/Developer/Toolchains`.

@jdx what do you think if I use the former to prevent having to use `sudo`, and then copy the `*.xctoolchain` into the Mise directory?

src/forge/mod.rs
```diff
@@ -79,7 +79,7 @@ pub fn get(fa: &ForgeArg) -> AForge {
                 ForgeType::Asdf => Arc::new(ExternalPlugin::new(name)),
                 ForgeType::Cargo => Arc::new(CargoForge::new(fa.clone())),
                 ForgeType::Npm => Arc::new(npm::NPMForge::new(fa.clone())),
-                ForgeType::Go => Arc::new(go::GoForge::new(fa.clone())),
+                ForgeType::Go => Arc::new(go::GoForge::new(fa.clone()))
```
jdx: nitpick: I'd prefer keeping the trailing comments. `rust fmt` might enforce this—not sure.

src/plugins/core/mod.rs
```diff
@@ -36,6 +38,7 @@ mod ruby;
 
 pub static CORE_PLUGINS: Lazy<ForgeList> = Lazy::new(|| {
     let mut plugins: Vec<Arc<dyn Forge>> = vec![
+        Arc::new(SwiftPlugin::new()),
```
jdx: nitpick: keep these alphabetized

src/plugins/core/swift.rs
```diff
@@ -0,0 +1,216 @@
...
+            "https://download.swift.org/swift-{}-release/{}/swift-{}-RELEASE/swift-{}-RELEASE-{}{}.{}",
+            tv.version,
+            match os.as_str() {
+                "osx" => "xcode".to_string(), // Apple uses "xcode" instead of "osx" for this path segment
+                os => os.replace(".", "").to_string(),
+            },
+            tv.version,
+            tv.version,
+            os,
+            match os.as_str() {
```
jdx: nitpick: just to make this a bit cleaner I would do this above:

```rust
let arch = match os.as_str() {
// ...
```

ditto for the extension

src/plugins/core/swift.rs
```diff
@@ -0,0 +1,216 @@
...
+            }
+        )
+    }
+
+    fn download(&self, tv: &ToolVersion, pr: &dyn SingleReport) -> Result<PathBuf> {
+        let url = self.download_url(&tv);
+        let filename = url.split('/').last().unwrap();
+        let download_path = tv.download_path().join(filename);
+        pr.set_message(format!("downloading {filename}"));
+        HTTP.download_file(&url, &download_path, Some(pr))?;
```
jdx, line 104: nice, glad you put the "pr" in on this since that enables progress bar reporting

src/plugins/core/swift.rs
```diff
@@ -0,0 +1,216 @@
...
+        Ok(download_path)
+    }
+
+    fn install(&self, ctx: &InstallContext, download_path: &Path) -> Result<()> {
+        let filename = download_path.file_name().unwrap().to_string_lossy();
+        ctx.pr.set_message(format!("installing {filename}"));
+        file::remove_all(ctx.tv.install_path())?;
```
jdx, line 111: I'm not 100% sure if this is necessary. I think mise actually does this automatically. Not a big deal though, I'm happy to leave it in just to be safe.

src/plugins/core/swift.rs
```diff
@@ -0,0 +1,216 @@
...
+        let bin_directory = ctx.tv.install_path().join("bin");
+        file::create_dir_all(&bin_directory)?;
+
+        if os() == "osx" {
```
jdx: not a huge deal but `if cfg!(target_os = "macos")` might be preferred here. I think it has some performance benefits which are clearly useless here but a good habit for us to be in.

src/plugins/core/swift.rs
jdx: using sudo is kind of a dealbreaker for anything in mise since doing that screws up permissions—so yeah I think what you said about CurrentUserHomeDirectory is better, then we move the xctoolchain into place sounds like the right approach

src/plugins/core/swift.rs
jdx: in general in mise we always just leave all the files that tools put into their tarballs and don't prune anything unless we have a good reason to prune files we explicitly don't want.

In regards to make_executable—yeah you'll likely need that since zip files don't have permissions. I believe we do this for bun which also uses zip.

src/plugins/core/swift.rs
jdx: yeah exactly, you'll need to export full strings. rust makes you do a bit more work around that since it technically is less efficient (the idea in rust is that anything that has a performance cost you will see in your code), but clearly for what we're doing in this file that's a completely irrelevant cost.

In fact there is probably not a single place in our entire codebase where string instantiation is a cost worth considering. I use `&'static str` purely just to stay in the habit in case code I write ever _does_ benefit from that consideration but it's probably not something in this project is ever (or at least very rarely) important.

src/plugins/core/mod.rs
jdx: we should also put this under the experimental list (see erlang below) for a few months while we beta test it

jdx:
@pepicrft really sorry it took me so long to get around to this. It looks excellent though. I had some feedback that really is just nitpicks pretty much. The one thing we should get is an e2e test though, https://github.com/jdx/mise/blob/main/e2e/test_erlang is a good example of one you can use as a reference.

You can either just copy `test_erlang` over to `test_swift` and push changes here, you will need to ping me to approve the test to execute I think since I don't think you've contributed to mise directly yet but of course I'm happy to do that as often as you need. You can also run the e2e tests via docker, see https://mise.jdx.dev/contributing.html for instruction on that.

If you have any trouble with the e2e test I'd be happy to add that to your PR too.

src/plugins/core/swift.rs
pepicrft: Good idea! Let me change it

src/plugins/core/swift.rs
pepicrft: It was a bit of copy-pasting 😬. I'm still getting familiar with the codebase constructs.

src/plugins/core/swift.rs
pepicrft: Thanks! I'll do that then.

pepicrft:
No worries @jdx. I appreciate your review of the PR.
The implementation is still incomplete, so as soon as I finish it, I'll add an e2e test to ensure everything works as expected.

I have one question. Xcode installations come with a Swift toolchain embedded in the application bundle, which is the one used by default when that version of Xcode is selected via `xcode-select`. I was wondering if `mise ls swift` should include that version too. Maybe with a different formatting:

```
$ mise ls swift
  5.9.0
  5.10.0 (/Applications/Xcode.app)
```

If you agree, how do you think I should approach it from the `Forge` trait. I proposed [here](https://github.com/jdx/mise/pull/1708/commits/9c37f8bae0327273910e253941cc3f55899d5c25), but I'm not entirely convinced that's the best idea. In essence, some plugins would be able to provide an additional set of versions that are managed outside of Mise.

johnpyp:
Having recently gone through installing swift on linux, there's a couple of issues on different distributions it seems (primarily due to linking to native libraries). Have you been able to test this on multiple distributions?

For example, on arch the most popular [package to install swift](https://aur.archlinux.org/packages/swift-bin) uses patchelf because the source swift distribution uses incorrect paths (at least for arch). Using https://github.com/swift-server/swiftly as an alternative, has similar issues where just running `swift` will cause some library not found errors.

I'm not very familiar with swift itself or the best way to resolve this across different distributions, but it would be great if it could accommodate them.

I know there are options when building swift to statically link its dependency libraries too, which might be another avenue to go down?

pepicrft:
Thanks @johnpyp! I haven't tested it yet in Linux distributions, but I'll make sure I do before considering the PR ready. What other contributions come to your mind?

shassard:
For Linux support, the Ubuntu 24.04 builds of Swift 5.10.1 work perfectly under Arch Linux without modification. The Ubuntu 24.04 builds don't seem to exist for all versions yet, but that might be a good option to use if they're available.

Issue #1711, opened 2024-02-26
Show more install logs (ruby in this case)


erickguan:
Hello,

When I build ruby, I would like to see more information during configuration. Currently, `mise` shows a collapsed log:

`mise ruby@3.2.3 -> ./configure "--prefix=$HOME/.local/share/mise/installs/ruby/3.2.3" --with-openssl-dir=/opt/homebrew/opt/openssl@3 --enable-shared --with-readline-dir=/opt/hom ⠐ 13s`

This is super cool and I like more advanced UX. Though I would need to validate the configuration when I skim through the logs.

The specific reason is that I need to enable yjit. I run `CONFIGURE_OPTS="--enable-yjit" mise install ruby@3.2.3`. In this case, I would need to validate that yjit option is part of the configure options:
`-> ./configure "--prefix=$HOME/.local/share/mise/installs/ruby/3.2.3" --with-openssl-dir=/opt/homebrew/opt/openssl@3 --enable-shared --with-readline-dir=/opt/homebrew/opt/readline --with-libyaml-dir=/opt/homebrew/opt/libyaml --with-gmp-dir=/opt/homebrew/opt/gmp --with-ext=openssl,psych,+ --enable-yjit`

This is shown when the build fails of course. But it is great that I can see them when the build is running smoothly.

jdx:
just use `--verbose`

erickguan:
Thanks! My bad. I didn't understand what the verbose would do in this case. Very cool that `mise` matches my intuitive habits. I made a small PR for some text update.

PR #1712, opened 2024-02-26
Update a few phrases for mise install

erickguan:
A minor effort in updating the doc. I can see more venues for improvements on:
- mentioning more UX features. 
- clarifying what the default tool's version

I will leave it to future work for now. Perhaps soon! 

The text should be more intuitive for my understanding on #1711

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -425,12 +425,12 @@ Options:
 ```text
 Install a tool version
 
-This will install a tool version to `~/.local/share/mise/installs/<PLUGIN>/<VERSION>`
-It won't be used simply by being installed, however.
-For that, you must set up a `.mise.toml`/`.tool-version` file manually or with `mise use`.
-Or you can call a tool version explicitly with `mise exec <TOOL>@<VERSION> -- <COMMAND>`.
+`install` will install a tool version to `~/.local/share/mise/installs/<PLUGIN>/<VERSION>`
+Tools are inactive without being enabled, however.
+To enable a tool, set up a `.mise.toml`/`.tool-version` file or `mise use`.
+Another way is to use a tool explicitly with `mise exec <TOOL>@<VERSION> -- <COMMAND>`.
 
-Tools will be installed in parallel. To disable, set `--jobs=1` or `MISE_JOBS=1`
+`install` installs tools in parallel, by default. To disable, set `--jobs=1` or `MISE_JOBS=1`
 
 Usage: install [OPTIONS] [TOOL@VERSION]...
 
@@ -452,7 +452,9 @@ Options:
           Directly pipe stdin/stdout/stderr from plugin to user Sets --jobs=1
 
   -v, --verbose...
-          Show installation output
+          Show installation output. This argument will print plugins' logs.
+          When plugin downloads assets, mise will show download logs.
+          When plugin configures and compiles tools, mise will show configuration and compilation logs.
 
 Examples:
 
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -425,12 +425,12 @@ Options:
 ```text
 Install a tool version
 
-`install` will install a tool version to `~/.local/share/mise/installs/<PLUGIN>/<VERSION>`
-Tools are inactive without being enabled, however.
-To enable a tool, set up a `.mise.toml`/`.tool-version` file or `mise use`.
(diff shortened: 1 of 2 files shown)


docs/cli-reference.md
```diff
@@ -425,12 +425,12 @@ Options:
 ```text
 Install a tool version
 
-This will install a tool version to `~/.local/share/mise/installs/<PLUGIN>/<VERSION>`
-It won't be used simply by being installed, however.
-For that, you must set up a `.mise.toml`/`.tool-version` file manually or with `mise use`.
-Or you can call a tool version explicitly with `mise exec <TOOL>@<VERSION> -- <COMMAND>`.
+`install` will install a tool version to `~/.local/share/mise/installs/<PLUGIN>/<VERSION>`
```
jdx:
```suggestion
Installs a tool version to `~/.local/share/mise/installs/<PLUGIN>/<VERSION>`
```

docs/cli-reference.md
```diff
@@ -425,12 +425,12 @@ Options:
...
+Tools are inactive without being enabled, however.
+To enable a tool, set up a `.mise.toml`/`.tool-version` file or `mise use`.
+Another way is to use a tool explicitly with `mise exec <TOOL>@<VERSION> -- <COMMAND>`.
```
jdx:
```suggestion
Alternatively, run `mise exec <TOOL>@<VERSION> -- <COMMAND>` to execute a tool without creating config files.
```

docs/cli-reference.md
```diff
@@ -425,12 +425,12 @@ Options:
 ```text
 Install a tool version
 
-This will install a tool version to `~/.local/share/mise/installs/<PLUGIN>/<VERSION>`
-It won't be used simply by being installed, however.
-For that, you must set up a `.mise.toml`/`.tool-version` file manually or with `mise use`.
-Or you can call a tool version explicitly with `mise exec <TOOL>@<VERSION> -- <COMMAND>`.
+`install` will install a tool version to `~/.local/share/mise/installs/<PLUGIN>/<VERSION>`
+Tools are inactive without being enabled, however.
```
jdx:
```suggestion
Installing alone will not activate the tools so they won't be in PATH.
```

I'm a little uncomfortable with this language since it still will create shims, so shim users _will_ have it in PATH but the shims won't work. I think that this language does make sense for the majority of users though and that subtle behavior difference won't make any practical difference since the way shims behave is if they're not active it falls back to the system bin (if any). So it behaves basically as if it "wasn't in PATH".

docs/cli-reference.md
```diff
@@ -425,12 +425,12 @@ Options:
...
+To enable a tool, set up a `.mise.toml`/`.tool-version` file or `mise use`.
```
jdx:
```suggestion
To install and/or activate in one command, use `mise use` which will create a `.mise.toml` file in the current directory to activate this tool when inside the directory.
```

docs/cli-reference.md
```diff
@@ -452,7 +452,9 @@ Options:
           Directly pipe stdin/stdout/stderr from plugin to user Sets --jobs=1
 
   -v, --verbose...
-          Show installation output
+          Show installation output. This argument will print plugins' logs.
+          When plugin downloads assets, mise will show download logs.
+          When plugin configures and compiles tools, mise will show configuration and compilation logs.
```
jdx:
```suggestion
          Show installation output.
          
          This argument will print plugin output such as download, configuration, and compilation output.
```

jdx:
so I really appreciate your attention on this. I agree that this language can be improved (like a lot of stuff in mise). Apologies if some of my feedback seems nitpicky. I'm trying to be careful about being consistent and using "activated" instead of "enabled" everywhere I describe this behavior.

Unfortunately this isn't the right place to make this change. This file is autogenerated, you'll need to make these changes in [this file's comments](https://github.com/jdx/mise/blob/main/src/cli/install.rs) instead. Then on the next release this markdown file will be updated.

docs/cli-reference.md
erickguan: Good point!

erickguan (commented):
Thanks for your review. I would guess the doc comes from a script but I didn't browser through code yet. Writing a PR is probably better than forgetting about it after getting your help (from an issue). 

mise is a nice tool. I am using it instead of asdf. I understand the consistency you want to achieve. I respect the effort you had put into mise already. The next patch will be better!

PR #1719, opened 2024-02-29
fix(java): sdkmanrc zulu JVMs are missing in mise

roele:
SDKMAN uses a different Java versioning scheme as the core plugin, therefore versions in `.sdkmanrc` are reported missing even if they're installed.

SDKMAN is using `{java-version}` ( e.g. 21.0.2) as opposed to vendor versions and different `{vendor}` strings. In most cases the `{vendor-version}` matches the Java Version with an additional build suffix. SDKMAN also supports vendors (bisheng, dragonwell, graalvm, liberica nik, mandrel, trava) that are not available/mappable in the core plugin.

ASDF: `{vendor}-{vendor-version}` (e.g. zulu-21.30.15)
SDKMAN: `{java-version}-{vendor}` (e.g. 21.0.2-zulu)

This PR is an approach to map the SDKMAN version. In case of "zulu" only the major version could be matched as we dont have the java version while matching. Other option would be to remove support for `.sdkmanrc` files completely.

### DIFF
--- a/src/forge/mod.rs
+++ b/src/forge/mod.rs
@@ -376,7 +376,7 @@ fn fuzzy_match_filter(versions: Vec<String>, query: &str) -> eyre::Result<Vec<St
     if query == "latest" {
         query = "[0-9].*";
     }
-    let query_regex = Regex::new(&format!("^{}([-.].+)?$", query))?;
+    let query_regex = Regex::new(&format!("^{}([+-.].+)?$", query))?;
     let versions = versions
         .into_iter()
         .filter(|v| {
--- a/src/forge/mod.rs
+++ b/src/forge/mod.rs
@@ -376,7 +376,7 @@ fn fuzzy_match_filter(versions: Vec<String>, query: &str) -> eyre::Result<Vec<St
     if query == "latest" {
         query = "[0-9].*";
     }
-    let query_regex = Regex::new(&format!("^{}([+-.].+)?$", query))?;
+    let query_regex = Regex::new(&format!("^{}([-.].+)?$", query))?;
     let versions = versions
         .into_iter()
         .filter(|v| {
--- a/src/plugins/core/java.rs
+++ b/src/plugins/core/java.rs
@@ -266,7 +266,7 @@ impl JavaPlugin {
     }
 
     fn tv_to_metadata(&self, tv: &ToolVersion) -> Result<&JavaMetadata> {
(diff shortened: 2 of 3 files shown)


jdx:
I'm wondering if it would be better to simply ignore zulu if that's what we see in parse_legacy_file

roele:
For zulu i think having a mapping to a major version mapping is better than having none at all. Zulu will always be treated specially (see installation) since Azul seem to do things differently than anybody else. Minor versions to not exist to my knowledge and using patch versions is probably an edge-case anyways in Java. Usually you want to develop against to newest major version.

src/plugins/core/java.rs
```diff
@@ -354,7 +354,27 @@ impl Forge for JavaPlugin {
                 .split_once('=')
                 .unwrap_or_default()
                 .1;
-            Ok(version.to_string())
+            if !version.contains('-') {
+                return Ok(version.to_string());
+            }
+            let (version, vendor) = version.rsplit_once('-').unwrap_or_default();
+            let vendor = match vendor {
+                "amzn" => "corretto",
```
jdx, line 374: this logic I think should go into parse_legacy_file

src/forge/mod.rs
```diff
@@ -376,7 +376,7 @@ fn fuzzy_match_filter(versions: Vec<String>, query: &str) -> eyre::Result<Vec<St
     if query == "latest" {
         query = "[0-9].*";
     }
-    let query_regex = Regex::new(&format!("^{}([-.].+)?$", query))?;
+    let query_regex = Regex::new(&format!("^{}([+-.].+)?$", query))?;
```
jdx: mind explaining what this is for?

roele: This is used for ignoring version suffixes added with `+` for example `temurin-jre-21.0.0+35.0.LTS`. Maybe it would be safer to implement/override `list_installed_versions_matching` and `list_versions_matching` in the Java Plugin with its own `fuzzy_match_filter`.

src/plugins/core/java.rs
roele: this is already part of `parse_legacy_file`

roele:
will also add a test-case and a hint in the documentation explaining the behaviour for `.sdkmanrc`, zulu major version mapping and potentially unsupported vendors.

PR #1739, opened 2024-03-03
Update java.rs to change symlink dir

MustCodeAl:
Change symlink comment to add a symlink to users home directory instead of globally on mac since mise will usually install the languages at the users home directory when installed through brew 
Screenshot 2024-03-02 at 8 54 45 PM

Now this should showup: `sudo ln -s /Users/notlaggy/.local/share/mise/installs/java/latest/Contents ~/Library/Java/JavaVirtualMachines/latest.jdk/Contents/`

### DIFF
--- a/src/plugins/core/java.rs
+++ b/src/plugins/core/java.rs
@@ -228,8 +228,8 @@ impl JavaPlugin {
                 "{}",
                 formatdoc! {r#"
                 To enable macOS integration, run the following commands:
-                sudo mkdir /Library/Java/JavaVirtualMachines/{version}.jdk
-                sudo ln -s {path}/Contents /Library/Java/JavaVirtualMachines/{version}.jdk/Contents
+                sudo mkdir ~/Library/Java/JavaVirtualMachines/{version}.jdk
+                sudo ln -s {path}/Contents ~/Library/Java/JavaVirtualMachines/{version}.jdk/Contents
                 "#,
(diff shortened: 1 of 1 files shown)


roele:
This has nothing to do with `brew` but with macOS system integration of JDKs. JDKs installed by package installers usually are expected at `/Library/Java/JavaVirtualMachines` to integrate with the macOS `/usr/libexec/java_home` tool.

Also note the caveat for `brew` Java installations. That is basically what we are doing with the mise installations here as well.
```
==> Caveats
For the system Java wrappers to find this JDK, symlink it with
  sudo ln -sfn /opt/homebrew/opt/openjdk/libexec/openjdk.jdk /Library/Java/JavaVirtualMachines/openjdk.jdk
```

See 
* https://docs.oracle.com/en/java/javase/21/install/installation-jdk-macos.html#GUID-EB197354-E07E-4C6A-8AF6-642E23241D39
* https://docs.aws.amazon.com/corretto/latest/corretto-11-ug/macos-install.html

jdx:
yeah my understanding aligns with what @roele said. The root directory is where `java_home` expects these to exist. Even if the home directory also works (not sure), it still probably makes sense to use the root directory since that seems more idiomatic—even if it certainly is a pain with sudo.

I'm going to close, but happy to chat more about this if you still think it would be helpful.

PR #1753, opened 2024-03-06
fix(go): go backend supports versions prefixed with 'v'

### LINKED ISSUE #1637: go backend supports versions prefixed with "v"
As policy, mise does not allow `v` prefixes for versions just to keep things consistent and the user doesn't need to remember whether their language uses that prefix or not. In many places it can optionally be added like in an `.nvmrc` file (where "v" prefixes are idiomatic), but it's optional. In fact, allowing an optional "v" across mise is something that probably should be looked at improving support in general for—that's not what this task is about though.

The go backend (and perhaps others), will return "v" prefixes for versions like the following (this is bad):

```
$ mise ls-remote go:github.com/DarthSim/hivemind
v1.0.1
v1.0.2
v1.0.3
v1.0.4
v1.0.5
v1.0.6
v1.1.0
```

notably, this prefix is also _required_ when installing go CLIs:

```
❯ mise i go:github.com/DarthSim/hivemind@1.1.0
go: github.com/DarthSim/hivemind@1.1.0: github.com/DarthSim/hivemind@1.1.0: invalid version: unknown revision 1.1.0
mise go failed
go: github.com/DarthSim/hivemind@1.1.0: github.com/DarthSim/hivemind@1.1.0: invalid version: unknown revision 1.1.0
mise go exited with non-zero status: exit code 1
mise Run with --verbose or MISE_VERBOSE=1 for more information
❯ mise i go:github.com/DarthSim/hivemind@v1.1.0
go: downloading github.com/DarthSim/hivemind v1.1.0
go: downloading github.com/pkg/term v1.2.0-beta.2
go: downloading github.com/urfave/cli v1.22.5
go: downloading github.com/DarthSim/godotenv v1.3.1
go: downloading github.com/joho/godotenv v1.4.0
go: downloading golang.org/x/sys v0.0.0-20211124211545-fe61309f8881
go: downloading github.com/cpuguy83/go-md2man/v2 v2.0.0-20190314233015-f79a8a8ca69d
go: downloading github.com/russross/blackfriday/v2 v2.0.1
go: downloading github.com/shurcooL/sanitized_anchor_name v1.0.0
mise go:github.com/DarthSim/hivemind@v1.1.0 ✓ installed
```

So 2 things likely need to happen:

- [ ] "v" prefix should be pruned from ls-remote output when using the go backend
- [ ] "v" prefix should be optional for `mise install` with the go backend

Now the question is whether this should be generic logic applied to all backends or just the go backend. In the past, I've been hesitant about aggressively pruning "v" prefixes on asdf plugins because I don't know if any might break some of them. I feel like the best solution is likely making this generic for all backends _except_ when using asdf plugins to avoid any problems.

I'm open to other ideas around this though especially if someone dives into it and has differing ideas.

roele:
Attempt to fix #1637 by removing the `v` prefix (if its the only character before the semantic version) from the `ToolVersion::version`.

For Go the prefix is stripped in `GoForge::list_remote_versions` and added for installation in `GoForge::install_version_impl`  since its required for installation.

### DIFF
--- a/src/forge/go.rs
+++ b/src/forge/go.rs
@@ -1,5 +1,7 @@
 use std::fmt::Debug;
 
+use regex::Regex;
+
 use crate::cache::CacheManager;
 use crate::cli::args::ForgeArg;
 use crate::cmd::CmdLineRunner;
@@ -32,7 +34,13 @@ impl Forge for GoForge {
                     let res = cmd!("go", "list", "-m", "-versions", "-json", cur_mod_path).read();
                     if let Ok(raw) = res {
                         let res = serde_json::from_str::<GoModInfo>(&raw);
-                        if let Ok(mod_info) = res {
+                        if let Ok(mut mod_info) = res {
+                            // remove the leading v from the versions
+                            mod_info.versions = mod_info
+                                .versions
+                                .into_iter()
+                                .map(|v| v.trim_start_matches('v').to_string())
+                                .collect();
                             return Ok(mod_info.versions);
                         }
                     };
@@ -50,9 +58,17 @@ impl Forge for GoForge {
         let settings = Settings::get();
         settings.ensure_experimental("go backend")?;
 
+        // if the (semantic) version has no v prefix, add it
+        let version_regex = Regex::new(r"^\d+(\.\d+)*([+-.].+)?$").unwrap();
+        let version = if version_regex.is_match(&ctx.tv.version) {
+            format!("v{}", ctx.tv.version)
+        } else {
+            ctx.tv.version.clone()
+        };
+
         CmdLineRunner::new("go")
             .arg("install")
-            .arg(&format!("{}@{}", self.name(), ctx.tv.version))
+            .arg(&format!("{}@{}", self.name(), version))
             .with_pr(ctx.pr.as_ref())
             .envs(config.env()?)
             .env("GOBIN", ctx.tv.install_path().join("bin"))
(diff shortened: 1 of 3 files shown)


src/forge/go.rs
```diff
@@ -50,9 +58,17 @@ impl Forge for GoForge {
         let settings = Settings::get();
         settings.ensure_experimental("go backend")?;
 
+        // if the (semantic) version has no v prefix, add it
+        let version_regex = Regex::new(r"^\d+(\.\d+)*([+-.].+)?$").unwrap();
```
roele: Not sure if that regex is good enough, since hashes seem also to be supported any hash starting with 7 digits, ~4% as per [[1]](https://medium.com/@jonathan_finch/git-commit-hash-number-theory-770f67ec492d), would match and be prefixed with `v`.

Maybe we could say at max. first 6 digits before minor version e.g. `^\d{1,6}(\.\d+)*([+-.].+)?$`.

src/plugins/mod.rs
```diff
@@ -30,6 +30,9 @@ pub static VERSION_REGEX: Lazy<regex::Regex> = Lazy::new(|| {
         .unwrap()
 });
 
+pub static VERSION_V_PREFIX_REGEX: Lazy<regex::Regex> =
+    Lazy::new(|| Regex::new(r"^v(\d+(\.\d+)*([+-.].+)?)$").unwrap());
```
jdx: There is a `regex!` macro that does this

PR #1763, opened 2024-03-10
Refactor Nushell script

### LINKED ISSUE #589: Nushell config breaks repeated tab completion of executables on PATH
**Describe the bug**

The generated `nu` config for rtx breaks second time tab completion of custom executables on the PATH.

**To Reproduce**
Steps to reproduce the behavior.

- include rtx config in nu config
- create new shell instance
- tab complete an entry on the path and `Ctrl + c`
- tab complete the same entry on the path
- will result in `NO RECORDS FOUND`

Removal of rtx from nu config fixes the issue.

**Expected behavior**

PATH entry should remain found.

**`rtx doctor` output**

```
rtx version:
  1.30.4 linux-x64 (7bcbe6d 2023-05-23)

build:
  Features: SELF_UPDATE
  Built: Tue, 23 May 2023 00:51:54 +0000
  Rust Version: rustc 1.69.0 (84c898d65 2023-04-16)
  Profile: release

shell:
  nu
  0.80.0

rtx environment variables:
  RTX_SHELL=nu

settings:
  {"experimental": "false", "missing_runtime_behavior": "warn", "always_keep_download": "false", "always_keep_install": "false", "legacy_version_file": "true", "plugin_autoupdate_last_check_duration": "10080", "trusted_config_paths": "[]", "verbose": "false", "asdf_compat": "false", "jobs": "4", "disable_default_shorthands": "false", "log_level": "INFO", "raw": "false"}

config files:
  /home/jjones/.tool-versions

plugins:
  golang   https://github.com/kennyp/asdf-golang.git#f006a12
  gradle   https://github.com/rfrancis/asdf-gradle.git#e47df00
  java     https://github.com/halcyon/asdf-java.git#1cbc9b4
  kotlin   https://github.com/asdf-community/asdf-kotlin.git#c2536cf
  node     (core)

toolset:
  java@temurin-17.0.6+10, kotlin@1.8.20, gradle@8.0.1, golang@1.20.2

No problems found
```

**Additional context**

Commenting out the line:
```nu
let-env $var.name = $"($var.value)"
```

in the activation script stops the prompt from breaking.

texastoland:
Closes #589 and fixes https://github.com/jdx/mise/issues/589#issuecomment-2002000307.

- [x] Test script locally
- [x] Test saving script locally
- [x] Check unit tests
- [x] Update docs (post merge/release) jdx/mise-docs#46

Resolves the following issues:

- `source` will eventually be deprecated
  - `export def main` makes `mise` the default import from `use mise.nu`
- `mise.nu` mistakenly overrides user `$env.config.hooks.display_output` (others too)
  - proposed nushell/nu_scripts#787 for built-in solution to `$env` collisions
  - `add-hook` workaround explained below
- `PATH` configuration more resilient in case it's a list
  - using `std path add` API (still undocumented nushell/nushell.github.io#1210)

```nushell
# usage
add-hook env_change.PWD {
  condition: { "MISE_SHELL" in $env }
  code: { mise_hook }
}

def --env add-hook [field hook] {
  # prepend 'hooks.' to the cell path enabling deep updates
  let field = $field | split row . | prepend hooks | into cell-path

  # get current hooks or empty list
  let hooks = $env.config | get --ignore-errors $field | default []

  # gracefully append new hook without erroring on missing cell paths
  $env.config = ($env.config | upsert $field ($hooks ++ $hook))
}
```

It would facilitate maintenance if the script could be extracted with `include_bytes!` and use vanilla `str::replace` instead of `formatdoc!` to avoid escaping braces.


texastoland:
@jdx Any feedback to progress?

toadslop:
Nice improvement!

src/shell/nushell.rs
```diff
@@ -96,8 +95,8 @@ impl Shell for Nushell {
         EnvOp::Set { key: &k, val: &v }.to_string()
     }
 
-    fn prepend_env(&self, k: &str, v: &str) -> String {
-        format!("$env.{k} = ($env.{k} | prepend '{v}')\n")
+    fn prepend_env(&self, _k: &str, v: &str) -> String {
+        format!("use std; std path add '{v}'\n")
```
jdx: since this is undocumented I'm a little uncomfortable adopting this since there might be a breaking change around it. I think I'd prefer the old logic unless there is a benefit to this I'm not aware of.

src/shell/nushell.rs
```diff
@@ -26,31 +26,30 @@ impl Shell for Nushell {
...
           }}
-            
+
           def "parse vars" [] {{
             $in | lines | parse "{{op}},{{name}},{{value}}"
           }}
-            
-          def --wrapped mise [command?: string, --help, ...rest: string] {{
+
+          export def --wrapped main [command?: string, --help, ...rest: string] {{
```
jdx, line 50: why are we renaming this? are functions global in nushell? if so does that mean it would conflict with something else named `main`? I don't know much about nushell though so I may very well be mistaken here.

src/shell/nushell.rs
```diff
@@ -26,31 +26,30 @@ impl Shell for Nushell {
...
+              condition: {{ "MISE_SHELL" in $env }}
+              code: {{ mise_hook }}
+            }}
+            add-hook env_change.PWD {{
+              condition: {{ "MISE_SHELL" in $env }}
+              code: {{ mise_hook }}
+            }}
+          }}
+
+          def --env add-hook [field hook] {{
```
jdx, line 40: similar to below, if these are exporting global functions I wouldn't want to pollute that namespace and face potential conflicts with another function named "add-hook"

jdx:
thanks for the contribution, and sorry for the delay, I have a lot on my plate right now. That said, I think this generally looks good, I had some clarifying questions and there is a test failure probably related to snapshots that need to be updated.

src/shell/nushell.rs
texastoland: (In theory) the `std` namespace is stable. It's undocumented because of the file it's in (the community needs a documentation generator). Apart from DX it mostly unifies OS differences which you're already doing. Will revert 👌🏼

src/shell/nushell.rs
texastoland: Replying in main comments so it doesn't get hidden after being resolved 👇🏼

src/shell/nushell.rs
texastoland: Same reply below 👇🏼

src/shell/nushell.rs
jdx: I think this is the only modification I'd like to see. It's a good change, and I'm sure it probably will work fine, but I would hate for things to just quit working at some point.

texastoland:
> sorry for the delay, I have a lot on my plate right now.

No worries I just didn't want to leave the impression this was a drive-by PR!

> > ```diff
> > -          def --wrapped mise [command?: string, --help, ...rest: string] {{
> > +          export def --wrapped main [command?: string, --help, ...rest: string] {{
> > ```

> are functions global in nushell? if so does that mean it would conflict with something else named `main`?

> > ```diff
> > +          def --env add-hook [field hook] {{
> > ```

> similar to [above], if these are exporting global functions I wouldn't want to pollute that namespace and face potential conflicts with another function named "add-hook"

My apology this deserved more documentation! Writing it up outside review comments so it doesn't get lost for future maintenance. Nu originally used `source mise.nu` similar to other shells. As it evolved they adopted first-class modules like other programming environments and will ultimately deprecate the `source` builtin.

`use mise.nu` can only impact the target file in the following ways:

1. Exclusively definitions (like `main`) marked `export` (like `pub` in Rust) is importable.
2. Any block in the source file wrapped in `export-env` can modify the current `$env` (shared namespace like environment variables).
3. Any command annotated `def --env` (like `add-hook`) can also modify the current `$env`.

Files are modules and commands named `main` are their default export. Therefore `use mise` would be similar to `use mise::{main as mise}` in Rust.

> It would facilitate maintenance if the script could be extracted with `include_bytes!` and use vanilla `str::replace` instead of `formatdoc!` to avoid escaping braces.

I also wanted to add types (for integration like `help` and completions) but it wouldn't be maintainable inlined in Rust code. Separate PR?

jdx:
thanks for the explanation!

jdx:
> I also wanted to add types to mise (for integration like help and completions) but it wouldn't be maintainable inlined in Rust code. Separate PR?

yeah I think that's quite a bit beyond "refactor"

src/shell/nushell.rs
texastoland: Dropped the relevant commit ✂️

texastoland:
Requested change complete and added related issue to the description :shipit:

> yeah I think that's quite a bit beyond "refactor"

Would you be open to such a change? Extract other scripts too or only Nu?

jdx:
> Would you be open to such a change? Extract other scripts too or only Nu? Happy to continue on Discord 💬

possibly... though it may conflict with some of the work I'm doing on [usage](https://github.com/jdx/usage) so I'm not sure right now

jdx (approved):

texastoland:
Then I'll start with a minimal draft and continue if it plays nice with your usage ✌🏼

texastoland:
PS ready to merge I believe and linked a docs PR when it's released :shipit:

texastoland:
I think I fixed the snapshot but the test says `This workflow is awaiting approval from a maintainer`.

Issue #1769, opened 2024-03-12
Watch env._.source files


nicolas-geniteau:
**Describe the bug**
Mise watches file in `env._.file` but not in `env._.source`. If we change content of one of  `source`, environment is not reloaded.

**To Reproduce**

With `mise.toml`:
```toml
[env]
_.source = ".env"
```
```bash
echo $FOO
# No output
```

```bash
echo "export FOO=bar" > .env
echo $FOO
# No output
```

**Expected behavior**

```bash
echo "export FOO=bar" > .env
echo $FOO
bar
```

PR #1770, opened 2024-03-12
fix: watch env._.source files

nicolas-geniteau:
Fix https://github.com/jdx/mise/issues/1769


nicolas-geniteau:
I've fixed tests, it didn't make sense to watch the `env._._path`.

@jdx , could you take another look?

Thanks !

PR #1845, opened 2024-03-30
feat(task): extend mise tasks output

### LINKED ISSUE #1623: Extend `mise tasks` output
Since all tasks share a global scope, it can get a little messy in the tasks list. 
I would suggest a way to sort/group the tasks by project_root or config file with a --group flag, for example.

Additionally, it would be nice to see aliases of tasks there too.

roele:
Adds the following options to `mise tasks ls` to enable output of all columns and sorting.
```
--full - Show all columns
--sort <COLUMN> - Sort by column. Default is name. [possible values: name, alias, description, source]
--sort-order <SORT_ORDER> - Sort order. Default is asc. [possible values: asc, desc]
```
Resolves #1623


jdx:
Not sure I like `--full`, maybe `-x,--extended`? I'm wondering if there are any conventions out there

jdx:
As a random aside, I'm not totally happy with our table library or how we use it in mise. I would like to refactor all the table code and possibly extract it like we did for demand.

It's not _terrible_ but I would like a simpler interface. I'm wondering if we could find a way to have nice conventions like you did for sorting here on the front-end that cleanly integrate with table output so we could use it everywhere in mise.

roele:
Agree, `--full` sounds odd, maybe `-e` or `-x` for `--extended` is better.

Regarding the output, most tools nowadays provide an `--output <FORMAT>` option which lets you even choose the format like `json`, `yaml` or `table`. Not sure if that makes sense in context of `mise` though.

jdx:
let's go with `-x,--extended`. I feel like I've seen that in other CLIs. We do have json output in a few commands, `mise ls` is one.

PR #1851, opened 2024-03-31
fix: stuck mise with flag --verbose

### LINKED ISSUE #1830: Stuck Mise with flag --verbose
**Describe the bug**

If you add an environment variable in the `env` table (in the .toml file), the value of which is computed through the [command](https://mise.jdx.dev/templates.html), then when you call any `mise` command with the `--verbose` flag, everything hangs on this line:
```bash
[DEBUG] ARGS: /Users/ernest0n/.local/bin/mise install --verbose
```

If you remove the `--verbose` flag, there are no problems, the values of computed variables are determined correctly.

**To Reproduce**

1. Add a line to the tool file:
```toml
[env]
MY_VARIABLE = "{{ exec(command:'echo 123') }}"
```
2. Run `mise install --verbose`

**Expected behavior**

Executing the command does not freeze

**`mise doctor` output**

```text
version: 2024.3.1 macos-arm64 (0d95fde 2024-03-04)
activated: yes
shims_on_path: no

build_info: 
  Target: aarch64-apple-darwin
  Built: Mon, 4 Mar 2024 22:02:06 +0000
  Rust Version: rustc 1.76.0 (07dca489a 2024-02-04)
  Profile: release

shell: 
  /bin/zsh
  zsh 5.9 (x86_64-apple-darwin22.0)

dirs: 
  data: ~/.local/share/mise
  config: ~/.config/mise
  cache: ~/Library/Caches/mise
  state: ~/.local/state/mise

config_files: 
  ~/.config/mise/config.toml
  ~/Projects/example/Gemfile
  ~/Projects/example/.mise/config.toml

plugins: 
  bun        (core)
  bundler    https://github.com/jonathanmorley/asdf-bundler.git#a461ac4
  deno       (core)
  go         (core)
  java       (core)
  node       (core)
  python     (core)
  ruby       (core)
  swiftgen   https://github.com/younke/asdf-swiftgen.git#072ce84
  swiftlint  https://github.com/klundberg/asdf-swiftlint.git#79f2c6e
  tuist      https://github.com/asdf-community/asdf-tuist.git#6bfab6d
  xcodes     https://github.com/younke/asdf-xcodes.git#9f037bf

toolset: 
  tuist@4.6.0       
  swiftlint@0.54.0  
  swiftgen@6.6.3    
  ruby@3.2.2        
  xcodes@1.4.1      
  bundler@2.4.13    

env_vars: 
  MISE_SHELL=zsh

settings: 
  activate_aggressive = true
  all_compile = false
  always_keep_download = false
  always_keep_install = false
  asdf_compat = false
  cargo_binstall = true
  color = true
  disable_tools = []
  experimental = false
  go_set_gopath = false
  go_set_goroot = true
  go_skip_checksum = false
  jobs = 4
  legacy_version_file = true
  node_compile = false
  not_found_auto_install = true
  paranoid = false
  python_compile = false
  python_default_packages_file = "/Users/ernest0n/.default-python-packages"
  raw = false
  trusted_config_paths = []
  quiet = false
  verbose = false
  yes = true
  ci = false
  debug = false
  trace = false
  log_level = "info"

  [status]
  show_env = false
  show_tools = false
```

**Additional context**

The reason why I need to use computed env variables:
I want to install ruby by explicitly passing the path to openssl installed via brew.

To do this I need to set the value of the `RUBY_CONFIGURE_OPTS` variable (using ruby-build) in the value of which I need to insert the result of the command `brew --prefix openssl@3`

roele:
@jdx This is an odd one, i would expect that the `env()` call in the Debug implementation of Config would already initialize the env but for some reason this leads to described behaviour. If i call `env()` explicitly before the `debug!` macro it works as expected. Any idea why that might be?

Resolves #1830


jdx:
it looks like this broke the python test for some reason. I tried running it again but it failed.

roele:
It seems the following code (`env()` and `path_dirs()`)  from `impl Debug for Config` causes a deadlock. Need to dig deeper to figure out why exactly...

```rust
if let Ok(env) = self.env() {
    if !env.is_empty() {
        s.field("Env", &env);
        // s.field("Env Sources", &self.env_sources);
    }
}
let path_dirs = self.path_dirs().cloned().unwrap_or_default();
if !path_dirs.is_empty() {
    s.field("Path Dirs", &path_dirs);
}
```

roele:
So far what i figured out is that `OneCell::get_or_try_init` is no-reentrant and might cause a deadlock if called recursively. Reworking the code to initialize `env` seems to mitigate that issue but causes another one related to Tera template resolution using exec. 

Locally i could only reproduce this once, where the template resolution (e.g. `{{exec(command='echo \"test\"')}}`) would execute (print out the command) in an endless loop, after touching the TOML file it was gone and non-reproducible though.

roele:
@jdx I assume the following code resembles a global singleton pattern in Rust?

```rust
pub fn try_get() -> Result<Arc<Self>> {
    if let Some(config) = &*CONFIG.read().unwrap() {
        return Ok(config.clone());
    }
    let config = Arc::new(Self::load()?);
    *CONFIG.write().unwrap() = Some(config.clone());
    Ok(config)
}
```

It seems that this does not work as intended when the `EnvDirective::PythonVenv` does call `Config::get()`. So `Config` uses `EnvDirective` on initialization (`load_env()`) to resolve the environment but also relies on `Config` to create a `venv`. This causes the reentrant behaviour on `OnceCell` and when mitigated a stackoverflow due to endless `load_env()` calls.

jdx:
> @jdx I assume the following code resembles a global singleton pattern in Rust?

yeah, I'm not using `Lazy` here because in the unit tests I need the ability to clear the value in a single process. It seems I recognized this was an issue when I wrote it: https://github.com/jdx/mise/blob/a8ac6cdca6d3322b6d406fa32d27c89ec488a9ee/src/config/env_directive.rs#L167

I'm not sure what to do here. We have a high-level paradox: env loading happens before we load the tools, but we need to load the tools so we know which python version to load.

roele:
I also don't see a straight forward solution right now. 

What could be done as a workaround is simply comment out the entire `debug!` macro or the parts (`Env`, `Path Dirs`) of the `impl Debug for Config` causing the deadlock.

jdx:
converted to draft until one of us figures out a better solution here

jdx:
where is this debug call? could we just only show the env if it is already loaded?

roele:
The `debug!` call causing this is at https://github.com/jdx/mise/blob/main/src/config/mod.rs#L94

jdx:
I'm thinking we change this conditional: https://github.com/jdx/mise/blob/main/src/config/mod.rs#L606

instead of calling `.env()` we call `self.env_with_sources.get()`

jdx:
I fixed this issue today

Issue #1836, opened 2024-03-25
mise ignores the environments `CLICOLOR_FORCE`


yggdr:
**Describe the bug**
Mise's documentation claims to respect the `CLICOLOR` and `CLICOLOR_FORCE` environment variables,  and `FORCE_COLOR` according to code comments. But with the color setting set to false, it never outputs color.

**To Reproduce**

```
with __xonsh__.env.swap({'FORCE_COLOR': '1', 'CLICOLOR_FORCE': '1', 'CLICOLOR': '1'}):
......     ~/local/bin/mise ls
......     env | egrep "FORCE_COLOR|CLICOLOR"
......     echo $FORCE_COLOR $CLICOLOR $CLICOLOR_FORCE
......     
Plugin       Version  Config Source              Requested
atlas        0.20.0                                       
cidr-merger  1.1.3                                        
go           1.22.1                                       
python       2.7.18                                       
python       3.9.18                                       
python       3.11.8   MISE_PYTHON_VERSION=3.11.8 3.11.8   
zig          0.11.0                                       
FORCE_COLOR=1
CLICOLOR_FORCE=1
CLICOLOR=1
1 1 1
```

Goes similarly in other shells.

**Expected behavior**
`CLICOLOR_FORCE` should make mise output color to stdout no matter what.

**`mise doctor` output**

```text
version: 2024.3.9 freebsd-x64 (5186f29 2024-03-25)
activated: yes
shims_on_path: no

build_info: 
  Target: x86_64-unknown-freebsd
  Features: DEFAULT, NATIVE_TLS
  Built: Mon, 25 Mar 2024 18:45:24 +0000
  Rust Version: rustc 1.76.0 (07dca489a 2024-02-04) (built from a source tarball)
  Profile: release

shell: 
  xonsh
  xonsh/0.15.1

dirs: 
  data: ~/.local/share/mise
  config: ~/.config/mise
  cache: ~/.cache/mise
  state: ~/.local/state/mise
  shims: ~/.local/share/mise/shims

config_files: 
  ~/.config/mise/config.toml

plugins: 
  atlas        https://github.com/pbr0ck3r/asdf-atlas.git#a2ed8f1
  bun          (core)
  cidr-merger  https://github.com/ORCID/asdf-cidr-merger.git#5522711
  deno         (core)
  go           (core)
  java         (core)
  node         (core)
  python       (core)
  ruby         (core)
  rust         https://github.com/code-lever/asdf-rust.git#95acf4f
[...]

PR #1862, opened 2024-04-05
fix: incorrect stdout for mise shell function in xonsh

yggdr:
The old implementation had issues with piping the result since stdout/stderr was not acutally captured in the shell but directly printed to the terminal.
The fix here unfortunately effectivly turns off colors in mise, as mise seems to not honor CLICOLOR_FORCE et. al. environment variables as per #1836


jdx:
I think it'd be better if we just found out why CLICOLOR_FORCE isn't working and fixed that directly

yggdr:
The coloring problem is only a side-effect of this fix, so it should imho have no bearing on whether this fix should be accepted or not. The fix addresses stdio within the xonsh shell which I handled incorrectly, thereby breaking pipe-ability. You can read about that here: https://github.com/xonsh/xonsh/discussions/5316

booniepepper:
I'd like to be able to fix the xonsh piping issue without regressing on color output.

yggdr:
I'd rather have a functionally working shell incl. pipeability than colours, nice though they may be. That the colours work at all in xonsh with the current code is a mere accident and side effect of my buggy original implementation.

jdx:
closing due to inactivity

PR #1869, opened 2024-04-06
fix(uninstall): fix uninstall completions

### LINKED ISSUE #1861: `mise remove tool@` completion suggests all available (not installed) versions
**Describe the bug**

`mise remove tool@` completion suggests all available versions of the tool, no matter if they're installed.

**To Reproduce**
Steps to reproduce the behavior.

```
mise install go@1.22.2
mise remove go@<TAB>
```

**Expected behavior**

Only installed versions of the tool would be suggested as completions.

**`mise doctor` output**

```text
version: 2024.4.0 linux-x64 (bcd3592 2024-04-02)
activated: yes
shims_on_path: no

build_info: 
  Built: Tue, 2 Apr 2024 01:29:57 +0000
  Rust Version: rustc 1.77.1 (7cf61ebde 2024-03-27)
  Profile: release

shell: 
  /bin/bash
  GNU bash, version 5.2.15(1)-release (x86_64-pc-linux-gnu)
  Copyright (C) 2022 Free Software Foundation, Inc.
  License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

  This is free software; you are free to change and redistribute it.
  There is NO WARRANTY, to the extent permitted by law.

dirs: 
  data: ~/.local/share/mise
  config: ~/.config/mise
  cache: ~/.cache/mise
  state: ~/.local/state/mise

config_files: 
  ~/.config/mise/config.toml

plugins: 
  bun     (core)
  deno    (core)
  go      (core)
  java    (core)
  node    (core)
  python  (core)
  ruby    (core)
  usage   https://github.com/jdx/mise-usage.git#fe3888a

toolset: 
  go@1.22.2     
  usage@0.1.17  

env_vars: 
  MISE_USE_TOML=true
  MISE_GO_SET_GOBIN=false
  MISE_SHELL=bash

settings: 
  activate_aggressive = false
  all_compile = false
  always_keep_download = false
  always_keep_install = false
  asdf_compat = false
  cargo_binstall = true
  color = true
  disable_tools = []
  experimental = false
  go_set_gobin = false
  go_set_gopath = false
  go_set_goroot = true
  go_skip_checksum = false
  jobs = 4
  legacy_version_file = true
  node_compile = false
  not_found_auto_install = true
  paranoid = false
  python_compile = false
  python_default_packages_file = "/home/scop/.default-python-packages"
  raw = false
  trusted_config_paths = []
  quiet = false
  verbose = false
  yes = false
  ci = false
  debug = false
  trace = false
  log_level = "info"

  [status]
  show_env = false
  show_tools = false

No problems found
```

**Additional context**
N/A

jdx:
Fixes #1861

### DIFF
--- a/src/assets/mise-extra.usage.kdl
+++ b/src/assets/mise-extra.usage.kdl
@@ -27,3 +27,28 @@ case $cur in
     ;;
 esac
 "#
+
+complete "installed_tool@version" run=r#"
+cur="{{words[CURRENT]}}"
+case $cur in
+  *@*)
+    tool="$(echo "$cur" | cut -d'@' -f1)"
+    prefix="$(echo "$cur" | cut -d'@' -f2)"
+
+    if [ ! -z "$prefix" ]; then
+      prefix="--prefix $prefix"
+    fi
+    versions=$(mise ls --installed $tool $prefix | awk '{print $2}' | sed '1!G;h;$!d')
+
+    for version in $versions; do
+      echo "$tool@$version"
+    done
+    ;;
+  *)
+    plugins=$(mise plugins --core --user)
+    for plugin in $plugins; do
+      echo "$plugin@"
+    done
+    ;;
+esac
+"#
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -1471,17 +1471,17 @@ Examples:
     $ mise trust
 ```
 
-## `mise uninstall [OPTIONS] [TOOL@VERSION]...`
+## `mise uninstall [OPTIONS] [INSTALLED_TOOL@VERSION]...`
 
 **Aliases:** `remove, rm`
 
 ```text
 Removes runtime versions
 
-Usage: uninstall [OPTIONS] [TOOL@VERSION]...
+Usage: uninstall [OPTIONS] [INSTALLED_TOOL@VERSION]...
 
 Arguments:
-  [TOOL@VERSION]...
+  [INSTALLED_TOOL@VERSION]...
           Tool(s) to remove
 
 Options:
--- a/mise.usage.kdl
+++ b/mise.usage.kdl
@@ -1059,7 +1059,7 @@ cmd "uninstall" help="Removes runtime versions" {
 "
     flag "-a --all" help="Delete all installed versions"
     flag "-n --dry-run" help="Do not actually delete anything"
-    arg "[TOOL@VERSION]..." help="Tool(s) to remove" var=true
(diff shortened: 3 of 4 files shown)


src/assets/mise-extra.usage.kdl
```diff
@@ -27,3 +27,28 @@ case $cur in
...
+cur="{{words[CURRENT]}}"
+case $cur in
+  *@*)
+    tool="$(echo "$cur" | cut -d'@' -f1)"
+    prefix="$(echo "$cur" | cut -d'@' -f2)"
+
+    if [ ! -z "$prefix" ]; then
+      prefix="--prefix $prefix"
+    fi
+    versions=$(mise ls --installed $tool $prefix | awk '{print $2}' | sed '1!G;h;$!d')
```
roele, line 41: Was also playing around with this. Does this work? `ls` returns another format than `ls-remote`. Also it potentially list missing versions.

My approach would have been 
```rust
complete "tool@version" run=r#"
prev="{{words[PREV]}}"
cur="{{words[CURRENT]}}"
declare -a rm=("uninstall" "remove" "rm")
case $cur in
  *@*)
    tool="$(echo "$cur" | cut -d'@' -f1)"
    prefix="$(echo "$cur" | cut -d'@' -f2)"

    if [[ " ${rm[@]} " =~ " ${prev} " ]]; then
      versions=$(mise ls --no-header $tool | grep -v '(missing)' | awk '{print $2}' | sed '1!G;h;$!d')
    else
      versions=$(mise ls-remote $tool $prefix | sed '1!G;h;$!d')
    fi

    for version in $versions; do
      echo "$tool@$version"
    done
    ;;
  *)
    if [[ " ${rm[@]} " =~ " ${prev} " ]]; then
      plugins=$(mise plugins --core --user)
    else
      plugins=$(mise plugins --all)
    fi

    for plugin in $plugins; do
      echo "$plugin@"
    done
    ;;
esac
"#
```

jdx: it would work, but I'd be hesitant to hardcode the aliases in the extras file. One problem with my solution is it shows plugins that have no installed versions but I feel it's at least an improvement.

roele: ah i see, the `--installed` flag filters the missing already. agree that hardcoding aliases is indeed not nice.

jdx: another note is you never need to pass`--no-header`, the header is omitted when it isn't a tty

PR #1873, opened 2024-04-08
Explorations into task dir issues

jatinn:
Been using mise for tool management and had started trying to use `tasks` but ran into some issues when trying to run from nested dirs.

Have a minimal test setup with 3 cases and was able to figure out some changes for 2 of the cases.

dir setup and all `mise tasks run` commands should be executed from within the `runner` dir
```
project
├── .mise.toml
├── scripts
│  └── hello.sh
└── runner
```

.mise.toml contents
```
[settings]
experimental = true

[tasks.case1]
# should print the project root dir not cwd
run = "pwd"

[tasks.case2]
run = "ls"
dir = "{{config_root}}/scripts"

[tasks.case3]
file = "scripts/hello.sh"
```

Leaving this as a draft pr as not very familiar with rust or the codebase so there are likely better ways of fixing these.
Hope this helps make it easier for folks to investigate further.

case1 successfully executes without any changes however it seems to use the current directory instead of the project root as mentioned in the [documentation](https://mise.jdx.dev/tasks/toml-tasks.html).

root cause for case2 seems to be that `dir` is not rendered using `tera` when loading tasks from the toml config (it works for script tasks) so the path substitutions do not happen within the dir path.

Believe for case3 it is something similar where the project root is not being added as a prefix to the file path for the script, however wasn't able to find or confirm if that is the case. Had tried a variation of case3 with specifying the `dir = "{{config_root}}"` to check if the `case2` fix would have helped with no luck.


jatinn:
@jdx are any of these approaches to the fix worth cleaning up?

jdx:
Sorry just haven't found a time to review yet, I'll give it a pass soon

PR #1885, opened 2024-04-10
feat: add option to list installed (backend) binaries

### LINKED ISSUE #1879: Add option to list installed (backend) binaries
Currently (backend) binaries can be installed via `mise use -g npm:prettier` but `mise ls` does not show them unless they're in a config or you specify the exact name like `mise ls npm:prettier`. This makes it hard to see what's there and remove certain binaries or versions. 

It would be helpful to show (all) installed binaries with versions (by default, via optional flag or by using a backend prefix e.g `npm:`).

roele:
Fixes #1879

### DIFF
--- a/src/cli/args/forge_arg.rs
+++ b/src/cli/args/forge_arg.rs
@@ -47,6 +47,11 @@ impl ForgeArg {
             downloads_path: dirs::DOWNLOADS.join(&pathname),
         }
     }
+
+    pub fn from_pathname(name: &str) -> Self {
+        let fa_name = name.replacen('-', ":", 1);
+        ForgeArg::from_str(&fa_name).unwrap()
+    }
 }
 
 impl Display for ForgeArg {
--- a/src/forge/mod.rs
+++ b/src/forge/mod.rs
@@ -9,6 +9,7 @@ use clap::Command;
 use console::style;
 use eyre::WrapErr;
 use itertools::Itertools;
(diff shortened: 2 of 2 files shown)


src/forge/mod.rs
```diff
@@ -62,6 +64,22 @@ fn load_forges() -> ForgeMap {
     plugins
 }
 
+fn list_installed_forges() -> eyre::Result<ForgeList> {
+    Ok(file::dir_subdirs(&dirs::INSTALLS)?
+        .into_par_iter()
+        .map(|name| {
+            let fa = ForgeArg::from_pathname(name.as_str());
+            match fa.forge_type {
+                ForgeType::Asdf => Arc::new(ExternalPlugin::new(name)) as AForge,
```
roele, line 73: I wonder if there is a way to ignore `Asdf` instead of filtering it out afterwards.

src/cli/args/forge_arg.rs
```diff
@@ -47,6 +47,11 @@ impl ForgeArg {
             downloads_path: dirs::DOWNLOADS.join(&pathname),
         }
     }
+
+    pub fn from_pathname(name: &str) -> Self {
+        let fa_name = name.replacen("-", ":", 1);
```
roele: I'm just assuming here that the first `-` is the separator for the forge type.

PR #1889, opened 2024-04-11
fix: mise panics if prefix: is used on certain core plugins

### LINKED ISSUE #1888: mise panics if prefix: is used on certain core plugins
**Describe the bug**
When using `prefix:` on certain plugins (bun, deno, java, ruby,..) the installation panics with following error.

**To Reproduce**
`mise i java@prefix:sapmachine-17`

**Expected behavior**
Installation is successful with latest version matching the prefix.

**Additional context**
Most likely related to following assertion used in the mentioned plugins.

```rust
 fn install_version_impl(&self, ctx: &InstallContext) -> Result<()> {
 ...
    assert!(matches!(
        &ctx.tv.request,
        ToolVersionRequest::Version { .. }
    ));
...
```

roele:
Fixes #1888

### DIFF
--- a/src/plugins/core/bun.rs
+++ b/src/plugins/core/bun.rs
@@ -112,15 +112,8 @@ impl Forge for BunPlugin {
         Ok(vec![".bun-version".into()])
     }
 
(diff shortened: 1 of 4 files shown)


jdx:
I switched it over to use the [contracts](https://docs.rs/contracts/latest/contracts/) crate to make it cleaner

Issue #1904, opened 2024-04-15
Go backend naming inconsistency (in `mise ls` and `mise prune`)


dhanak:
**Describe the bug**
I have some packages installed with the go backend, e.g., `go:github.com/tomnomnom/gron`. It is listed as `go:github.com/tomnomnom/gron = 'latest'` in my `mise/config.toml`, it installs as expected and works fine. However, when I list the installed packages, I see this:

```
go:github.com-tomnomnom-gron       latest                                                                      
go:github.com/tomnomnom/gron       latest         ~/.config/mise/config.toml                      latest
```

In other words, it shows up as I entered it, but also with dashes instead of slashes, and the latter is shown as missing. Furthermore, if I run `mise prune`, it asks me if I want to remove the dashed "variant". Answering yes removes the `gron` installation, and now the slashed version shows up as missing. Running `mise install` again restarts the cycle.

**To Reproduce**

```
$ mise use go:github.com/tomnomnom/gron
$ mise ls
$ mise prune
```

**Expected behavior**

Only `go:github.com/tomnomnom/gron` is shown in the listing.

**`mise doctor` output**

```text
version: 2024.4.4 linux-x64 (37b44e2 2024-04-11)
activated: yes
shims_on_path: no

build_info: 
  Target: x86_64-unknown-linux-gnu
  Features: DEFAULT, NATIVE_TLS, OPENSSL
  Built: Thu, 11 Apr 2024 17:14:58 +0000
  Rust Version: rustc 1.77.2 (25ef9e3d8 2024-04-09)
  Profile: release

shell: 
  /bin/zsh
  zsh 5.9 (x86_64-debian-linux-gnu)

dirs: 
  data: ~/.local/share/mise
  config: ~/.config/mise
  cache: ~/.cache/mise
  state: ~/.local/state/mise
  shims: ~/.local/share/mise/shims

config_files: 
  ~/.config/mise/config.toml

plugins: 
  bun                                (core)
  cargo:mdcat                        (core)
  cargo:ouch                         (core)
  cmake                              https://github.com/srivathsanmurali/asdf-cmake.git#3811ad7
  deno                               (core)
  elixir                             https://github.com/asdf-vm/asdf-elixir.git#a4c42e1
  erlang                             (core)
  eza                                https://github.com/lwiechec/asdf-eza.git#dedb2d1
  fd                                 https://gitlab.com/wt0f/asdf-fd.git#def644b
[...]

PR #1905, opened 2024-04-15
fix: go backend naming inconsistency (in mise ls and mise prune)

roele:
Mitigates #1904 which was introduced with #1885. 

- Cargo crates use kebab-case (or snake_case?) to they should not cause issues.
- Go packages cannot have dashes so reverting `-` to `/` fixes the issue.
- NPM backend is still problematic as scopes can have `-` and end with `/`, so reverting from for example `npm-@adobe-aio-cli` to `npm:@adobe/aio-cli` can only happen on best guess replacing first dash after the `@`.

### DIFF
--- a/src/cli/args/forge_arg.rs
+++ b/src/cli/args/forge_arg.rs
@@ -49,7 +49,18 @@ impl ForgeArg {
     }
 
     pub fn from_pathname(name: &str) -> Self {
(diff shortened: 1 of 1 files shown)


jdx:
I think this is fine for now but this should probably dispatch to the forge to figure out how to un-pathname itself

PR #1909, opened 2024-04-16
refactor: use a metadata file for forges

roele:
This is a follow up on #1905 attempting to move the "un-dirname" logic out of the `ForgeArg` into the `Forge` implementations.

Additionally it seems for `Go` while package names cannot have `-`, module names can and often do. So the logic for `Go` will also not work reliably. For example https://pkg.go.dev/search?q=http reveals all sorts of exceptions.

Using a `+` as separator would yield better results but i'm afraid thats not easy to change now is it?

### DIFF
--- a/src/forge/forge_meta.rs
+++ b/src/forge/forge_meta.rs
@@ -8,15 +8,16 @@ use super::ForgeType;
 pub struct ForgeMeta {
     pub id: String,
     pub name: String,
+    pub forge_type: String,
 }
 
-const FORGE_META_FILENAME: &str = ".mise.forge.toml";
+const FORGE_META_FILENAME: &str = ".mise.forge.json";
 
 impl ForgeMeta {
     pub fn read(dirname: &str) -> ForgeMeta {
         let meta_path = &dirs::INSTALLS.join(dirname).join(FORGE_META_FILENAME);
-        let toml = file::read_to_string(meta_path).unwrap_or_default();
-        toml::from_str(&toml).unwrap_or(Self::default_meta(dirname))
+        let json = file::read_to_string(meta_path).unwrap_or_default();
+        serde_json::from_str(&json).unwrap_or(Self::default_meta(dirname))
     }
 
     pub fn write(fa: &ForgeArg) -> eyre::Result<()> {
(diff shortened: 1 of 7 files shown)


jdx:
I think we need to modify our strategy. Maybe we need a metadata file to store the real name

roele:
Agreed, the path is not suitable and will most likely continue to cause problems no matter what separator we use.

roele:
Gave it a first shot but i'm not sure if we should use a structured file instead? Also the `upgrade` does not seem to work for backends but i'm not sure if this is related or is caused by something else yet.

src/forge/forge_meta.rs
```diff
@@ -0,0 +1,64 @@
...
+
+use super::ForgeType;
+
+#[derive(Debug, Default, Serialize, Deserialize)]
+pub struct ForgeMeta {
+    pub id: String,
+    pub name: String,
+}
+
+const FORGE_META_FILENAME: &str = ".mise.forge.toml";
```
jdx: can we use json? that's typically what I prefer for generated files. I have a dumb reason for this too, which is that `.mise.forge.toml` would be used as a mise config if `MISE_ENV=forge`. While I don't see that as _likely_, using json will also avoid that edge case.

src/forge/forge_meta.rs
```diff
@@ -0,0 +1,64 @@
...
+                ForgeMeta { id, name }
+            }
+            None => ForgeMeta {
+                id: id.to_string(),
+                name: id.to_string(),
+            },
+        }
+    }
+
+    fn name_for_type(name: String, forge_type: &str) -> String {
```
jdx, line 61: I'm inclined to just not even bother including this logic to keep things simpler, but I suppose it makes sense at least in the short term when people don't have the file

src/forge/forge_meta.rs
roele: i usually prefer JSON as well but have not seen any in context of mise. was not aware of the `MISE_ENV` use-case, very good point.

src/forge/forge_meta.rs
roele: was wondering about that too but tools without a meta file would not show up without a fallback to the current logic so i thought it would make the transition a bit smoother keeping it for now. maybe adding a comment to remove this later could help.

PR #1926, opened 2024-04-20
fix(exec): default to @latest version

zph:
Addresses a bug with `mise exec` where using an unversioned tool reference causes the underlying binary to not be installed on first use.

Instead we treat an unqualified tool as having tool@latest and handle that in the else clause.

Discord discussion/question:

```
What's the expected behavior for mise backends when the desired tool@version does not exist in a config file and is not installed and mise x tool@version -- ... is run? I assumed it to be like npx behavior of install if missing then run, but it isn't for most backends other than npm.

For example, if I lack cargo:eza in any of my config files nor installed and run mise x cargo:eza -- which -a eza , I see that it does not install cargo:eza + run the --command with eza install folder in the path. In that case the command fails.

But if I run mise x npm:prettier@3.1.0 -- which -a prettier in similar circumstances I see that prettier is available and executed.

The behavior I expect is the npm case and the cargo case is surprising and undesirable for me, because I'm thinking of mise x as being like npx where it will setup dependency if missing then run it.  I see that if I first mise install cargo:eza followed by mise x cargo:eza it works as expected.

Is the npm behavior your expected/desired behavior?

(this is in the context of debugging my PR for a pipx backend and realizing that mise x and mise install followed by mise x have different behavior in my e2e tests)

jdx: I think mise x with any tool on any backend should autoinstall it
```

Closes #1247

### DIFF
--- a/e2e/test_system
+++ b/e2e/test_system
@@ -9,4 +9,4 @@ system_node="$(which node)"
 mise_node="$(mise which node)"
 
 assert "mise x node@system nodejs@system -- which node" "$system_node"
-assert "mise x node -- which node" "$mise_node"
+assert "mise x -- which node" "$mise_node"
--- a/src/cli/exec.rs
+++ b/src/cli/exec.rs
@@ -51,7 +51,10 @@ pub struct Exec {
 impl Exec {
     pub fn run(self) -> Result<()> {
         let config = Config::try_get()?;
-        let mut ts = ToolsetBuilder::new().with_args(&self.tool).build(&config)?;
+        let mut ts = ToolsetBuilder::new()
+            .with_args(&self.tool)
+            .with_default_to_latest(true)
+            .build(&config)?;
         let opts = InstallOptions {
             force: false,
             jobs: self.jobs,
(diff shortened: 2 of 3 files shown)


jdx:
it seems I had been working on this myself a while ago but for some reason I couldn't get it to work correctly https://github.com/jdx/mise/pull/1247/files

zph:
Will check on test results a bit later.

> it seems I had been working on this myself a while ago but for some reason I couldn't get it to work correctly https://github.com/jdx/mise/pull/1247/files

I see why it's tough to change. I spent more time on it yesterday but didn't come to a good solution.

I'm going to focus on the pipx backend PR and the gh releases backend and then this is my next interest to fix.

jdx:
Yeah apologies for leading you down this path without warning you that it's a lot tougher than it appears but I completely forgot about these issues. I'll try to find some time this week myself to poke around with it and see if I can either come up with a solution or at least a sketch of one.

jdx:
glanced at this a bit and noticed I can't get the unit tests to pass locally either but wasn't able to figure out why. Only had a few minutes to spare so wasn't able to dive in further but hopefully one of us can track down what is going on in the unit tests. I suspect in some of the setup `mise x` is being called in a way that is causing this to not work properly.

jdx:
I don't love this solution since it makes things quite complex to reason about but I think it's the behavior we want

PR #1932, opened 2024-04-22
Spike on building out a releases backend called `bin`

zph:
(edited for rename from bin->ubi)

Allows for installing:
`ubi:goreleaser/goreleaser`
`ubi:goreleaser/goreleaser@1.16.2`
`ubi:https://github.com/goreleaser/goreleaser/releases/download/v1.16.2/goreleaser_Darwin_arm64.tar.gz`

Uses `houseabsolute/ubi` under the hood for initial version as a binary but should use it as a library for portability.


jdx:
we should call it `ubi` in case we want to use an alternative later

PR #1942, opened 2024-04-23
Ensure the e2e tests pass in the dev container

Adirelle:
This is an early, draft pull request.

As discussed on discord, its purpose is to have the e2e tests passing in the dev container, and also to use this container to run all the e2e tests in CI.

### DIFF
--- a/.mise.toml
+++ b/.mise.toml
@@ -112,7 +112,7 @@ run = 'docker build $root -f $root/packaging/dev/Dockerfile -t ghcr.io/jdx/mise:
 
 [tasks."docker:cargo"]
 description = "run cargo inside of development docker container"
-run = 'docker run -ti --rm -v $root:/mise -w /mise ghcr.io/jdx/mise:dev cargo'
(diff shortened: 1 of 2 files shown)


.mise.toml
```diff
@@ -112,7 +112,7 @@ run = 'docker build $root -f $root/packaging/dev/Dockerfile -t ghcr.io/jdx/mise:
 
 [tasks."docker:cargo"]
 description = "run cargo inside of development docker container"
-run = 'docker run -ti --rm -v $root:/mise -w /mise ghcr.io/jdx/mise:dev cargo'
+run = 'docker run -ti --rm -e "GITHUB_API_TOKEN=${GITHUB_API_TOKEN:-}" -v $root:/mise -w /mise ghcr.io/jdx/mise:dev cargo'
```
jdx, line 115: I may have merged too early, I think this would've done the same thing but cleaner: 

```
docker run -ti --rm -e GITHUB_API_TOKEN
```

https://docs.docker.com/reference/cli/docker/container/run/#env

PR #1948, opened 2024-04-24
refactor: converted just tasks in mise tasks.

### LINKED ISSUE #1567: convert just tasks to mise tasks
I've been working to convert everything in the [justfile](https://github.com/jdx/mise/blob/main/justfile) over to mise tasks but there are still a few tasks that need to be converted

Adirelle:
Resolves #1567


Adirelle:
The test worklow now installs the latest release of mise to run the test tasks, which in turn will build and run mise from source.

jdx:
nice cleanup! 👍

Issue #1710, opened 2024-02-25
Not generating mise completions despite usage CLI is installed


filipeltsilva:
Hi!

I migrated from rtx to mise and I'm trying to generate the completions through `mise completion fish` command, but I got this return:

```
# if "usage" is not installed show an error
if ! command -v usage &> /dev/null
    echo "Error: usage not found. This is required for completions to work in mise. https://usage.jdx.dev" >&2
    return 1
end

set _usage_spec_mise (mise usage | string collect)
complete -xc mise -a '(usage complete-word -s "$_usage_spec_mise" -- (commandline -cop) (commandline -t))
```

I installed the `usage` plugin through `mise use -g usage` command and the system detected its presence when I run the `command -v usage` above.

Here is my `mise doctor` log:

```
version: 2024.2.18 linux-x64 (4893e09 2024-02-24)
activated: yes
shims_on_path: no

build_info: 
  Target: x86_64-unknown-linux-musl
  Features: DEFAULT, NATIVE_TLS, OPENSSL
  Built: Sat, 24 Feb 2024 06:33:38 +0000
  Rust Version: rustc 1.76.0 (07dca489a 2024-02-04)
  Profile: release

shell: 
  /usr/bin/fish
  fish, version 3.7.0

dirs: 
  data: ~/.local/share/mise
  config: ~/.config/mise
  cache: ~/.cache/mise
  state: ~/.local/state/mise
  shims: ~/.local/share/mise/shims

config_files: 
  ~/.config/mise/config.toml

plugins: 
  bun         (core)
  deno        (core)
  erlang      (core)
  go          (core)
  java        (core)
  neovim      https://github.com/richin13/asdf-neovim.git#0b7b611
  node        (core)
  python      (core)
  ruby        (core)
  shellcheck  https://github.com/luizm/asdf-shellcheck.git#b160a52
  usage       https://github.com/jdx/mise-usage.git#fe3888a

toolset: 
  go@1.22.0         
  neovim@0.9.5      
  node@20.11.1      
  ruby@3.3.0        
  shellcheck@0.9.0  
  usage@0.1.9       

env_vars: 
  MISE_NODE_DEFAULT_PACKAGES_FILE=/home/filipelemos/.config/mise/.default-npm-packages
  MISE_SHELL=fish
  MISE_GO_DEFAULT_PACKAGES_FILE=/home/filipelemos/.config/mise/.default-go-packages
[...]

PR #1950, opened 2024-04-25
Use usage through mise automatically if it's not installed in fish shell

onnimonni:
Hey!

I tried to read the documentation but the installing the completions was not straightforward and I ended up seeing the same error as in: https://github.com/jdx/mise/issues/1710

```
Error: usage CLI not found. This is required for completions to work in mise.
See https://usage.jdx.dev for more information.
```

By enabling `usage` automatically through mise this problem goes away 💪 I think it's quite sensible default to run this automatically if `usage` can't be found?

If my installation steps below don't make sense and I should have installed `mise` in other way let me know and I will be happy to update the docs too. Thanks for creating this amazing piece of software 💪🙇

## Steps to reproduce the issue and how I installed `mise`

I installed fish in the following way to my MacBook:
```zsh
brew install mise
```

I then added following lines into my `~/.config/fish/config.fish` to enable mise automatically:

```fish
if type -q mise
  mise activate | source
end
```

I still wasn't able to get the mise autocompletions by pressing the tab so I then used following command:
```zsh
mise completion fish > ~/.config/fish/completions/mise.fish
```

Which gave me the original issue above which I was able to fix by using the `mise use -g usage` command.


jdx:
sorry for being late on this but this is a generated file so we can't just edit it. I also am not sure I like the idea of just auto-installing like that

PR #1958, opened 2024-04-26
fix(zig): can't install zig@master from v2024.4.6

### LINKED ISSUE #1956: Can't install `zig@master` from v2024.4.6
**Describe the bug**

```shell
❯ mise install zig@master --verbose
[DEBUG] ARGS: mise install zig@master --verbose
[DEBUG] Config {
    Config Files: [
        "~/.mise.toml",
        "~/.config/mise/config.toml",
    ],
}
[DEBUG] Toolset: dotnet-core@latest, lua@latest, luajit@latest, tig@latest, vim@latest, ruby@latest, neovim@nightly, node@20, powershell-core@7.4, python@3, zig@latest
downloading zig-linux-x86_64-master.tar.xz
[DEBUG] GET Downloading https://ziglang.org/download/master/zig-linux-x86_64-master.tar.xz to ~/.local/share/mise/downloads/zig/master/zig-linux-x86_64-master.tar.xz
[DEBUG] GET https://ziglang.org/download/master/zig-linux-x86_64-master.tar.xz
[DEBUG] starting new connection: https://ziglang.org/
[DEBUG] GET https://ziglang.org/download/master/zig-linux-x86_64-master.tar.xz 403 Forbidden
Error:
   0: HTTP status client error (403 Forbidden) for url (https://ziglang.org/download/master/zig-linux-x86_64-master.tar.xz)

Location:
   /home/mimikun/.cargo/registry/src/index.crates.io-6f17d22bba15001f/mise-2024.4.8/src/http.rs:73

Version:
   2024.4.8 linux-x64 (2024-04-23)

Suggestion: Run with --verbose or MISE_VERBOSE=1 for more information.

```

**To Reproduce**

1. Install mise and setup
2. Run `mise install zig@master` 

**Expected behavior**
Install success

**`mise doctor` output**

```text
❯ mise doctor
version: 2024.4.8 linux-x64 (2024-04-23)
activated: yes
shims_on_path: yes

build_info:
  Features: DEFAULT, NATIVE_TLS
  Built: Tue, 23 Apr 2024 10:31:55 +0000
  Rust Version: rustc 1.79.0-nightly (7f2fc33da 2024-04-22)
  Profile: release

shell:
  /usr/bin/fish
  fish, version 3.7.1

dirs:
  data: ~/.local/share/mise
  config: ~/.config/mise
  cache: ~/.cache/mise
  state: ~/.local/state/mise

config_files:
  ~/.config/mise/config.toml
  ~/.mise.toml

plugins:
  bun              (core)
  deno             (core)
  dotnet-core      https://github.com/emersonsoares/asdf-dotnet-core.git#4583e4b
  erlang           (core)
  go               (core)
  java             (core)
  lua              https://github.com/Stratus3D/asdf-lua.git#36fae6e
  luajit           https://github.com/smashedtoatoms/asdf-luaJIT.git#bbfe3cd
  neovim           https://github.com/richin13/asdf-neovim.git#0b7b611
  node             (core)
  powershell-core  https://github.com/mimikun/asdf-powershell-core.git#84d8930
  python           (core)
  ruby             (core)
  tig              https://github.com/koketani/asdf-tig.git#9ac08f3
  vim              https://github.com/mimikun/asdf-vim.git#5563975
  zig              (core)

toolset:
  dotnet-core@8.0.204
  lua@5.4.6
  luajit@2.0.5--2.4.4
  tig@2.5.9
  vim@9.1.0373
  ruby@3.3.1
  neovim@nightly
  node@20.12.2
  powershell-core@7.4.2
  python@3.12.3
  zig@0.12.0

env_vars:
  MISE_SHELL=fish
  MISE_FISH_AUTO_ACTIVATE=1
  MISE_EXPERIMENTAL=1
  MISE_NOT_FOUND_AUTO_INSTALL=false
  MISE_USE_VERSIONS_HOST=true
  MISE_GLOBAL_CONFIG_FILE=/home/mimikun/.config/mise/config.toml
  MISE_JOBS=4
  MISE_DATA_DIR=/home/mimikun/.local/share/mise
  MISE_CONFIG_DIR=/home/mimikun/.config/mise
  MISE_LEGACY_VERSION_FILE=1
  MISE_CACHE_DIR=/home/mimikun/.cache/mise
  MISE_LOG_FILE=/home/mimikun/.cache/mise/mise.log
  MISE_YES=1
  MISE_ASDF_COMPAT=1

settings:
  activate_aggressive = false
  all_compile = false
  always_keep_download = false
  always_keep_install = false
  asdf_compat = true
  cargo_binstall = true
  color = true
  disable_tools = []
  experimental = true
  go_set_gopath = false
  go_set_goroot = true
  go_skip_checksum = false
  jobs = 4
  legacy_version_file = true
  node_compile = false
  not_found_auto_install = false
  paranoid = false
  python_compile = false
  python_default_packages_file = "/home/mimikun/.default-python-packages"
  raw = false
  trusted_config_paths = []
  quiet = false
  verbose = false
  yes = true
  ci = false
  debug = false
  trace = false
  log_level = "info"

  [status]
  show_env = false
  show_tools = false

No warnings found
No problems found
```

**Additional context**
None

roele:
Fixes #1956 by fetching the current version for `master` from `https://ziglang.org/download/index.json` on demand without caching (nightly build) and using the `builds` instead of `downloads` context.

cc/ @MustCodeAl

### DIFF
--- a/src/plugins/core/zig.rs
+++ b/src/plugins/core/zig.rs
@@ -40,7 +40,6 @@ impl ZigPlugin {
     }
 
     fn fetch_remote_versions(&self) -> Result<Vec<String>> {
-        // let json = HTTP_FETCH.get_text("https://ziglang.org/download/index.json")?;
         match self.core.fetch_remote_versions_from_mise() {
             Ok(Some(versions)) => return Ok(versions),
             Ok(None) => {}
@@ -59,7 +58,7 @@ impl ZigPlugin {
     }
 
     fn download(&self, tv: &ToolVersion, pr: &dyn SingleReport) -> Result<PathBuf> {
-        let url = if tv.version == "master" {
+        let url = if tv.version == "ref:master" {
             format!(
                 "https://ziglang.org/builds/zig-{}-{}-{}.tar.xz",
                 os(),
@@ -95,7 +94,7 @@ impl ZigPlugin {
                 "zig-{}-{}-{}",
                 os(),
                 arch(),
-                if ctx.tv.version == "master" {
+                if ctx.tv.version == "ref:master" {
                     self.get_master_version()?
                 } else {
                     ctx.tv.version.clone()
@@ -142,7 +141,7 @@ impl Forge for ZigPlugin {
     fn legacy_filenames(&self) -> Result<Vec<String>> {
         Ok(vec![".zig-version".into()])
     }
-    #[requires(matches ! (ctx.tv.request, ToolVersionRequest::Version { .. } | ToolVersionRequest::Prefix { .. }), "unsupported tool version request type")]
+    #[requires(matches ! (ctx.tv.request, ToolVersionRequest::Version { .. } | ToolVersionRequest::Prefix { .. } | ToolVersionRequest::Ref { .. }), "unsupported tool version request type")]
     fn install_version_impl(&self, ctx: &InstallContext) -> Result<()> {
         let tarball_path = self.download(&ctx.tv, ctx.pr.as_ref())?;
         self.install(ctx, &tarball_path)?;
--- a/src/plugins/core/zig.rs
+++ b/src/plugins/core/zig.rs
@@ -26,6 +26,7 @@ impl ZigPlugin {
         let core = CorePlugin::new("zig");
         Self { core }
     }
+
     fn zig_bin(&self, tv: &ToolVersion) -> PathBuf {
         tv.install_path().join("zig")
     }
@@ -57,14 +58,23 @@ impl ZigPlugin {
         Ok(versions)
     }
 
-    fn downlaod(&self, tv: &ToolVersion, pr: &dyn SingleReport) -> Result<PathBuf> {
-        let url = format!(
-            "https://ziglang.org/download/{}/zig-{}-{}-{}.tar.xz",
-            tv.version,
-            os(),
-            arch(),
-            tv.version
-        );
+    fn download(&self, tv: &ToolVersion, pr: &dyn SingleReport) -> Result<PathBuf> {
+        let url = if tv.version == "master" {
+            format!(
+                "https://ziglang.org/builds/zig-{}-{}-{}.tar.xz",
+                os(),
+                arch(),
+                self.get_master_version()?
+            )
+        } else {
+            format!(
+                "https://ziglang.org/download/{}/zig-{}-{}-{}.tar.xz",
+                tv.version,
+                os(),
+                arch(),
+                tv.version
+            )
+        };
 
         let filename = url.split('/').last().unwrap();
         let tarball_path = tv.download_path().join(filename);
@@ -74,15 +84,23 @@ impl ZigPlugin {
 
         Ok(tarball_path)
     }
+
     fn install(&self, ctx: &InstallContext, tarball_path: &Path) -> Result<()> {
         let filename = tarball_path.file_name().unwrap().to_string_lossy();
         ctx.pr.set_message(format!("installing {filename}"));
         file::remove_all(ctx.tv.install_path())?;
         untar_xy(tarball_path, &ctx.tv.download_path())?;
         file::rename(
-            ctx.tv
-                .download_path()
-                .join(format!("zig-{}-{}-{}", os(), arch(), ctx.tv.version)),
+            ctx.tv.download_path().join(format!(
+                "zig-{}-{}-{}",
+                os(),
+                arch(),
+                if ctx.tv.version == "master" {
+                    self.get_master_version()?
+                } else {
+                    ctx.tv.version.clone()
+                }
+            )),
             ctx.tv.install_path(),
         )?;
         file::create_dir_all(ctx.tv.install_path().join("bin"))?;
@@ -96,6 +114,16 @@ impl ZigPlugin {
 
     fn verify(&self, ctx: &InstallContext) -> Result<()> {
         self.test_zig(ctx)
+    }
+
+    fn get_master_version(&self) -> Result<String> {
+        let version_json: serde_json::Value =
+            HTTP_FETCH.json("https://ziglang.org/download/index.json")?;
+        let master_version = version_json
+            .pointer("/master/version")
+            .and_then(|v| v.as_str())
+            .ok_or_else(|| eyre::eyre!("Failed to get master version"))?;
+        Ok(master_version.to_string())
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/plugins/core/zig.rs
```diff
@@ -57,14 +58,23 @@ impl ZigPlugin {
...
-    fn downlaod(&self, tv: &ToolVersion, pr: &dyn SingleReport) -> Result<PathBuf> {
-        let url = format!(
-            "https://ziglang.org/download/{}/zig-{}-{}-{}.tar.xz",
-            tv.version,
-            os(),
-            arch(),
-            tv.version
-        );
+    fn download(&self, tv: &ToolVersion, pr: &dyn SingleReport) -> Result<PathBuf> {
+        let url = if tv.version == "master" {
```
jdx: "master" shouldn't be a valid version, it should use `ref:master` I think to make it clear that you want to use the tip of the repo if I'm understanding what this means

roele: Both the website https://ziglang.org/download/ and http://mise-versions.jdx.dev/zig have a `master` version which seems to be a binary build of the tip of the repo.

Wouldn't `ref:master` imply that you compile/build a certain branch/tag locally from source?

roele: Other option would be to filter out `master` from the http://mise-versions.jdx.dev/zig response and not allow bleeding-edge version installations. Not sure how common it is to use this, maybe someone from the Zig community can chime in here.

jdx: In case you didn't know, mise-versions gets its data from listing the versions with `mise ls-remote zig` so if we change mise to remove the version it will be removed from that API as well (eventually, it runs on a cron job).

I think we probably should, @booniepepper may be able to weigh in on zig generally. I suspect it probably _is_ common since in younger languages users are more likely to adopt into bleeding edge features.

booniepepper: Zig folks will absolutely need/want bleeding-edge versions. People who are developing software at the tip version of Zig will need to be able to know and display the version _("master" is not accurate enough to be useful in any way. It's like buying a brand-new car: it loses tremendous value almost immediately after installation)_

"master" from the Zig project perspective is not a version, the current "master" version as of posting is `0.13.0-dev.46+3648d7df1`

I'd recommend when we're parsing `https://ziglang.org/download/index.json` we extract the version for `master`.

```
{
  "master": {
    "version": "0.13.0-dev.46+3648d7df1",
    ...
```

Longer term it may be useful to keep a list of some trailing N build versions, but that list is not available through the json file we're parsing.

jdx: here's my question since it probably wasn't clear from earlier conversation. By convention in mise we typically use `ref:<sha>` for when we're building from a repo, e.g.: `mise install node@ref:master`. So `mise install zig@master` doesn't follow that because "master" is not a version.

So what I'm proposing is we ignore the "master" that's coming from zig's json and instead manually allow users to install with `mise install zig@ref:master`.

There are benefits beyond just following convention, one I can think of is `mise upgrade` which will behave differently if installed as `ref:master` over a "version" like `master`. In the case of `ref:` it might always upgrade because `ref:master` one day might not be the same as `ref:master` the next, but mise would assume `master` is always the same. (note this functionality isn't actually implemented yet, but I will get to it one day, there are probably other assumptions about ref: and not ref: versions that have similar expectations.

Anyways, with all of that context, my question is whether `ref:master` actually should use this json or if it should pull down the project repo and compile from scratch if you specify `ref:master`. I can thing of arguments on both sides here.

OR maybe option #3 which would be for mise to properly deal with "loose" versions like `master` which may point to different real versions. We might be able to do this today if we treated this as an alias to `0.13.0-dev.46+3648d7df1` that we update when fetching versions.

This is all pretty nitpicky stuff that doesn't have a lot of real impact today, but I suspect they will be important decisions in the future.

roele: Agree that `ref:master` would make much more sense in this case. As per docs it compiles from source so that's what i would expect as of now. Having said that compiling from source always comes with lots of complexity, dependencies and dealing with possible platform specifics etc. This would clearly have lots of advantages but not sure if its worth the effort doing this for every core plugin.

I am not in favour of option 3) as we should enforce/stick to semantic versioning as much as possible.

I guess (at least for now) i would prefer a more pragmatic approach like having a `0.13.0-dev.46+3648d7df1` version in the `ls-remote` output which can be installed as binary. For this to work we might need to keep a cache of the last 5-10 `0.13.0-dev.` versions and also evict those if `0.13.0` is released and move to `0.14.0-dev.`. Installing an older `-dev.` version should still be possible even if not shown in `ls-remote` by simply checking for the version to contain `-dev.` and use as is.

jdx: > Agree that ref:master would make much more sense in this case. As per docs it compiles from source so that's what i would expect as of now. Having said that compiling from source always comes with lots of complexity, dependencies and dealing with possible platform specifics etc. This would clearly have lots of advantages but not sure if its worth the effort doing this for every core plugin.

This is correct, and I don't think any core plugins currently support `ref:` installs though that's definitely something I consider a gap—we should support this for all of the core plugins eventually I think.

I think in the short term all we need to do is remove `master` from `mise ls-remote` and `mise install zig@ref:master` _should_ work with this PR—though of course I haven't tested that so I can't say for sure.

roele: I think adding support for this might prove problematic. How would we build from source for `java` with all the various different vendor builds?

I assume you refer to changing `master` to `ref:master` to install the pre-compiled binaries?
Removing `master` from `ls-remote` might require to remove the first block as we might have a chicken-egg issue with following code (unless http://mise-versions.jdx.dev sets `MISE_USE_VERSIONS_HOST`?) 
```rust
fn fetch_remote_versions(&self) -> Result<Vec<String>> {
    // let json = HTTP_FETCH.get_text("https://ziglang.org/download/index.json")?;
    match self.core.fetch_remote_versions_from_mise() {
        Ok(Some(versions)) => return Ok(versions),
        Ok(None) => {}
        Err(e) => warn!("failed to fetch remote versions: {}", e),
    }

    let releases: Vec<GithubRelease> =
        HTTP_FETCH.json("https://api.github.com/repos/ziglang/zig/releases?per_page=100")?;
    let versions = releases
        .into_iter()
        .map(|r| r.tag_name)
        .unique()
        .sorted_by_cached_key(|s| (Versioning::new(s), s.to_string()))
        .collect();
    Ok(versions)
}
```

jdx: ah yeah, java would be an exception to compiling

PR #1962, opened 2024-04-27
chore: use self-hosted runners

jdx:


johnpyp:
Is moving this to a self hosted runner a good idea? Just wondering if it might unnecessarily expose the CI to availability or supply chain risk potentially that staying on Github Actions wouldn't. Are github actions not sufficient for something (I believe they're completely free for OSS projects right)?

PR #1986, opened 2024-04-29
security: avoid security problems by using https

artificial-intelligence:
some URLs are currently fetched via insecure http
avoid this by using https.


jdx:
this isn't a security issue

PR #1989, opened 2024-04-29
feat: cli command for listing backends

### LINKED ISSUE #1983: cli command for listing backends
we should have a command like `mise backends` that lists what is available. This information should also be added to `mise dr` if it's not already there.

This is a pretty easy change if someone wants to pick it up, I'd look at how `mise plugins ls` works since it'll be almost the same code but list the "forges" (what backends are called internally because reasons) instead.

roele:
Resolves #1983.

**New Command**
Adds a command `backends` with sub-command `ls` to list available backend types.
```bash
$ mise backends
cargo
go
npm
pipx
ubi
```

**New Doctor Section**
Adds a section `backends` to the `doctor` command output and filters out (installed) backends from the `plugins` section since these are shown in `toolset`.
```bash
...

backends: 
  cargo
  go
  npm
  pipx
  ubi

plugins: 
  bun         (core)
  deno        (core)
  erlang      (core)
...
```

### DIFF
--- a/src/cli/backends/ls.rs
+++ b/src/cli/backends/ls.rs
@@ -1,0 +1,41 @@
+use eyre::Result;
+
+use crate::forge::{self, ForgeType};
+
+/// List built-in backends
+#[derive(Debug, clap::Args)]
+#[clap(visible_alias = "list", after_long_help = AFTER_LONG_HELP, verbatim_doc_comment)]
+pub struct BackendsLs {}
+
+impl BackendsLs {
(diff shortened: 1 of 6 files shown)


jdx (commented):
that was quick!

dhanak:
Please also add completions for the new command.

roele:
@jdx on another note, i assume backends are still experimental should we mark the command as such?

jdx:
well what we should probably do is just show "asdf" if experimental is not set

PR #1995, opened 2024-04-29
test(java): added e2e test for corretto-8 shorthand

jdx:


dkowis:
Thanks so much for fixing this so quickly :D

jdx:
heh all I did is hit the "revert" button

dkowis:
And add a e2e test to catch the bug :D And it was all resolved the same day!

PR #2018, opened 2024-05-02
fix: backend install directory not removed if empty

### LINKED ISSUE #2017: backend install directory not removed if empty
**Describe the bug**
After removing a backend binary the install directory is not deleted because of the new metadata file (directory not empty).

**To Reproduce**
Install a single version of a backend binary and remove afterwards.

```bash
mise i npm:@angular/cli
mise rm npm:@angular/cli
```

**Expected behavior**
Directory `~.local/share/mise/installs/npm:angular-cli` should not exist anymore after removal.

**Additional context**
Introduced by https://github.com/jdx/mise/pull/1909

roele:
Adds some new functions to remove a directory passing in a list of files to ignore for the empty dir check. This allows to remove the (empty) backend directory despite the new hidden `.mise.forge.json` file.

Fixes #2017

### DIFF
--- a/src/file.rs
+++ b/src/file.rs
@@ -49,6 +49,21 @@ pub fn remove_dir<P: AsRef<Path>>(path: P) -> Result<()> {
     .wrap_err_with(|| format!("failed to remove_dir: {}", display_path(path)))
 }
 
+pub fn remove_dir_ignore<P: AsRef<Path>>(
+    path: P,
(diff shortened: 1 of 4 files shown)


roele:
Have experienced some flakiness in the unit tests, even locally they seemed to fail somewhat randomly 🤔

jdx:
Hmm, doesn't look related, I'll check it out tomorrow

PR #2024, opened 2024-05-04
Handle errors from the underlying plugin updates

offbyone:
This wraps things up so there's no panic when the plugin update fails.

This isn't perfect; it'd be nice to find a way to report the underlying
cause, if possible, but it beats a panic.

## The exception

``` 
Message:  called `Result::unwrap()` on an `Err` value:
   0: git failed: Cmd(["git", "-C", "/Users/offby1/.local/share/mise/plugins/action-validator", "-c", "safe.directory=/Users/offby1/.local/share/mise/plugins/action-validator", "fetch", "--prune", "--update-head-ok", "origin", "main:main"]) From https://github.com/mpalmer/action-validator
       ! [rejected]        main       -> main  (non-fast-forward)
   0:

Location:
   src/git.rs:41

Location: src/cli/plugins/update.rs:55
```

Note that this change loses the _cause_, but survives without a crash. I'm open to suggestions about how to format the error.

### DIFF
--- a/src/cli/plugins/update.rs
+++ b/src/cli/plugins/update.rs
@@ -1,6 +1,8 @@
 use console::style;
 use eyre::Result;
+use itertools::Itertools;
 use rayon::prelude::*;
+use std::sync::Mutex;
 
 use crate::config::Settings;
 use crate::plugins;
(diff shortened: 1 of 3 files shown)


src/cli/plugins/update.rs
```diff
@@ -49,12 +51,28 @@ impl Update {
...
             .install(|| {
+                let errors = Mutex::new(Vec::new());
+
                 plugins.into_par_iter().for_each(|(plugin, ref_)| {
                     let prefix = format!("plugin:{}", style(plugin.id()).blue().for_stderr());
                     let pr = mpr.add(&prefix);
-                    plugin.update(pr.as_ref(), ref_).unwrap();
+                    match plugin.update(pr.as_ref(), ref_) {
+                        Ok(_) => (),
+                        Err(e) => {
```
jdx:
```suggestion
                    if let Err(e) = plugin.update(pr.as_ref(), ref_) {
```

jdx: I think this could be further refactored into having `plugins.into_par_iter().for_each(|(plugin, ref_)| {` instead be something like `plugins.into_par_iter().map(|(plugin, ref_)| {` and returns the list of errors but that's not a huge deal

jdx:
I have not used this yet but I believe eyre supports multiple errors which could be used instead of manually formatting: https://github.com/eyre-rs/color-eyre/blob/master/examples/multiple_errors.rs

src/cli/plugins/update.rs
offbyone: I am very much a Rust newbie, so I'll take any suggestions you have, especially if they're accompanied by a "here's why it's better"; this is only the second bit of Rust I've ever written outside of the early stages of an AoC a few years ago. Thanks :D

offbyone:
Do you want me to take a kick at that, first?

jdx:
Yeah I think the multiple errors would be good, I can have a look at it next week if it ends up being tricky

offbyone:
Will do. I'm currently fighting with the borrow checker on that, sort of unsurprisingly. I've got some time on a plane today, perhaps I'll figure something out through trial and error. Thanks!

jdx:
It might help to just submit how you think it should work (broken) and I can sort out the borrow checker

offbyone:
There ya go.

jdx:
thanks for this, it's a good change!

PR #2025, opened 2024-05-04
Add support for injecting extra options to `pipx`

joshbode:
Allow setting extra options for `pipx` installs, e.g.

```toml
"pipx:some-package" = { version = "latest", preinstall = "foo|bar" }
```

which would translate to:

```shell
pipx install some-package --preinstall foo --preinstall bar
```


jdx:
These should be scoped, maybe in an "args" object

roele:
Guess this would be handy for other backends like `cargo` as well. Only question is how this could work with `install` and `use` where you can potentially define multiple tools at once without complicating the `[TOOL@VERSION]...` pattern. Maybe use an `--args` option which is applied to all specified tools or only allow one tool if `args` is set?

PR #2036, opened 2024-05-06
feat: cleanup invalid symlinks in .local/state/mise/(tracked|trusted)-configs

### LINKED ISSUE #2020: cleanup invalid symlinks in .local/state/mise/(tracked|trusted)-configs
While fixing #1746 i noticed that my `~/.local/state/mise/(tracked|trusted)-configs` directories had a lot of invalid symlinks. This is kinda expected as `mise` probably cannot know if you remove a `.mise.toml` file somewhere which is no longer needed.

It would be handy to have a cleanup routine which removes invalid symlinks from time to time. I guess this could happen after either a trust hash is checked or a new one is created.

roele:
Fixes #2020 by introducing an experimental `mise config prune` command to prune invalid config file references from `.local/state/mise/(tracked|trusted)-configs`.

### DIFF
--- a/src/cli/prune.rs
+++ b/src/cli/prune.rs
@@ -5,11 +5,14 @@ use console::style;
 use eyre::Result;
 
 use crate::cli::args::ForgeArg;
+use crate::config::tracking::Tracker;
 use crate::config::{Config, Settings};
 use crate::forge::Forge;
 use crate::toolset::{ToolVersion, ToolsetBuilder};
 use crate::ui::multi_progress_report::MultiProgressReport;
 use crate::ui::prompt;
+
+use super::trust::Trust;
 
 /// Delete unused versions of tools
 ///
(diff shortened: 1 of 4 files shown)


jdx:
I think I'd prefer `mise prune --config` but I'm open to a reason why we would put it here instead

jdx:
Ah yeah I did not think of that

src/cli/config/prune.rs
```diff
@@ -0,0 +1,32 @@
+use crate::{
+    cli::trust::Trust,
+    config::{tracking::Tracker, Settings},
+};
+use eyre::Result;
+
+/// [experimental]  Prune tracked and trusted config file references
```
jdx: Add a note that this only deletes references that no longer point to existing configs. It might appear this deletes all of them regardless.

jdx (changes requested):

roele:
We probably also should also ensure in `mise prune` that both are cleaned up.

jdx:
I'm not sure. I think there probably will be some small subset of users with directories that come and go which also use prune to clear out old tools—they may want references that don't exist

roele:
Up until now, the tracked configs have been cleanup up as part of `Tracker::list_all()`. I moved that out to make it more explicit but we can skip the cleanup of those for `prune` completely.

src/cli/config/prune.rs
```diff
@@ -0,0 +1,28 @@
...
+/// [experimental]  Prune tracked and trusted config file links that point to non-existent configs
+#[derive(Debug, clap::Args)]
+#[clap(verbatim_doc_comment, after_long_help = AFTER_LONG_HELP)]
+pub struct ConfigPrune {}
+
+impl ConfigPrune {
+    pub fn run(self) -> Result<()> {
+        let settings = Settings::try_get()?;
+        settings.ensure_experimental("`mise config prune`")?;
+        Tracker::clean()?;
```
jdx: You know I think for parity we really should support --dry-run for this. Perhaps we could think about making a generic dry run function since we do that a lot in the codebase

jdx:
Here's a thought, maybe we put everything under `mise prune` but if you specify `--tools` or `--config` that would filter so as to only prune that type. No args would essentially be like a (nonexistent) `--all`.

It might be worth looking at what `docker system prune` is doing for a potential reference.

src/cli/prune.rs
```diff
@@ -27,10 +30,39 @@ pub struct Prune {
...
+    }
+
+    fn prune_configs(&self) -> Result<()> {
+        if !self.dry_run {
+            Tracker::clean()?;
+            Trust::clean()?;
+            info!("pruned configuration links");
+        } else {
+            info!("pruned configuration links {}", style("[dryrun]").bold());
+        }
```
jdx:
```suggestion
        if self.dry_run {
            info!("pruned configuration links {}", style("[dryrun]").bold());
        } else {
            Tracker::clean()?;
            Trust::clean()?;
            info!("pruned configuration links");
        }
```

PR #2090, opened 2024-05-13
feat: pass github token to UBI and cargo-binstall backends.

### LINKED ISSUE #2067: GITHUB_TOKEN for cargo and ubi backends
The cargo-binstall and ubi backends delegates the work to the tool binaries, which may have to query the GitHub API to fetch the informations about the latest releases.  Right now, mise itself can use the content of the GITHUB_API_TOKEN variable to access the API but these tools expect the token to be stored in a variable named GITHUB_TOKEN.

The backends could use GITHUB_API_TOKEN to set GITHUB_TOKEN when it is not defined.

Adirelle:
The environment variable GITHUB_TOKEN is passed through calls to ubi and cargo-binstall. If GITHUB_TOKEN is not defined but GITHUB_API_TOKEN is, it is passed GITHUB_TOKEN .

Resolves #2067.

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -141,6 +141,11 @@ pub static PATH_NON_PRISTINE: Lazy<Vec<PathBuf>> = Lazy::new(|| match var("PATH"
 pub static DIRENV_DIFF: Lazy<Option<String>> = Lazy::new(|| var("DIRENV_DIFF").ok());
 #[allow(unused)]
 pub static GITHUB_API_TOKEN: Lazy<Option<String>> = Lazy::new(|| var("GITHUB_API_TOKEN").ok());
+pub static GITHUB_TOKEN: Lazy<Option<String>> = Lazy::new(|| {
+    var("GITHUB_TOKEN")
+        .or_else(|_| var("GITHUB_API_TOKEN"))
+        .ok()
+});
 
 pub static MISE_USE_VERSIONS_HOST: Lazy<bool> =
     Lazy::new(|| !var_is_false("MISE_USE_VERSIONS_HOST"));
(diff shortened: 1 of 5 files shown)


src/env.rs
```diff
@@ -141,6 +141,12 @@ pub static PATH_NON_PRISTINE: Lazy<Vec<PathBuf>> = Lazy::new(|| match var("PATH"
 pub static DIRENV_DIFF: Lazy<Option<String>> = Lazy::new(|| var("DIRENV_DIFF").ok());
 #[allow(unused)]
 pub static GITHUB_API_TOKEN: Lazy<Option<String>> = Lazy::new(|| var("GITHUB_API_TOKEN").ok());
+#[allow(unused)]
```
jdx: why?

src/forge/cargo.rs
```diff
@@ -54,7 +55,12 @@ impl Forge for CargoForge {
         let settings = Settings::get();
         settings.ensure_experimental("cargo backend")?;
         let cmd = if self.is_binstall_enabled() {
-            CmdLineRunner::new("cargo-binstall").arg("-y")
+            let runner = CmdLineRunner::new("cargo-binstall").arg("-y");
+            if let Some(token) = &*GITHUB_TOKEN {
+                runner.env("GITHUB_TOKEN", token)
```
jdx:
```suggestion
                runner = runner.env("GITHUB_TOKEN", token)
```

I prefer making it mutable and doing it this way, makes later edits simpler

src/env.rs
Adirelle: Honestly: copypasta of the previous variable.

PR #2116, opened 2024-05-14
feat(tasks): add --json flag

### LINKED ISSUE #2043: mise tasks: machine-readable (json) output
I know tasks are experimental and this is probably coming anyway but just to be sure:
I'd love `mise tasks` to be able to generate machine-readable output, like the json output of `mise env` (so other tools can pick up the listing and invoke tasks through mise)

vrslev:
Fixes #2043


jdx:
very well done!

PR #2123, opened 2024-05-15
feat: allow install of specific version from https://mise.run #1800

### LINKED ISSUE #1800: Add variable for specific mise version during installation
Hello :wave: I would like to be able to install a specific version of `mise` with the official install script (https://mise.run/).

Is it possible to create an environment variable, for example `MISE_VERSION` that can be defined during installation [here](https://github.com/jdx/mise/blob/4f13c6329f358bb0ff9dc750984ddeef14d1f4e4/packaging/standalone/install.envsubst#L151) ?

Thanks in advance.

Its-Alex:
Closes #1800 

I tested the script after rendering with `scripts/render-install.sh`. Can you tell me if anything is missing?


packaging/standalone/install.envsubst
```diff
@@ -73,50 +73,29 @@ shasum_bin() {
 }
 
 get_checksum() {
+	version=$1
 	os="$(get_os)"
 	arch="$(get_arch)"
+	url="https://github.com/jdx/mise/releases/download/${version}/SHASUMS256.txt"
 
-	checksum_linux_x86_64="$MISE_CHECKSUM_LINUX_X86_64"
```
jdx, line 79: removing these makes it so the gpg signed version of the installer is no longer secure. We still need these but obviously that won't work with specifying the version. We should have both logic based on whether we're fetching the current version or an earlier one.

jdx: here's the docs around install.sh.sig: https://mise.jdx.dev/getting-started.html#alternate-installation-methods

Its-Alex: @jdx Didn't see that feature, thanks for the documentation. I think it's patched, however, I'm not sure how to test this

jdx: not a good way to do it, we'll just have to check after release

jdx (approved):

jdx:
nice work!

rakuyoMo:
@Its-Alex Hello, excuse me, I saw the possibility of fixing the mise version locally through this PR, as I started the [discussion](https://github.com/jdx/mise/discussions/1946) before. Could you please take some time to look at the possibility of making it happen? Thank you so much!

Its-Alex:
@rakuyoMo This PR aims to only install a specific version. It is up to the user to prevent updates.

As far as I know, I don't think mise has this option. I don't have experience with Rust, and since this is a Rust feature, I don't think I can tackle this easily. :sweat_smile:

PR #2182, opened 2024-05-25
fix: sort pipx forge remote versions semantically

jasisk:
Seems pypi sorts `releases` lexically which can result in the wrong version being selected for `latest` in the pipx forge. [Example](https://pypi.org/pypi/magic-wormhole/json)

Noticed this when attempting to use `pipx:magic-wormhole` which would install `v0.9.2` instead of `v0.14.0`.

Just like the last time I was in here, I'm not a rust person and just kinda fumbled my way through (and didn't write any tests!) so please destroy at will.

Previous behavior:
```
$ mise x pipx:magic-wormhole -- wormhole --version
creating virtual environment...
creating shared libraries...
upgrading shared libraries...
installing magic-wormhole from spec 'magic-wormhole==0.9.2'...
done! ✨ 🌟 ✨
mise pipx:magic-wormhole@0.9.2 ✓ installed                                                                                                                                                                                         magic-wormhole 0.9.2
```

New behavior:
```
$ cargo run x pipx:magic-wormhole -- wormhole --version
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.11s
     Running `/tmp/mise/target/debug/mise x 'pipx:magic-wormhole' -- wormhole --version`
creating virtual environment...
creating shared libraries...
upgrading shared libraries...
installing magic-wormhole from spec 'magic-wormhole==0.14.0'...
done! ✨ 🌟 ✨
mise pipx:magic-wormhole@0.14.0 ✓ installed
magic-wormhole 0.14.0
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1646,6 +1646,7 @@ dependencies = [
  "reqwest",
  "rmp-serde",
  "self_update",
+ "semver",
  "serde",
  "serde_derive",
  "serde_json",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,6 +97,7 @@ self_update = { version = "<0.40.0", default-features = false, features = [
     "compression-flate2",
     "signatures",
 ] }
+semver = "1.0.23"
 serde = "1.0.199"
 serde_derive = "1.0.199"
 serde_json = { version = "1.0.116", features = [] }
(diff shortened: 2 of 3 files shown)


jasisk:
... thank goodness for those e2e tests.
As `mkdocs` shows, assuming they're all semantic versions is a bad assumption. :\
Next approach might be to add a version entry for `latest` based on what `.info.version` is set to from pypi.

jasisk:
Fixed by #2209

PR #2199, opened 2024-05-27
feat: Mint(swift) backend

kattouf:
### Implementation of custom backend based on [Mint](https://github.com/yonaskolb/Mint):  A package manager that installs and runs executable Swift packages

Motivated by [discussion](https://github.com/jdx/mise/discussions/1250#discussioncomment-8117373)

`Mint` works in a simple way: it looks for release tags, checkout the specified one, and compiles the executable target.

The following environment variables are available to configure the installation path: `MINT_PATH`, `MINT_LINK_PATH`.

From the README:
> You can set `MINT_PATH` and `MINT_LINK_PATH` environment variables to configure where Mint caches builds, and where it symlinks global installs. These default to ~/.mint and ~/.mint/bin respectively.

Result:
mise-use
mise-bin-paths

### WIP / Questions:

- [ ] Need to update documentation / cli help?

- [ ] Dependencies: The `Mint backend` depends on `Mint`, which at runtime, in turn, depends on `swift-lang`. Currently, neither can be installed by `mise`:
    - Mint does not have a plugin and cannot be installed by any other backend.
    - Swift has a [plugin](https://github.com/fcrespo82/asdf-swift), but the last time I tried it, it didn't work. This needs to be double-checked. 
    UPD: also [pr](https://github.com/jdx/mise/pull/1708) can be useful here
    
    **So the question is: do we need to implement plugins for all dependencies first, to use them from the backend? Or dependencies free implementation fits too?**
- [ ] End-to-end test:
    - Can I pre-install `Mint` this way? (For example, like `test_ubi`).
    (Currently, `Mint` doesn't have any installation scripts for use outside the repository).
    - Should I pre-install `Mint` dependencies as well? (Swift language).


pepicrft:
I wonder if we should have the logic in Mise instead of proxying through another tool. The logic would be comprised of:
- Running `swift build` to generate the binaries.
- Look up the binaries in the conventional build directory
- Copy those artifacts to the location where Mise expects them.

Thoughts @jdx ?

kattouf:
>  if we should have the logic in Mise instead of proxying through another too

It looks like interesting task! 

But in my opinion, `mise` should be a "proxy" and contain only knowledge on how to work with other tools. This approach ("proxy") makes it easy for everyone to switch to and from `mise` and prevents `mise` from becoming a "super builder" with a lot of knowledge about different technology stacks details, duplicating logic already written in other community respected and used tools.

In our case, I think we could contribute to Mint and get profits for both Mise and Mint users at once 💪 
For example, I'm already have some todo-list for `Mint` contributing:
1. Install script with configuration like in mise.run (now I have some of this script logic in e2e test)
2. Add more ENV settings that mirrors cli flags, like `MINTFILE_PATH`, `SILENT_OUTPUT` etc

pepicrft:
I have a different opinion, but I'm not too intense about it.

Mise should not replace an underlying build tool, like Cargo, SPM, or Go's toolchain, because it's not its responsibility, but shelling out to a tool that does the same as Mise, invoking a build tool, SPM, in this case, feels unnecessary indirection, which comes with some drawbacks:
- Issues are harder to debug.
- Mise loses control over the output formatting and errors, which are now propagated through an additional layer.
- It requires users to have an additional tool in their system.

I think the [Go](https://github.com/jdx/mise/blob/main/src/forge/go.rs) and [Cargo](https://github.com/jdx/mise/blob/main/src/forge/cargo.rs) are good examples of that, and there could be an `SPM` forge that takes a similar role for Swift packages. If it was costly, I'd disagree with re-building the functionality, but most of what Mint does, is already implemented by Mise, so it's very low cost.

kattouf:
I see, thank you for clarification. We agree about what mise should do 🤝 

At first I was a little confused by the fact that Rust and Go have built-in methods for installing executable files, unlike SPM. But after thinking a little, I realized that `swift install` is `swift build && copy built binary`, simplifying.

Then I looked at your original message from a new perspective and I think your idea is promising 👍

jdx:
this is a tangent, but my experience with swift is very basic and mostly limited to reading about the basic design, I think it's one of the most incredible languages out there and I don't understand why it seemingly isn't used out of the apple ecosystem. It really seems like it could be an ideal language for CLI tools—far better than Go and much easier to learn than Rust while I assume has similar performance characteristics.

pepicrft:
>  Ok, I think it is, so I've already started work on new backend 🙂

Oh! This is very nice :). Let me know when it's ready, and I can review it. That work, along with [this one](https://github.com/jdx/mise/pull/1708), which I started some time ago, should make Swift well-supported by Mise.

kattouf:
Closed in favor of https://github.com/jdx/mise/pull/2241

PR #2223, opened 2024-05-30
remove deleted files in default.nix

FelisNivalis:


jdx:
looks like this was added in

PR #2241, opened 2024-06-02
feat: SPM(Swift Package Manager) backend

kattouf:
Implementation of backend for install executables managed by [Swift Package Manager](https://www.swift.org/documentation/package-manager/)

### Description 

[Discussion](https://github.com/jdx/mise/discussions/1250#discussioncomment-8117373) 
Compared to [previous PR](https://github.com/jdx/mise/pull/2199) there are no dependencies on third parties other than [Swift](https://www.swift.org) itself.

SPM does not support installing an executable, only the build command. This backend implementation itself covers this functionality and is described by a flow:
1. Clone the repository with the SPM package into a temporary directory.
2. Read `Package.swift` for a list of executables
3. Build each executable and copy the binary with associated resources to `mise` install directory

Demo screenshots:
Screenshot 2024-06-02 at 14 14 53

Screenshot 2024-06-02 at 14 17 09

P.S. I'm not full-time Rust developer, feel free to make comments if something is not done according to best practices in Rust community

### Requires attention

1. Missed `swift`dependency
    I had some problems installing `swift` with the current asdf plugin so I guess we'll just have to wait for https://github.com/jdx/mise/pull/1708
    Now missed dependency leads to:
    - Unable to run e2e test
    - Users are responsible for installing swift
2. Support for build artifacts on other platforms
    I have no experience using `swift` on other platforms (Linux, Windows) and can't be sure about finite list of resource and dynamic libraries formats we should support to copy together with binary.
    I decide dive into it after we are sure that everyone is happy with the feature and we will discuss merging it.
3. Name of package to install
   1. I don't sure about how flexible it should be, so for now I limited it to two options:
        - GitHub repo name: `spm:nicklockwood/SwiftFormat@0.53.10`
        - GitHub https url: `spm:https://github.com/nicklockwood/SwiftFormat.git@0.53.10`
         Maybe we should add other options by community requests
    2. When we use different formats for the same package repo, they are perceived as different packages:    
        mise-spm-name
         I got some errors when trying to change the BackendArg name to generic format (shortened) in `SPMBackend::new(name)`
        Are there any other options to solve this problem?
4. Versioning
     Current implementation looking for GitHub release tags for installation candidates. Installation by branch/commit can also be supported, but I'm not sure if this should be done now.


src/backend/spm.rs
```diff
@@ -0,0 +1,327 @@
...
+        debug!(
+            "Cloning swift package repo: {}, revision: {} to path: {}",
+            package_repo.url.as_str(),
+            revision,
+            tmp_repo_dir.display()
+        );
+        let repo = Repository::clone(package_repo.url.as_str(), &tmp_repo_dir)?;
+        let (object, reference) = repo.revparse_ext(&revision)?;
+        repo.checkout_tree(&object, None)?;
+        repo.set_head(reference.unwrap().name().unwrap())?;
```
kattouf: FYI
I tried to use for clone + checkout flow project git client by example in `core/python.rs`, but ran into problems and so far used git2 directly.
image
mise-git-update

src/backend/spm.rs
```diff
@@ -0,0 +1,322 @@
...
+    /// https://github.com/owner/repo.git
+    url: Url,
+    /// owner/repo_name
+    shorthand: String,
+}
+
+impl SwiftPackageRepo {
+    fn new(name: &str) -> Result<Self, eyre::Error> {
+        let shorthand_regex = regex!(r"^[a-zA-Z0-9_-]+/[a-zA-Z0-9_-]+$");
+        let shorthand_in_url_regex =
```
pepicrft, line 218: Would it make sense to generalize this such that it also supports other Git platforms, for example Gitlab?

src/backend/spm.rs
```diff
@@ -0,0 +1,322 @@
...
+            regex!(r"https://github.com/([a-zA-Z0-9_-]+/[a-zA-Z0-9_-]+)\.git");
+
+        let shorthand =
+            if let Some(Some(m)) = shorthand_in_url_regex.captures(name).map(|c| c.get(1)) {
+                m.as_str()
+            } else if shorthand_regex.is_match(name) {
+                name
+            } else {
+                Err(eyre::eyre!("Invalid swift package repo: {}", name))?
```
pepicrft:
```suggestion
                Err(eyre::eyre!("Invalid Swift package repository: {}. The repository should either be a GitHub repository slug, owner/name, or the complete URL, https://github.com/owner/name.", name))?
```

src/backend/spm.rs
```diff
@@ -0,0 +1,322 @@
...
+#[derive(Deserialize)]
+struct PackageDescription {
+    products: Vec<PackageDescriptionProduct>,
+}
+
+#[derive(Deserialize)]
+struct PackageDescriptionProduct {
+    name: String,
+    #[serde(deserialize_with = "PackageDescriptionProductType::deserialize_product_type_field")]
+    r#type: PackageDescriptionProductType,
```
pepicrft, line 284: Out of curiosity, what's the meaning of this `r#type` annotation? (I'm not that versed at Rust)

jdx (approved):

e2e/backend/test_spm_slow
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+# Enable test when `Swift` installation will be fixed on Linux (for docker run)
+# Wait for https://github.com/jdx/mise/pull/1708
+exit 0
+
+mise use swift
+
+assert "mise x spm:nicklockwood/SwiftFormat@0.53.10 -- swiftformat --version" "0.53.10"
+assert "mise x spm:https://github.com/nicklockwood/SwiftFormat.git@0.53.10 -- swiftformat --version" "0.53.10"
```
pepicrft, line 10: Would you mind adding https://github.com/tuist/tuist?

pepicrft (approved):
Logic-wise it looks prefect. Nicely done @kattouf.
Regarding the Rust code and patterns, I'd wait for @jdx's review.

src/backend/spm.rs
kattouf: It's `raw identifier`, used to define user data names, same as language keywords (like \`type\` in swift)
https://stackoverflow.com/a/69871965

src/backend/spm.rs
kattouf: Yes, it makes sense globally, I just decided not to spend too much time on unpopular cases right now.

I think we could cover almost everything else cases by implementing:
1. support for any URLs up to `.git`
2. support for various git revisions as versions: tag, commit, branch

and divide the versioning flow into:
1. github releases
2. specific git revision

But is it worth doing this right away, I'm not sure

e2e/backend/test_spm_slow
kattouf: Ha, with pleasure! But the installation time seems too long for testing, don't you think?
Maybe we could to this with some `install specific executable` feature, but now it takes about 5m and looks not tests friendly.
image

pepicrft: 5m is terrible! So feel free to skip it then :)

src/backend/spm.rs
pepicrft: Gotcha! Thanks for the reference to Swift, it helped :)

src/backend/spm.rs
pepicrft: I think this can be done in a follow-up PR. It seems a good candidate to be extracted into a separate function that can be reused across the backend. I'm sure @jdx has already put some thoughts there.

kattouf:
@pepicrft Could you also check `Requires attention` section in PR description? None of this seems like a blocker to you right now?
We have already partially discussed points 3 and 4, so we can focus on 1, 2 and 3.ii?

pepicrft:
> 1.  Missed swift dependency

I'd be consistent with what the other backends do. If they install the required toolchain absent, we can error out telling users that "Swift was not found in the environment" until we install it ourselves. This is a scarce scenario because most developers using this backend will likely be Xcode developers with the toolchain available in their systems.

> 2. Support for build artifacts on other platforms

I'm in a similar spot here. At least on Linux, if everything is statically linked, the binary should work, so the issues will likely arise when using dynamic dependencies, which I'm not entirely sure are supported on Linux. Similarly, this is a bit uncommon, so I'd ship this as it is and iterate on it as feedback comes our way.

> 3. When we use different formats for the same package repo, they are perceived as different packages

This is a bit confusing, especially considering that they are referring to the same package. @jdx might be able to help with this one.

kattouf:
> I'd be consistent with what the other backends do.

Currently `mise` doesn't automatically install backend dependencies, instead the [documentation suggests](https://mise.jdx.dev/dev-tools/backends/ubi.html) installing them using `mise` or any other convenient method and gives an error message:
image
It seems that the backend is a feature that expects that it will be used by those people who already have the entire necessary tools, they just want to describe them in a single config

> Similarly, this is a bit uncommon, so I'd ship this as it is and iterate on it as feedback comes our way.

Agree 🤝

pepicrft:
Overall, fantastic work @kattouf 👏🏼. As one of the advocates of Mise in the Swift ecosystem, I couldn't be more excited of this work landing. 
We have to wait for @jdx to do the merge.

PR #2266, opened 2024-06-07
fix: Replaced http with https for security

shirayu:


jdx:
this doesn't impact security and it also makes mise significantly slower

PR #2295, opened 2024-06-16
docs: Mention the fish shell's automatic activation in the Quickstart section

rmacklin:
Since folks might not read the whole page before they start following the steps in the Quickstart section, I think it's helpful to mention this detail sooner.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -36,6 +36,11 @@ Now that `mise` is installed, you can optionally activate it or add its [shims](
 For interactive shells, `mise activate` is recommended. In non-interactive sessions, like CI/CD, IDEs, and scripts, using `shims` might work best. You can also not use any and call `mise exec/run` directly instead.
 See [this guide](dev-tools/shims.md) for more information.
 
+:::info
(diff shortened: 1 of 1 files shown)


rmacklin:
Hey @jdx, thanks for merging!

I have to ask because it seems like a significant rewrite (and force-push) of the git history of the `main` branch took place (over 1300 fewer commits exist in the new history and the rewrite appears to go as far back as the initial commit): Have you explained the what/why behind that rewrite anywhere? I'm curious to know more details if you can share them.

I tried searching for issues that mention this, as well as checking your blog and twitter feed, but didn't find anything. Not suggesting anything malicious happened, but it does raise some eyebrows when a very popular repository/tool has such a significant history rewrite.

rmacklin:
Ah, I hadn't thought this would be mentioned as a release, but I just spotted https://github.com/jdx/mise/releases/tag/v2024.11.27

> I rewrote the git history to remove what were mostly commits from the asdf registry repo that was included in this repo at one point so from now on the history should be clear and the CHANGELOG will reflect everything from this point. Older release information will still be available at https://github.com/jdx/mise/releases

Issue #1912, opened 2024-04-17
Lazily evaluated env template variables in path entries


josb:
Setup:
* `HOME=/Users/john`
* A global `config.toml`file with:
   *  a `_.path`  template entry with value `"{{env.GEM_HOME}}/bin" `
   * an env entry: `GEM_HOME = "{{env.HOME}}/.gem" `
* A `.mise.toml` file in some directory under the home directory with:
   * an env entry: `GEM_HOME = "{{env.HOME}}/.mise-gemsets/foo" `

Today, env entries in path templates are expanded using the environment variables defined in the same mise config file, either global or local.

The idea behind a lazily expanded reference in path entries is to only expand the path env reference after both the global and local mise configs have been read, and use whatever value the env variable has at that point.

In the above example, this would mean that in the specific directory a path entry valued `/Users/john/.mise-gemsets/foo/bin` would be added, whereas elsewhere the entry `/Users/john/.gem/bin` would be added instead.

Today this doesn't work because the path entry has to be set in the global config where it is always added with the global value, which is not what we want here.

If there's another way to achieve this with existing mise facilities I'd love to learn about it. Thanks!

josb:
Here's a demo of my current [branch](https://github.com/josb/mise/tree/josb/lazy-path-entries), which allows lazy path entries to be specified using `{ value = "lazy_path_entry" }` in addition to the existing `"normal_path_entry"`. The effect is that the same positional `PATH` entry is either controlled by the global mise config, or by a local mise config in some directory, as can be seen below:

`~/.config/mise/config.toml`:
```toml
[env]
GEM_HOME = "{{env.HOME}}/.gem"
_.path = [
  { value = "{{env.GEM_HOME}}/bin" },
]
```

`~/github/josb/.mise.toml`:
```toml
[env]
GEM_HOME = "{{env.HOME}}/.mise-gemsets/foo"
```

```shell
~% mise env
export GEM_HOME=/Users/josb/.gem
export PATH='/Users/josb/.gem/bin:/opt/homebrew/opt/ruby/bin:...
~% cd ~/github/josb
~/github/josb% mise env
export GEM_HOME=/Users/josb/.mise-gemsets/foo
export PATH='/Users/josb/.mise-gemsets/foo/bin:/opt/homebrew/opt/ruby/bin:...
~/github/josb% 
```

Normal `_.path` entries also still work, of course.

Feedback welcome.

jdx:
[...]

PR #2310, opened 2024-06-20
feat: Lazily evaluated env template variables in path entries

josb:
Implements #1912.

Please note that there's a TODO entry in `mise::EnvMapVisitor.visit_map`. The issue is that `deserialize_with = "deserialize_arr"` can either be left off or added to`path: Vec<PathEntry>`:
* If it is left off, the feature works as demonstrated but 3 tests fail;
* if it is added, lazy path entries (`{ value = "{{ env.GEM_HOME }}/bin" }`) cause a parse error (`invalid type: map, expected a string`) but the 3 failing tests now pass.

I am hoping you can help me address this issue, as I don't understand Serde / Rust well enough yet.
This is also the reason why I didn't add any tests for this feature yet.

Thanks!


src/config/env_directive.rs
```diff
@@ -79,6 +137,8 @@ impl EnvResults {
     ) -> eyre::Result<Self> {
         let settings = Settings::get();
         let mut ctx = BASE_CONTEXT.clone();
+        debug!("resolve: input: {:#?}", &input);
```
jdx: these should all be trace level

josb: Will fix, I just added these while trying to understand the code. If you feel some serve little purpose I am happy to remove them.

Also, if you have suggestions for better identifier names, I will happily change them. 🙏🏻

josb: I changed all debug calls to trace, and added a custom deserializer to get both the tests to pass and the lazy path entries to be parsed correctly.

josb:
Any feedback, @jdx?

Issue #2337, opened 2024-06-28
Docs is not clear enough


hexWars:
In https://mise.jdx.dev/lang/java.html, I can clearly see how to get the remote Java versions for download using `mise ls-remote java`.

However, in https://mise.jdx.dev/lang/node.html, I did not see how to get the remote versions. It seems that a description should be added.

If possible, I can make a simple PR.

PR #2340, opened 2024-06-30
docs: update go.md,node.md,python.md,ruby.md

hexWars:
If you have time, could you please take a look and see if this is appropriate?

about: #2337


jdx:
this doesn't really seem like the right place for this sort of thing. Seems like this should be somewhere more generic since it's not specific to languages.

hexWars:
I know this isn't based on a specific language, but when I want to use Node and get the remote version of Node, I encountered the mise ls-remote command on a Java page.

This seems rather odd.

Perhaps it can be extracted and placed in a more prominent location.

Do you have any better suggestions?

Issue #2049, opened 2024-05-10
implement more tera functions


jdx:
[just has quite a few functions](https://github.com/casey/just?tab=readme-ov-file#system-information) which is a good starting point for us to improve what can be done with templates in mise

tfh-cri:
some equivalent to the tera built-in `slugify` would be useful (in my current need, adding a git-branch specific suffix to a python virtualenv managed by mise).

I assumed the tera builtin would already work, but looks like mise is using the 
```
tera = { ..., default-features = false }
```
definition which removes various functions including `slugify`.

A haphazard local rebuild suggests `default-features = true` adds about 3MB to the final binary (23->26MB), so might be excessive for just that function.

jdx:
I'd accept a pr to add that feature for sure

jdx:
in fact all the default ones seem like they could be useful

erickguan:
I plan to implement this in 3 steps.

1. I will implement functions with "strict" definitions. They are;
  - system information
  - Environment Variables
  - Invocation Directory
  - Source and Source Directory
  - String Manipulation
  - Path Manipulation
  - UUID and Hash Generation
  - Semantic Versions
2. After that, we can discuss how to handle errors for other functions generating runtime failures and filesystem, shell access. I have not thought about security, etc.
3. Then the other functions you consider make sense. e.g. no point implementing `justfile`.

What do you think about it?

jdx:
tera already supports some of this but for what it doesn't have support this sounds fine

erickguan:
Now we have added most of `just` functions, at least within PRs. A quick question and some progress updates.

Do you want to improve error handling or it's a future improvment? You don't use `tera::Error` which generally provides a few more error information. The current way of throwing error via CLI is fine judging no one complains.

You mentioned `invocation_directory` and other paths in #2569. I saw some XDG paths in `env.rs` but not more than that. I will leave them to you.

Then the last few functions are:
- Error Reporting
- Random
- Datetime

I don't see a point implementing `just_exexcutable` nor `just_pid` functions. `is_dependency` is not relevant because mise is not a build system.

jdx:
> Do you want to improve error handling

yeah, doesn't have to be a part of this issue necessarily though

> You mentioned invocation_directory and other paths in https://github.com/jdx/mise/pull/2569. I saw some XDG paths in env.rs but not more than that. I will leave them to you.

[...]

PR #2354, opened 2024-07-06
feat(template): enable more tera builtin funcs/filters

tfh-cri:
enables `default-features` build arg for Tera, which adds the following additional filters/functions:

Filters:

* `truncate`
* `date`
* `filesizeformat`
* `slugify`
* `urlencode`
* `urlencode_strict`

Functions:

* `now`

Details at https://keats.github.io/tera/docs/#built-ins

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -175,12 +175,6 @@ dependencies = [
 
 [[package]]
 name = "base64"
-version = "0.21.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9d297deb1925b89f2ccc13d7635fa0714f12c87adce1c75356b39ca9b7178567"
-
-[[package]]
-name = "base64"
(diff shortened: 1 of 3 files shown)


tfh-cri:
This is a (somewhat belated) followup to https://github.com/jdx/mise/issues/2049

jdx:
Isn't default features on by default?

tfh-cri:
> Isn't default features on by default?

Huh, it does seem to be, although I could have sworn when I posted that original comment to #2049 it was as a result of wanting to use `{{ foo | slugify }}` and it not working, which sent me down the rabbithole of why not.

...Doh, looks like it got fixed in 48ca74043e21fe12de18a8457e4554ac2cadb17b and I hadn't realised.

So yeah, this is unnecessary, although I do now know how to build mise 😄

PR #2363, opened 2024-07-10
feat: LuaJIT core plugin

cstrahan:
Here's a core plugin for LuaJIT. Let me know if there's anything I can fix/improve.

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -20,7 +20,6 @@ pub struct GithubTag {
 }
 
 #[derive(Debug, Deserialize)]
-#[serde(rename_all = "camelCase")]
 pub struct GithubCommit {
     pub sha: String,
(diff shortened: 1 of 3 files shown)


cstrahan:
The macOS specific bits were inspired by: https://github.com/Homebrew/homebrew-core/blob/e1f3f72d7b4e73c4daf62c78017d4d0cf9727c63/Formula/l/luajit.rb

Tested on my M1 MacBookPro.

cstrahan:
The https://github.com/smashedtoatoms/asdf-luaJIT plugin also installs luarocks -- perhaps we should do that too?

jdx:
Sorry but core plugins are only for the most popular languages. I don't have the bandwidth to maintain a lot of them.

PR #2370, opened 2024-07-12
feat: support env vars in plugin urls

### LINKED ISSUE #2364: Support envars in plugin urls
Hello,

We have internal mise plugin repos that require authentications like:

[plugins]
release-cli = 'https://$TOKEN_USERNAME:$TOKEN@gitlab.com/foo/bar/asdf/release-cli.git'

We would line to request support for interpolating the envvars like $TOKEN_USERNAME

roele:
Fixes #2364

### DIFF
--- a/src/config/config_file/mise_toml.rs
+++ b/src/config/config_file/mise_toml.rs
@@ -219,8 +219,15 @@ impl ConfigFile for MiseToml {
         }
     }
 
-    fn plugins(&self) -> HashMap<String, String> {
-        self.plugins.clone()
+    fn plugins(&self) -> eyre::Result<HashMap<String, String>> {
+        self.plugins
+            .clone()
+            .into_iter()
+            .map(|(k, v)| {
+                let v = self.parse_template(&v)?;
+                Ok((k, v))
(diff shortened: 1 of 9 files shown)


src/config/config_file/mise_toml.rs
```diff
@@ -324,8 +331,23 @@ impl ConfigFile for MiseToml {
...
+                let k = k.clone();
+                let v: Result<BTreeMap<String, String>, eyre::Error> = v
+                    .clone()
+                    .into_iter()
+                    .map(|(k, v)| {
+                        let v = self.parse_template(&v)?;
+                        Ok((k, v))
+                    })
+                    .collect();
+                v.map(|v| (k, v))
```
roele, line 348: despite my efforts to re-map, this still seems to panic, not quite sure why

roele: culprit seems to be https://github.com/jdx/mise/blob/65313580a4aadecc8c118f2beb59c475172318c5/src/config/mod.rs#L57 which is called in case of alias but not for plugins

PR #2378, opened 2024-07-15
Update getting-started.md to work with homebrew

boris-smidt-klarrio:
Make the install readme also work with the homebrew install.


boris-smidt-klarrio:
Not sure why it was closed? 
FYI, I had a few issues when people installed it and this solved the issue making the setup less brittle.
Especially if people manage there installs manually themselfs and don't use `~/.local/bin`.

PR #2406, opened 2024-07-28
fix: mise use does not create a local .mise.toml anymore

### LINKED ISSUE #2405: mise use does not create a local .mise.toml anymore
**Describe the bug**
As per documentation http://localhost:5173/dev-tools/#mise-use the `use` command without `--global` flag should create a local `.mise.toml` file. This seems not to be the case and the global file gets overwritten with the project specific version.

**To Reproduce**
```sh
mise use --global node@latest
mise ~/.config/mise/config.toml tools: node@22.5.1

mkdir -p ~/myproj
cd ~/myproj
mise use node@20
mise ~/.config/mise/config.toml tools: node@20.16.0
```
Note that in both cases the global configuration file is modified.

**Expected behavior**
A local `.mise.toml` file is created if none exists. Global configuration does NOT get overwritten without the `--global` flag.

**Additional context**
Likely related to changes in https://github.com/jdx/mise/pull/2207

roele:
Fixes #2405 

Not quite sure if this is the correct approach. The [`LOCAL_CONFIG_FILENAMES`](https://github.com/jdx/mise/blob/b9bc7dfef5f804e5c6bb8bf474d857d74ee6eb92/src/config/mod.rs#L458) is very extensive and also contains the path of the global configuration file. So traversing up directories from a directory within user home will at some point find the global configuration. Am i assuming correctly that you could also have the path `.config/mise/config.toml` in a specific project directory?

### DIFF
--- a/src/cli/use.rs
+++ b/src/cli/use.rs
@@ -6,8 +6,10 @@ use itertools::Itertools;
 
 use crate::cli::args::{BackendArg, ToolArg};
 use crate::config::config_file::ConfigFile;
-use crate::config::{config_file, Config, Settings, LOCAL_CONFIG_FILENAMES};
-use crate::env::{MISE_DEFAULT_CONFIG_FILENAME, MISE_GLOBAL_CONFIG_FILE};
+use crate::config::{config_file, is_global_config, Config, Settings, LOCAL_CONFIG_FILENAMES};
+use crate::env::{
+    MISE_DEFAULT_CONFIG_FILENAME, MISE_DEFAULT_TOOL_VERSIONS_FILENAME, MISE_GLOBAL_CONFIG_FILE,
+};
 use crate::file::display_path;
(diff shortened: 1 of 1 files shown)


roele:
After looking at the failing tests i wonder how config file creation is supposed to work. How does know if a tool-version file should be created? There is a settings asdf_compat, a MISE_USE_TOML environment variable and multiple CONFIG_FILE_NAME variables.

Additionally the tests seem to use an artificial setup where the global config is in `~/config` rather than `~/.config`. Hence the issue of the global config file getting found when traversing up does not show.

src/cli/use.rs
```diff
@@ -254,7 +271,44 @@ mod tests {
     }
 
     #[test]
-    fn test_use_local_tool_versions() {
+    fn test_use_local_create() {
+        reset();
+        let _ = file::remove_file(env::current_dir().unwrap().join(".test-tool-versions"));
```
roele, line 276: we have to explicitly remove the .test-tool-versions as the test setup creates both

PR #2430, opened 2024-08-03
feat: allow installing cargo packages with `--git`

### LINKED ISSUE #2420: Install from cargo backend using `--git`
For one reason or another, some Rust projects are not published on crates.io. For example, [zenith](https://github.com/bvaisvil/zenith) does not have a crate. Instead, they ask their users to install via `--git` flag. I could not see any way to specify this using Mise.

It would be nice to be able to specify this as an option in the config file. Or better yet, support a similar syntax to the Go backend where it would just install using `--git` if it detects a path instead of a package name.

jdx:
Fixes #2420

### DIFF
--- a/src/cli/args/tool_arg.rs
+++ b/src/cli/args/tool_arg.rs
@@ -22,7 +22,7 @@ pub struct ToolArg {
 pub enum ToolVersionType {
     Path(PathBuf),
     Prefix(String),
-    Ref(String),
+    Ref(String, String),
     Sub { sub: String, orig_version: String },
     System,
     Version(String),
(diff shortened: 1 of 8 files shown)


src/cli/args/tool_arg.rs
```diff
@@ -57,7 +57,9 @@ impl FromStr for ToolVersionType {
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         Ok(match s.split_once(':') {
-            Some(("ref", r)) => Self::Ref(r.to_string()),
+            Some((ref_type @ ("ref" | "tag" | "branch" | "rev"), r)) => {
```
jdx, line 60: this is my first time ever using this syntax—I never saw the use-case until today!

PR #2477, opened 2024-08-19
docs(readme): remove failing green color

duhow:
image


jdx:
I don't like that it's a different shade of green from the other green badges

PR #2479, opened 2024-08-19
feat: mise hints

### LINKED ISSUE #2460: mise hints
I think it would be great to add some "hints" to mise—like homebrew has if you're familiar. Here's a few ideas:

```
$ mise use python@3.9
[mise hint] install multiple python versions simultaneously with `mise use python@3.9 python@3.10`
```

multiple versions is useful specifically for python

```
$ mise plugins
[mise hint] see available plugins with `mise registry`
```

I see a lot of confusion that mise is more than just the set of core plugins.

Of course we can add more as we go. We could back this with a setting, maybe for each one as well as a global one to disable all of them.

roele:
First draft for a very pragmatic approach on mise hints. Just adds a simple `hint!` macro with static args which respects the setting. Since i wasn't able to figure out yet how to have a multi-type setting i thought it might be handy to use `disable_hints= ["*"] ` as a way to disable all commands (IIRC we use `*` for running all tasks as well?)

It might be nicer to have something which we could pin to the run method such as `#[hint(...)]` (derive, proc_macro?). Not sure if Rust has something similar to Aspects in Java which allows you to run stuff after/before method calls etc.

Still a bit undecided about the styling. We probably should stick to a `dim` prefix like info messages but should highlight the command somehow with `bold`?

Screenshot 2024-08-20 at 00 25 26

Fixes #2460

### DIFF
--- a/schema/settings.json
+++ b/schema/settings.json
@@ -31,6 +31,14 @@
       "description": "disables built-in shorthands",
       "type": "boolean"
     },
+    "disable_hints": {
+      "description": "disable hints for mise commands",
+      "type": "array",
+      "items": {
+        "description": "command name or * for all commands",
+        "type": "string"
+      }
+    },
     "disable_tools": {
       "description": "tools that should not be used",
       "items": {
@@ -42,6 +50,34 @@
     "experimental": {
       "description": "enable experimental features",
       "type": "boolean"
+    },
+    "go_default_packages_file": {
+      "description": "path to file containing default go packages",
+      "type": "string"
+    },
+    "go_download_mirror": {
+      "description": "url to use as a mirror for go downloads",
+      "type": "string"
+    },
+    "go_repo": {
+      "description": "url to use as a mirror for go modules",
+      "type": "string"
+    },
+    "go_set_gopath": {
+      "description": "set GOPATH to the directory containing the go tool",
+      "type": "boolean"
+    },
+    "go_set_goroot": {
+      "description": "set GOROOT to the directory containing the go tool",
+      "type": "boolean"
+    },
+    "go_skip_checksum": {
+      "description": "skip checksum verification for go downloads",
+      "type": "boolean"
+    },
+    "http_timeout": {
+      "description": "timeout for http requests in seconds",
+      "type": "integer"
     },
     "jobs": {
       "description": "number of tools to install in parallel, default is 4",
@@ -58,6 +94,11 @@
         "type": "string"
       },
       "type": "array"
+    },
+    "libgit2": {
+      "description": "use libgit2 for git operations",
+      "type": "boolean",
+      "default": true
     },
     "node_compile": {
       "description": "do not use precompiled binaries for node",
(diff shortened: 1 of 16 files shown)


src/cli/install.rs
```diff
@@ -47,6 +47,11 @@ pub struct Install {
 
 impl Install {
     pub fn run(self) -> Result<()> {
+        hint!(
```
roele: hint should probably be moved into the match Some(runtime) statement

jdx:
I like the *!

jdx:
I think a great addition to this would be the current warning about precompiled python binaries which is probably annoying to people that already are aware of its limitations

schema/settings.json
```diff
@@ -31,6 +31,14 @@
       "description": "disables built-in shorthands",
```
jdx, line 31: We can actually probably just remove this file

roele: was already wondering if this is still useful since all settings can also be part of `.mise.toml`

roele:
@jdx tried to disable the hints for the tests but one test keeps failing and i can't figure out why. it passes in my IDE and when run via cargo test but not via nextest.

```
--- TRY 3 STDOUT:        mise::bin/mise cli::install::tests::test_install_nothing ---

running 1 test
test cli::install::tests::test_install_nothing ... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot file: src/cli/snapshots/mise__cli__install__tests__install_nothing.snap
Snapshot: install_nothing
Source: src/cli/install.rs:193
────────────────────────────────────────────────────────────────────────────────
Expression: output
────────────────────────────────────────────────────────────────────────────────
+new results
────────────┬───────────────────────────────────────────────────────────────────
          0 │+mise hint install multiple versions simultaneously with mise install python@3.8 python@3.9
────────────┴───────────────────────────────────────────────────────────────────
To update snapshots run `cargo insta review`
Stopped on the first failure. Run `cargo insta test` to run all snapshots.
FAILED

failures:

failures:
    cli::install::tests::test_install_nothing

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 233 filtered out; finished in 0.12s
```

roele:
Is it just me or does this feel a bit annoying and intrusive? How would we educate users about disabling the hints?
Maybe something interactive (confirmation/dialog) which disables it after confirmation would be easier to use and more tolerable?

src/plugins/core/python.rs
```diff
@@ -154,9 +158,12 @@ impl PythonPlugin {
             }
         };
 
-        warn!("installing precompiled python from indygreg/python-build-standalone");
-        warn!("if you experience issues with this python (e.g.: running poetry), switch to python-build");
-        warn!("by running: mise settings set python_compile 1");
+        hint!(
+            "install",
```
roele, line 162: i wonder if we need more fine grained control here

jdx:
yeah I think we should show a second line like: `mise hint disable this thing with "mise settings set ..."`

roele:
it's a bit unfortunate that array settings are not additive, so we need to include existing settings `mise settings set disable_hints install,plugins` otherwise it would overwrite a `disable_hints = ["plugins"]` instead adding to it.

jdx:
yeah maybe at some point we could add a `mise settings add disable_hints` or something

PR #2505, opened 2024-08-27
Use HTTPS for downloading precompiled pythons

maxnordlund:
In this day and age _not_ using TLS is just begging for malicious actors to inject malware. This fixes that by adding an `s` (and removing the comment).

I couldn't measure any real difference between HTTP and HTTPS (~10 ms), so I wonder how you, @jdx, came to that conclusion?

### DIFF
--- a/src/plugins/core/python.rs
+++ b/src/plugins/core/python.rs
@@ -104,8 +104,7 @@ impl PythonPlugin {
     fn fetch_precompiled_remote_versions(&self) -> eyre::Result<&Vec<(String, String, String)>> {
         self.precompiled_cache.get_or_try_init(|| {
             let settings = Settings::get();
-            // using http is not a security concern and enabling tls makes mise significantly slower
-            let raw = HTTP_FETCH.get_text("http://mise-versions.jdx.dev/python-precompiled")?;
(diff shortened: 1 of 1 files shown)


jdx:
this is just selecting version numbers, it's not possible to inject malwayre

maxnordlund:
Massive compared to what? Downloading the python binary/ies will dwarf this, unless how course the list of versions is much larger then the binaries themselves.

If it can be hijacked, it can be used to inject malware. I know we cannot know how something can be exploited now or in the  future. So it's better to prevent any and all injection.

If you really think you need those 10 ms (and to be clear HTTP was not strictly faster than HTTPS, on my machine/connection it varied), add an environmental variable. Maybe `MISE_USE_INSECURE_CONNECTION`? I don't know enough rust to implement it, but I imagine reading an environmental variable is easy.

jdx:
Version checking is done for commands other than installs

jdx:
If you'd like me to reconsider show me how this could be an attack vector.

maxnordlund:
left-pad

maxnordlund:
You didn't answer my question on how you measured the download speed difference? Do you have a script I can run so we can compare?

PR #2509, opened 2024-08-27
feat: add support for using uv tool (uvx) as a replacement for pipx

### LINKED ISSUE #2506: Add support for uvx backend
[uvx](https://docs.astral.sh/uv/concepts/tools/) is a  `pipx` replacement built in Rust (on top of `uv`). I think for people using `uv` as the package manager, it makes sense to also use `uvx` for tools.

I can contribute to this feature, but I may need some guidance on best practices for this use case. I think the implementation would be pretty similar to the `pipx` backend, so I am not sure if adding a new backend is recommended, or maybe we can extend the `pipx` module.

antoniomdk:
Closes #2506

### DIFF
--- a/src/backend/pipx.rs
+++ b/src/backend/pipx.rs
@@ -9,6 +9,7 @@ use crate::cache::CacheManager;
 use crate::cli::args::BackendArg;
 use crate::cmd::CmdLineRunner;
 use crate::config::{Config, Settings};
+use crate::file;
 use crate::github;
 use crate::http::HTTP_FETCH;
 use crate::install_context::InstallContext;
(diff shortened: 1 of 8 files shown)


src/backend/pipx.rs
```diff
@@ -84,23 +85,42 @@ impl Backend for PIPXBackend {
...
-            .with_pr(ctx.pr.as_ref())
-            .env("PIPX_HOME", ctx.tv.install_path())
-            .env("PIPX_BIN_DIR", ctx.tv.install_path().join("bin"))
-            .envs(ctx.ts.env_with_path(&config)?)
-            .prepend_path(ctx.ts.list_paths())?
-            // Prepend install path so pipx doesn't issue a warning about missing path
-            .prepend_path(vec![ctx.tv.install_path().join("bin")])?
-            .execute()?;
-
+        if self.is_uvx_enabled() {
```
antoniomdk, line 88: For the moment, there's not much discrepancies between `uv tool` and `pipx`, we can refactor it so most of the args are added outside the if-else conditions. In the future, there may be more differences and we may need to roll back to this version.

src/cli/settings/set.rs
```diff
@@ -48,6 +48,7 @@ impl SettingsSet {
             "node_compile" => parse_bool(&self.value)?,
             "not_found_auto_install" => parse_bool(&self.value)?,
             "paranoid" => parse_bool(&self.value)?,
+            "pipx_uvx" => parse_bool(&self.value)?,
```
antoniomdk, line 51: Not sure if this is the best name, or if we can use something like `python_use_uvx`. There are other issues associated with using `uv` for venv management, so I just wanted to avoid conflicting names.

jdx:
I think you missed the JSON schema

antoniomdk:
@jdx Thanks a lot for the prompt review. It should be fixed now :)

PR #2510, opened 2024-08-27
feat(pipx): add support for specifying package extras

### LINKED ISSUE #2491: pipx: support features
pipx has a concept of optional features upon package installation, via `pipx install package[feature,…]`.  This is used, for example, by [Harlequin](https://harlequin.sh/docs/getting-started).

To my knowledge, mise has no support for this.  Would it be possible to add it?

As a work-around, for the example of Harlequin, one can install things via `pip install harlequin parlequin-postgress` etc. instead.

antoniomdk:
Closes #2491

Some Python tools support optional features upon package installation, via `pipx install package[feature,…]`. This is used, for example, by [Harlequin](https://harlequin.sh/docs/getting-started).

This PR adds support for specifying extras via

```toml
mytool = { version = '3.10', extras = 'foo,bar' }
```

NOTE: It only works for tools published to a registry (e.g PyPI), git-based tools require a different syntax than the one already implemented. (`git+...#egg=project[...]`).

NOTE: I had some issues setting up the local docker environment (unrelated to this project), so I couldn't run the e2e tests. There may be other parts of the code that could complain if using `[, ]` in the tool name, but the PyPI resolution error is the first one that popped up when trying to use package extras.

### DIFF
--- a/e2e/backend/test_pipx_extras
+++ b/e2e/backend/test_pipx_extras
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+require_cmd python3
+
+# Create a system pipx that always fail and push it to the front of PATH
+cat >"$HOME/bin/pipx" <<'EOF'
+#!/usr/bin/env bash
+echo "CALL TO SYSTEM pipx! args: $*" >&2
+exit 1
+EOF
+chmod +x "$HOME"/bin/pipx
+export PATH="$HOME/bin:$PATH"
+
+# Just to be sure...
+assert_fail "pipx"
+
+# Use precompiled python
+export MISE_PYTHON_COMPILE=0
+
+# Set up a 2-step installation: pipx@1.5.0 > pipx:mkdocs@1.6.0
+cat >.mise.toml <<EOF
+[tools]
+pipx = "1.5.0"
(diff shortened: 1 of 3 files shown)


jdx:
I think Tool Options would be a better solution as it wouldn't require inventing a new convention https://mise.jdx.dev/dev-tools/#tool-options

antoniomdk:
@jdx How would you propose the syntax for tool options in this case? Python extras are a list rather than a map, maybe something like 
```
mytool = { version = '3.10', extras = ['bar'] }
```

jdx:
I think for now unfortunately it only supports strings so you will need to do this:

```toml
mytool = { version = '3.10', extras = 'foo,bar' }
```

antoniomdk:
That's fine, I'll update the PR to accommodate the tool options convention!

antoniomdk:
@jdx I've updated the code to use `ToolVersions`. Not sure if we have a way to add validation for the extra field, but probably the pipx/uvx error will be informative enough.

jdx:
you've got a test fialure

antoniomdk:
@jdx this is ready for review. The git history is kind of messy, let me know if you want me to clean it up before merging 👍

Issue #2504, opened 2024-08-27
mise run fails when adding env directives


mdesouky:
**Describe the bug**
When adding a python env line to  `[env]` block in `.mise.toml` file as follows:
```
[env]
_.file = 'server/.env'
_.python.venv = { path = ".venv", create = true } # create the venv if it doesn't exist
```
and trying to run tasks I get the following error
```
The application panicked (crashed).
Message:  called `Result::unwrap()` on an `Err` value: JoinPathsError { inner: JoinPathsError }
Location: src/path_env.rs:50
```
Full trace
```
❯ RUST_BACKTRACE=full mise run install
The application panicked (crashed).
Message:  called `Result::unwrap()` on an `Err` value: JoinPathsError { inner: JoinPathsError }
Location: src/path_env.rs:50

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                                ⋮ 9 frames hidden ⋮                               
  10: mise::toolset::Toolset::env_with_path::ha352fc7531460682
      at <unknown source file>:<unknown line>
  11: mise::cli::run::Run::run::h58120052c1c0853f
      at <unknown source file>:<unknown line>
  12: mise::cli::Cli::run::hf6cb0a1443c80811
      at <unknown source file>:<unknown line>
  13: mise::main::h4b059410b007a228
      at <unknown source file>:<unknown line>
  14: std::sys_common::backtrace::__rust_begin_short_backtrace::h59d757c488a0c1a5
      at <unknown source file>:<unknown line>
  15: std::rt::lang_start::{{closure}}::hc34123bf95d8d56d
      at <unknown source file>:<unknown line>
  16: std::panicking::try::h0cf3a98f51cee1e9
      at <unknown source file>:<unknown line>
  17: std::rt::lang_start_internal::h07dd9d9140cc1a1e
      at <unknown source file>:<unknown line>
  18: _main<unknown>
      at <unknown source file>:<unknown line>

Run with COLORBT_SHOW_HIDDEN=1 environment variable to disable frame filtering.
```
**To Reproduce**

- Add the following line under the `[env]` block

`_.python.venv = { path = ".venv", create = true }`

**Expected behavior**
Mise tasks should run as normal which happens when you remove that line

**`mise doctor` output**

```text
❯ mise doctor     
version: 2024.7.4 macos-arm64 (2024-07-19)
[...]

PR #2511, opened 2024-08-27
fix(src/path_env.rs): Issue 2504: Fix for JoinPathsError

mcallaway:
The original submitter of Issue 2504 triggered the JoinPathsError when adding a python env in the [env] section:

    _.python.venv = { path = ".venv", create = true } # create the venv if it doesn't exist

I noticed the same error independently when using the `poetry` plugin and attempting to run `[tasks]`. Consider the following `.mise.toml` file that works fine:

```
[tools]
[tools.poetry]
version = 'latest'
pyproject = 'pyproject.toml'

[tools.python]
version = '3.11'
# Note this is commented out
#virtualenv = '.venv'

[tasks.foo]
run = "echo foo"
```

Running the `foo` task works:

```
❯ mise r foo
[foo] $ echo foo
foo
```

But now if I activate that `virtualenv` config line:

```
[tools]
[tools.poetry]
version = 'latest'
pyproject = 'pyproject.toml'

[tools.python]
version = '3.11'
virtualenv = '.venv'

[tasks.foo]
run = "echo foo"
```

Now running the task triggers the same JoinPathsErr:

```
❯ mise r foo
Message:  called `Result::unwrap()` on an `Err` value: JoinPathsError { inner: JoinPathsError }
Location: src/path_env.rs:50

```

In exploring the bug I observed that path_env.add() was being called with two copies of the `.venv` directory concatenated with a colon ":". So my fix is to update the add() function to check for this, split the path argument on ":" and add each element.

My proposed fix is in `./src/path_env.rs`

### DIFF
--- a/src/path_env.rs
+++ b/src/path_env.rs
@@ -139,4 +139,20 @@ mod tests {
             format!("/1:/2:/3:/before-1:/before-2:/before-3:/after-1:/after-2:/after-3")
         );
     }
+    #[test]
+    fn test_path_env_with_colon() {
+        reset();
+        let mut path_env = PathEnv::from_iter(
+            [
+                "/before1",
+                "/before2"
+            ]
+            .map(PathBuf::from),
+        );
+        path_env.add("/after1:/after2".into());
+        assert_eq!(
+            path_env.to_string(),
+            format!("/before1:/before2:/after1:/after2")
+        );
+    }
 }
--- a/src/path_env.rs
+++ b/src/path_env.rs
@@ -27,7 +27,15 @@ impl PathEnv {
     }
 
     pub fn add(&mut self, path: PathBuf) {
-        self.mise.push(path);
+        let path_string: &str = &path.to_string_lossy();
+        if path_string.contains(':') {
+            let parts: Vec<&str> = path_string.split(':').collect();
+            for part in parts {
+                self.mise.push(PathBuf::from(part));
+            }
+        } else {
+            self.mise.push(path);
+        }
     }
 
     pub fn to_vec(&self) -> Vec<PathBuf> {
@@ -144,15 +152,15 @@ mod tests {
         reset();
         let mut path_env = PathEnv::from_iter(
             [
-                "/before1",
-                "/before2"
+                "/item1",
+                "/item2"
             ]
             .map(PathBuf::from),
         );
-        path_env.add("/after1:/after2".into());
+        path_env.add("/1:/2".into());
         assert_eq!(
             path_env.to_string(),
-            format!("/before1:/before2:/after1:/after2")
+            format!("/1:/2:/item1:/item2")
         );
     }
 }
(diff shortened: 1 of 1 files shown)


mcallaway:
This PR addresses jdx/mise#2504

src/path_env.rs
```diff
@@ -27,7 +27,15 @@ impl PathEnv {
     }
 
     pub fn add(&mut self, path: PathBuf) {
-        self.mise.push(path);
+        let path_string: &str = &path.to_string_lossy();
+        if path_string.contains(':') {
```
jdx: This conditional isn't necessary if I'm reading the code right, if there are no colons then splitting would just give us a Vec of length 1

src/path_env.rs
```diff
@@ -139,4 +147,11 @@ mod tests {
             format!("/1:/2:/3:/before-1:/before-2:/before-3:/after-1:/after-2:/after-3")
         );
     }
+    #[test]
+    fn test_path_env_with_colon() {
+        reset();
+        let mut path_env = PathEnv::from_iter(["/item1", "/item2"].map(PathBuf::from));
+        path_env.add("/1:/2".into());
```
jdx, line 149: I was concerned that we may need to handle escaping colons but my reading is that escaping colons actually is not allowed. So nothing to be done here but if anyone else wants to confirm this I would appreciate the input.

jdx (changes requested):

src/path_env.rs
```diff
@@ -27,7 +27,15 @@ impl PathEnv {
     }
 
     pub fn add(&mut self, path: PathBuf) {
-        self.mise.push(path);
+        let path_string: &str = &path.to_string_lossy();
```
jdx: I think we should use https://doc.rust-lang.org/std/env/fn.split_paths.html that way you don't need to convert to lossy

jdx:
Some minor feedback but this seems like a solid patch!

src/path_env.rs
mcallaway: The source of the bug is that we are in some cases that I do not understand seeing "path" arguments that *do* contain colons. Something is calling:

    path_env.add("/Users/me/git/code/.venv:/Users/me/git/code/.venv")

This isn't what I would expect, but it happens. So this conditional says, "If there's a colon, split the item on the colon and then handle each element. If there's no colon, do the normal thing.

src/path_env.rs
mcallaway: This test is ensuring that if some caller says `path_env.add("foo:bar")` that the right thing happens, even though we don't expect that to be the normal case.

src/path_env.rs
mcallaway: I see what you mean. I pushed an update.

src/path_env.rs
mcallaway: I think I see what you mean. I pushed an update.

jdx (approved):

PR #2513, opened 2024-08-27
feat(zig): make dev builds installable

buraksirma:
Hello, I need to switch between dev builds of `zig` because of the compatibility issues. I made dev-builds not listed on `ls-remote` command since there are many of them. I don't have rust experience, sorry if I mutated some function that I should not.

You can see the commands I executed for testing purposes below.
image


jdx:
closing in favor of #2514

Issue #2257, opened 2024-06-05
When will "mise install" not install tools?


bhollis:
This has been happening on our GitHub runners and I'm not sure why. We clearly have `ruby@3.3.0` in the set of tools to be installed, `mise install` says `all  runtimes are installed`, but then `mise ls` says Ruby is missing:

```
Run mise install --verbose
[DEBUG] ARGS: mise install --verbose
[DEBUG] Config {
    Config Files: [
        "~/actions-runner/_work/stately/stately/.mise.toml",
        "~/actions-runner/_work/stately/stately/.ruby-version",
        "~/actions-runner/_work/stately/stately/.nvmrc",
        "~/.config/mise/config.toml",
    ],
}
[DEBUG] ToolRequestSet.build(102.106µs): ToolRequestSet: cargo:cargo-binstall@latest node@20.13.1 ruby@3.3.0 go@1.22.3 java@temurin-21 cargo:cargo-cranky@0.3.0 cargo:protoc-gen-prost-crate@0.3.1 golangci-lint@1.58.1 go:github.com/segmentio/golines@0.12.2 go:github.com/drshriveer/gtools/genum/cmd/genum@v0.0.0-20240314211858-40dccd828943 cargo:cross@0.2.5 cargo:wasm-bindgen-cli@0.2.92 buf@1.32.2 sccache@0.8.1 npm:ts-proto@1.176.1 npm:@bufbuild/protoc-gen-es@1.10.0 go:google.golang.org/protobuf/cmd/protoc-gen-go@1.34.1 go:github.com/planetscale/vtprotobuf/cmd/protoc-gen-go-vtproto@0.6.0 cargo:protoc-gen-prost-serde@0.3.0 cargo:protoc-gen-tonic@0.4.0 cargo:protoc-gen-prost@0.3.1

[INFO] mise all runtimes are installed
Run mise ls
buf                                                             1.32.2                              ~/actions-runner/_work/stately/stately/.mise.toml    1.32.2                            
cargo:cargo-binstall                                            1.6.8                               ~/.config/mise/config.toml                           latest                            
cargo:cargo-cranky                                              0.3.0                               ~/actions-runner/_work/stately/stately/.mise.toml    0.3.0                             
cargo:cross                                                     0.2.5                               ~/actions-runner/_work/stately/stately/.mise.toml    0.2.5                             
cargo:protoc-gen-prost                                          0.3.1                               ~/actions-runner/_work/stately/stately/.mise.toml    0.3.1                             
[...]

PR #2524, opened 2024-08-29
fix: ensure `mise install` installs missing runtimes listed in `mise ls`

stanhu:
Previously `mise ls` could show missing runtimes, but `mise install` would report all runtimes were installed. The latter only checked if the install directory existed, which could be present if `mise install` aborted early. However, the former also checked the existence of the runtime symlink and the `incomplete` file in the cache.

To make this behavior consistent, dispatch the tool request to the backend. Exclude the runtime symlink check since this breaks the install check for `node@latest`.

Closes https://github.com/jdx/mise/issues/2257

### DIFF
--- a/src/toolset/tool_request.rs
+++ b/src/toolset/tool_request.rs
@@ -128,11 +128,10 @@ impl ToolRequest {
     }
 
     pub fn is_installed(&self) -> bool {
-        // TODO: dispatch to backend
-        match self {
-            Self::System(_) => true,
-            _ => self.install_path().is_some_and(|p| p.exists()),
-        }
+        let backend = backend::get(self.backend());
+        let tv = ToolVersion::new(backend.as_ref(), self.clone(), self.version());
+
+        backend.is_version_installed(&tv, false)
     }
 
     pub fn install_path(&self) -> Option<PathBuf> {
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -219,13 +219,15 @@ pub trait Backend: Debug + Send + Sync {
             false => vec![],
         })
     }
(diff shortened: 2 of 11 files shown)


stanhu:
Looks like https://github.com/jdx/mise/pull/2524 is failing for the NPM backend.

src/toolset/tool_request.rs
```diff
@@ -128,11 +128,10 @@ impl ToolRequest {
...
     pub fn is_installed(&self) -> bool {
-        // TODO: dispatch to backend
-        match self {
-            Self::System(_) => true,
-            _ => self.install_path().is_some_and(|p| p.exists()),
-        }
+        let backend = backend::get(self.backend());
+        let tv = ToolVersion::new(backend.as_ref(), self.clone(), self.version());
+
+        backend.is_version_installed(&tv)
```
stanhu: @jdx This appears to break the NPM test that runs `mise x npm:prettier@3.1.0 -- prettier -v` since `node@latest` does not appeared to be installed since it has a runtime symlink: https://github.com/jdx/mise/blob/ba90c3bbe71bd33d628df607326da9f0cf363af1/src/backend/mod.rs#L228

I'm not completely sure why that symlink check is needed. Another alternative might be add a separate method for `is_version_avaiable` that excludes the symlink check?

jdx: Yeah this is tricky for this exact reason. There are situations where we want to differentiate and some we don't. Let me do a bit of auditing and figure out what we should do.

jdx: Without diving too deep I'm thinking this could be a flag for the existing install check

stanhu: Thanks, I added the flag.

Issue #2458, opened 2024-08-15
Ubi fails to install a binary when the bin path does not exist


qbedard:
**Describe the bug**

Ubi fails to install binary because the bin path does not exist. This probably means that mise needs to create the path before running Ubi.

```
mise failed to execute command: ubi --in /Users/[user]/.local/share/mise/installs/ubi-evilmartians-lefthook/1.7.12/bin --project evilmartians/lefthook --tag 1.7.12
mise No such file or directory (os error 2)
mise Run with --verbose or MISE_VERBOSE=1 for more information
```

**To Reproduce**

```toml
[tools]
rust = "latest"
"cargo:ubi" = "latest"
"ubi:evilmartians/lefthook" = "latest"

[settings]
experimental = true
```

**Expected behavior**

Successfully installs lefthook.

**`mise doctor` output**

```text
version: 2024.8.6 macos-x64 (2024-08-12)
activated: yes
shims_on_path: no

build_info: 
  Target: x86_64-apple-darwin
  Features: DEFAULT, NATIVE_TLS
  Built: Mon, 12 Aug 2024 07:50:25 +0000
  Rust Version: rustc 1.80.1 (3f5fd8dd4 2024-08-06) (Homebrew)
  Profile: release

shell: 
  fish
  fish, version 3.7.1

dirs: 
  data: ~/.local/share/mise
  config: ~/.config/mise
  cache: ~/Library/Caches/mise
  state: ~/.local/state/mise
  shims: ~/.local/share/mise/shims

config_files: 
  ~/.config/mise/config.toml
  ~/ghq/github.com/[user]/dotfiles/.mise.toml

backends: 
  cargo
  core
  go
  npm
  pipx
  spm
  ubi

plugins: 
  action-validator  https://github.com/mpalmer/action-validator.git#b54fac9
  bun               (core)
  deno              (core)
  erlang            (core)
  go                (core)
  java              (core)
  node              (core)
  python            (core)
  ruby              (core)
  rust              https://github.com/code-lever/asdf-rust.git#95acf4f
  terraform         https://github.com/asdf-community/asdf-hashicorp.git#197e3ec
  usage             https://github.com/jdx/mise-usage.git#fe3888a
  zig               (core)

toolset: 
  go@1.22.6             
  node@20.16.0          
  node@18.20.4          
  python@3.11.9         
  rust@1.80.1           
  terraform@1.9.4       
  terraform@1.5.1       
  usage@0.3.0           
[...]

Issue #2457, opened 2024-08-15
Pipx install instructions not working: no such file or directory


eproxus:
**Describe the bug**
When following [the instructions](https://mise.jdx.dev/dev-tools/backends/pipx.html#dependencies) to get the `pipx` backend working, an `No such file or directory (os error 2)` error is issued when trying to install something.

**To Reproduce**

```console
$ mise use -g python
$ pip install --user pipx
$ mise install pipx:psf/black
mise failed to execute command: pipx install git+https://github.com/psf/black.git@22.1.0
mise No such file or directory (os error 2)
mise Run with --verbose or MISE_VERBOSE=1 for more information
```

This is also true for normal PyPi packages.

**Expected behavior**
Packages install correctly.

**`mise doctor` output**

```text
version: 2024.8.6 macos-arm64 (2024-08-12)
activated: yes
shims_on_path: no

build_info: 
  Target: aarch64-apple-darwin
  Features: DEFAULT, NATIVE_TLS
  Built: Mon, 12 Aug 2024 07:50:25 +0000
  Rust Version: rustc 1.80.1 (3f5fd8dd4 2024-08-06) (Homebrew)
  Profile: release

shell: 
  /opt/homebrew/bin/fish
  fish, version 3.7.1

dirs: 
  data: ~/.local/share/mise
  config: ~/.config/mise
  cache: ~/Library/Caches/mise
  state: ~/.local/state/mise
  shims: ~/.local/share/mise/shims

config_files: 
  ~/.config/mise/config.toml

backends: 
  cargo
  core
  go
  npm
  pipx
  spm
  ubi

plugins: 
  bun        (core)
  deno       (core)
  elixir     https://github.com/asdf-vm/asdf-elixir.git#a4c42e1
  erlang     (core)
  go         (core)
  java       (core)
  node       (core)
  php        https://github.com/asdf-community/asdf-php.git#1eaf4de
  python     (core)
  rebar      https://github.com/Stratus3D/asdf-rebar.git#4af0770
  ruby       (core)
  usage      https://github.com/jdx/mise-usage.git#fe3888a
  watchexec  https://github.com/nyrst/asdf-watchexec.git#b667a08
  zig        (core)

toolset: 
  bun@1.1.22            
  elixir@1.17.2-otp-27  
  erlang@27.0.1         
  erlang@26.2.5.2       
  erlang@25.3.2.13      
  go@1.22.6             
  node@22.6.0           
  python@3.12.5         
[...]

Issue #2315, opened 2024-06-20
mise fails to install rust library using cargo-binstall


lunks:
**Describe the bug**
When I try to install any cargo plugin via cargo-binstall, it fails. Using `cargo-binstall` directly works fine.

**To Reproduce**
This is my config.toml:
```toml
[settings]
experimental = true
cargo_binstall = true
[tools]
1password-cli = "latest"
awscli = "latest"
aws-copilot = "latest"
aws-sso-cli = "latest"
aws-vault = "latest"
bat = "latest"
bottom = "latest"
"cargo:cargo-binstall" = "latest"
"cargo:cargo-update" = "latest"
"cargo:eza" = "latest"
"cargo:fd-find" = "latest"
"cargo:ripgrep" = "latest"
delta = "latest"
github-cli = "latest"
node = "latest"
"npm:cdk" = "latest"
"pipx:powerline-status" = "latest"
python = "latest"
rust = "latest"
usage = "latest"
```
Run `mise -i cargo:csvlens@latest`

**Expected behavior**

It should install csvlens

**`mise doctor` output**

```text
version: 2024.6.6 linux-x64 (409d6e4 2024-06-20)
activated: yes
shims_on_path: no

build_info: 
  Target: x86_64-unknown-linux-gnu
  Features: DEFAULT, NATIVE_TLS, OPENSSL
  Built: Thu, 20 Jun 2024 16:13:11 +0000
  Rust Version: rustc 1.79.0 (129f3b996 2024-06-10)
  Profile: release

shell: 
  /usr/bin/zsh
  zsh 5.9 (x86_64-pc-linux-gnu)

dirs: 
  data: ~/.local/share/mise
  config: ~/.config/mise
  cache: ~/.cache/mise
  state: ~/.local/state/mise
  shims: ~/.local/share/mise/shims

config_files: 
  ~/.tool-versions
  ~/.config/mise/config.toml
  ~/code/allegro/.tool-versions

backends: 
  cargo
  core
  go
  npm
  pipx
  spm
  ubi

plugins: 
  1password-cli  https://github.com/NeoHsu/asdf-1password-cli.git#f5d5aab
  amass          https://github.com/dhoeric/asdf-amass.git#17cec1f
  aws-copilot    https://github.com/NeoHsu/asdf-copilot#1341900
  aws-sso-cli    https://github.com/adamcrews/asdf-aws-sso-cli.git#169a1a3
[...]

Issue #2174, opened 2024-05-23
`mise install` fails to install missing `npm:...` package


xavdid:
**Describe the bug**

When the active configuration file lists an npm package that isn't installed, `mise install` fails with an error:

```
Error:
   0: failed to execute command: npm install -g tldr@3.4.0 --prefix /Users/david/.local/share/mise/installs/npm-tldr/3.4.0
   1: No such file or directory (os error 2)

Location:
   src/cmd.rs:242

Version:
   2024.5.21 macos-arm64 (2024-05-23)

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```

**To Reproduce**

1. In an empty directory, create the following config:

```toml
[tools]
"npm:tldr" = "latest"

[settings]
# needed for npm packages
experimental = true
```

2. run `mise install`

**Expected behavior**

The `npm` package `tldr` is installed and the `tldr` command works.

**`mise doctor` output**

```text
version: 2024.5.21 macos-arm64 (2024-05-23)
activated: yes
shims_on_path: no

build_info: 
  Target: aarch64-apple-darwin
  Features: DEFAULT, NATIVE_TLS
  Built: Thu, 23 May 2024 03:16:31 +0000
  Rust Version: rustc 1.78.0 (9b00956e5 2024-04-29) (Homebrew)
  Profile: release

shell: 
  /bin/zsh
  zsh 5.9 (x86_64-apple-darwin22.0)

dirs: 
  data: ~/.local/share/mise
  config: ~/.config/mise
  cache: ~/Library/Caches/mise
  state: ~/.local/state/mise
  shims: ~/.local/share/mise/shims

config_files: 
  ~/.config/mise/config.toml
  ~/Desktop/tt/mise.toml

backends: 
  cargo
  core
  go
  npm
  pipx
  ubi

plugins: 
  bun     (core)
  deno    (core)
  elixir  https://github.com/mise-plugins/mise-elixir.git#7b06439
  erlang  (core)
  go      (core)
  java    (core)
  node    (core)
  python  (core)
  ruby    (core)
  zig     (core)

toolset: 
  npm:tldr@3.4.0        (missing)
  python@3.12.3         
  node@20.13.1          
  go@1.22.3             
  elixir@1.16.3-otp-26  
  npm:trash-cli@5.0.0   
  npm:epub-wordcount@2.1.2  
  npm:will-it-play@1.0.1  

[...]

PR #2532, opened 2024-09-02
fix: Ensure dependencies are available for alternative backends

xavdid:
When installing a tool via one of the experimental backends (`npm`, `pipx`, etc), the install would fail if the backend was managed by `mise` itself. 

When setting up the paths for the installation command (`npm install tldr`) to run, `mise` wasn't making the backends specified in `get_dependencies` available to the subshell. So, they'd fail with a cryptic error:

```
No such file or directory (os error 2)
```

This was because `npm` wasn't in the path, despite it being installed. 

I restructured some existing code to make creating a Toolset for a backend's dependents simple, then prepended those paths to the resulting command. I've confirmed locally that installations commands that fail without the

I also added a line to an e2e test, but I wasn't able to get those running locally in docker, so it's worth double checking that works as intended before merge.

---

fixes https://github.com/jdx/mise/issues/2174 
fixes https://github.com/jdx/mise/issues/2315
fixes https://github.com/jdx/mise/issues/2457
fixes https://github.com/jdx/mise/issues/2458

### DIFF
--- a/e2e/backend/test_npm
+++ b/e2e/backend/test_npm
@@ -6,3 +6,4 @@ export NPM_CONFIG_FUND=false
 mise use node
 assert "mise x npm:prettier@3.1.0 -- prettier -v" "3.1.0"
 assert "FORCE_COLOR=0 mise x npm:@antfu/ni@0.21.12 -- ni -v 2>/dev/null | head -n1" "@antfu/ni  v0.21.12"
+assert "mise install npm:tldr"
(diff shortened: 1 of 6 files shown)


jdx:
I believe this has caused a failure in e2e/backend/test_go_install_slow, currently investigating why

xavdid:
Looks like that release branch was failing before this was merged (based on [this CI run](https://github.com/jdx/mise/actions/runs/10658542241)), which makes me think it's https://github.com/jdx/mise/pull/2524 instead. The test [uses a symlink](https://github.com/jdx/mise/blob/732fc58deda43339e5dd0e5136c5b71dab275232/e2e/backend/test_go_install_slow#L10) and that changed how symlinks are handled across a number of operations.

Like I said, I wasn't able to get the e2e tests running locally using docker (mise wouldn't compile; `mise run docker:cargo test` worked fine though), so it's possible it's here! But nothing in this PR would have touched `go`- only the changed backends are affected at all.

jdx:
yep you are correct, verified via bisect

PR #2561, opened 2024-09-11
feat: implement a few tera functions for mise toml config

erickguan:
I implemented these functions:
- arch
- num_cpus
- os
- os_family
- invocation_directory
- quote
- kebabcase
- lowercamelcase
- shoutykebabcase
- shoutysnakecase
- snakecase

Since there are difference from `just`, I figure it's good to check in before I continue implementing more and add documentation.

1. `PlatformInfo` provides "runtime" information as `uname`. `just` returns `target` information. I don't agree on correctness since it will break if you cross-compile. It's always a hassle for package managers. So `PlatformInfo` API provides a different output than `just`'s.
   I took inspiration from [YADM's template design](https://yadm.io/docs/templates). I provide the result of `arch`, `os`, ... as is. This behavior aligns with `direnv` where `direnv` exposes no arch information. It's `direnv`'s users job to parse platform information. And for mise, users can use tera filter to transform text.
   What do you think of this?
3. `invocation_directory` is annoying to fix. `BASE_CONTEXT` is initialized lazily so I have difficulty testing other functions (I wrote one test per one custom function). We can come back to this later or refactor the context a bit. I prefer skipping inconsistency for now.
4. `just` implements functions like `kebabcase` as function. I implement them as tera filters. Any objections?
5. Where should I add documentation? Also, any other thoughts?

Part of #2049

### DIFF
--- a/src/tera.rs
+++ b/src/tera.rs
@@ -75,7 +75,7 @@ pub fn get_tera(dir: Option<&Path>) -> Tera {
     tera.register_function(
         "invocation_directory",
         move |_args: &HashMap<String, Value>| -> tera::Result<Value> {
-            let path = env::current_dir().unwrap_or(PathBuf::default());
+            let path = env::current_dir().unwrap_or_default();
 
             let result = String::from(path.to_string_lossy());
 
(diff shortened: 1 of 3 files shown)


src/tera.rs
```diff
@@ -78,12 +124,257 @@ pub fn get_tera(dir: Option<&Path>) -> Tera {
...
+    #[test]
+    #[cfg(target_family = "unix")]
+    fn test_render_with_custom_function_os_family_unix() {
+        let mut tera = get_tera(Option::default());
+
+        let result = tera
+            .render_str("{{ os_family() }}", &Context::default())
+            .unwrap();
+
+        assert_eq!("Linux", result);
```
jdx, line 260: already merged this, but it probably could've been DRYed up with `if cfg!()` macros

PR #2576, opened 2024-09-14
feat: add blake & uuid func

erickguan:
Part of #2049


jdx:
I don't this feature is worth adding crates for, if we were already using them it'd be fine but I don't think this would be used enough to warrant adding a dependency for either

PR #2578, opened 2024-09-15
feat: add semver matching in mise templates

erickguan:
Part of #2049

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,6 +92,7 @@ self_update = { version = "0.41", default-features = false, features = [
     "compression-flate2",
     "signatures",
 ] }
+semver = "1"
 serde = "1.0.199"
 serde_derive = "1.0.199"
 serde_ignored = "0.1"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -92,6 +92,7 @@ self_update = { version = "0.41", default-features = false, features = [
     "compression-flate2",
     "signatures",
 ] }
+semver = "1"
```
jdx: we already use the `versions` crate for semver comparisons

docs/templates.md
```diff
@@ -52,6 +52,7 @@ And these testers:
 - `if path is dir` – if the path is a directory
 - `if path is file` – if the path is a file
 - `if path is exists` – if the path exists
+- `if version is semver_matching` - if a version matches semantically (e.g. `if "1.10.2" is semver_matching("~ 1.10.0")` ).
```
jdx:
```suggestion
- `if version is semver_matching(String)` - if a version matches semantically (e.g. `if "1.10.2" is semver_matching("~ 1.10.0")` ).
```

Cargo.toml
erickguan: Ok.

PR #2582, opened 2024-09-15
feat: add rest of tera features for templates

### LINKED ISSUE #2049: implement more tera functions

erickguan:
Fixes #2049

I missed a few XDG paths and `absolute`. Ping me if we want the rest of XDG directories in mise.

For `absolute`, do we wait rust stable feature to add it? (I left a TODO in the `tera.rs`)

### DIFF
--- a/src/tera.rs
+++ b/src/tera.rs
@@ -22,6 +22,10 @@ pub static BASE_CONTEXT: Lazy<Context> = Lazy::new(|| {
     if let Ok(dir) = env::current_dir() {
         context.insert("cwd", &dir);
     }
+    context.insert("xdg_cache_home", &*env::XDG_CACHE_HOME);
+    context.insert("xdg_config_home", &*env::XDG_CONFIG_HOME);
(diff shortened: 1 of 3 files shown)


src/tera.rs
```diff
@@ -72,6 +80,58 @@ pub fn get_tera(dir: Option<&Path>) -> Tera {
...
+                    let n = n.as_u64().unwrap();
+                    match args.get("alphabet") {
+                        Some(Value::String(alphabet)) => {
+                            let alphabet = alphabet.chars().collect::<Vec<char>>();
+                            let mut rng = thread_rng();
+                            let result =
+                                (0..n).map(|_| alphabet.choose(&mut rng).unwrap()).collect();
+                            Ok(Value::String(result))
+                        }
+                        _ => Err("choice alphtbet must be an string".into()),
```
roele: a typo seems to have sneaked in here

erickguan: Fixed!

PR #2586, opened 2024-09-16
feat: add Atmos

mtweeman:
Atmos was already added in #2577, but removed in #2573

### DIFF
--- a/docs/registry.md
+++ b/docs/registry.md
@@ -36,6 +36,7 @@ editLink: false
 | asdf-plugin-manager | [asdf:asdf-community/asdf-plugin-manager](https://github.com/asdf-community/asdf-plugin-manager) |
 | assh | [asdf:zekker6/asdf-assh](https://github.com/zekker6/asdf-assh) |
 | atlas | [asdf:komi1230/asdf-atlas](https://github.com/komi1230/asdf-atlas) |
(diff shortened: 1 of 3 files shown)


docs/registry.md
```diff
@@ -36,6 +36,7 @@ editLink: false
 | asdf-plugin-manager | [asdf:asdf-community/asdf-plugin-manager](https://github.com/asdf-community/asdf-plugin-manager) |
 | assh | [asdf:zekker6/asdf-assh](https://github.com/zekker6/asdf-assh) |
 | atlas | [asdf:komi1230/asdf-atlas](https://github.com/komi1230/asdf-atlas) |
+| Atmos | [asdf:cloudposse/asdf-atmos](https://github.com/cloudposse/asdf-atmos) |
```
roele: maybe this should also be lower case

mtweeman: it seems that all names follow lowercase convention in this file. corrected. thanks for remark!

jdx:
you need to change it here https://github.com/mise-plugins/registry

mtweeman:
> you need to change it here https://github.com/mise-plugins/registry or it will just get overwritten on the next release

I wasn't aware of it. Thanks for letting me know! PR already created (https://github.com/mise-plugins/registry/pull/34)

PR #2606, opened 2024-09-18
fix: go backend can't install tools without 'v' prefix in git repo tags

roele:
@jdx The infamous `v` prefix topic again. While general [go module convention](https://go.dev/doc/modules/version-numbers) is to use `v` prefix there are several modules around without. Since versions can not be easily queried (for example via curl HEAD request) one approach i can think of handling this is a simple retry.


src/backend/go.rs
```diff
@@ -63,21 +65,29 @@ impl Backend for GoBackend {
...
-        };
+        if ctx.tv.version.starts_with("v") {
+            bail!("versions should not start with a 'v', please remove it");
+        }
 
-        CmdLineRunner::new("go")
+        let result = CmdLineRunner::new("go")
             .arg("install")
-            .arg(format!("{}@{}", self.name(), version))
+            .arg(format!("{}@v{}", self.name(), ctx.tv.version.clone()))
```
jdx:
```suggestion
            .arg(format!("{}@v{}", self.name(), &ctx.tv.version))
```

I know this is nitpicky but it bugs me

src/backend/go.rs
```diff
@@ -63,21 +65,29 @@ impl Backend for GoBackend {
...
             .with_pr(ctx.pr.as_ref())
+            .with_no_output()
```
jdx, line 76: why?

src/backend/go.rs
```diff
@@ -63,21 +65,29 @@ impl Backend for GoBackend {
...
             .envs(self.dependency_env()?)
             .env("GOBIN", ctx.tv.install_path().join("bin"))
-            .execute()?;
+            .execute();
+
+        if result.is_err() {
+            error!("Failed to install, trying again without added 'v' prefix");
```
jdx:
```suggestion
            warn!("Failed to install, trying again without added 'v' prefix");
```

src/backend/go.rs
jdx: this seems not great, wouldn't this hide the output if there is a missing prefix?

roele: we do not allow missing `v` anymore and bail. what i've seen from the output is, that it seems to be duplicated because of the `combined_output` in case of error.

roele: you get...
```
  ❯ mise-debug i go:github.com/dadav/helm-schema/cmd/helm-schema@0.13.1-2
go: invalid github.com import path "github.com/dadav"
go: invalid github.com import path "github.com"
go: github.com/dadav/helm-schema/cmd/helm-schema@v0.13.1-2: github.com/dadav/helm-schema/cmd/helm-schema@v0.13.1-2: invalid version: unknown revision cmd/helm-schema/v0.13.1-2
mise go failed
go: github.com/dadav/helm-schema/cmd/helm-schema@v0.13.1-2: github.com/dadav/helm-schema/cmd/helm-schema@v0.13.1-2: invalid version: unknown revision cmd/helm-schema/v0.13.1-2
mise Failed to install, trying again without added 'v' prefix
mise go:github.com/dadav/helm-schema/cmd/helm-schema@0.13.1-2 ✓ installed  
```

roele: as opposed to 
```
  ❯ mise-debug i go:github.com/dadav/helm-schema/cmd/helm-schema@0.13.1-2
mise go failed
go: github.com/dadav/helm-schema/cmd/helm-schema@v0.13.1-2: github.com/dadav/helm-schema/cmd/helm-schema@v0.13.1-2: invalid version: unknown revision cmd/helm-schema/v0.13.1-2
mise Failed to install, trying again without added 'v' prefix
mise go:github.com/dadav/helm-schema/cmd/helm-schema@0.13.1-2 ✓ installed                                                                   
```

src/backend/go.rs
```diff
@@ -63,21 +65,28 @@ impl Backend for GoBackend {
...
-            format!("v{}", ctx.tv.version)
-        } else {
-            ctx.tv.version.clone()
-        };
+        if ctx.tv.version.starts_with("v") {
+            bail!("versions should not start with a 'v', please remove it");
+        }
 
-        CmdLineRunner::new("go")
+        let result = CmdLineRunner::new("go")
```
jdx: can you make this a little closure like this:

```rust
let install = |v| {
            CmdLineRunner::new("go")
                .arg("install")
                .arg(format!("{}@{v}", self.name())
                .with_pr(ctx.pr.as_ref())
                .envs(self.dependency_env()?)
                .env("GOBIN", ctx.tv.install_path().join("bin"))
                .execute()
};
```

then just call it like this:

```rust
if let Err(_) = install(format!("{}", &ctx.tv.version)) {
  install(format!("v{}", &ctx.tv.version))?;
}
```

roele: neat!

sgtsquiggs:
Consider reverting this PR:

```
❯ mise install go:golang.org/x/tools/cmd/stringer@v0.21.0
mise versions should not start with a 'v', please remove it
mise Run with --verbose or MISE_VERBOSE=1 for more information
```

This is a breaking change!

jdx:
the go backend is experimental so breaking changes should be expected

PR #2628, opened 2024-09-23
fix: fix failing test for macos in tera.rs

roele:


jdx:
this isn't failing in main

PR #2635, opened 2024-09-24
docs: document java shorthand and its limitations

### LINKED ISSUE #2620: Create a setting for preferred java vendor
It would be nice if there was a way to set the preferred java vendor.

I would suggest two options:

* `java_preferred_vendor` (string) to set the preferred java vendor, so "21" when `java_preferred_vendor` is set to `temurin`, then `java@21` would resolve to `temurin-21`.
* `java_allow_preferred_vendor` (boolean) if false, then 21 would resolve to `openjdk-21` rather than `temurin-21`.

The purpose of the former is to allow a user or project maintainer to generally set a preferred vendor.

The purpose of the latter is to allow a project maintainer to override a user's settings and require the vendor to be explicitly set by the project. The reason for this is that a project may have licensing restrictions, have explicitly purchased a commercial sdk, or so on. In this case, a user's local setting for `java_preferred_vendor` should not override the desired sdk.

These comments in a different issue are linked here for reference:

* https://github.com/jdx/mise/issues/2618#issuecomment-2364115841
* https://github.com/jdx/mise/issues/2618#issuecomment-2364121337
* https://github.com/jdx/mise/issues/2618#issuecomment-2364554415

roele:
Fixes #2620.

### DIFF
--- a/docs/lang/java.md
+++ b/docs/lang/java.md
@@ -11,26 +11,31 @@ The code for this is inside the mise repository at
 
 ## Usage
 
-The following installs the latest version of java-openjdk-17.x (if some version of openjdk-17.x is
+The following installs the latest version of openjdk-21.x (if some version of openjdk-21.x is
 not already installed) and makes it the global default:
 
 ```sh
-mise use -g java@openjdk-17
-mise use -g java@17         # alternate shorthands for openjdk-only
+mise use -g java@openjdk-21
+mise use -g java@21         # alternate shorthands for openjdk-only
 ```
 
 See available versions with `mise ls-remote java`.
+
+::: warning
+Note that shorthand versions (like `21` in the example) use `OpenJDK` as the vendor.
+The OpenJDK versions will only be updated for a 6-month period. Updates and security patches will not be available after this short period. This also applies for LTS versions. Also see <https://whichjdk.com> for more information.
+:::
 
 ## macOS JAVA_HOME Integration
 
 Some applications in macOS rely on `/usr/libexec/java_home` to find installed Java runtimes.
 
 To integrate an installed Java runtime with macOS run the following commands for the proper
-version (e.g. openjdk-20).
+version (e.g. openjdk-21).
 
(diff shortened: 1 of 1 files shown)


docs/lang/java.md
```diff
@@ -21,6 +21,11 @@ mise use -g java@17         # alternate shorthands for openjdk-only
 
 See available versions with `mise ls-remote java`.
 
+::: warning
+Note that the shorthand version `17` uses `OpenJDK` as the vendor.
+The OpenJDK versions will only be updated for a 6-month period. Updates and security patches will not be available after this short period. This also applies for LTS versions. Also see https://whichjdk.com for more information.
```
jdx:
```suggestion
The OpenJDK versions will only be updated for a 6-month period. Updates and security patches will not be available after this short period. This also applies for LTS versions. Also see <https://whichjdk.com> for more information.
```

docs/lang/java.md
```diff
@@ -11,7 +11,7 @@ The code for this is inside the mise repository at
 
 ## Usage
 
-The following installs the latest version of java-openjdk-17.x (if some version of openjdk-17.x is
+The following installs the latest version of openjdk-17.x (if some version of openjdk-17.x is
```
jdx: what's the latest lts version? should we use that as the example?

docs/lang/java.md
```diff
@@ -21,6 +21,11 @@ mise use -g java@17         # alternate shorthands for openjdk-only
 
 See available versions with `mise ls-remote java`.
 
+::: warning
+Note that the shorthand version `17` uses `OpenJDK` as the vendor.
```
jdx:
```suggestion
Note that shorthand versions (like `17` in the example) uses `OpenJDK` as the vendor.
```

docs/lang/java.md
roele: latest is 23 but we probably should stick to LTS (Long Term Support) versions so 21

PR #2636, opened 2024-09-24
fix: release 2024.9.7 breaks configurations that were using v in version names with go backend

### LINKED ISSUE #2634: Release 2024.9.7 breaks configurations that were using `v` in version names with go backend
**Describe the bug**
Since the release of 2024.9.7 we're seeing:
```mise versions should not start with a 'v', please remove it```
in our CICD runs during `mise install`

**To Reproduce**

Use mise 2024.9.7 with config
```text
[tools]
go = "1.22.4"
node = "lts-hydrogen"
python = "3.10"
# tool pins should be more strict
golangci-lint = "1.54.2"
just = "1.25.2"
pnpm = "8"
poetry = "1.4.2"
sqlc = "1.21.0"
jq = "1.7.1"
yq = "4.43.1"
"go:github.com/clear-street/reinforcer/cmd/reinforcer" = "v0.3.1"
"go:github.com/golang/mock/mockgen" = "v1.6.0"
"go:github.com/sgtsquiggs/gta/cmd/gta" = "v0.3.1"
"go:github.com/sgtsquiggs/tools/cmd/structtagger" = "v0.1.0"
"go:github.com/vektra/mockery/v2" = "v2.43.0"
"go:golang.org/x/tools/cmd/stringer" = "v0.21.0"
"go:golang.org/x/vuln/cmd/govulncheck" = "v1.0.1"
"go:github.com/dmarkham/enumer" = "v1.5.9"

[plugins]
sqlc = "https://github.com/schmir/asdf-sqlc.git"
jq = "https://github.com/lsanwick/asdf-jq.git"
yq = "https://github.com/sudermanjr/asdf-yq.git"
```

in cicd we overlay this local config:
```text
[tools]
go = "system"
node = "system"

[settings]
python_compile = true
jobs = 2
```

**Expected behavior**

mise should install dependencies without error

**`mise doctor` output**

```text
will need to capture this from cicd job runner...
```

**Additional context**
Add any other context about the problem here. Consider running mise with `--debug` or `--trace` for extra debug info.

roele:
Seems bailing was a bit harsh, just allow both variants but keep the fallback.

Fixes #2634

### DIFF
--- a/src/backend/go.rs
+++ b/src/backend/go.rs
@@ -1,7 +1,5 @@
 use std::fmt::Debug;
 
-use eyre::bail;
-
 use crate::backend::{Backend, BackendType};
 use crate::cache::CacheManager;
 use crate::cli::args::BackendArg;
@@ -65,9 +63,12 @@ impl Backend for GoBackend {
         let settings = Settings::get();
         settings.ensure_experimental("go backend")?;
 
-        if ctx.tv.version.starts_with("v") {
-            bail!("versions should not start with a 'v', please remove it");
-        }
+        let version = if ctx.tv.version.starts_with("v") {
+            warn!("usage of a 'v' prefix in the version is discouraged");
+            ctx.tv.version.to_string().replacen("v", "", 1)
+        } else {
+            ctx.tv.version.to_string()
+        };
 
         let install = |v| {
             CmdLineRunner::new("go")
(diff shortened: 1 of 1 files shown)


src/backend/go.rs
```diff
@@ -65,9 +63,11 @@ impl Backend for GoBackend {
         let settings = Settings::get();
         settings.ensure_experimental("go backend")?;
 
-        if ctx.tv.version.starts_with("v") {
-            bail!("versions should not start with a 'v', please remove it");
-        }
+        let version = if ctx.tv.version.starts_with("v") {
+            ctx.tv.version.to_string().replacen("v", "", 1)
```
roele, line 68: should we print a hint or warning to not use `v` prefixes here?

jdx: yeah show a warning

PR #2654, opened 2024-09-25
feat: added postinstall hook

### LINKED ISSUE #1014: Provide way to specify simple post-install hooks/commands to run for tools (e.g NodeJS corepack)
Currently every time I install or upgrade a NodeJS version I need to remember to do a `corepack enable` afterwards since [corepack](https://nodejs.org/api/corepack.html) is now recommended for use with yarn, pnpm etc - despite the functionality being considered experimental from Node's perspective. At least with `mise` there is no need to `reshim` :-)

Not quite sure where/how might make sense to specify this, or perhaps whether something specific could be done for the Node plugin, e.g

```toml
node = {version='20', corepack: true}
node = [{version='16' corepack: true}, 'prefix:16', 'ref:master', 'path:~/.nodes/16']
```

This may be a bad idea, or there may be an alternate/existing approach that I could not find.

jdx:
Fixes #1014

### DIFF
--- a/docs/dev-tools/index.md
+++ b/docs/dev-tools/index.md
@@ -155,5 +155,12 @@ mise plugins may accept configuration in the form of tool options specified in `
 mytool = { version = '3.10', foo = 'bar' }
 ```
 
+All tools can accept a `postinstall` option which is a shell command to run after the tool is installed:
+
+```toml
+[tools]
+node = { version = '20', postinstall = 'corepack enable' }
+```
+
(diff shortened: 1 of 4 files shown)


jasisk:
@jdx I'd added the tiny bit of [config](https://mise.jdx.dev/lang/node.html#configuration) ages ago for `corepack` specifically in the `node` plugin (`MISE_NODE_COREPACK`)—would you like me to remove that bit of functionality now that this much more generally-useful hook is in place?

jdx:
nah I think that can stay for now

PR #2684, opened 2024-09-29
feat: enable Java core plugin on Windows

TobiX:
I have no idea what I'm doing, but this looked easy enough :D

It seems the Windows downloads for JDKs are ZIP instead of tar.gz, so I just
switched out the filter and the uncompress step and everything started working
nicely.


roele (approved):
Looks good to me.

Something unrelated that caught my eye is that the `musl` feature is probably only relevant in context of (Alpine) Linux. We could also put that behind a unix config.

Issue #788, opened 2023-08-10
Add `~/.default-cargo-crates` configuration support


mehalter:
I've been really loving `rtx` and specifically the ability to configure default package files such as `RTX_PYTHON_DEFAULT_PACKAGES_FILE` and `RTX_NPM_DEFAULT_PACKAGES_FILE` to put configuration files in the same location not in my home directory. `asdf-rust` also has a similar feature to put a `~/.default-cargo-crates` file (https://github.com/code-lever/asdf-rust/issues/20). It would be awesome to have a similar environment variable so I can put default cargo packages in the same location (maybe `RTX_CARGO_DEFAULT_CRATES`).

Thanks for all your hard work on this project!

jdxcode:
I don't understand, aren't you using asdf-rust in rtx?

mehalter:
I am, and it works great and I can make the file `~/.default-cargo-crates` and it works perfectly. I noticed that rtx adds environment variables for node and python for changing the location because `asdf-python` and `asdf-node` also only look at `~/.default-python-packages` and `~/.default-npm-packages` respectively. They don't have any knowledge of the `RTX_PYTHON_DEFAULT_PACKAGES_FILE` and `RTX_NPM_DEFAULT_PACKAGES_FILE` environment variables. I think it would be nice to have a uniform support with the `asdf-rust` as well to specify `RTX_CARGO_DEFAULT_CRATES` to change the location of that file in a similar manner.

mehalter:
Thanks for the information! I didn't realize that `python` and `node` were reimplemented here as core plugins. I thought this was adding shims into the `asdf-plugins`. Sorry for the noise!

farzadmf:
@mehalter Is there any documentation of using `asdf-rust` in mise? I really like the auto-install of packages feature, but I can't make it work by doing `mise use -g rust`

mehalter:
It looks like as of last week `rust` is a core plugin (https://github.com/jdx/mise/pull/3219). It looks like this new plugin doesn't support default crates yet. @jdx could this issue get reopened if this feature is planned similarly to how it's implemented in the core `python` and `node` plugins?

farzadmf:
@jdx I think it's mostly a matter of convenience TBH. Would be nice to have a consistent experience between all the tools and not having to consider Rust as an exception compared to Python and Node.

I do understand the core and non-core plugins, but I personally think it would be nice to achieve things in the tools in a similar manner.

[...]

PR #2701, opened 2024-10-03
docs: add Rust creates

yanskun:
Added a list of cargo create packages installed by default to the documentation

I did as described in this issue and it worked, but it was not yet mentioned in the documentation.
https://github.com/jdx/mise/issues/788


farzadmf:
@yanskun Question: how can we use the default crates feature? I do have a file in `~/.default-cargo-crates`, but when I do `mise use -g rust`, it doesn't take that file into consideration

Checking the docs for [Rust](https://mise.jdx.dev/lang/rust.html), I don't see any mention of this feature.

jdx:
Use the cargo backend

PR #2705, opened 2024-10-05
feat: use uv to create venv by default

### LINKED ISSUE #2279: Use uv for much faster Python virtual env creation
`virtualenv` and `uv venv` are much faster than `python -m venv`  at creating virtual envs:

```
❯ time python -m venv python-demo           # 2.79 secs

❯ time pyvenv/bin/virtualenv venv-demo     #  224.15 millis

❯ time pyvenv/bin/uv venv uv-venv-demo  # 19.79 millis
```

I'm not sure how best to ensure those tools are installed and can be used.  Should mise offer an "install command" type option ([like tox](https://tox.wiki/en/latest/config.html#install_command)) or just use uv or virtualenv if its on the path and assume the user is taking care of that?  I'm inclined towards the latter and, if a user wants it to be more automatic, they should put their desired tool in [default-python-packages](https://mise.jdx.dev/lang/python.html#default-python-packages) or, as I would do, make it available using pipx.

erickguan:
Fixes #2279

### DIFF
--- a/.mise.toml
+++ b/.mise.toml
@@ -52,8 +52,8 @@ depends = ["build"]
 env = { CLICOLOR_FORCE = "0" }
 run = [
     "mise usage > mise.usage.kdl",
-    "mise g task-docs > tasks.md",
-    "usage g md -f mise.usage.kdl -m --out-dir docs/cli --url-prefix /cli --html-encode",
+    "mise generate task-docs > tasks.md",
+    "usage generate markdown -m --out-dir docs/cli --url-prefix /cli --html-encode --file mise.usage.kdl",
     "markdownlint --fix docs/cli",
 ]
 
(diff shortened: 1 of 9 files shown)


Aeron:
All Python-related environment variables have the `MISE_PYTHON_*` prefix, but the new option has lost the `PYTHON` part. It could be unclear or unexpected for users.

Maybe the `MISE_PYTHON_VENV_STDLIB` would be more suitable here if we want that option explicitly available. It would better correlate with the setting key name as well.

erickguan:
Thanks for the review. Fixed.

honzajavorek:
I upgraded mise and suddenly any pip command stopped to work in new virtualenvs. I had to dive deep down to figure out that now uv is used by default. uv creates virtualenvs without pip. Yes, I have uv installed, because I played around with it, but it doesn't mean I want all my virtualenvs be created with uv, because there are many existing workflows based on pip being present. I think this should have been opt-in, not opt-out. I bet I'm not the only one who got all their work broken with the update.

erickguan:
I'm sorry this causes troubles on your end. Happy to follow up with a new issue.

PR #2707, opened 2024-10-05
feat: git.core.autocrlf=true causes the install script to fail

### LINKED ISSUE #2660: git.core.autocrlf=true causes the install script to fail
Thanks for maintaining `mise`! 🙌 🚀 

I've experienced a situation, where `~/.gitconfig` had an `autocrlf=true` on a MacOS machine(don't ask).

This caused bash scripts, responsible for installation to fail with an error, which is not user-friendly:

```
mise ~/.local/share/mise/plugins/cljstyle/bin/download failed
env: bash\r: No such file or directory
```

I propose showing a warning, to the users, who might find themselves, in a similar situation.

I'm up for working on a PR for this 👍 

Let me know what you think!

roele:
Fixes #2660

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -107,6 +107,14 @@ impl Git {
         if let Some(parent) = self.dir.parent() {
             file::mkdirp(parent)?;
         }
+        if cfg!(not(target_os = "windows")) {
+            warn_about_autocrlf(
+                git2::Config::open_default()?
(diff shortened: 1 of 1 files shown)


jdx:
I don't want to put this in

roele:
Understandable, it's a rather odd edge-case.

PR #2708, opened 2024-10-05
fix: use join_paths to create new_path

finalchild:
Fixes the issue that hook_env tries to join paths with ':' on Windows.


finalchild:
@jdx fixed the code :sweat_smile:
The test will work now

PR #2720, opened 2024-10-09
fix: don't skip the latest version fetch

finalchild:
Removed the code that skips the latest version check if the version is already installed.
I don't know the exact purpose of the code, so please double-check.

The reasoning is this:
* If latest_versions is false, we already don't fetch the remote versions.
* If latest_versions is true, we want the up-to-date version. Due to the fuzzy-matching, we can't be sure that this is the latest version even if the version is present in the installation directory. Thus, we should not skip the remote fetch.

This fixes the bug where the environment variables are not properly set when installing "latest" versions with the vfox backend on windows. (In env_hook, vfox.rs expects the resolved versions like "1.20.0" as the argument, but due to this skip mise is passing "latest". We could fix vfox.rs as jdx maintains this, but IMO this patch makes resolving more correct.)

### DIFF
--- a/src/toolset/tool_version.rs
+++ b/src/toolset/tool_version.rs
@@ -184,12 +184,6 @@ impl ToolVersion {
             }
         }
 
-        let existing = build(v.clone())?;
-        if backend.is_version_installed(&existing, true) {
-            // if the version is already installed, no need to fetch all the remote versions
-            return Ok(existing);
-        }
-
         if v == "latest" {
             if !latest_versions {
                 if let Some(v) = backend.latest_installed_version(None)? {


finalchild:
#347 revived this check, but the `latest_versions` flag was not present then.

Maybe the flag is enough to provide the performance benefit.

finalchild:
I can't figure out the test fail reason. Is this just random timeout?

finalchild:
Yes it was

jdx:
this removes expected behavior. `latest` should be the latest _installed_ version in situations where we want to simply use whatever was most recently installed like running `mise x` or using shims. They shouldn't eagerly install new versions except when running commands like `mise install`.

finalchild:
@jdx Isn't it covered by the `latest_versions` flag? I thought general commands call with `latest_versions = false` to only use the installed versions.

jdx:
let me trace the code a bit, it's possible this was needed at some point but is no longer

PR #2755, opened 2024-10-14
docs: http_timeout should be a duration string

### LINKED ISSUE #2752: Documentation still says http_timeout should be an integer even though it should be a duration string now
I've just updated to the latest release and i'm getting the error:
```
mise --version
2024.10.4 linux-x64 (6c21c0a 2024-10-14)
Error loading settings file: TOML parse error at line 29, column 16
   |
29 | http_timeout = 300
   |                ^^^
invalid type: integer `300`, expected a string
```

It looks like you updated it to be a duration string [here](https://github.com/jdx/mise/commit/7d1b6ac772ffd33dcb9d28cbacd86e60f7928368) but theres a few places you still refer to it as an integer:
- https://github.com/jdx/mise/blob/7248fcccb43361979f112268fc15ecd54f2a7344/schema/settings.json#L80
- https://github.com/jdx/mise/blob/7248fcccb43361979f112268fc15ecd54f2a7344/docs/configuration.md?plain=1#L147

roele:
Fixes #2752 

@jdx IIRC you mentioned once that `schema/settings.json` can be removed, is this correct?


jdx:
yeah we should get rid of it, unfortunately making http_timeout a duration is a breaking change I'm not sure how to avoid but given it was the only duration we used in seconds I think it's probably best to roll forward. I did make sure that the env var continues to work, but the integer in the config file won't

PR #2786, opened 2024-10-21
feat: add argc plugin to registry

rgeraskin:
Adds an asdf plugin for Argc, an open-source Bash CLI framework: https://github.com/sigoden/argc

Also PR'd to asdf-plugins: https://github.com/asdf-vm/asdf-plugins/pull/1061


jdx:
Does it work with ubi?

roman-geraskin-m10:
yes, it works with ubi. But to use it with ubi I have to add extra tools to install ubi itself
```
[tools]
rust = "1.82.0"
ubi = "0.1.2"
"ubi:sigoden/argc" = "v1.20"
```

Also, I can just use rust to install argc without ubi
```
[tools]
rust = "1.82"
"cargo:argc" = "v1.20"
```

But I think a much simpler way to install argc is to use asdf backend
```
[tools]
argc = "1.20"
[plugins]
argc = "https://github.com/rgeraskin/asdf-argc"
```

This PR is aimed to drop the need of `[plugins]` section in `.mise.toml`

jdx:
ubi is built into mise itself so it does not need to be separately installed

roman-geraskin-m10:
thanks, I've used an outdated mise. Works great with ubi!

jdx:
You can still add it to the registry with ubi

PR #2795, opened 2024-10-23
plugin: add germanium and puffin (asdf plugins)

saul-salazar-dotcom:


jdx:
Have you checked if these can be installed with ubi instead of asdf? If it’s compatible, change this to use ubi.

If ubi is not compatible, and this is something that can be fixed in ubi, then it should be patched there first.

saul-salazar-dotcom:
damn you got me there, I didn't know about ubi, I would have to research about it first

PR #2801, opened 2024-10-25
feat: add ubi installers for various tools

BurnerWah:
Doing a lot of these in one PR so I can make the edits faster

Tools included:

- act
- actionlint
- apko
- chamber
- changie
- ctop
- dive
- doctl
- flyctl
- grype
- hadolint (mac os arm64 needs rosetta, but that's true with the asdf plugin too)
- helmfile
- helmsman
- kind
- kompose
- lefthook
- nfpm
- pluto
- shellcheck
- yq

Personal testing on ubi indicates that all of these should install compatible versions of each tool for supported platforms


BurnerWah:
I had not accounted for the the tests expecting specific backends while working on these, unfortunately

jdx:
Can we just change the tests to use some other random (preferably lightweight) CLI?

jdx:
If you don't have time I can probably look into the tests tomorrow, thanks for doing this! Definitely a boost to security for us to move away from these plugins where we can.

BurnerWah:
yeah changing the tests is probably the easiest solution, since they're not failing due to a bug or anything like that
i'd appreciate if you could do that since i can't actually get the tests to run on my laptop for reasons entirely unrelated to mise lol

PR #2805, opened 2024-10-25
feat: Add support for generating completions for Amazon CloudWhisperer CLI

miguelmig:
Related to #2804
This PR still has a bit of naming confusion between `Amazon Q`, `Amazon Code Whisperer CLI` and even `Fig` so I'm looking for some guidance on how to keep the naming in the code. 
It's my first attempt at contributing and my first try at Rust as well so feel free to criticize 

This still misses the automated publish step as I'm waiting for feedback before moving forward with that

### DIFF
--- a/.mise.toml
+++ b/.mise.toml
@@ -83,6 +83,13 @@ depends = ["build"]
 env = { NO_COLOR = "1" }
 run = "mise render-mangen"
 
+[tasks."render:q"]
+description = "Generates the Amazon Q spec file for auto completion"
+run = [
+    "mise generate q-spec -o completions/mise.ts",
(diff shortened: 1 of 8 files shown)


jdx:
I would accept this as an extension to usage but not into mise directly

miguelmig:
@jdx Could you give some more context? Usage appears to have its own spec format, and Amazon Cloud Whisperer CLI has a different spec format for completions

jdx:
Usage could output fig completions like it does for bash/zsh/fish.

jdx:
Simple clap completions like this would be incomplete, missing things like completing tool names/versions

PR #2822, opened 2024-10-26
feat: add zed-editor with ubi backend

yodatak:


yodatak:
For now i see its don't work i report the issue to zed upstream sorry !

jimeh:
I think your issue is that Zed isn't a single standalone binary CLI tool, but rather a complex graphical application. The linux tar archive for example contains a `zed.app` folder with lots of things that make up the application.

However, they do ship `zed-remote-server` which can be installed via mise, as it is a single standalone binary.

For example, in a `.mise.toml`:

```toml
[tools]
"ubi:zed-industries/zed" = { version = "latest", matching = "zed-remote-server", exe = "zed-remote-server" }
```

Or in the `registry.toml` file, I think something like this should work:

```toml
zed-remote-server = ["ubi:zed-industries/zed[matching=zed-remote-server,exe=zed-remote-server]"]
```

PR #2823, opened 2024-10-26
feat: add ubi backend for eza

yodatak:
by the way should we remove exa because its unmaintend ? https://github.com/ogham/exa


jdx:
> by the way should we remove exa because its unmaintend ? https://github.com/ogham/exa

yes

jdx:
btw if you wouldn't mind, run `mise run render` to generate the docs when making these changes. Not super important because it'll happen later, but this groups the change in the correct pr.

jimeh:
This sadly seems to break the `eza` shorthand on macOS, as the official eza repository does not provide binaries for macOS in their releases.

There's been issue open about it since September 2023: https://github.com/eza-community/eza/issues/228

The asdf plugin worked around this by downloading binaries from [`cargo-bins/cargo-quickinstall`](https://github.com/cargo-bins/cargo-quickinstall), but I doubt that's a viable solution here.

@jdx do you have any minimum set of supported platforms/architectures in mind for tools before ubi as the default backend is viable? Personally I'd assume at least linux and macOS would be desirable, though that's mostly because I use mise on both daily.

PR #2861, opened 2024-10-29
perf(shell/zsh.rs): avoid hook-env execution on Enter without command

powerman:
Some zsh configuration are intentionally optimized to work really fast and it's important to keep it feels this way. And while `mise hook-env` is really fast (0.008 sec on my system) and slowdown isn't noticeable when running any command (even builtin one like `echo` or `true`) it still result in _noticeable_ slowdown when just pressing Enter without any command.

At a glance proposed optimization shouldn't break any functionality, and it fixes mentioned above noticeable slowdown.


src/shell/zsh.rs
```diff
@@ -20,6 +20,8 @@ impl Shell for Zsh {
             export MISE_SHELL=zsh
             export __MISE_ORIG_PATH="$PATH"
 
+            eval "$({exe} hook-env{flags} -s zsh)"
```
powerman: This fixes two issues:
- One is related to this PR: first command in a new zsh won't get right env because of this PR's optimization.
- Another issue is not related to this PR and fixes `zsh -i -c '...'` - current mise implementation won't provide right env to the command executed inside child zsh.

powerman:
Also, if `mise activate zsh` will now output code which runs `mise hook-env`, then this doc should be updated: https://mise.jdx.dev/dev-tools/shims.html#zshrc-bashrc-files

src/shell/zsh.rs
jdx: I can't remember why but there is a reason I chose not to run hook-env automatically

powerman: > I can't remember why but there is a reason I chose not to run hook-env automatically

Sometimes only way to recall it is to change a code and wait until something breaks.

I'm not sure is missing env in a `zsh -i -c '...'` considered a real issue. This is probably a real "corner" case and might not worth supporting: without `-i` it won't load Mise anyway, but `-i` with `-c` makes not much sense and in my experience it used mostly for testing zsh performance - how fast it loads user configs in interactive mode.

Still, your doc example in https://mise.jdx.dev/dev-tools/shims.html#zshrc-bashrc-files ask user to run it manually after activation - which has exactly same effect as my change to run it automatically but require extra manual change in user's config. Also, same 2-command example is repeated in https://mise.jdx.dev/troubleshooting.html#mise-isn-t-working-when-calling-from-tmux-or-another-shell-initialization-script.

So, I'm not joking here: if you can't remember why you chose not to run hook-env automatically then in current situation I think it makes sense to start running it automatically - this way it either will work and let you simplify UX and clean up docs, or breaks and this time you'll add a comment somewhere describing why it's bad idea to run it automatically on activation.

powerman: Maybe this was the reason: https://github.com/jdx/mise/pull/204#issuecomment-1449061691?

powerman: If that dotenv conflict is still actual, then you probably need to mention needs to manually add second line with hook-env in shell rc files at https://mise.jdx.dev/getting-started.html#_2-activate-mise (and maybe with a note about direnv and what to do between these two lines to avoid conflict with dotenv.

joshbode:
It could be my particular setup, but I'm finding that the `chpwd` and `precmd` hooks are both executing whenever I change directory (in that order), so I've replaced the `_mise_hook` in my `.zshrc` as follows to ensure they're mutually exclusive:

```zsh
source <(mise activate zsh)

function _mise_hook {
  source <(mise hook-env -s zsh)
  ZLAST_COMMANDS=""  # prevent precmd hook also triggering
}
ZLAST_COMMANDS="_"  # ensure the precmd hook executes initially
```

powerman:
> ```shell
> ZLAST_COMMANDS="_"  # ensure the precmd hook executes initially
> ```

I think it's better to do `eval "$(mise hook-env -s zsh)"` instead here (after `mise activate`) - this way is mentioned in the docs, it doesn't mangle ZLAST_COMMANDS and it doesn't makes any assumptions about current hook optimizations.

> ```shell
>   ZLAST_COMMANDS=""  # prevent precmd hook also triggering
> ```

This may breaks other things which use `ZLAST_COMMANDS`. So, it's more safe to use own flag instead:

```diff
+typeset -g mise_skip_hook_cmd
 _mise_hook() {
   eval "$(mise hook-env -s zsh)";
+  mise_skip_hook_cmd=1
 }
 _mise_hook_cmd() {
-  if [[ -n "$ZLAST_COMMANDS" ]]; then
+  if [[ -n "$ZLAST_COMMANDS" && -z "$mise_skip_hook_cmd" ]]; then
     _mise_hook;
   fi
+  mise_skip_hook_cmd=
 }
```

If you like to avoid this extra complexity then this one will works too in most (but not all) cases:

```diff
-  if [[ -n "$ZLAST_COMMANDS" ]]; then
+  if [[ -n "$ZLAST_COMMANDS" && "$ZLAST_COMMANDS" != "cd" ]]; then
```

joshbode:
Good point, @powerman

I ended up with this solution, which seems to work well for me, and doesn't rely on `ZLAST_COMMANDS` (I don't always use `cd` to change dirs, sometimes just directly typing the directory name as I have the `AUTO_CD` option enabled):

```zsh
_last_mise_hook_trigger=""

function _mise_hook_chpwd {
  _mise_hook
  _last_mise_hook_trigger="chpwd"
}
chpwd_functions[1]=_mise_hook_chpwd

function _mise_hook_precmd {
  if [[ ${_last_mise_hook_trigger} == "precmd" ]]; then
    _mise_hook
  fi
  _last_mise_hook_trigger="precmd"
}
precmd_functions[1]=_mise_hook_precmd
```

PR #2881, opened 2024-11-01
feat(registry): add ubi (+ go & cargo) installers for tools

BurnerWah:
Tools included:

- action-validator
- agebox
- chezmoi
- chisel (ubi & go; on my laptop the asdf plugin wouldn't even run)
- choose (ubi & cargo; asdf plugin doesn't work unless on amd64 linux, cargo-binstall works)
- cilium-cli
- circleci-cli
- colima
- docuum (ubi & cargo; adds proper arm64 support, cargo binstall works)
- dotenv-linter
- drone
- gator
- jx (adds proper arm64 support)
- k0sctl
- k2tf
- k3d
- k6
- mkcert (adds proper arm64 support)
- newrelic-cli (adds proper arm64 support on mac os)
- ruff
- task
- zellij
- zola

Also, I ran the pre-commit task so some other files got generated


BurnerWah:
There's a few more things here that _could_ be given a cargo or go installer, but I'm a little hesitant on including them when building either takes a long time, or is prone to failure

Issue #2704, opened 2024-10-04
`upgrade --bump ubi:rust-analyzer/rust-analyzer` removes old but doesn't install new


liskin:
**Describe the bug**
The new `upgrade --bump` functionality behaves strangely on `ubi:rust-analyzer/rust-analyzer` — it offers the upgrade when run with `--interactive`, but doesn't bump the version in the config file, and only removes the old version without installing the new one.

**To Reproduce**
```console
$ mise use -g --pin ubi:rust-analyzer/rust-analyzer@2024-09-16
mise ubi:rust-analyzer/rust-analyzer@2024-09-16 ✓ installed
mise ~/.config/mise/config.toml tools: ubi:rust-analyzer/rust-analyzer@2024-09-16
$ mise upgrade --bump ubi:rust-analyzer/rust-analyzer
mise Uninstalling rust-analyzer/rust-analyzer@2024-09-16 ✓ removing ~/.cache/mise/ubi-rust-analyzer-rust-analyzer/2024-09-16
$ mise ls
Tool                             Version               Config Source              Requested
ubi:rust-analyzer/rust-analyzer  2024-09-16 (missing)  ~/.config/mise/config.toml 2024-09-16
```

```console
$ mise upgrade --bump --interactive
mise upgrade
Select tools to upgrade
 > • rust-analyzer/rust-analyzer 2024-09-16 -> 2024-09-30 (~/.config/mise/config.toml)
↑/↓/k/j up/down • x/space toggle • a toggle all • / filter • enter confirm
```

**Expected behavior**
I expected the pinned version in ~/.config/mise/config.toml to be bumped (as it is for all other tools) and the new version to be installed (likewise).

**`mise doctor` output**

```text
version: 2024.10.0 linux-x64 (de95bc9 2024-10-03)
activated: no
shims_on_path: yes

build_info:
  Target: x86_64-unknown-linux-gnu
  Features: DEFAULT, NATIVE_TLS, OPENSSL
  Built: Thu, 3 Oct 2024 18:34:02 +0000
  Rust Version: rustc 1.81.0 (eeb90cda1 2024-09-04)
  Profile: release

shell:
  /bin/bash
  GNU bash, version 5.2.32(1)-release (x86_64-pc-linux-gnu)
  Copyright (C) 2022 Free Software Foundation, Inc.
  License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

  This is free software; you are free to change and redistribute it.
[...]

PR #2889, opened 2024-11-01
fix: handle General/Complex Versioning in --bump

liskin:
#### [fix: handle General/Complex Versioning in --bump](../commit/6b183df29cc8f61ab00dc29f0d7e084c71cc820b)

The previous fix to handle non-semver versions in `--bump` used a crude heuristic - a check whether there's a "." in the version string - to decide whether to attempt to `chunkify` or just fall back to returning the new version unmodified. This, however, resulted in no bump happening (`None` returned from `check_semver_bump`) when either of the old/new versions looked like "v1.2.3".

Additionally, versions like "1.2a1" and "1.2a2" were considered the same, because: a) the heuristic failed to recognise they won't be parsed/chunkified correctly, and b) `chunkify` used `nth` that silently throws parts of the version away.

I'm proposing to fix this by reimplementing `chunkify` using the much more generic versions::Mess format, which simply splits the version number into chunks and separators. We then convert it into a simpler format of just chunks (first chunk as is, further chunks with a separator prepended).

This design has an issue: we don't recognise a change in the versioning scheme. A bump from "\<commit sha\>" to "v1.2.3" will result in just "v1", because the commit sha is parsed as a single chunk. The most obvious case of this, "latest" being parsed as a single chunk, is handled explicitly in the code, as it would just be a mess otherwise.

A potential workaround for this issue would be to add a flag (e.g. `--pin`) that would make `--bump` skip the `check_semver_bump` logic and always use the full new version (as suggested in https://github.com/jdx/mise/issues/2704#issuecomment-2399624421). This would also help in the following case: a project using variable length version numbers instead of the full 3-chunk semver. Trying to follow this sequence of bumps: "20.0", "20.0.1", "20.1" isn't possible with the current logic.

Related: 0b2c2aa67730 ("fix: upgrade --bump with non-semver versions (#2809)")

#### [fix: Allow --bump from 20.0.1 to 20.1](../commit/d937241a622e1e7216ef4ed54666b83816bf666d)

It's weird so we still warn, but returning `None` from `check_semver_bump` only makes sense if the versions are deemed to be the same. Otherwise it's just confusion for the user — the UI presents this as an upgrade, proceeds to uninstall the old version, but fails to do the actual bump and no new version is installed.


jdx:
I think right now it probably doesn't work right with `ref:` or `prefix:` so we should probably just ignore those in `mise up` altogether, maybe with a warning, until we have a better solution for them.

src/toolset/mod.rs
```diff
@@ -818,10 +823,30 @@ mod tests {
...
+        );
         std::assert_eq!(
             check_semver_bump("2024-09-16", "2024-10-21"),
             Some("2024-10-21".to_string())
         );
+        std::assert_eq!(
+            check_semver_bump("20.0a1", "20.0a2"),
+            Some("20.0a2".to_string())
+        );
+        std::assert_eq!(check_semver_bump("v20", "v20.0.0"), None);
```
jdx, line 944: are you actually seeing `v` prefixed versions somewhere? we should be removing that prefix everywhere we can by this point

liskin: They're in my config so I thought it'd be nice if `mise upgrade --bump` didn't choke on them. I can remove them manually but it'd be nice if others didn't need to.

liskin: But you've got a point that the test cases don't reflect reality. There might be a "v" in `old` but not in `new`.

liskin:
> I think right now it probably doesn't work right with `ref:` or `prefix:` so we should probably just ignore those in `mise up` altogether, maybe with a warning, until we have a better solution for them.

Actually, the code already kinda tries to ignore them: https://github.com/jdx/mise/blob/9bedf175172cef3872c34469a2f8c94b2a8a3294/src/toolset/mod.rs#L402-L420
But it's half-assed — only skips the bump, but still performs the uninstall. And then, as I mentioned earlier in https://github.com/jdx/mise/issues/2704#issuecomment-2399555891, the dry-run output doesn't match what's actually going to happen, so the output in the failing test is "Would install dummy@2.0.0" but that's not what would happen in practice, as there'd be no bump.

Oh, what a can of worms. I think I give up for today :-D

jdx:
I did some work around prefix: in https://github.com/jdx/mise/pull/3054

PR #2908, opened 2024-11-04
feat: add autocompletion for tasks run

### LINKED ISSUE #2903: `mise tasks run`: add autocompletion
Autocompletion already works for `mise run`, but not for `mise tasks run`.

roele:
Fixes #2903

### DIFF
--- a/src/cli/usage.rs
+++ b/src/cli/usage.rs
@@ -20,6 +20,12 @@ impl Usage {
             run: "mise tasks --usage".to_string(),
         });
 
+        let tasks = spec.cmd.subcommands.get_mut("tasks").unwrap();
+        let tasks_run = tasks.subcommands.get_mut("run").unwrap();
+        tasks_run.mounts.push(usage::SpecMount {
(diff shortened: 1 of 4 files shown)


jdx:
that file is generated, it needs to be added here https://github.com/jdx/mise/blob/36a19f87b80aafb15f33dd818f3a4b4a0277c295/src/cli/usage.rs#L19

roele:
was wondering about that, thanks for the hint @jdx

src/cli/usage.rs
```diff
@@ -20,6 +20,12 @@ impl Usage {
             run: "mise tasks --usage".to_string(),
         });
 
+        let tasks = spec.cmd.subcommands.get_mut("tasks").unwrap();
```
jdx, line 23: I actually think this can/should go into the extras file since usage _should_ merge everything, but I also suspect there is a reason I didn't do that before so I think this is fine.

jdx (approved):

PR #2930, opened 2024-11-05
docs: move alternative ways to install mise to a separate page

hverlin:
As suggested in https://github.com/jdx/mise/discussions/2440, the `getting-started` page might be too intimidating for someone new to `mise`.

> 2/3rd of the page is filled up with: Alternate Installation Methods, Shells, and Uninstalling. I think it makes sense to link to this content in other pages but for the majority of users this is only adding noise and making them want to leave the docs or search for more examples.

To address this, this PR adds an `installing-mise.md` page which lists all the ways that `mise` can get installed.

> A little into the page it offers you the choice about activate vs shims vs none. Most people new to Mise won't want to make choices like this right away since it requires research to understand the consequences.

> There is very little explanation about what activate actually does at this point. It doesn't need to be in-depth like the FAQ, just something like "it updates your env vars and path every time your prompt is run to ensure you use the correct versions"

Moved the `shims` activation method to `shims` and added a short explanation.

Other changes:
- Made all the code blocks easier to copy-paste by removing the `$`
- Updated the `using mise` section of the getting started so that users try to use both `mise exec -- node -v` and `node -v`
- Add section on how to set up the autocompletion

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -1,44 +1,40 @@
 # Getting Started
 
-Using mise typically involves 3 steps:
+This will show you how to install mise and get started with it. This is a suitatable way when using an interactive shell like `bash`, `zsh`, or `fish`.
 
-1. Installing the CLI
-2. Activating mise or adding its shims to PATH <Badge type="tip" text="optional" />
-3. Adding tools to mise
+## 1. Install `mise` CLI
 
-## Quickstart
+See [installing mise](/installing-mise) for other ways to install mise (`apt`, `yum`, `nix`, etc.).
 
-### 1. Install `mise` CLI
-
-First we need to download the mise CLI.
-See [below](#alternate-installation-methods) for alternate installation methods.
-This directory is simply a suggestion.
-mise can be installed anywhere.
-
-```sh
-$ curl https://mise.run | sh
-$ ~/.local/bin/mise --version
-mise 2024.x.x
+```shell
+curl https://mise.run | sh
 ```
 
-::: tip
-"~/.local/bin" does not need to be in PATH. mise will automatically add its own directory to PATH
-when activated.
-:::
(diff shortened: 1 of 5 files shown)


docs/getting-started.md
```diff
@@ -54,395 +50,55 @@ echo 'eval "$(~/.local/bin/mise activate zsh)"' >> ~/.zshrc
...
 ```sh
-$ mise use --global node@20
-$ mise exec -- node -v
-v20.x.x
+mise exec -- node -v
+# v22.x.x
 ```
 
 :::tip
 Use `mise x -- node -v` or set a shell alias in your shell's rc file like `alias x="mise x --"` to
```
hverlin, line 82: This tip is nice, however I would perhaps move it to another page?

docs/installing-mise.md
```diff
@@ -0,0 +1,371 @@
...
+```
+
+```sh [zsh]
+# This requires zsh to be configured. If you e.g. use oh-my-zsh, it should work out of the box.
+mkdir -p /usr/local/share/zsh/site-functions
+mise completion zsh  > /usr/local/share/zsh/site-functions/_mise
+```
+
+```sh [fish]
+mise completion fish > ~/.config/fish/completions/mise.fish
```
hverlin, line 353: I don't use `fish`. However, I still tried it to make sure it works. I am getting the following error (even with usage installed)
```
root@2c295de84df0 /# source ~/.config/fish/completions/mise.fish
~/.config/fish/completions/mise.fish (line 6): 'return' outside of function definition
    return 1
    ^
from sourcing file ~/.config/fish/completions/mise.fish
```

removing the `return 1` fixes it

jdx: you don't `source` completion files in fish

PR #2938, opened 2024-11-05
docs: Add a guide to help to migrate from nvm to mise

hverlin:
This is my understanding how one would move from `nvm` to `mise`.
I have added to the `node.md` page, but it could be moved to another page, also.


hverlin:
On a personal website or dev.to and link it from the docs?

jdx:
yeah

hverlin:
Created https://dev.to/hverlin/migrating-from-nvm-to-mise-4mfp
Seems that it will also be easy to find with a Google search so I will leave it up to you if you think if it's worth it to add a link to it

Issue #2595, opened 2024-09-18
windows alternate tasks


jdx:
because shell commands work quite differently in windows, we could allow configuration for tasks specific to OS, such as:

```toml
[tasks.echo]
run = 'echo $FOO'
run_windows = 'echo %FOO%'
```

or maybe something like this:

```toml
[tasks.echo]
run = 'echo $FOO'
[tasks-windows.echo]
run = 'echo %foo%'
```

I think the former is what I prefer, though I can think of a few different ways to present this like maybe a `tasks.os.windows` key, though that would prevent anyone from having a task labeled "os" so it would have drawbacks.

rodvieirasilva:
I am using Mise on Windows, and I encountered some difficulties with shell command compatibility. To address the differences in shell commands across operating systems, I propose adding an os.windows/os.unix section in the TOML configuration that can override any configuration. This approach allows users to specify OS-specific settings not only for tasks but for other sections as well, such as [settings], [tools], or any other sections that may benefit from customization.

The configuration would work as follows: when an OS-specific section (e.g., [os.windows]) is provided, it will inherit properties from the main configuration, but any properties specified in the OS-specific section will override the main configuration values. This enables users to customize tasks and tools with a high degree of flexibility while keeping platform-specific variations organized and easy to manage.

Example of Proposed Structure:
Base Configuration:
```toml
[tasks.clean]
description = "Clean the build directory before building"
run = '''
    echo "Cleaning..." > clean.log
'''
```
OS-Specific Override:

```toml
[os.windows]
[tasks.clean]
description = "Windows-specific clean task"
shell = "PowerShell" # specify shell for Windows

[os.unix]
[tasks.clean]
shell = "bash" # specify shell for Unix
```
Benefits:
Consistency and Clarity: Each task can inherit tags and configurations from the base [tasks] section, while also applying OS-specific overrides as needed.
Flexibility: Beyond tasks, any other configuration sections (e.g., [settings], [tools]) can be similarly customized with platform-specific values without impacting the main configuration.
Ease of Use: This structure keeps OS-specific commands and settings consolidated, making it straightforward to understand and modify configurations for different environments.
This approach balances flexibility and clarity, allowing projects to handle platform-specific configurations more seamlessly while minimizing redundancy.

If I receive an OK with this proposal, I can implement the issue.

Thanks.

jdx:
you could do this today with `MISE_ENV=windows` and have a `mise.windows.toml` file

[...]

PR #2967, opened 2024-11-08
feat(tasks): add shell settings for Windows compatibility in tasks

rodvieirasilva:
Adds settings for Windows compatibility in tasks:

Like discuss in https://github.com/jdx/mise/issues/2595

New settings:

- windows_executable_extensions
- windows_default_inline_shell_args
- windows_default_file_shell_args
- use_file_shell_for_executable_tasks
- unix_default_inline_shell_args
- unix_default_file_shell_args

example of a settings working on windows for shell scripts:

```toml
[settings]
windows_executable_extensions = ["sh"]
windows_default_file_shell_args = ["D:\\Program Files\\Git\\usr\\bin\\bash.exe"]
use_file_shell_for_executable_tasks = true
```

As discussed in the issue, some changes were necessary to enable running tasks in shell script or nushell formats.

This PR introduced a side effect that I’d like to ask about. Is this e2e test correct?

```
[tasks.test]
run = 'echo "testing!"'
---
assert "mise run test arg1 arg2 arg3" "testing! arg1 arg2 arg3"
```

Should it print the other arguments even though they weren’t directly passed to `echo`?  
If so, further modification will be needed.


src/task/mod.rs
```diff
@@ -396,9 +396,15 @@ where
         Ok(self
             .iter()
             .filter(|(k, _)| {
-                let p: PathBuf = k.split(':').collect();
-
-                matcher.is_match(p)
+                let path: PathBuf = k.split(':').collect();
```
rodvieirasilva, line 399: This enable to run tasks without extension:
```
mise r test.bat
or
mise r test
```

jdx: shouldn't we only do this check on windows?

jdx:
this seems worthwhile regardless, but it wouldn't resolve #2595 since I don't see a way to make it use a different task when on windows (if I'm just missing that please point it out). I think one solution this could support for file tasks would be allowing a project with the following:

```
tasks/run
tasks/run.bat
```

where `run` would be used on unix and `run.bat` would be used on windows. I feel like with my suggestion it would make that work. I'd also like to see an e2e test both on windows and unix here that has 2 files tasks for unix and windows but uses the correct one on their respective platforms. I created a windows test here you can use as a starting point: https://github.com/jdx/mise/pull/2972

src/task/mod.rs
rodvieirasilva: I think this check could apply to both Unix and Windows systems. Personally, I always use the .sh extension for my script files. It would feel a bit inconsistent to run `mise r test.sh` on Unix and optional `mise r test` on Windows.

However, I’m open to whichever approach you think is best—let me know your preference, and I can make the necessary adjustments.

rodvieirasilva:
> This PR introduced a side effect that I’d like to ask about. Is this e2e test correct?
> 
> ```
> [tasks.test]
> run = 'echo "testing!"'
> ---
> assert "mise run test arg1 arg2 arg3" "testing! arg1 arg2 arg3"
> ```
> 
> Should it print the other arguments even though they weren’t directly passed to `echo`? If so, further modification will be needed.

rodvieirasilva:
> this seems worthwhile regardless, but it wouldn't resolve #2595 since I don't see a way to make it use a different task when on windows (if I'm just missing that please point it out). I think one solution this could support for file tasks would be allowing a project with the following:
> 
> ```
> tasks/run
> tasks/run.bat
> ```
> 
> where `run` would be used on unix and `run.bat` would be used on windows. I feel like with my suggestion it would make that work. I'd also like to see an e2e test both on windows and unix here that has 2 files tasks for unix and windows but uses the correct one on their respective platforms. I created a windows test here you can use as a starting point: #2972

In the example you mentioned, by setting `windows_executable_extensions` to `.bat`, Windows would run `run.bat` and ignore `run`. 

Understood; I'll add the new e2e test.

Regarding my previous comment:
> This PR introduced a side effect that I’d like to ask about. Is this e2e test correct?
> 
> ```
> [tasks.test]
> run = 'echo "testing!"'
> ---
> assert "mise run test arg1 arg2 arg3" "testing! arg1 arg2 arg3"
> ```
> 
> Should it print the other arguments even though they weren’t directly passed to `echo`?  
> If so, further modification will be needed.

Any thoughts on this?  
Thank you!

jdx:
> In the example you mentioned, by setting windows_executable_extensions to .bat, Windows would run run.bat and ignore run.

ah ok sounds good, I'm cool with this

> Should it print the other arguments even though they weren’t directly passed to echo?

yes

rodvieirasilva:
updated @jdx.
Added the tests and the fix for inline commands tasks.

PR #2985, opened 2024-11-10
feat: Add Fig completions to Mise

miguelmig:
Requires a manual accept of fig's CLA by the mise-en-place account (see [here](https://github.com/withfig/autocomplete/pull/2500#issuecomment-2466850991))
Re-used the `release-plz` workflow, let me know if that secret shouldn't be used for this.


miguelmig:
Will reopen with correct base commit

PR #2986, opened 2024-11-10
feat: Add Fig completions

miguelmig:
Relates to #2804 
Requires a manual accept of fig's CLA by the mise-en-place account (see test example [here](https://github.com/withfig/autocomplete/pull/2500#issuecomment-2466850991))
Re-used the `release-plz` workflow, let me know if that secret shouldn't be used for this.

The action should probably be updated not to run every patch version update as mise is currently very active on releasing


miguelmig:
Needs fix, is outputting twice into `tasks/fig/src/mise.ts`

miguelmig:
Done

PR #2989, opened 2024-11-11
docs: document and enable MISE_MIN_VERSION setting

### LINKED ISSUE #2933: Document `min_version` config setting
If you want to take advantage of a newish feature of mise in a shared project, it would be nice to have a way to specify a minimum mise version and at least get a warning or maybe error if it isn't matched. (Having mise actually control its own version seems overkill, but a warning would be nice.)

erickguan:
Fixes #2933


schema/mise.json
```diff
@@ -295,6 +295,10 @@
           "type": "string",
           "enum": ["trace", "debug", "info", "warn", "error"]
         },
+        "min_version": {
+          "description": "Minimal version of mise to use this config.",
```
jdx:
```suggestion
          "description": "Minimal version of mise to use with this config",
```

settings.toml
```diff
@@ -367,6 +367,12 @@ enum = [
     ["error"]
 ]
 
+[min_version]
```
jdx: what is this for? this config isn't a setting. If you'd like to document it, it should go into docs/configuration.md

hverlin:
https://github.com/jdx/mise/blob/e13543aa938f0c64709c9b0bcf72c48084353f9d/schema/mise.json#L732-L735
It's already in the schema so we just need to add it to the docs?

settings.toml
erickguan: Ooops, my mistake. Updated.

schema/mise.json
```diff
@@ -788,7 +788,7 @@
       ]
     },
     "min_version": {
-      "description": "minimum version of mise required to use this config",
+      "description": "Minimal version of mise to use with this config",
```
jdx:
```suggestion
      "description": "minimum version of mise required to use this config",
```

docs/configuration.md
```diff
@@ -124,6 +124,14 @@ note adding an alias will also add a symlink, in this case:
...
 ```
 
+### Other configurations
+
+** Define minimal required mise version for this config **
+
+```toml
+min_version = '2024.11.1'
+```
+
```
jdx:
```suggestion
### Minimum mise version

Define a minimum supported version of mise for the config file. mise will ignore config files that use too new of a version.

```toml
min_version = '2024.11.1'
```

```

File: docs/configuration.md [hunk 5]
```diff
@@ -124,6 +124,14 @@ note adding an alias will also add a symlink, in this case:
 my_custom_node = '20'
 ```
 
+### Minimum mise version
+
+Define a minimum supported version of mise for the config file. mise will ignore config files that use too new of a version.
```
glasser, line 129: Reading these docs (not trying it myself), it's unclear to me what "ignore" means — will there be a warning or will this be invisible to users?  Will mise generally function just ignoring that single file?

erickguan: Added #2994

PR #2994, opened 2024-11-11
docs: describe mise behavior when mise version is lower than min_version

erickguan:
Thanks for fixing my previous PR. This PR addresses [@glasser's comment](https://github.com/jdx/mise/pull/2989/files#r1837006850).

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -130,7 +130,9 @@ my_custom_node = '20'
 
 ### Minimum mise version
 
-Define a minimum supported version of mise for the config file. mise will ignore config files that use too new of a version.
+Specify the minimum supported version of mise required for the configuration file.
+If the configuration file specifies a version of mise that is higher than
+the currently installed version, mise will error out.
 
 ```toml
 min_version = '2024.11.1'


glasser:
Thanks. Also it's not clear to me what section this key goes in — the other `###` headings are section names?

glasser:
Oh, I see, it's actually not in a section — that might be worth stating explicitly!

glasser:
Actually experimenting with this for once, it looks like the impact of this (similar to, say, a TOML parse error) is that mise does not update PATH (etc) but also doesn't actively clear PATH entries from previous mise runs.

erickguan:
> it's actually not in a section — that might be worth stating explicitly!

I would keep this out of this PR because other sections follow similar style.

> Actually experimenting with this for once, it looks like the impact of this (similar to, say, a TOML parse error) is that mise does not update PATH (etc) but also doesn't actively clear PATH entries from previous mise runs.

Good find. This looks like a problem but a trivial one. It's easy to reload the interactive shell for mise users. If there are more complains, we could look into it. What do you think?

docs/configuration.md
```diff
@@ -2,23 +2,23 @@
 
 ## `mise.toml`
 
-`mise.toml` is the config file for mise. They can be at any of the following file paths (in order of precedence, top overrides configuration of lower paths):
```
jdx: change these back

docs/configuration.md
```diff
@@ -130,7 +130,13 @@ my_custom_node = '20'
...
 ### Minimum mise version
 
-Define a minimum supported version of mise for the config file. mise will ignore config files that use too new of a version.
+Specify the minimum supported version of mise required for the configuration file.
+If the configuration file specifies a version of mise that is higher than
+the currently installed version, mise will:
+
+- Ignore the configuration file.
+- Output a warning message.
+- Terminate execution.
```
jdx:
```suggestion
the currently installed version, mise will error out.
```

PR #3014, opened 2024-11-13
fix(python): use consistent conditions for precompiled python

risu729:
Follow up of https://github.com/jdx/mise/commit/dfb13de8ebc1da2542ab67ba2b81e8b26a085c20

### DIFF
--- a/src/plugins/core/python.rs
+++ b/src/plugins/core/python.rs
@@ -141,7 +141,7 @@ impl PythonPlugin {
         let (tag, filename) = match precompile_info {
             Some((_, tag, filename)) => (tag, filename),
             None => {
-                if cfg!(windows) || SETTINGS.python.compile == Some(false) {
+                if cfg!(windows) || SETTINGS.python.compile != Some(true) {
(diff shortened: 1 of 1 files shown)


jdx:
this behavior is tricky but I think this is not what we want. It should do the following:

* None - install precompiled if available
* false - never use precompiled
* true - always use precompiled

I believe with your change the first would never happen

risu729:
I'm so sorry for not checking well. I now understand what you mean.

Please let me share my understanding.

`_list_remote_versions` uses `python-build` unless it's not disabled like in Windows or by `python.compile=true` because all precompiled versions are included in versions from `python-build`.

I found dev versions have different syntaxes about the latest dev version, `3.13-dev` in `python-build`, and `3.13.0rc3` and `3.13.0rc2` in `python-build-standalone`. But this shouldn't cause any errors as versions cannot be resolved by `python-build-standalone` fall back to `python-build`.

jdx:
you got it, it's definitely strange logic but I think this is the best

PR #3047, opened 2024-11-16
feat: added `os` option to mise.toml

jdx:


schema/mise.json
```diff
@@ -763,16 +763,35 @@
...
               "type": "string"
+            },
+            "os": {
+              "oneOf": [
+                {
+                  "description": "operating system to install on",
+                  "type": "array"
+                },
+                {
+                  "description": "option to pass to tool",
```
powerman, line 774: Looks like a typo, probably should be same as for array.

jdx: not a typo

Issue #3094, opened 2024-11-20
Feature Request: How to migrate from ASDF


muescha:
The "old" README included a section titled "How do I migrate from asdf?" (see [here](https://github.com/jdx/mise/blob/10c6da9932a848f1e6bde6b2ad646893ad9a867c/README.md#how-do-i-migrate-from-asdf)), which was removed in [this commit](https://github.com/jdx/mise/commit/b76888ffa4553c21f2e08dcaf20a2bc4fa00d608).

I couldn’t find this section in the new documentation site. It would be helpful to have a "Migrate" section that explains how to transition from asdf or other tools.

muescha:
ok found it via google but not via the included search:

https://mise.jdx.dev/faq.html#how-do-i-migrate-from-asdf

![Bildschirmfoto 2024-11-20 um 14 30 53](https://github.com/user-attachments/assets/b4d0a9c3-34c5-46c2-a7ec-6512b8148c2b)

I would suggest to add a link to this section at the page:

https://mise.jdx.dev/dev-tools/comparison-to-asdf.html

muescha:
created pull request:

#3095

PR #3095, opened 2024-11-20
docs: update comparison-to-asdf.md - add link to faq: migration from asdf

muescha:
- add link to faq: migration from asdf


muescha:
see #3094

muescha:
instead of a `##` header it can also styled a tip - I am not sure what is better :)

PR #3128, opened 2024-11-21
feat: ignore config files after not trusting them

### LINKED ISSUE #761: Ability to permanently ignore/untrust a config
There are some specific dirs with configs in them that I would like to be permanently set in untrusted/ignored mode, and not be bothered about when I cd to the dir with them in effect. I couldn't find an option to do that in the CLI or in the config, so it seems the feature does not exist.

Something like `rtx trust --untrust --permanent` could be it. `rtx trust` would then make such marked one trusted though.

jdx:
Fixes #761

### DIFF
--- a/docs/cli/trust.md
+++ b/docs/cli/trust.md
@@ -26,9 +26,13 @@ The config file to trust
 
 Trust all config files in the current directory and its parents
 
+### `--ignore`
+
+Do not trust this config and ignore it in the future
+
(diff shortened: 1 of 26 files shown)


joshuataylor:
Does this include global mise config, e.g under `$HOME/.config/mise/config.toml`?

After updating, I received a message saying this is no longer trusted, globally. `mise trust` fixed it.

PR #3136, opened 2024-11-22
docs: add more detail about task vars

jdx:


docs/tasks/index.md
```diff
@@ -40,3 +43,25 @@ task3 = "echo task3"
...
+run = './scripts/test-e2e.sh {{vars.e2e_args}}'
+```
+
+Like most configuration in mise, vars can be defined across several files. So for example, you could
+put some vars in your global mise config `~/.config/mise/config.toml`, use them in a task at
+`~/src/work/myproject/mise.toml`. You can also override those vars in "later" config files such
+as `~/src/work/myproject/mise.local.toml` and they will be used inside tasks of any config file.
+
+As of this writing vars are only supported in TOML tasks. I want to add support for file tasks, but
+I don't want to turn all file tasks into tera templates just for this feature.
```
syhol, line 67: Might be worth exposing them as prefixed env vars like how in usage `flag "-u --user <user>"` becomes `$usage_user`

jdx: my first thought is that vars could just be put into tera templates anywhere in mise (though right now it only works in tasks) and we could make the mise/usage header for file tasks rendered with mise (unsure if it is right now but I think they are not)

so you would do this:

```bash
#!/usr/bin/env bash
#USAGE flag "-u --user <user>" default="{{vars.foo}}"

echo $usage_foo
```

what I don't love is that they are _still_ env vars even if they're scoped down a bit, but that's a problem for usage generally since the way it works is via env vars.

PR #3185, opened 2024-11-24
feat(env): Allow exporting env vars as dotenv format

miguelmig:
As the title says, this allows exporting currently active environment variables(excluding `PATH`) in a [dotenv](https://hexdocs.pm/dotenvy/dotenv-file-format.html) format.


src/cli/env.rs
```diff
@@ -34,8 +38,10 @@ impl Env {
 
         if self.json {
             self.output_json(&config, ts)
-        } else {
+        } else if self.shell.is_some() {
```
jdx: does this still default to `$SHELL`?

src/cli/env.rs
```diff
@@ -55,6 +61,15 @@ impl Env {
         }
         Ok(())
     }
+
+    fn output_dotenv(&self, config: &Config, ts: Toolset) -> Result<()> {
```
jdx, line 65: needs an e2e test

src/cli/env.rs
miguelmig: Probably not, will fix that!

PR #3197, opened 2024-11-25
Make Rebar backend depend on Erlang

eproxus:
Erlang is needed to install Rebar (or at least to run it).


eproxus:
@jdx Thanks for the quick merge! 🙏 Sorry I didn't have time to fix the commit message 😅

PR #3206, opened 2024-11-26
docs: remove shebang from tasks run doc

### LINKED ISSUE #3174: Shebang in `run=` has no effect, but using it in example creates impression it will work:
Shebang in `run=` has no effect, but using it in example creates impression it will work:

https://github.com/jdx/mise/blob/36a19f87b80aafb15f33dd818f3a4b4a0277c295/docs/tasks/toml-tasks.md?plain=1#L33-L36

_Originally posted by @powerman in https://github.com/jdx/mise/issues/2873#issuecomment-2453996738_

erickguan:
Fixes #3174

### DIFF
--- a/docs/tasks/toml-tasks.md
+++ b/docs/tasks/toml-tasks.md
@@ -5,7 +5,7 @@ Tasks can be defined in `mise.toml` files in different ways:
 ```toml
 [tasks.cleancache]
 run = "rm -rf .cache"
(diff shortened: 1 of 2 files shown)


jdx:
the documentation is correct, but it's functionality I just haven't built yet. What the shebang should do (I think) is change the `shell` of the task.

PR #3234, opened 2024-11-27
docs: document aqua/ubi better

jdx:


docs/dev-tools/backends/aqua.md
```diff
@@ -1,8 +1,25 @@
...
+mise has a reimplementation of aqua that knows how to work with these files to install tools.
+
+As of this writing, aqua is relatively new to mise and because a lot of tools are being converted from
+asdf to aqua, there may be some configuration in aqua tools that need to be tightened up. I put some
+common issues below and would strongly recommend contributing changes back to the aqua registry if you
+notice problems. The maintainer is super responsive and great to work with.
+
+If all else fails, you can disable aqua entirely with [`MISE_DISABLE_BACKENDS=aqua`](/configuration/settings.html#disable_backends).
+
+Currently aqua plugins don't support setting environment variables or doing more than simply downloading
```
syhol, line 20: @jdx Rather than aqua plugins, do you perhaps mean "aqua tools" or "the aqua backend"?

In the first paragraph you mention that aqua doesn't require plugins:

> aqua is the ideal backend to use for new tools since they don't require plugins

jdx: yes thank you

docs/dev-tools/backends/aqua.md
```diff
@@ -32,3 +49,31 @@ to use the aqua backend. To see these tools, run `mise registry | grep aqua:`.
...
+
+The aqua registry defines supported envs for each tool of the os/arch. I've noticed some of these
+are simply missing os/arch combos that are in fact supported—possibly because it was added after
+the registry was created for that tool.
+
+The fix is simple, just edit the `supported_envs` section of `registry.yaml` for the tool in question.
+
+### Using `version_filter` instead of `version_prefix`
+
+This is a weird one that doesn't causes weird issues in mise. In general in mise we like versions like
```
scop, line 67: Wondering if this was meant to be "that causes" or "that doesn't cause", now it's a bit between the two :)

PR #3240, opened 2024-11-27
feat: standalone tasks

jdx:
Allows running tasks with `mise run ./foo` or fetching tasks via http


powerman:
How is `mise run ./foo` differs from `mise exec -- ./foo`?

jdx:
it's run as a task so you get usage parsing

PR #3244, opened 2024-11-27
refactor: use aqua for shfmt

scop:


registry.toml
```diff
@@ -801,7 +801,7 @@ shell2http.backends = ["aqua:msoap/shell2http", "asdf:ORCID/asdf-shell2http"]
 shellcheck.backends = ["ubi:koalaman/shellcheck", "asdf:luizm/asdf-shellcheck"]
 shellcheck.os = ["linux", "macos"]
 shellspec.backends = ["aqua:shellspec/shellspec", "asdf:poikilotherm/asdf-shellspec"]
-shfmt.backends = ["asdf:luizm/asdf-shfmt"]
+shfmt.backends = ["aqua:mvdan/sh", "asdf:luizm/asdf-shfmt"]
 shfmt.os = ["linux", "macos"]
```
scop, line 805: Being new to editing the registry, I'm not sure what this `shfmt.os` is for, but it seems to me as if it could be unnecessary and shfmt on windows would work through aqua just fine (I've no way to test that right now though).

jdx: it will skip the release tests but you haven't defined any for shfmt so it's fine

PR #3262, opened 2024-11-28
fix(registry/github-cli): set exe=gh for aquq backend entry

jimeh:
Fixes error when trying to install/upgrade `gh`/`github-cli`, for example:

```
$ mise upgrade gh
mise ERROR failed to install aqua:cli/cli@2.63.0
mise ERROR Failed to install with ubi 'aqua:cli/cli@2.63.0': could not find any files named cli in the downloaded zip file
```


jdx:
`exe=` is a setting for ubi, not aqua

jimeh:
Interesting, because my local build of mise with the change in this PR did successfully upgrade `gh` to 2.63.0, while the latest release (2024.11.32) cannot upgrade it, complaining it couldn't find a file named `cli` in the zip file.

jdx:
it looks like what's happening here is an issue with the state, it's a bigger problem. I think the `~/.local/share/mise/installs/gh/.mise.backend` file is reporting ubi but we changed this to aqua. What mise should do is continue to use ubi (because you may have other versions installed and the versions installed with one backend may not be compatible with another so we keep you on whatever backend you used for the first install).

If you delete `~/.local/share/mise/installs/gh` it should fix this though, but let me know what is in that file to verify my suspicion first.

jimeh:
Ah, yeah that makes more sense. My `~/.local/share/mise/installs/gh/.mise.backend` file contains:

```
gh
ubi:cli/cli
```

And I can confirm that fully removing `gh` and re-isntalling 2.62.0 and upgrading to 2.63.0 works fine via aqua, and the state file is updated to contain details for aqua instead.

So I guess the issue might be that with the state set to ubi, but registry pointing to aqua (without the `exe`) option, it ends up running ubi without the exe option? Since me adding the exe option to aqua in registry, made the upgrade via ubi work again.

PR #3307, opened 2024-11-30
feat: Add `--json-extended` option to settings ls 

### LINKED ISSUE #3301: mise settings improvements
`mise settings` has been improved in the latest releases and now offers `json` output. Thank you!

```shell
❯ mise settings --json                    
{
  "disable_backends": "[\"java\"]",
  "experimental": true
}

❯ mise settings --toml                    
disable_backends = ["java"]
experimental = true

❯ mise settings                           
Key               Value    Source                    
disable_backends  ["java"] ~/.config/mise/config.toml
experimental      true     ~/.config/mise/config.toml
```

Would it be possible to make the following changes? (I can create PRs)
- `toml` and/or `json` output could include the source of the settings, like it does for the table output (I have a similar request for `mise env --json` in https://github.com/jdx/mise/issues/3012)
```json
{
  "experimental": { "value": true, "source": "~/.config/mise/config.toml" }
}
```
- arrays are shown as strings in the json output, but not in the toml output. I believe this might be a bug?

For context, I am trying to display them in https://github.com/hverlin/mise-vscode:
image

I have also noticed that `python.default_packages_file` does not show in the default output, even it's different from the default value. I might use the default output from `mise` going forward, but just curious why isn't it shown?

hverlin:
Fix #3301
- Add `--json-extended` output option to `mise settings ls`
- Update `--json` output to print arrays as json arrays
- ensure only one output option can be set

### DIFF
--- a/docs/cli/settings.md
+++ b/docs/cli/settings.md
@@ -31,6 +31,10 @@ List all settings
 
 Output in JSON format
 
+### `--json-extended`
+
+Output in JSON format with sources
+
 ### `-T --toml`
 
 Output in TOML format
--- a/docs/cli/settings/ls.md
+++ b/docs/cli/settings/ls.md
@@ -31,6 +31,10 @@ Use the local config file instead of the global one
 
 Output in JSON format
 
+### `--json-extended`
+
+Output in JSON format with sources
+
 ### `-T --toml`
 
 Output in TOML format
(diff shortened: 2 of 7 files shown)


jdx:
Maybe we want to call this --json-extended in case we want to add more to it later

hverlin:
> Maybe we want to call this --json-extended in case we want to add more to it later

Yeah, good idea. Updated it. I think that the setting `description` could be included later, for example

Is it good to merge?

Issue #3362, opened 2024-12-05
Cargo compilation of v2024.12.0 fails


avysk:
**Describe the bug**
I have v2024.11.37 installed via `cargo install`; and installation of v2024.12.0 via cargo fails with `error[E0425]: cannot find value OS in this scope`; the compiler complains about this: `let tarball_name = format!("otp-{ARCH}-{OS}.tar.gz");` in `erlang.rs`. Same with compiling from git. Running on FreeBSD 14.2-RELEASE.

**To Reproduce**

`cargo install mise`

**Expected behavior**

Compilation succeeds.

**`mise doctor` output**

```text
version: 2024.11.37 freebsd-x64 (2024-11-30)
activated: no
shims_on_path: yes

build_info:
  Target: x86_64-unknown-freebsd
  Features: DEFAULT, NATIVE_TLS
  Built: Sat, 30 Nov 2024 08:42:09 +0000
  Rust Version: rustc 1.83.0 (90b35a623 2024-11-26)
  Profile: release                                                          

shell:
  /usr/local/bin/bash
  GNU bash, version 5.2.37(0)-release (amd64-portbld-freebsd14.1)
  Copyright (C) 2022 Free Software Foundation, Inc.
  License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>                                                                                                                                                       This is free software; you are free to change and redistribute it.          There is NO WARRANTY, to the extent permitted by law.                                                                                                 dirs:                                                                         cache: ~/.cache/mise                                                        config: ~/.config/mise                                                      data: ~/.local/share/mise                                                   shims: ~/.local/share/mise/shims                                            state: ~/.local/state/mise

config_files:
  /usr~/.config/mise/config.toml

backends:
  aqua
  asdf
  cargo
  core
  gem
  go
  npm
  pipx
  spm
  ubi
  vfox

plugins:
  pipx  https://github.com/yozachar/asdf-pipx.git#81cb547

toolset:
  [...]

env_vars:
  (none)

settings:
  experimental  true /usr~/.config/mise/config.toml

1 problem found:

1. new mise version 2024.12.0 available, currently on 2024.11.37
```

avysk:
I think PR #3363 fixes this.

PR #3363, opened 2024-12-05
fix: allow compilation not on macOS, Linux, or Windows

avysk:
Allow erlang.rs work if compiled not on Mac, Linux or Windows.


avysk:
I think it fixes #3362

PR #3374, opened 2024-12-05
feat: add mise-task.json schema

### LINKED ISSUE #3369: Json schema does not support tasks files
**Describe the bug**

From the [docs](https://mise.jdx.dev/tasks/#task-configuration):

> If using a toml file for tasks, the file should be the same format as the [tasks] section of mise.toml but without the [task] prefix

```toml
task1 = "echo task1"
task2 = "echo task2"
task3 = "echo task3"

[task4]
run = "echo task4"
```

However, this is not supported in the current json schema, and the ide will return errors in those files:

image

**Expected behavior**

Json schema should not report errors in separate toml files for tasks where `[tasks]` prefix needs to be omitted

**`mise doctor` output**

```text
version: 2024.12.0 linux-x64 (20525d2 2024-12-04)
activated: yes
shims_on_path: no

build_info:
  Built: Wed, 4 Dec 2024 19:00:28 +0000
  Rust Version: rustc 1.83.0 (90b35a623 2024-11-26)
  Profile: release

shell:
  /usr/bin/zsh
  zsh 5.8.1 (x86_64-ubuntu-linux-gnu)

dirs:
  cache: ~/.cache/mise
  config: ~/.config/mise
  data: ~/.local/share/mise
  state: ~/.local/state/mise

config_files:
  ~/.config/mise/config.toml
  ~/Projects/.../.mise.local.toml

backends:
  aqua
  asdf
  cargo
  core
  gem
  go
  npm
  pipx
  spm
  ubi
  vfox

plugins:

toolset:
  ubi:jdx/usage@1.3.4

env_vars:
  MISE_SHELL=zsh

settings:

No problems found

```

hverlin:
Fix #3369


jdx:
I think there is one more thing we need to do: https://github.com/jdx/mise/blob/3222e48c35b915b2a396ceed9207554a78c0ae2d/scripts/publish-s3.sh#L15

jdx:
one more thing, we should probably add the schema comment to the docs—otherwise nobody will know it exists

docs/tasks/index.md
```diff
@@ -10,36 +10,77 @@ Here's my favorite features about mise's task runner:
...
 - last-modified checking to avoid rebuilding when there are no changes—requires minimal config
-- `mise watch` to automatically rebuild on changes—no configuration required, but it helps
+- [mise watch](./running-tasks.html#watching-files) to automatically rebuild on changes—no configuration required, but it helps
 - ability to write tasks as actual bash script files and not inside yml/json/toml strings that lack
   syntax highlighting and linting/checking support
 
 There are 2 ways to define tasks: [inside of `mise.toml` files](./toml-tasks.html) or as [standalone shell scripts](./file-tasks.html).
 
-## Task Environment Variables
+## Tasks in `mise.toml` files
```
hverlin, line 19: Hope it's ok to add these sections. I found that maybe having a quick example of the 2 sorts of tasks next to each on this page would help before they dive into the other pages

PR #3382, opened 2024-12-06
fix: Fix build on FreeBSD

yurivict:


yurivict:
Is this fixed in the master branch?

Issue #3012, opened 2024-11-13
Additional json output for mise doctor and mise env


hverlin:
Hi, would it be possible to add `json` output for more commands?
- `mise doctor --json` (this would allow one to extract `settings`, `dirs`, `warnings`, ...)
- `mise env --json` exists, however it does not indicate which config file was used to load each environment variable (like `mis ls --json` or `mise tasks --json`). Can we add another option to this command to indicate where the variable comes from?

hverlin:
For `mise doctor`, I think the only thing that's needed are the state directories (otherwise, everything is exposed through other commands like `version`, `backends`, `plugins`, `settings`...

One can use a similar logic as in the code to get the state dirs, though just wondering if having a hidden command to get them as json would work (maybe `mise dirs --json`)? I can create a PR

(unless `mise doctor --json` is better?)

The main reason is to know where the `shims` are located without to parse `mise doctor` output

PR #3389, opened 2024-12-07
feat: Add --json-extended option to mise env

hverlin:
Add `mise env --json-extended` which shows where each environment variable is sourced from.

see #3012


src/cli/env.rs
```diff
@@ -51,6 +58,61 @@ impl Env {
...
     }
 
+    fn output_extended_json(&self, config: &Config, ts: Toolset) -> Result<()> {
+        let mut res = BTreeMap::new();
+
+        ts.env_with_path(config)?.iter().for_each(|(k, v)| {
+            res.insert(k.to_string(), BTreeMap::from([("value", v.to_string())]));
+        });
+
+        config.env_with_sources()?.iter().for_each(|(k, v)| {
```
hverlin, line 68: This is iterating twice as `env_with_path` already does the same. If this is an issue, we would need to move the logic to add source/tool into `env_with_path` which I was not sure about given that this is used everywhere

jdx: these things almost never make any practical difference in mise

jdx:
your test is failing. I would pipe it to jq and just parse some of it

src/cli/env.rs
hverlin: Is it ok to keep as is, or should I move the logic in `env_with_path`?

jdx: it's fine like this

PR #3405, opened 2024-12-08
docs: Update index page

hverlin:
Just an idea of things which could be added on the home page to get a super quick idea of what `mise` can do.
This is all in Markdown, but it could be possible to build something a bit prettier, more like on the https://vite.dev homepage (obviously not as advanced, but for example, presenting some information into two columns would help)

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -32,7 +32,8 @@ features:
 ---
 
 ## Powerful polyglot tool version manager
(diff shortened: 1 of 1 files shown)


hverlin:
Maybe something like this could work well:

image

hverlin:
@jdx Would you be interested in having something like in the screenshot?

jdx:
yeah I think that's excellent, the landing page is definitely too spartan at the moment

hverlin:
will re-open another later

PR #3457, opened 2024-12-10
chore: Update usage-lib dependency

miguelmig:
Upgrades usage-lib dependency to benefit of the latest bugfixes


jdx:
we're already on the latest

https://github.com/jdx/mise/blob/195039ff2bbe702c7e80ace3fcaeb95cb02d018b/Cargo.lock#L3996

miguelmig:
Interesting, I need to figure out why mise tasks --usage still marks arguments with default as required then

PR #3459, opened 2024-12-10
docs: fixing elvish install instructions

ejrichards:
Previous instructions were not working on the latest version of Elvish, giving the following error:

```
Exception: wrong type: need ns, got map
```

I changed it so the namespace is updated in the `eval` callback.

I also changed the alias to use `edit:add-var` since the `fn` syntax is contained to scope, so if the previous code was in, for example, an if statement, it would not export the `mise` alias.

### DIFF
--- a/docs/installing-mise.md
+++ b/docs/installing-mise.md
@@ -334,14 +334,15 @@ make sure they match)
 Add following to your `rc.elv`:
 
 ```shell
-var mise: = (eval &ns=[&] &on-end=$put~ (mise activate elvish | slurp))
+var mise: = (ns [&])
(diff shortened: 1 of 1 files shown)


docs/installing-mise.md
```diff
@@ -334,14 +334,15 @@ make sure they match)
 Add following to your `rc.elv`:
 
 ```shell
-var mise: = (eval &ns=[&] &on-end=$put~ (mise activate elvish | slurp))
```
SolitudeSF, line 337: you coudve just wrapped [&] in (ns [&]) and it would've had same result, instead of splitting it into multiple statements, which has not relation to the issue.

PR #3466, opened 2024-12-11
fix: improve hook-env

jdx:
* check config filetime modification in millis so scripts work better
* only warn if files are untrusted
* added a `--no-hook-env` flag for activate for debugging


src/hook_env.rs
```diff
@@ -114,7 +114,7 @@ fn have_mise_env_vars_been_modified(watches: &HookEnvWatches) -> bool {
 
 #[derive(Debug, Default, Serialize, Deserialize)]
 pub struct HookEnvWatches {
-    latest_update: u64,
+    latest_update: u128,
```
jdx, line 117: my understanding is this shouldn't work because we serialize to messagepack which doesn't have bigint... but it makes the test pass so ¯\\\_(ツ)_/¯

PR #3489, opened 2024-12-12
feat: add settings for python venv creation

### LINKED ISSUE #1720: Add `without_pip` option to `_.python.venv`
Hi,

The new `_.python.venv` mechanism sounds very good.

Would it make sense to add a `without_pip` option for creation, mirroring `python -m venv --without-pip` ?

Virtual environment without pip are viable nowadays, since pip has a `--python` option.

### LINKED ISSUE #1721: Add `python` option to `_.python.venv`
When working in a virtual environment, it is sometimes useful to have the other global pythons to remain available.

For instance, when in a project directory with, say, a python 3.12 venv active, I sometimes need to run `tox -e py310` to run tests with another version.

When the global pythons are managed with mise and `tools.python` is set, the other global pythons are not available.

A possible solution would be to have `python` as an option of `_.python.venv` that would be used over `tools.python` which could then be absent.

Assuming that makes sense and fits with the path you are following, of course.

jdx:
Fixes #1721
Fixes #1720

### DIFF
--- a/docs/lang/python.md
+++ b/docs/lang/python.md
@@ -136,6 +136,9 @@ _.python.venv = ".venv" # relative to this file's directory
 _.python.venv = "/root/.venv" # can be absolute
 _.python.venv = "{{env.HOME}}/.cache/venv/myproj" # can use templates
 _.python.venv = { path = ".venv", create = true } # create the venv if it doesn't exist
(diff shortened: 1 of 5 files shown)


sbidoul:
Wonderful, thanks!

PR #3491, opened 2024-12-12
docs: Add more examples for toml tasks

hverlin:


docs/tasks/toml-tasks.md
```diff
@@ -143,10 +141,60 @@ run = 'echo {{flag(name=("myflag")}}'
 # runs: echo true
 ```
 
+```toml
+[tasks.maybeClean]
+run = """if [ '{{flag(name='clean')}}' = 'true' ]; then echo 'cleaning' ; fi""",
```
jdx: seems like a missed opportunity to show how multiline strings work in toml. The comma at the end is also a syntax error

docs/tasks/toml-tasks.md
```diff
@@ -143,10 +141,60 @@ run = 'echo {{flag(name=("myflag")}}'
...
+
+Here is another example with `deno`:
+
+```toml
+[tools]
+deno = 'latest'
+
+[tasks.download_task]
+description = "Shows that you can use deno in a task"
+shell = 'deno eval'
```
jdx: while this works, it can also be done with a shebang:

```toml
run = """
#!/usr/bin/env -S deno run --allow-env
"""
```

which I prefer even though users need to understand `env -S` (which I don't see as a downside, `env -S` is great). I think we should show both methods.

jdx: I don't know this for sure, but I suspect a shebang would also be easier for IDEs to syntax highlight, so I think that should be our preferred method here

docs/tasks/toml-tasks.md
```diff
@@ -143,10 +141,64 @@ run = 'echo {{flag(name=("myflag")}}'
...
+
+```toml
+[tools]
+deno = 'latest'
+
+[tasks.download_task]
+description = "Shows that you can use deno in a task"
+shell = 'deno eval' # or use a shebang: #!/usr/bin/env -S deno run
+run = """
+import ProgressBar from "jsr:@deno-library/progress";
```
jdx: sorry, this is what I meant

```suggestion
run = """
#!/usr/bin/env -S deno run
import ProgressBar from "jsr:@deno-library/progress";
```

docs/tasks/toml-tasks.md
hverlin: I was thinking that this solution is more portable, maybe? Need to try on windows to see what happens.

I am not sure if it makes a diff for syntax highlighting for languages that don't have `#` comments
image

hverlin: Ah wait, this is `mise` parsing the script actually, so it should work on window then.
https://github.com/jdx/mise/blob/d5781beedf06c75c3e977e3a0ee26c2d7f80958e/src/task/task_script_parser.rs#L312-L314

jdx: I think that is only partially done, the shell_type there is only used for knowing whether or not it should escape args but I plan to soon make it actually parse and execute the shebang directly instead of going through `sh -c`. I would still document it this way, the windows issue I'll fix soon.

jdx: I get that it doesn't work for deno, but what about `#!/usr/bin/env node`? I like the deno example, I'm just curious what vscode does.

hverlin: I am trying with bun (https://bun.sh/docs/runtime/shell) and it only works with `shell`

```toml
[tasks.bun]
run = """
#!/usr/bin/env -S bun -e

import { $ } from "bun";
const response = await fetch("https://example.com");
await $`cat < ${response} | wc -c`; // 1256
"""
```

```
1 | /var/folders/b9/_gbvqlqd4338k6c_t3jzbrzc0000gq/T/.tmpPjDr0I/script
          ^
error: Invalid flag "f" in regular expression
    at /private/tmp/test-mise1/[eval]:1:7
```
With shell, it works
```toml
[tasks.bun_shell]
shell = "bun -e"
run = """
import { $ } from "bun";
const response = await fetch("https://example.com");
await $`cat < ${response} | wc -c`; // 1256
"""
```
```
[bun_shell] $ import { $ } from "bun";
    1256
```

jdx:
this is just a note, but there actually is a scenario where you may prefer `shell = ` to the shebang, the shebang only works on a single run command, e.g.:

```toml
[tasks.foo]
run = [
"""
#!/usr/bin/env node
console.log("node")
""",
"""
#!/usr/bin/env bash
echo "foo"
""""
```

but of course `shell` would work on all of them. Why would someone want multiple run commands with different shells? no idea, but it's permitted.

docs/tasks/toml-tasks.md
hverlin: Seems that `just` is extracting the shebang for windows: https://just.systems/man/en/shebang-recipes.html
I almost feel like the `shell` option is more reliable then?

jdx: I think for bun you want to use `#!/usr/bin/env bun`

hverlin: For VSCode, it only works if you create a separate file. For inline strings, it appears that it can only be implemented at the plugin level (https://github.com/microsoft/vscode/issues/1751)

jdx: if only I knew the mise-vscode maintainer ;)

jdx: I _do_ think syntax highlighting with shebangs would be a killer feature, or even if you got it to default to multiline strings as bash

jdx: btw we should add your extension to the docs

hverlin: Sure, I will make a PR next week. Want to check the windows support first

jdx: might be worth waiting until I get the shebang parsing fully in place—right now it almost certainly will fail with any shebang—though a user could work around that with run_windows

docs/tasks/toml-tasks.md
```diff
@@ -143,10 +141,125 @@ run = 'echo {{flag(name=("myflag")}}'
...
+
+::: code-group
+
+```toml [python]
+[tools]
+python = 'latest'
+
+[tasks.python_task]
+run = """
+#!/usr/bin/env -S python
```
jdx: we should reserve `env -S` for shebangs that have multiple args (like `deno run`). We should also document on this page what the heck `env -S` does in a callout of some kind. I suspect most users won't have a clue what it means. I think I already do that in the tips and tricks doc but this page is going to be far more heavily trafficked.

jdx:
sorry for a bit of back and forth on this, but I think I know what I want to do in regards to documenting `shell` vs `shebang`. I think it's great to have multiple languages but I think for a few reasons shebangs should be the golden path. One is that it's more flexible, another is we should (at some point) get syntax highlighting for it, so I think we shouldn't document `shell` at all in the sort of introduction part of this.

What we do need on this page though is an exhaustive list of all the properties you can pass to tasks like we have for [settings](https://mise.jdx.dev/configuration/settings.html). That could be done here or in a follow-up at some point.

docs/tasks/toml-tasks.md
hverlin: Sounds good, I will update 👍

jdx:
the tricky thing about that list though is almost all properties are associated with file and toml tasks. So I'm not sure where it should live.

jdx:
one more thing, I've been using `"""`, but I think we actually want to use `'''`

jdx:
oh whoops, seems I am not using `sh -c` when there is a shebang:

https://github.com/jdx/mise/blob/usage-upgrade/src/cli/run.rs#L375

I forgot about that

hverlin:
Still had one commit to make 😅

docs/tasks/toml-tasks.md
```diff
@@ -60,6 +85,246 @@ dir = "{{cwd}}"
...
+
+```toml
+[tasks.build]
+raw = true
+run = 'echo "Enter your name:"; read name; echo "Hello, $name!"'
+```
+
+## quiet
+
+Set `quiet = false` to supress mise additional output.
```
syhol, line 326: @hverlin should this be `quiet = true`?

jdx: wait I think I missed this, we don't actually have this lol, this was a feature idea

jdx: maybe I'll just implement it so we don't need to remove it

jdx: fixed in #3514

hverlin: yeah, this is why I was asking in the discord

hverlin: Other fixes: https://github.com/jdx/mise/pull/3501

PR #3492, opened 2024-12-12
fix: Use arguments for to pass staged filenames to pre-commit task

joshbode:


src/cli/generate/git_pre_commit.rs
```diff
@@ -51,10 +51,23 @@ impl GitPreCommit {
...
+  rm -f "${{PIPE}}"
+}}
+trap 'cleanup' EXIT INT TERM
+
+git diff-index --cached --name-only HEAD > "${{PIPE}}" &
+while read -r ARG; do
+  set -- "$@" "${{ARG}}"
+done < "${{PIPE}}"
+
+exec mise run "{task}" "$@"
```
jdx, line 70: I already merged this but realized this should be `exec mise run "{task}" -- "$@"` though the only time that would matter is if you added a flag to the pre-commit task that had the exact string as a filename

joshbode: ah - good point, but yes... pretty unlikely - and hopefully people avoid filenames with leading dashes :)

PR #3494, opened 2024-12-12
feat(registry): add aqua as primary backend for tools using ubi currently

jimeh:
Add aqua as the primary backend for all possible tools that are using ubi as the primary backend currently.

This was achived by comparing `mise registry` and `aqua list` against each other with the following command:

    comm -12 \
      <(mise registry | awk '{ print $2 }' | cut -d'[' -f1 | grep -E '^ubi:' | sed 's/^ubi://' | sort -u) \
      <(aqua list | cut -d',' -f2 | sort -u)

I have not tested all of these, so I sadly offer no guarantee every single tool installs correctly from aqua, but my assumption is that we assume aqua's registry works.


jdx:
well you certainly broke something given the test failures. If you want to do this you'll need to test one individually, we can't just assume they work.

jdx:
@jimeh I'm going to close just to keep things tidy, but if you want to pick this up again I'm happy to reopen

jimeh:
@jdx sure thing, I haven't gotten back around to looking at this properly, cause I'm having issues running the e2e tests locally on macOS (it either complains it can't find a `mise` executable, or complains `mise` is a macOS MachO executable).

I'll try in a VM when I have a chance, or digging deeper into what's causing the issues.

PR #3501, opened 2024-12-12
docs: add explanation about shebang

hverlin:
See #3491

### DIFF
--- a/docs/tasks/toml-tasks.md
+++ b/docs/tasks/toml-tasks.md
@@ -137,13 +137,7 @@ You can hide a task from the list of available tasks by setting `hide = true`:
 ```toml
 [tasks.cleancache]
 run = "rm -rf .cache"
-hide = true # hide this task from mise tasks / mise run
-```
-
-Note that if you want to see hidden tasks, you can use the `--hidden` flag:
-
-```shell
-mise tasks ls --hidden
+hide = true # hide this task from mise tasks ls / mise run
 ```
 
 ## alias
@@ -208,7 +202,7 @@ shell = 'bash -c'
 run = "cargo clippy"
 ```
 
-or use a `shebang`:
+or use a shebang:
 
 ```toml
 [tasks.lint]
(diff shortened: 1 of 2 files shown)


docs/tasks/toml-tasks.md
```diff
@@ -282,12 +298,28 @@ ruby = 'latest'
...
 """
 ```
 
 :::
 
+::: details What's a shebang? What's the difference between `#!/usr/bin/env` and `#!/usr/bin/env -S`
+
+A shebang is the character sequence `#!` at the beginning of a script file that tells the system which program should be used to interpret/execute the script.
+The [env command](https://manpages.ubuntu.com/manpages/jammy/man1/env.1.html) comes from GNU Coreutils.
+
```
hverlin, line 306: Should I indicate here that `mise` is going to create a file and run it instead?
https://github.com/jdx/mise/blob/224a0b523ba63f54b5eaf89a1fc5722188ce3f99/src/cli/run.rs#L375-L383

docs/tasks/toml-tasks.md
```diff
@@ -137,7 +137,13 @@ You can hide a task from the list of available tasks by setting `hide = true`:
...
 [tasks.cleancache]
 run = "rm -rf .cache"
-hide = true # hide this task from the list
+hide = true # hide this task from mise tasks / mise run
+```
+
+Note that if you want to see hidden tasks, you can use the `--hidden` flag:
+
+```shell
+mise tasks ls --hidden
```
jdx: I think we can take this out, it's already documented in the reference and I think easy to figure out

docs/tasks/toml-tasks.md
```diff
@@ -180,20 +186,29 @@ outputs = ['target/debug/mycli']
...
+You can specify a `shell` command to run the script with (default is [`sh -c`](/configuration/settings.html#unix_default_inline_shell_args) or [`cmd /c`](/configuration/settings.html#windows_default_inline_shell_args)):
 
 ```toml
 [tasks.lint]
 shell = 'bash -c'
 run = "cargo clippy"
 ```
 
-or use a shebang:
+or use a `shebang`:
```
jdx: Shebang shouldn't be in backticks

docs/tasks/toml-tasks.md
jdx: So far I don't believe anyone has needed to be aware of this behavior for any reason so I think no

Issue #3451, opened 2024-12-10
powershell


jdx:
we should add powershell (see [zsh](https://github.com/jdx/mise/blob/main/src/shell/zsh.rs) as an example) which would allow windows users to not need shims and they'd be able to have env vars without running tasks

Mathieson:
Is this feature released? I am not having luck with it.

thomasaarholt:
The documentation has not been updated. You can add this to your $PROFILE:
`Invoke-Expression (& { (mise activate pwsh | Out-String) })`

airtonix:
@thomasaarholt  does this require a certain version of windows or powershell? 

```pwsh
Unable to find type [System.Management.Automation.LocationChangedEventArgs].
At line:54 char:30
+ ... pwd_hook = [EventHandler[System.Management.Automation.LocationChanged ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (System.Manageme...hangedEventArgs:TypeName) [], RuntimeException
    + FullyQualifiedErrorId : TypeNotFound

The property 'LocationChangedAction' cannot be found on this object. Verify that the property exists and can be set.
At line:66 char:13
+             $ExecutionContext.SessionState.InvokeCommand.LocationChan ...
+             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
    + FullyQualifiedErrorId : PropertyAssignmentException

```

thomasaarholt:
Huh, I'm sorry that I didn't think to check that. I'm not a power user on PowerShell, so I pasted this context and that error into ChatGPT and got back

> The error you're encountering, Unable to find type `[System.Management.Automation.LocationChangedEventArgs]`, typically arises when attempting to use mise with Windows PowerShell 5.1. This version lacks certain types and features present in newer releases, leading to compatibility issues.
> 
> Solution: Upgrade to PowerShell 7.

I believe it is generally recommended to upgrade from PowerShell 5. Do you have reasons not to do this?

hollowshiroyuki:
On Windows there's different "Powershells", the default installed with the system is called "Windows Powershell" by Microsoft, it's Powershell 5.1 and you can't upgrade it.
[...]

PR #3506, opened 2024-12-12
feat: Add Powershell support

fgilcc:
Addresses issue #3451 

I was just thinking of doing this and seeing the issue gave me the rest of the motivation I needed.

The implementation takes some inspiration from direnv. https://github.com/direnv/direnv/pull/1171/files#diff-0ae809380696f548da1b4494718ec87231367232b99ca0c9472021f95645e1d3

Couldn't find a non intrusive way to emulate precmd and cmd_not_found functions. The workarounds I found would involve overwriting existing functions or events and possibly break user configurations and other scripts. 

Seems to work in both Linux and windows.


jdx:
wow awesome! code seems good to me, but I have nowhere near the powershell experience to know. lmk when you want me to test it on my windows laptop.

We will also need to update the docs to show how to install it.

src/shell/pwsh.rs
```diff
@@ -0,0 +1,198 @@
...
+
+            function mise {{
+
+                $code = [System.Management.Automation.Language.Parser]::ParseInput($MyInvocation.Line.Substring($MyInvocation.OffsetInLine - 1), [ref]$null, [ref]$null)
+                $myLine = $code.Find({{ $args[0].CommandElements }}, $true).CommandElements | ForEach-Object {{ $_.ToString() }} | Join-String -Separator ' '
+                $command, [array]$arguments = Invoke-Expression ('Write-Output -- ' + $myLine)
+                if ($null -eq $arguments) {{ $arguments = @() }} 
+
+
+                if ($arguments.Lenght -eq 0) {{
```
jdx: typo?

jdx:
we will need some equivalent of precmd though, otherwise if you edit a config file—even with `mise use`—it won't update PATH. The not_found handler isn't a big deal though.

jdx:
seems there are some workarounds here but in general it isn't supported: https://github.com/PowerShell/PowerShell/issues/14484 what a shame

If nothing else, we could call `hook-env` whenever `mise` is run. That would make `mise use` work, but editing outside of mise would not work.

fgilcc:
> seems there are some workarounds here but in general it isn't supported: [PowerShell/PowerShell#14484](https://github.com/PowerShell/PowerShell/issues/14484) what a shame
> 
> If nothing else, we could call `hook-env` whenever `mise` is run. That would make `mise use` work, but editing outside of mise would not work.

I also ran into this issue while testing. Installed python but the path was not updated. Thought of adding hook-env at the end of the function but did not want to deviate from how it works in other shells. If its fine I will add it.

jdx:
I think it's ok if it's the best we can do. It's worth exploring the alternatives too, I'll have a look at them later. Even if they're a bit hacky we might be able to make it optional behavior.

jdx:
I feel like we should do something like this:

```powershell
function PSConsoleHostReadLine {
    [Microsoft.PowerShell.PSConsoleReadLine]::ReadLine($Host.Runspace, $ExecutionContext)

    _mise_hook
}
```

and put it behind a flag: `mise activate pwsh --ps-readline`

that way people can have this behavior but since it's hacky they'll have the opportunity to opt-in

jdx:
that said, maybe we should do the prompt strategy? I just remembered we're already doing that for bash and nobody has complained about it:

https://github.com/jdx/mise/blob/9365ad12fc1178fab0f1a82659a7406d320edb04/src/shell/bash.rs#L51

jdx:
out of these 3 options, I'm cool with whatever you think is best

fgilcc:
> that said, maybe we should do the prompt strategy? I just remembered we're already doing that for bash and nobody has complained about it:
> 
> https://github.com/jdx/mise/blob/9365ad12fc1178fab0f1a82659a7406d320edb04/src/shell/bash.rs#L51
> 
> I think the annoying thing is likely to be that we'd need to be careful to reuse the existing prompt which is similarly to the annoying logic we have in other shells for the not_found handler

Went with the prompt. 

Didn't like the solution of restoring the previous chpwd and prompt functions on mise deactivate as it would remove any function that was combined on top later on so I just decided to make _mise_hook do nothing on mise deactivate

fgilcc:
Added some simple docs where I could find. They're not great for powershell; profile location is different on windows and by default the .config/powershell folder is not created, which will make the line append fail.

hyunbinseo:
- Windows 11
- PowerShell 7.4.6
- Installed using winget
- Updated using `mise self-update`

Following command simply logged text in the terminal:

```powershell
mise activate pwsh
```

```powershell
$env:MISE_SHELL = 'pwsh'
$env:__MISE_ORIG_PATH = $env:PATH

function mise {
    # Read line directly from input to workaround powershell input parsing for functions
    $code = [System.Management.Automation.Language.Parser]::ParseInput($MyInvocation.Statement.Substring($MyInvocation.OffsetInLine - 1), [ref]$null, [ref]$null)
    $myLine = $code.Find({ $args[0].CommandElements }, $true).CommandElements | ForEach-Object { $_.ToString() } | Join-String -Separator ' '
    $command, [array]$arguments = Invoke-Expression ('Write-Output -- ' + $myLine)

    if ($null -eq $arguments) {
        & C:\Users\seohy\AppData\Local\Microsoft\WinGet\Packages\jdx.mise_Microsoft.Winget.Source_8wekyb3d8bbwe\mise\bin\mise.exe
        return
    }

    $command = $arguments[0]
    $arguments = $arguments[1..$arguments.Length]

    if ($arguments -contains '--help') {
        return & C:\Users\seohy\AppData\Local\Microsoft\WinGet\Packages\jdx.mise_Microsoft.Winget.Source_8wekyb3d8bbwe\mise\bin\mise.exe $command $arguments
    }

    switch ($command) {
        { $_ -in 'deactivate', 'shell', 'sh' } {
            if ($arguments -contains '-h' -or $arguments -contains '--help') {
                & C:\Users\seohy\AppData\Local\Microsoft\WinGet\Packages\jdx.mise_Microsoft.Winget.Source_8wekyb3d8bbwe\mise\bin\mise.exe $command $arguments
            }
            else {
                & C:\Users\seohy\AppData\Local\Microsoft\WinGet\Packages\jdx.mise_Microsoft.Winget.Source_8wekyb3d8bbwe\mise\bin\mise.exe $command $arguments | Out-String | Invoke-Expression -ErrorAction SilentlyContinue
            }
        }
        default {
            & C:\Users\seohy\AppData\Local\Microsoft\WinGet\Packages\jdx.mise_Microsoft.Winget.Source_8wekyb3d8bbwe\mise\bin\mise.exe $command $arguments
            $status = $LASTEXITCODE
            if ($(Test-Path -Path Function:\_mise_hook)){
                _mise_hook
            }
            # Pass down exit code from mise after _mise_hook
            pwsh -NoProfile -Command exit $status
        }
    }
}

function Global:_mise_hook {
    if ($env:MISE_SHELL -eq "pwsh"){
        & C:\Users\seohy\AppData\Local\Microsoft\WinGet\Packages\jdx.mise_Microsoft.Winget.Source_8wekyb3d8bbwe\mise\bin\mise.exe hook-env $args -s pwsh | Out-String | Invoke-Expression -ErrorAction SilentlyContinue
    }
}

function __enable_mise_chpwd{
    if (-not $__mise_pwsh_chpwd){
        $Global:__mise_pwsh_chpwd= $true
        $_mise_chpwd_hook = [EventHandler[System.Management.Automation.LocationChangedEventArgs]] {
            param([object] $source, [System.Management.Automation.LocationChangedEventArgs] $eventArgs)
            end {
                _mise_hook
            }
        };
        $__mise_pwsh_previous_chpwd_function=$ExecutionContext.SessionState.InvokeCommand.LocationChangedAction;

        if ($__mise_original_pwsh_chpwd_function) {
            $ExecutionContext.SessionState.InvokeCommand.LocationChangedAction = [Delegate]::Combine($__mise_pwsh_previous_chpwd_function, $_mise_chpwd_hook)
        }
        else {
            $ExecutionContext.SessionState.InvokeCommand.LocationChangedAction = $_mise_chpwd_hook
        }
    }
}
__enable_mise_chpwd
Remove-Item -ErrorAction SilentlyContinue -Path Function:/__enable_mise_chpwd

function __enable_mise_prompt {
    if (-not $__mise_pwsh_previous_prompt_function){
        $Global:__mise_pwsh_previous_prompt_function=$function:prompt
        function global:prompt {
            if (Test-Path -Path Function:\_mise_hook){
                _mise_hook
            }
            & $__mise_pwsh_previous_prompt_function
        }
    }
}
__enable_mise_prompt
Remove-Item -ErrorAction SilentlyContinue -Path Function:/__enable_mise_prompt

_mise_hook
if (-not $__mise_pwsh_command_not_found){
    $Global:__mise_pwsh_command_not_found= $true
    function __enable_mise_command_not_found {
        $_mise_pwsh_cmd_not_found_hook = [EventHandler[System.Management.Automation.CommandLookupEventArgs]] {
            param([object] $Name, [System.Management.Automation.CommandLookupEventArgs] $eventArgs)
            end {
                if ([Microsoft.PowerShell.PSConsoleReadLine]::GetHistoryItems()[-1].CommandLine -match ([regex]::Escape($Name))) {
                    if (& C:\Users\seohy\AppData\Local\Microsoft\WinGet\Packages\jdx.mise_Microsoft.Winget.Source_8wekyb3d8bbwe\mise\bin\mise.exe hook-not-found -s pwsh -- $Name){
                        _mise_hook
                        if (Get-Command $Name -ErrorAction SilentlyContinue){
                            $EventArgs.Command = Get-Command $Name
                            $EventArgs.StopSearch = $true
                        }
                    }
                }
            }
        }
        $current_command_not_found_function = $ExecutionContext.SessionState.InvokeCommand.CommandNotFoundAction
        if ($current_command_not_found_function) {
            $ExecutionContext.SessionState.InvokeCommand.CommandNotFoundAction = [Delegate]::Combine($current_command_not_found_function, $_mise_pwsh_cmd_not_found_hook)
        }
        else {
            $ExecutionContext.SessionState.InvokeCommand.CommandNotFoundAction = $_mise_pwsh_cmd_not_found_hook
        }
    }
    __enable_mise_command_not_found
    Remove-Item -ErrorAction SilentlyContinue -Path Function:/__enable_mise_command_not_found
}
```

Had to paste the above content to the `$PROFILE` for the activation to work:

```powershell
if (!(Test-Path -Path $PROFILE)) {
  New-Item -ItemType File -Path $PROFILE -Force
}

notepad $PROFILE
```

thomasaarholt:
Here's a simpler line, @hyunbinseo:

`Invoke-Expression (& { (mise activate pwsh | Out-String) })`

PR #3527, opened 2024-12-13
docs: add vitepress-plugin-group-icons

hverlin:
See https://vpgi.vercel.app/, this allows adding a title to the code blocks. (also used by vite: https://github.com/vitejs/vite/blob/main/docs/package.json#L15)
- remove unneeded code groups
- make tiles clickable on the home page
- add python + uv script example


hverlin:
```
 mise $ /home/runner/.local/share/mise/installs/firebase/13.29.1/firebase --version
  13.29.0
  mise ERROR firebase: 
     0: expected output not found: 13.29.1, got: 13.29.0
```
Unexpected indeed 😅

Issue #1898, opened 2024-04-13
Missing (any python version causes venv not be created


scop:
```toml
[tools]
python = [
   "3.12",
   "3.13-dev",
   "3.10",
]
[env]
_.python.venv = { path = "venv", create = true }
```

With this config, if venv does not exist and _any_ of the Python versions specified in the config are missing, the resulting error and behavior when `cd`ing to the dir is (using 3.10 missing as the example)
```
mise no venv found at: .../venv

To create a virtualenv manually, run:
python -m venv .../venv
mise missing: python@3.10
```

In combination with #1897, this seems a bit odd given that the missing version here would not even be the version with which the venv would be created.

Further, when not cd'ing to the dir but just hitting enter in it (bash integration), the error message is the same _except_ the last `mise missing: ...` line is not output. This got me quite confused because it was the scenario where I hit the issue in, and I did not know that one of the python versions was missing.

Maybe either output the "mise missing: ..." also when just hitting the enter and not cd'ing, and/or not bother with missing python versions regarding venv creation in case the venv would not be created or activated with the missing version?

mise version 2024.4.4 linux-x64 (37b44e2 2024-04-11)

ocervell:
Agree, this also blocks a normal `ls` command so it is a weird behavior.

syhol:
## The "no venv" warning

> this also blocks a normal ls command

It doesn't block anything, it just print the warning before every shell prompt inside the directory, which I think is a really good thing.

> Further, when not cd'ing to the dir but just hitting enter in it (bash integration), the error message is the same except the last mise missing: ... line is not output. This got me quite confused because it was the scenario where I hit the issue in, and I did not know that one of the python versions was missing.

I think this is a great point and should be addressed.
Today the warning is fine when `create = false` or `create` is not set. But when `create = true` is set, it should have better instructions on what is wrong and what should be done to resolve it. My proposal is:
```console
mise WARN  no venv found at: .../venv
mise will automatically create the venv once all requested python versions are installed.
To install the missing python tools and create the venv, please run:
mise install
```

I don't think we should change the behaviour of the existing `mise missing: python@3.10` warning because thats part of the `status.missing_tools` setting.

## Only block venv creation when missing the required python version.

> In combination with https://github.com/jdx/mise/issues/1897, this seems a bit odd given that the missing version here would not even be the version with which the venv would be created.

[...]

PR #3573, opened 2024-12-15
docs: better warning when venv auto create is skipped

syhol:
```console
mise is 📦 v2024.12.10 via 🥟 v1.1.38 via  v23.3.0 via 🐍 v3.13.0 (.venv) via 🦀 v1.83.0 
❯ rm -rf .venv
mise WARN  no venv found at: ~/Code/syhol/mise/.venv

mise will automatically create the venv once all requested python versions are installed.
To install the missing python versions and create the venv, please run:
mise install

mise is 📦 v2024.12.10 via 🥟 v1.1.38 via  v23.3.0 via 🐍 v3.13.0 (.venv) via 🦀 v1.83.0 
❯ vim mise.toml # <-- changing env._.python.venv.create from true to false
mise WARN  no venv found at: ~/Code/syhol/mise/.venv

To create a virtualenv manually, run:
python -m venv ~/Code/syhol/mise/.venv
mise WARN  missing: python@3.9.21
```

### DIFF
--- a/src/config/env_directive.rs
+++ b/src/config/env_directive.rs
@@ -290,12 +290,18 @@ impl EnvResults {
                             .into_iter()
                             .chain(env::split_paths(&env_vars[&*PATH_KEY]))
                             .collect::<Vec<_>>();
-                        if ts
+                        let missing = ts
                             .list_missing_versions()
                             .iter()
-                            .any(|tv| tv.ba().tool_name == "python")
-                        {
-                            debug!("python not installed, skipping venv creation");
+                            .any(|tv| tv.ba().tool_name == "python");
+                        if missing {
+                            warn!(
+                                "no venv found at: {p}\n\n\
+                                mise will automatically create the venv once all requested python versions are installed.\n\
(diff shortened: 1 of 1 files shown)


jdx:
So showing warnings for literally anything inside of `hook-env` I've found to be almost always a bad idea. If for some reason you don't want this to be displayed, you'll be annoyed with it constantly. I think we should keep this as debug level. If we _were_ to do it, it would certainly have to fit on a single line.

For example, I recently removed warnings for similar situations with `_.source` and `_.file` so mise just silently ignores their non-existence.

jdx:
I think the better thing to do here is put the information into `mise dr`

syhol:
I'm not actually adding a new warning here, just changing the existing warning if it's create = true. Sorry I should have included more context in my description.

It comes from this ticket: https://github.com/jdx/mise/issues/1898#issue-2241448074

If you still want to close this PR, that's no problem.

jdx:
oh I see, I saw the debug! and assumed you were increasing the visibility, it's probably fine. If people complain, we can always tweak it.

PR #3582, opened 2024-12-15
docs: add missing go_set_goroot setting to Go document

liangyi9812:
Reference: https://mise.jdx.dev/configuration/settings.html#go_set_goroot


jdx:
do this instead to generate it:

https://github.com/jdx/mise/blob/80d5b8d78dbc15e57751c518fd6693fb4c432ab5/docs/lang/erlang.md?plain=1#L24-L29

PR #3602, opened 2024-12-16
feat: add zizmore

boris-smidt-klarrio:


jdx:
should just use the cargo backend

Issue #3356, opened 2024-12-04
add brew/windows installs to getting started carousel


jdx:

hverlin:
I can take this. However, on the https://mise.jdx.dev/getting-started.html page, we will have to clarify that some installers will automatically [activate](https://github.com/Homebrew/homebrew-core/blob/ab9a669e4bbc9c5326714ff36891978de84b586b/Formula/m/mise.rb#L50) / use [shims](https://mise.jdx.dev/installing-mise.html#windows-scoop) etc. 

Some also automatically install the [autocompletion](https://github.com/Homebrew/homebrew-core/blob/ab9a669e4bbc9c5326714ff36891978de84b586b/Formula/m/mise.rb#L45) (also for [macport](https://github.com/macports/macports-ports/blob/master/sysutils/mise/Portfile#L34))

Based on this, I feel like maybe the `curl` install should perhaps offer you to install the autocompletion (or add a message about it at the end of the installation)

hverlin:
Maybe we can leverage something like https://vitepress-plugins.sapphi.red/tabs/ here (like on this [page](https://infisical.com/docs/cli/overview))

jdx:
> I can take this. However, on the https://mise.jdx.dev/getting-started.html page, we will have to clarify that some installers will automatically [activate](https://github.com/Homebrew/homebrew-core/blob/ab9a669e4bbc9c5326714ff36891978de84b586b/Formula/m/mise.rb#L50) / use [shims](https://mise.jdx.dev/installing-mise.html#windows-scoop) etc.
I believe just those 2.

> Some also automatically install the [autocompletion](https://github.com/Homebrew/homebrew-core/blob/ab9a669e4bbc9c5326714ff36891978de84b586b/Formula/m/mise.rb#L45) (also for [macport](https://github.com/macports/macports-ports/blob/master/sysutils/mise/Portfile#L34))

I think they all do except mise.run

> Based on this, I feel like maybe the curl install should perhaps offer you to install the autocompletion (or add a message about it at the end of the installation)

I'm cool with a message, but automating it is going to be impossible to get right.

> Maybe we can leverage something like https://vitepress-plugins.sapphi.red/tabs/ here (like on this [page](https://infisical.com/docs/cli/overview))

it's already included, see https://mise.jdx.dev/installing-mise.html#apt

jdx:
ah ok, sounds good to me—though I'm also content with code comments

PR #3613, opened 2024-12-16
docs: add getting-started carousel

hverlin:
Add getting-started carousel  (#3356)

The tabs sync across the page. This should make it easier for users to follow the activation step.

image

image

image

image


hverlin:
The idea is that the `tabs` will sync across the page. Example for curl vs windows:

image

image

hverlin:
Let me know if this approach works for you. We can also split Windows scoop/winget into two separate tabs

PR #3640, opened 2024-12-17
fix: sort keys in `__MISE_DIFF` to make the serialised value deterministic

joshbode:
`__MISE_DIFF` changes value randomly when updating if any environment variables
are set by `.mise.toml` since the keys (of `new` and `old`) are not being sorted
when serialised.

(I use the value of `__MISE_DIFF` to determine if `mise` has updated the
environment to colour my prompt)

### DIFF
--- a/src/env_diff.rs
+++ b/src/env_diff.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, HashMap};
+use std::collections::BTreeMap;
 use std::ffi::OsString;
 use std::fmt::Debug;
 use std::io::prelude::*;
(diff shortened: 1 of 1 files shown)


src/env_diff.rs
```diff
@@ -9,21 +9,33 @@ use eyre::Result;
...
 
 use crate::env::PATH_KEY;
 use crate::{cmd, file};
 
-#[derive(Default, Serialize, Deserialize)]
+#[derive(Default, serde_derive::Serialize, serde_derive::Deserialize)]
 pub struct EnvDiff {
-    #[serde(default)]
+    #[serde(serialize_with = "ordered_map")]
     pub old: HashMap<String, String>,
```
jdx: just change this to IndexMap or BTreeMap instead

joshbode: oh, that's easier :)

PR #3660, opened 2024-12-17
docs: add gem backend to index

syhol:
## Index and sidebar
image image

## Gem page
image

### DIFF
--- a/docs/dev-tools/backends/gem.md
+++ b/docs/dev-tools/backends/gem.md
@@ -41,12 +41,3 @@ Or you can reinstall all gems with:
 ```sh
 mise install -f "gem:*"
 ```
-
-## Settings
-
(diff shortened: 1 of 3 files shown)


docs/dev-tools/backends/gem.md
```diff
@@ -41,12 +41,3 @@ Or you can reinstall all gems with:
...
 ```
-
-## Settings
-
-Set these with `mise settings set [VARIABLE] [VALUE]` or by setting the environment variable listed.
-
-<script setup>
-import Settings from '/components/settings.vue';
-</script>
-<Settings child="gem" :level="3" />
```
syhol, line 52: I can't see any gem settings in http://localhost:5173/configuration/settings.html and it was causing the page to error.

syhol:
@jdx is too fast

PR #3670, opened 2024-12-18
feat: Allows control of config_root for global config

bnorick:
- Adds `MISE_GLOBAL_CONFIG_ROOT` to control config_root for global config file which previously was always `env::HOME` (which remains the default, when unconfigured)
- Allows `MISE_IGNORED_CONFIG_PATHS` to specify explicit paths of config files to ignore


bnorick:
Seems reasonable to be able to pass specific paths in `MISE_IGNORED_CONFIG_PATHS`, so I included that here too since I needed it at the same time. If you think this isn't the best way to handle that, happy to address it.

PR #3716, opened 2024-12-19
refactor(registry): use aqua:skaji/relocatable-perl for perl

scop:


registry.toml
```diff
@@ -1220,7 +1220,7 @@ pdm.backends = ["pipx:pdm", "asdf:1oglop1/asdf-pdm"]
 pdm.test = ["pdm --version", "PDM, version {{version}}"]
 peco.backends = ["aqua:peco/peco", "asdf:asdf-community/asdf-peco"]
 periphery.backends = ["asdf:MontakOleg/asdf-periphery"]
-perl.backends = ["asdf:ouest/asdf-perl"]
+perl.backends = ["aqua:skaji/relocatable-perl", "asdf:ouest/asdf-perl"]
```
scop, line 1223: relocatable-perls are available for Linux and macOS only. I wonder if that's a problem re this entry?

scop, line 1223: I did not add a perl.test, because I'm not sure how to match e.g. the perl --version output for those relocatable perls for their "package" versions, for example 5.40.0.1 contains
```
This is perl 5, version 40, subversion 0 (v5.40.0) built for x86_64-linux
```
(The `.1` is missing from the output)

jdx: No because asdf plugins don't work on windows anyways

PR #3718, opened 2024-12-19
docs: update cookbook

hverlin:
WIP, starting to collect ideas from #3645


hverlin:
CC @syhol in case you have other suggestions. Otherwise, we can make create additional PRs maybe

docs/mise-cookbook/nodejs.md
```diff
@@ -0,0 +1,109 @@
...
+
+[tasks.info]
+description = "Print project information"
+run = '''
+echo "Project: $PROJECT_NAME"
+echo "NODE_ENV: $NODE_ENV"
+'''
+```
+
+## Example with `pnpm`
```
syhol, line 79: Personally I don't use camelCase in my task names. You could use either:
- `[tasks."pnpm:install"]`
- `[tasks.pnpm-install]`

I don't feel strongly about it though, so just a suggestion.

docs/mise-cookbook/docker.md
```diff
@@ -0,0 +1,59 @@
...
+❯ mise docker
+[docker] $ docker run --pull=always -it --rm --entrypoint bash jdxcode/mise:latest
+# latest: Pulling from jdxcode/mise
+# Digest: sha256:eecc479b6259479ffca5a4f9c68dbfe8631ca62dc59aa60c9ab5e4f6e9982701
+# Status: Image is up to date for jdxcode/mise:latest
+root@75f179a190a1:/mise# eval "$(mise activate bash)"
+root@75f179a190a1:/mise# echo "" >/mise/config.toml
+root@75f179a190a1:/mise# mise prune --yes
+mise pruned configuration links
+mise python@3.13.1 ✓ remove /mise/cache/python/3.13.1
```
syhol, line 57: Might be worth putting a little comment  above this, mentioning that we overwrite and prune because jdx/mise docker image comes with python installed.
```console
root@75f179a190a1:/mise# # overwrite config to give us a clean state
```

docs/mise-cookbook/nodejs.md
```diff
@@ -0,0 +1,109 @@
...
+
+This example uses `pnpm` as the package manager. This will skip installing dependencies if the lock file hasn't changed.
+
+```toml [mise.toml]
+[tools]
+node = '22'
+
+[hooks]
+post_install = 'corepack enable'
```
syhol, line 88: Personally I think directly installing pnpm with mise is a nicer experience, but we have had a few people ask how to use corepack with mise.

Also corepack won't be built into newer versions of nodejs by default. It's worth thinking a bit more about this one.

syhol (approved):
Some notes but looks good

docs/mise-cookbook/nodejs.md
hverlin: Yeah, I could show both. I like `corepack` more since it's installed the exact version which is in the package.json (`pnpm` will do it in v10, but this avoids installing `pnpm` twice)

I can replace it with `npx corepack enable` to future-proof it :+1:

docs/mise-cookbook/nodejs.md
hverlin: Agree that it will be nicer. I will update to `pnpm-install`

PR #3753, opened 2024-12-21
feat: allow dotnet prerelease

acesyde:
Allow prerelease

### DIFF
--- a/src/backend/dotnet.rs
+++ b/src/backend/dotnet.rs
@@ -31,7 +31,10 @@ impl Backend for DotnetBackend {
             "{}?q={}&packageType=dotnettool&take=1&prerelease={}",
             feed_url,
             &self.tool_name(),
-            SETTINGS.dotnet.allow_prerelease
+            SETTINGS
+                .dotnet
+                .package_flags
+                .contains(&"prerelease".to_string())
         ))?;
 
         if feed.total_hits == 0 {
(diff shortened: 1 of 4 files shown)


src/backend/dotnet.rs
```diff
@@ -28,9 +28,10 @@ impl Backend for DotnetBackend {
         let feed_url = self.get_search_url()?;
 
         let feed: NugetFeedSearch = HTTP_FETCH.json(format!(
-            "{}?q={}&packageType=dotnettool&take=1",
+            "{}?q={}&packageType=dotnettool&take=1&prerelease={}",
```
jdx, line 31: are there other flags we might be able to send? I'm thinking of a setting that might allow more functionality like `MISE_DOTNET_PACKAGE_FLAGS=prerelease` or something

acesyde: There is no other parameters to send to the nuget api, I let you decide what you prefer and I can implement it 👍

acesyde: I will check something and if it's better to implement your solution for the future , do not merge it yet please

jdx: Looking at this again, should this actually be a tool option? Or maybe we support both?

https://mise.jdx.dev/dev-tools/#tool-options

acesyde: Hmm good idea, I think I'll implement it, it's better if it works the same everywhere

acesyde: After investigation the `prerelease` is at the `_list_remote_versions` level not at the `install_version_` so doesn't have access to `tv.request.options()`

PR #3762, opened 2024-12-21
fix: Improve fig spec with better generators

miguelmig:
Also prevents some issues that might trigger the autocomplete's linter.
This, together with the changes on usage-cli `1.7.3` version, should get the automated PR to pass the linter checks on the autocomplete repo


miguelmig:
There's still one nasty eslint complaint about descriptions that I can't get to popup on this repo, still struggling with figuring out how eslint imports rules from other modules..once I figure it out I'll leave a comment here.

docs/.vitepress/config.ts
```diff
@@ -264,7 +264,7 @@ function cliReference(commands: { [key: string]: Command }) {
         x.collapsed = true;
         x.items = Object.keys(command.subcommands)
           .filter(
-            (subcommand) => command.subcommands![subcommand].hide !== true,
+            (subcommand) => command.subcommands![subcommand].hide !== true
```
jdx: I'd prefer if we didn't need to enable es5 style everywhere

miguelmig: Fair point, will investigate if I can use `package.json` hierarchy to handle this

PR #3765, opened 2024-12-21
docs: update IDE integration page

hverlin:
- Add additional explanations + screenshots for shims
- mention that one can use `mise which`/`mise where` as a fallback as well
- Add URL to existing mise plugins


docs/intellij-sdk-selection.png
hverlin, line 1: @jdx Can I include assets in the repo? I have noticed that other assets are stored at `https://github.com/jdx/mise-docs/assets/...`

jdx: yeah

docs/ide-integration.md
```diff
@@ -37,10 +59,22 @@ end
...
 :::
 
-This assumes that `mise` is on PATH. If it is not, you'll need to use the absolute path (
+This assumes that `mise` is on `PATH`. If it is not, you'll need to use the absolute path (
 e.g.: `eval "$($HOME/.local/bin/mise activate zsh)"`).
 
-:::: tip
+Here is an example showing that VSCode will use `node` provided by `mise`:
+
+::: tabs
```
hverlin, line 67: We can add other IDEs here as well. I wanted to add a screenshot for `nvim` but there is actually a lot to configure to get a Node.js debugger running

134130:
Thanks!

docs/ide-integration.md
jdx: I suppose it depends on how people use it, but for me I have no need to configure neovim because I always run it from a project directory with the tools configured and they don't need to change in that session

hverlin: yeah, I meant I wanted a screenshot with the js debugger in nvim, but that's too much to configure. I usually only use it for light configuration edits

PR #3791, opened 2024-12-22
Revert "fix: Use arguments for to pass staged filenames to pre-commit task"

jdx:
Reverts jdx/mise#3492

this is creating a bunch of "mise.pre-commit' files that are annoying to deal with

### DIFF
--- a/docs/cli/generate/git-pre-commit.md
+++ b/docs/cli/generate/git-pre-commit.md
@@ -9,7 +9,7 @@
 This command generates a git pre-commit hook that runs a mise task like `mise run pre-commit`
 when you commit changes to your repository.
 
-Staged files are passed to the task via appended arguments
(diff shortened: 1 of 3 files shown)


joshbode:
I just noticed this revert - I can see what was wrong - the `mktemp` was missing the `-t` option so it was creating the pipe in the current folder (also this option is not available on the macOS (BSD) version of `mktemp`). Though, the trap should have been cleaning up the pipe, regardless...

This is what it should have been to be platform-independent (and correctly in the temp-dir):
```bash
: "${TMPDIR:=/tmp}"
PIPE=$(mktemp -u "${TMPDIR%/}/mise.{task}.XXXXXXXX")
```

I can make a PR again to fix this (and check the trap logic...), unless you want to stick with the `STAGED` approach (which as previously discussed won't work with files that have spaces).

PR #3823, opened 2024-12-26
fix: panic when setting config value

roele:
Rough draft around the discussion #3820. 

This handles the basic use-case and will behave like the following.
```
# set simple value or nested value
mise config set tools.node 20 --> node = "20"
mise config set tools.node.version 20 --> node = { version = "20" }

# transform version to inline table if its already set
mise config set tools.node 20 --> node = "20"
mise config set tools.node.postinstall 'corepack enable' --> node = { version = "20", postinstall = "corepack enable" }
```

### DIFF
--- a/src/cli/config/set.rs
+++ b/src/cli/config/set.rs
@@ -51,14 +51,27 @@ impl ConfigSet {
             let mut config: toml_edit::DocumentMut = std::fs::read_to_string(&file)?.parse()?;
             let mut container = config.as_item_mut();
             let parts = self.key.split('.').collect::<Vec<&str>>();
-            for key in parts.iter().take(parts.len() - 1) {
-                container = container.as_table_mut().unwrap().entry(key).or_insert({
-                    let mut t = toml_edit::Table::new();
-                    t.set_implicit(true);
-                    toml_edit::Item::Table(t)
-                });
-            }
             let last_key = parts.last().unwrap();
+            for (idx, key) in parts.iter().take(parts.len() - 1).enumerate() {
+                container = container
+                    .as_table_like_mut()
+                    .unwrap()
+                    .entry(key)
+                    .or_insert({
+                        let mut t = toml_edit::Table::new();
+                        t.set_implicit(true);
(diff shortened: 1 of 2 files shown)


src/cli/config/set.rs
```diff
@@ -99,7 +112,13 @@ impl ConfigSet {
                 TomlValueTypes::Infer => bail!("Type not found"),
             };
 
-            container.as_table_mut().unwrap().insert(last_key, value);
+            container
+                .as_table_like_mut()
+                .unwrap_or({
+                    let t = toml_edit::Table::new();
+                    toml_edit::Item::Table(t).as_table_like_mut().unwrap()
```
jdx, line 123: I think you have this in another PR but we should set the flag for implicit keys or whatever it's called

jdx:
mind adding an e2e test? it doesn't look like we have any for these commands, but I think it should be pretty simple. I would copy https://github.com/jdx/mise/blob/main/e2e/cli/test_settings_add

roele:
will add a test and look into the additional type inference issue but this might have to wait until tomorrow

jdx:
sounds good, I think what you have in the description is good enough. A pattern I've been doing is this which might be useful here:

https://github.com/jdx/mise/blob/b66f7cac26186b1f935c11e296647baebf1a20f1/e2e/lockfile/test_lockfile_use#L14-L16

it's basically a snapshot test

src/cli/config/set.rs
```diff
@@ -78,7 +91,10 @@ impl ConfigSet {
                             SettingsType::ListString => TomlValueTypes::List,
                             SettingsType::ListPath => TomlValueTypes::List,
                         },
-                        None => TomlValueTypes::String,
+                        None => match self.value.as_str() {
+                            "true" | "false" => TomlValueTypes::Bool,
```
roele, line 95: not sure if we also should treat `0,1,y,n` as `boolean`, would these even map to TOML values? mid-term the whole type inference logic should probably be moved to a package as it is also used by `settings set` for example

PR #3867, opened 2024-12-29
fix: mise settings unset does not seem to work

roele:
Draft for discussion #3865

### DIFF
--- a/src/cli/settings/unset.rs
+++ b/src/cli/settings/unset.rs
@@ -20,24 +20,38 @@ pub struct SettingsUnset {
 
 impl SettingsUnset {
     pub fn run(self) -> Result<()> {
-        let path = if self.local {
-            config::local_toml_config_path()
-        } else {
-            config::global_config_path()
-        };
-        let raw = file::read_to_string(&path)?;
-        let mut config: DocumentMut = raw.parse()?;
-        if !config.contains_key("settings") {
-            return Ok(());
-        }
-        let settings = config["settings"].as_table_mut().unwrap();
-        settings.remove(&self.key);
-
-        // validate
-        let _: SettingsFile = toml::from_str(&config.to_string())?;
-
-        file::write(&path, config.to_string())
+        unset(&self.key, self.local)
     }
+}
+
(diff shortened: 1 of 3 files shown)


src/cli/settings/unset.rs
```diff
@@ -20,26 +20,40 @@ pub struct SettingsUnset {
...
+        config::local_toml_config_path()
+    } else {
+        config::global_config_path()
+    };
+    let raw = file::read_to_string(&path)?;
+    let mut config: DocumentMut = raw.parse()?;
+    if !config.contains_key("settings") {
+        return Ok(());
+    }
+    let mut settings = config["settings"].as_table_mut().unwrap();
```
jdx:
```suggestion
    if let Some(mut settings) = config["settings"].as_table_mut() {
```

that way we can remove the unwrap

src/cli/settings/unset.rs
```diff
@@ -20,26 +20,40 @@ pub struct SettingsUnset {
...
+    if key.contains(".") {
```
jdx:
```suggestion
    if let Some((parent_key, child_key)) = key.split_once(".")
```

src/cli/settings/unset.rs
```diff
@@ -20,26 +20,40 @@ pub struct SettingsUnset {
...
+        let (parent_key, child_key) = key.split_once('.').unwrap();
+
+        key = child_key;
+        settings = settings
+            .entry(parent_key)
+            .or_insert(toml_edit::Item::Table(toml_edit::Table::new()))
```
jdx: do this here:

https://github.com/jdx/mise/blob/0cfdf92bb73f66a317fb83bb347ea7e88fd1c55a/src/cli/config/set.rs#L57

not that I think it's super important, but I'm trying to be consistent about always setting that flag when creating new tables

jdx:
could you also add a line to test this here?

https://github.com/jdx/mise/blob/0cfdf92bb73f66a317fb83bb347ea7e88fd1c55a/e2e/cli/test_settings_set#L13

jdx:
unrelated, I have a message for you in the Discord (since I think you don't check it often)

PR #3919, opened 2025-01-02
feat: Add setting description to mise settings --json-extended output

hverlin:
Add `description` and `type` to `settings --json-extended`. This way, one does not need to pull https://mise.jdx.dev/schema/mise.json (and the description is always accurate regardless of the current mise version)

```sh
❯ cargo run -q -- settings --json-extended
{
  "disable_tools": {
    "value": [
      "php"
    ],
    "type": "array",
    "description": "Tools defined in mise.toml that should be ignored",
    "source": "/home/hverlin/.config/mise/config.toml"
  },
  "experimental": {
    "value": true,
    "type": "boolean",
    "description": "Enable experimental mise features which are incomplete or unstable—breakings changes may occur",
    "source": "/home/hverlin/.config/mise/config.toml"
  }
}
```


hverlin:
Please let me know if this makes sense. I am thinking that I can add the `type` as well

PR #3920, opened 2025-01-02
docs: fix bash completion setup instructions

bestagi:
Fix this error
```
bash: /etc/bash_completion.d/mise: Permission denied
```


uofirob:
Thanks for this!  This worked for me.

docs/installing-mise.md
```diff
@@ -389,7 +389,7 @@ Then, run the following commands to install the completion script for your shell
 ```sh [bash]
 # This requires bash-completion to be installed
 mkdir -p /etc/bash_completion.d/
-mise completion bash --include-bash-completion-lib > /etc/bash_completion.d/mise
+sudo bash -c "mise completion bash --include-bash-completion-lib > /etc/bash_completion.d/mise"
```
malept: We should avoid running commands as the superuser when possible. I'd change this to something like:
```suggestion
mise completion bash --include-bash-completion-lib | sudo tee /etc/bash_completion.d/mise > /dev/null
```

jdx: yes, thank you

PR #3956, opened 2025-01-06
fix: pre-commit hook temporary file

joshbode:
- Create temporary files in correct location (`${TMPDIR}`) and make `mktemp` usage cross-platform
- Fix cleanup of temporary file (`exec` doesn't allow `trap` to be triggered)
- Allow pre-commit for the first commit (i.e. where `HEAD` isn't valid yet)

### DIFF
--- a/docs/cli/generate/git-pre-commit.md
+++ b/docs/cli/generate/git-pre-commit.md
@@ -9,7 +9,7 @@
 This command generates a git pre-commit hook that runs a mise task like `mise run pre-commit`
 when you commit changes to your repository.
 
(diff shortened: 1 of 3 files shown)


jdx:
I think having a 3 line script is just better

joshbode:
That's your choice, however, the current approach has got some issues:

- file names with spaces aren't supported
- the pre-commit hook won't work on the first commit because `HEAD` won't be defined yet
- dealing with the `STAGED` variable is potentially error-prone

jdx:
@joshbode I started working on a sister project for git hooks: https://github.com/jdx/hk

PR #3969, opened 2025-01-06
fix: mise install rust failed on windows

roele:
Draft for discussion #3960.

I currently don't have the setup (yet) to compile and test this properly.

### DIFF
--- a/e2e-win/rust.Tests.ps1
+++ b/e2e-win/rust.Tests.ps1
@@ -1,5 +1,13 @@
 
-Describe 'node' {
+Describe 'rust' {
+    It 'installs rust 1.83.0' {
+        $env:MISE_CARGO_HOME = "%TEMP%\.cargo"
+        $env:MISE_RUSTUP_HOME = "%TEMP%\.rustup"
+        mise x rust@1.83.0 -- rustc -V | Should -BeLike "rustc 1.83.0*"
+        Remove-Item Env:MISE_CARGO_HOME
+        Remove-Item Env:MISE_RUSTUP_HOME
+    }
+
     It 'executes rust 1.82.0' {
         mise x rust@1.82.0 -- rustc -V | Should -BeLike "rustc 1.82.0*"
     }
--- a/src/plugins/core/rust.rs
+++ b/src/plugins/core/rust.rs
@@ -32,7 +32,7 @@ impl RustPlugin {
             return Ok(());
         }
         ctx.pr.set_message("Downloading rustup-init".into());
-        HTTP.download_file("https://sh.rustup.rs", &rustup_path(), Some(&ctx.pr))?;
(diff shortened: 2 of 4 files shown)


src/plugins/core/rust.rs
```diff
@@ -203,6 +203,16 @@ const CARGO_BIN: &str = "cargo";
 #[cfg(windows)]
 const CARGO_BIN: &str = "cargo.exe";
 
+#[cfg(unix)]
+fn rustup_url() -> String {
```
roele, line 207: this is currently a function on purpose as we might need to pass the architecture via `https://win.rustup.rs/[aarch64|x86_64|i686]` for windows

jdx: I don't have it handy but I think there is a MISE_ARCH setting that should also be used here

roele: since `mise` does not support `i686` i guess we can simply drop that or throw an error in this case?

src/plugins/core/rust.rs
```diff
@@ -203,6 +203,21 @@ const CARGO_BIN: &str = "cargo";
...
+    "https://sh.rustup.rs".to_string()
+}
+
+#[cfg(windows)]
+fn rustup_url() -> String {
+    let arch = if TARGET.contains("aarch64") {
+        "aarch64"
+    } else {
+        "x86_64"
+    };
```
jdx:
```suggestion
    let arch = &*SETTINGS.arch;
```

src/plugins/core/rust.rs
jdx: so MISE_ARCH can be used to automatically use the right arch but it has another use-case, people can force the arch to be x86_64 even if they're on aarch64 which is sometimes necessary

e2e-win/rust.Tests.ps1
```diff
@@ -1,5 +1,5 @@
 
-Describe 'node' {
+Describe 'rust' {
     It 'executes rust 1.82.0' {
         mise x rust@1.82.0 -- rustc -V | Should -BeLike "rustc 1.82.0*"
```
jdx, line 12: I wonder if we could get this to actually test the rustup install as well. Maybe if we set [MISE_RUSTUP_HOME and MISE_CARGO](https://mise.jdx.dev/lang/rust.html#rust.cargo_home) to a temp dir?

roele: Wondered about this as well. It seems the windows image comes with Rust pre-installed so we would need to work around that somehow. Setting the env variables sounds like a possible approach...

jdx: I haven't looked thoroughly, but I think that if we just modify the default directories then mise will not attempt to use any existing rustup like the one on the GHA image

PR #3988, opened 2025-01-08
fix(schema): fix task schema

risu729:
Updates inconsistency between the schema and the docs.

Not sure whether the description is good for `outputs.auto`.
`automatically touch an internal tracked file instead of specifying outputs`
https://mise.jdx.dev/tasks/task-configuration.html#outputs

### DIFF
--- a/schema/mise-task.json
+++ b/schema/mise-task.json
@@ -161,12 +161,32 @@
               "type": "boolean"
             },
             "outputs": {
-              "description": "files created by this task",
-              "items": {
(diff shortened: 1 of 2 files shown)


risu729:
It seems the test is failing because of this.
https://github.com/editorconfig-checker/editorconfig-checker/issues/409

roele:
See https://github.com/editorconfig-checker/editorconfig-checker/issues/409

roele:
@risu729 My PR also just failed because of this...

PR #3992, opened 2025-01-08
feat: add kubectx/kubens to registry

roele:
As discussed at #3945


roele:
Seems there is an issue with the dependency `editorconfig-checker` installed via aqua. The [release 3.1.0](https://github.com/editorconfig-checker/editorconfig-checker/releases) does not seem to have any assets (besides source code)

```
failed to install aqua:editorconfig-checker/editorconfig-checker@3.1.0
```

PR #4017, opened 2025-01-09
feat: Add jlpkg via ubi

dhanak:


dhanak:
Valid point.

That being said, could you please add some guideline in `CONTRIBUTING.md`, or in `SECURITY.md` (where it is explicitly requested to make contributions to `registry.toml`) as to what is the approximate lower limit of stars for a tool to be considered notable enough?

jdx:
there's no number

Issue #3890, opened 2024-12-31
`mise` fails to install `kubectl` on Windows from Aqua registry


jdx:
### Discussed in https://github.com/jdx/mise/discussions/3885

<div type='discussions-op-text'>

<sup>Originally posted by **dangmai** December 30, 2024</sup>
Using this `mise.toml`:

```toml
[tools]
kubectl = '1.32.0'
```

`mise` will fail to install `kubectl` on Windows with the following message:

```
mise ERROR failed to install aqua:kubernetes/kubectl@1.32.0
mise ERROR HTTP status client error (404 Not Found) for url (https://cdn.dl.k8s.io/1.32.0/bin/windows/amd64/kubectl)        tl)                                                                                                                4/kubectl)   
mise ERROR HTTP status client error (404 Not Found) for url (https://cdn.dl.k8s.io/release/v1.32.0/bin/windows/amd6d64/kubectl)
mise ERROR Run with --verbose or MISE_VERBOSE=1 for more information
```

This is because `aqua` automatically adds the `.exe` extension on Windows platform, while `mise` does not. There was a PR on Aqua's repository to make the `.exe` extension explicit in the URL, but the maintainer has closed it: https://github.com/aquaproj/aqua-registry/pull/30398 (with a follow up note for `mise` users [here](https://github.com/aquaproj/aqua-registry/issues/30430)).

I think `mise` logic should follow `aqua` logic when determining the URL to use while downloading dependencies.</div>

jdx:
it'll be awhile until I get back home where my windows laptop is to be able to do this.

I believe the logic would need to be added somewhere around here though: https://github.com/jdx/mise/blob/cc88dca50e8e0dac94dbb83d0ce1ebcfc38a1ec4/src/backend/aqua.rs#L203

roele:
I think the `.exe` has probably to be added around here

https://github.com/jdx/mise/blob/5ca603583166cc23543a0c4dce7727b308f663f5/src/backend/aqua.rs#L535

PR #4024, opened 2025-01-09
fix: mise fails to install kubectl on windows from aqua registry

roele:
Another follow up for #3890

It seems for `http` package types with a `raw` format we might need to add the `.exe` extension in some cases (e.g kubernetes/kubectl). Correct approach would be to fix the registry and provide an override for the windows OS with proper url but since aqua has fixed this in code we might have to follow this approach i assume.

TODO: needs to be tested on a Windows machine


jdx:
I tested on my windows machine and it works

PR #4034, opened 2025-01-10
feat: update JSON output for task info/ls

hverlin:
Add additional properties to the JSON output of `mise tasks info <taskname> --json` and `mise tasks ls --json`. The main difference between the two is that `mise tasks info` is going to parse the `usage` specs.

I am not sure what's the expected serialization output for task with args. The default is that it returns an array like this:
```json
"depends" : [
  ["build", "-v" ],
 "other_task"
]
```

 but we could instead always return an object such as:
```json
"depends" : [
  { "name": "build", "args": ["-v"] },
  { "name": "other_task" }
]
```


hverlin:
test failure seems unrelated

src/cli/tasks/ls.rs
```diff
@@ -133,20 +133,27 @@ impl TasksLs {
...
-                if task.hide {
-                    inner.insert("hide".to_string(), task.hide.into());
-                }
-                inner.insert("description".to_string(), task.description.into());
-                inner.insert(
-                    "source".to_string(),
-                    task.config_source.to_string_lossy().into(),
-                );
-                inner
+                json!({
```
jdx, line 136: nice use of the macro

PR #4038, opened 2025-01-10
docs: set prose-wrap with prettier

jdx:


henrebotha:
I think something broke? This page is now this weird recursive layout.

mise jdx dev_dev-tools_

PR #4048, opened 2025-01-10
fix: mise does not operate well under Git Bash on Windows

### LINKED ISSUE #4011: mise does not operate well under GitBash on Windows
### Discussed in https://github.com/jdx/mise/discussions/3961

<sup>Originally posted by **zartc** January  6, 2025</sup>
As a user of GitBash (which comes with Git for Windows), I installed the 2025.1.0 Windows-x64 (ef6936a 2025-01-01). 
The program works fine, but it produces paths that are incompatible with GitBash. Any command that updates the PATH variable (including `eval $(mise activate bash)`) corrupts the GitBash terminal session, rendering it unusable because GitBash cannot find any commands anymore due to the corrupted PATH variable.

### The problem:
_mise_ outputs paths in Windows format: `C:\Program Files\Git\mingw64\bin;C:\Program Files\Git\usr\local\bin;C:\Program Files\Git\usr\bin` using the `C:` prefix, the `\` directory separator, and the `;` path separator. However, GitBash expects paths to be in Linux format: using `/c/` prefix, `/` directory separator, and `:` path separator.

### One element of solution:
When running under GitBash ($OSTYPE == "msys"), paths generated by _mise_ should be in Linux format using the `/` and `:` separators. In addition, paths pointing to directories inside the GitBash installation directory should be without the `C:\Program Files\Git` prefix, so that they are seen as absolute paths **inside** the Linux VFS. For example, `C:\Program Files\Git\mingw64\bin` => `/mingw64/bin`, `C:\Program Files\Git\usr\bin` => `/usr/bin`.
</div>

roele:
Fixes #4011 

Gave this a try but detection of `msys` or `cygwin` does not work via `$OSTYPE` as this is not available via `env` so `env::var()` does not return anything. Not sure how reliable my approach via `MSYSTEM` and `PWD` is though.

Another unresolved issue seems to be related to line breaks on cygwin. The `.bashrc` file does not source successfully until its opened and saved (without changes). Probably would require a `dos2unix` on cygwin, not sure why that is not an issue on msys.
```bash
$ source .bashrc
': not a valid identifiernctions
-bash: .bashrc: line 35: syntax error near unexpected token `$'\r''
'bash: .bashrc: line 35: `function __zsh_like_cd() {
```

### DIFF
--- a/src/shims.rs
+++ b/src/shims.rs
@@ -129,6 +129,22 @@ pub fn reshim(ts: &Toolset, force: bool) -> Result<()> {
 #[cfg(windows)]
 fn add_shim(mise_bin: &Path, symlink_path: &Path, shim: &str) -> Result<()> {
     let shim = shim.trim_end_matches(".cmd");
+    // write a shim file without extension for use in Git Bash/Cygwin
+    file::write(
+        symlink_path.with_extension(""),
+        formatdoc! {r#"
+        #!/bin/bash
+
+        exec mise x -- {shim} "$@"
+        "#},
+    )
+    .wrap_err_with(|| {
+        eyre!(
+            "Failed to create symlink from {} to {}",
+            display_path(mise_bin),
+            display_path(symlink_path)
+        )
+    })?;
     file::write(
         symlink_path.with_extension("cmd"),
         formatdoc! {r#"
@@ -180,7 +196,7 @@ pub fn get_shim_diffs(
 fn get_actual_shims(mise_bin: impl AsRef<Path>) -> Result<HashSet<String>> {
     let mise_bin = mise_bin.as_ref();
 
-    Ok(list_executables_in_dir(&dirs::SHIMS)?
+    Ok(list_shims()?
         .into_par_iter()
         .filter(|bin| {
             let path = dirs::SHIMS.join(bin);
@@ -211,6 +227,27 @@ fn list_executables_in_dir(dir: &Path) -> Result<HashSet<String>> {
         .collect())
 }
 
+fn list_shims() -> Result<HashSet<String>> {
+    Ok(dirs::SHIMS
+        .read_dir()?
+        .par_bridge()
+        .map(|bin| {
+            let bin = bin?;
+            // files and symlinks which are executable or extensionless files (Git Bash/Cygwin)
+            if (file::is_executable(&bin.path()) || bin.path().extension().is_none())
+                && (bin.file_type()?.is_file() || bin.file_type()?.is_symlink())
+            {
+                Ok(Some(bin.file_name().into_string().unwrap()))
+            } else {
+                Ok(None)
+            }
+        })
+        .collect::<Result<Vec<_>>>()?
+        .into_iter()
+        .flatten()
+        .collect())
+}
+
 fn get_desired_shims(toolset: &Toolset) -> Result<HashSet<String>> {
     Ok(toolset
         .list_installed_versions()?
@@ -222,9 +259,12 @@ fn get_desired_shims(toolset: &Toolset) -> Result<HashSet<String>> {
             });
             if cfg!(windows) {
                 bins.into_iter()
-                    .map(|b| {
+                    .flat_map(|b| {
                         let p = PathBuf::from(&b);
-                        p.with_extension("cmd").to_string_lossy().to_string()
+                        vec![
+                            p.with_extension("").to_string_lossy().to_string(),
+                            p.with_extension("cmd").to_string_lossy().to_string(),
+                        ]
                     })
                     .collect()
             } else {


src/assets/bash_zsh_support/chpwd/function.sh
```diff
@@ -1,5 +1,5 @@
 # shellcheck shell=bash
-export -a chpwd_functions
+declare -a chpwd_functions
```
roele: there is no `-a` flag for export i assume this is supposed to be `declare`?

jdx: I think it is necessary, I got it from here https://github.com/rvm/rvm/tree/master/scripts/extras/bash_zsh_support/chpwd

jdx:
for what it's worth, I'm not sure git bash is a necessary use-case for us. There are many alternatives like shims, powershell, and wsl for windows users.

roele:
I'm generally not a Windows user but I agree that shims and powershell seem to work well enough. Windows is a PITA anyways, in my last project Git Bash was my only option (was not allowed to use WSL or install anything else). While probably rare there might be some users restrained to such a setup.

jdx:
right but even then shims will work. What won't work is you won't see env vars unless you wrap things in tasks but as I'm sure you've heard me explain, that's kind of how I think projects should be setup anyways since it doesn't rely on shell extensions.

jdx:
isn't that already happening?

jdx:
is this something that can be configured with `PATHEXT`? or is there a git bash equivalent for that?

roele:
PATHEXT is a Windows environment variable and i am not aware of an equivalent in Git Bash/Cygwin. Emulating PATHEXT in Bash does also not seem to work.
```bash
export PATHEXT=.com:.exe:.bat:.cmd

if declare -f command_not_found_handle >/dev/null; then 
    eval "original_command_not_found_handle() $(declare -f command_not_found_handle|tail -n +2)"
fi
command_not_found_handle(){
    local PATHEXT_EXPANDED i
    IFS=: read -a PATHEXT_EXPANDED<<<"$PATHEXT"
    for i in "${PATHEXT_EXPANDED[@]}"; do
        if type "$1$i" &>/dev/null; then
            "$1$i" "${@:2}"
            return $?
        fi
    done
    if declare -f original_command_not_found_handle >/dev/null; then
        original_command_not_found_handle "$@"
    else
        return 127
    fi
}
```

While this PR makes `activate` work for Git Bash and Cygwin detection seems a bit hacky. I wonder if we should drop this for now. Having at least the shims work would be great though.

roele:
The following hook might be a workaround for the shims `.cmd` ending issue. We might simply document this as possible workaround in the troubleshooting docs "Windows problems" section.

```
command_not_found_handle()
{
   cmd=$1
   shift
   args=( "$@" )

   IFS=:
   for dir in $PATH; do
      for executable in $dir/$cmd.bat $dir/$cmd.cmd $dir/$cmd.exe; do
         if [ -f "$executable" ]; then
            "$executable" "${args[@]}"
            return
         fi
      done
   done

   orig_command_not_found_handle "$cmd" "${args[@]}"
} 
if declare -f command_not_found_handle >/dev/null; then 
    eval "orig_command_not_found_handle() $(declare -f command_not_found_handle|tail -n +2)"
fi
```

ffaen:
One possible fix for this is to add two shims on windows, one for cmd and one for git bash:

Tested with two shims for node

`node.cmd` (unchanged)
```
@echo off
setlocal
mise x -- node %*
```

`node` (no extension)
```
#!/usr/bin/env bash

mise x -- node "$@"
```

This works totally fine on Git Bash, Cmd, and Powershell (as cmd and powershell will ignore the file with no ext)

Unsure if this works well with the current implementation of [`reshim()`](https://github.com/jdx/mise/blob/main/src/shims.rs#L86) as at first glance it expects a one-to-one shim to app? (My primary language isn't rust so happy to be wrong!)

roele:
@ffaen Was also thinking  about something in that direction which should be doable.

src/shims.rs
```diff
@@ -190,14 +206,15 @@ fn get_actual_shims(mise_bin: impl AsRef<Path>) -> Result<HashSet<String>> {
         .collect::<HashSet<_>>())
 }
 
-fn list_executables_in_dir(dir: &Path) -> Result<HashSet<String>> {
+fn list_executables_in_dir(dir: &Path, include_extensionless: bool) -> Result<HashSet<String>> {
```
roele: not quite sure i like this flag but the alternative would be an almost duplicate method for listing executables in the shims dir

src/shims.rs
```diff
@@ -129,6 +129,22 @@ pub fn reshim(ts: &Toolset, force: bool) -> Result<()> {
 #[cfg(windows)]
 fn add_shim(mise_bin: &Path, symlink_path: &Path, shim: &str) -> Result<()> {
     let shim = shim.trim_end_matches(".cmd");
+    // write a shim file without extension for use in Git Bash/Cygwin
+    file::write(
+        symlink_path.with_extension(""),
+        formatdoc! {r#"
+        #!/usr/bin/bash
```
jdx: I'm not sure we can rely on this _always_ being there like `/bin/bash` would be

src/shims.rs
```diff
@@ -129,6 +129,22 @@ pub fn reshim(ts: &Toolset, force: bool) -> Result<()> {
...
+
+        mise x -- {shim} "$@"
```
jdx: can we use `exec` in git bash?

src/shims.rs
roele: good point!

roele: good point!

src/shims.rs
roele: you mean instead of the alias `x`? (if so we might also wanna use exec in the cmd files)

roele: you mean instead of the alias `x`? (if so we might also wanna use exec in the cmd files)

jdx: no I mean `exec mise x -- {shim} "$@"` so that things like ctrl-c pass to the actual tool and not mise

jdx: no I mean `exec mise x -- {shim} "$@"` so that things like ctrl-c pass to the actual tool and not mise

roele: mhhh... ~for me `exec mise x -- java -version` crashes my terminal emulator tab~~
seems to work fine in context of a bash script though

src/shims.rs
```diff
@@ -222,9 +259,12 @@ fn get_desired_shims(toolset: &Toolset) -> Result<HashSet<String>> {
             });
             if cfg!(windows) {
                 bins.into_iter()
-                    .map(|b| {
+                    .flat_map(|b| {
                         let p = PathBuf::from(&b);
-                        p.with_extension("cmd").to_string_lossy().to_string()
+                        vec![
+                            p.with_extension("").to_string_lossy().to_string(),
```
jdx, line 265: I bet we have a non-zero number of tools with bin names that have "." in them, this will probably break them—though I think that was the case before.

this is fine for now I think, but worth keeping in mind

zartc:
@jdx 
> is this something that can be configured with `PATHEXT`? or is there a git bash equivalent for that?

Git Bash/Cygwin only use the `$PATH` variable to find commands. The only thing to be aware of is that the $PATH variable must be in Unix format, not in Windows format, as mise is currently producing. To convert a path to Unix format one can use the `cygpath -u` command.
IMHO, mise-v2025.4.12-windows-x64.exe assumes that it will run exclusively in a pure Windows environment (PowerShell or Cmd) but this is not always true, it might also run under a cygwing environment on a Windows machine and it should check that.

PR #4059, opened 2025-01-11
feat: added `mise cfg --tracked-configs`

jdx:


jdx:
@pepicrft we will be able to use this to list all of the projects

PR #4068, opened 2025-01-12
feat: migrate eza to ubi

sassdavid:


jdx:
eza doesn't offer macos binaries

PR #4089, opened 2025-01-13
docs: warn that mise is not a dependency manager

syhol:
## Summary

- A response to https://github.com/jdx/mise/discussions/4043
- An extension of https://github.com/jdx/mise/commit/42dcb3bc5a6547d3d148c391ceccfd9228e34669
- Renaming `package` -> `tool`
- Adding warnings

### Warnings

Perhaps these warnings are too loud, we could either move them somewhere else in the page or convert them to normal paragraphs. But for now this is how they look:

image

image

image

image

image

image

### DIFF
--- a/docs/dev-tools/backends/cargo.md
+++ b/docs/dev-tools/backends/cargo.md
@@ -1,5 +1,10 @@
 # Cargo Backend
 
+::: warning
+Mise is not a replacement for a project dependency manager. Please continue to use `cargo`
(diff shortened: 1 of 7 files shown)


jdx:
I think the changes I put into pipx are plenty here. I don't won't to admonish this too much because mise _does_ manage dependencies with default packages, so we can't really say it's not a package manager

syhol:
No problem, I'll close this

PR #4094, opened 2025-01-14
fix: Fixes infinite loop in auto install not found bash function

bnorick:
This infinite loop bug is triggered when `mise` itself becomes not found, e.g., by deleting the directory `mise` is in.

Here's some output of the shell with `set -x` on when doing `rm -rf` on the directory (`workspace/`) containing `mise` in some subpath:
```
+ rm -rf workspace/
++ _mise_hook
++ local previous_exit_status=0
+++ mise hook-env -s bash
+++ local command
+++ command=hook-env
+++ '[' 3 = 0 ']'
+++ shift
+++ case "$command" in
+++ command mise hook-env -s bash
+++ mise hook-not-found -s bash -- mise
+++ local command
+++ command=hook-not-found
+++ '[' 5 = 0 ']'
+++ shift
+++ case "$command" in
+++ command mise hook-not-found -s bash -- mise
+++ mise hook-not-found -s bash -- mise
+++ local command
+++ command=hook-not-found
+++ '[' 5 = 0 ']'
+++ shift
+++ case "$command" in
+++ command mise hook-not-found -s bash -- mise
+++ mise hook-not-found -s bash -- mise
+++ local command
+++ command=hook-not-found
+++ '[' 5 = 0 ']'
+++ shift
+++ case "$command" in
+++ command mise hook-not-found -s bash -- mise
+++ mise hook-not-found -s bash -- mise
+++ local command
+++ command=hook-not-found
+++ '[' 5 = 0 ']'
+++ shift
+++ case "$command" in
+++ command mise hook-not-found -s bash -- mise
+++ mise hook-not-found -s bash -- mise
```

### DIFF
--- a/src/shell/bash.rs
+++ b/src/shell/bash.rs
@@ -71,7 +71,7 @@ impl Shell for Bash {
                 fi
 
                 command_not_found_handle() {{
-                    if [ "$1" != "mise" ] && {exe} hook-not-found -s bash -- "$1"; then
+                    if [[ "$1" != "mise" && "$1" != "mise-"* ]] && {exe} hook-not-found -s bash -- "$1"; then
                       _mise_hook
                       "$@"
                     elif [ -n "$(declare -f _command_not_found_handle)" ]; then
--- a/src/shell/bash.rs
+++ b/src/shell/bash.rs
@@ -56,7 +56,7 @@ impl Shell for Bash {
             {chpwd_load}
             chpwd_functions+=(_mise_hook)
             _mise_hook
-            "#, 
+            "#,
             chpwd_functions = include_str!("../assets/bash_zsh_support/chpwd/function.sh"),
             chpwd_load = include_str!("../assets/bash_zsh_support/chpwd/load.sh")
             });
@@ -71,7 +71,7 @@ impl Shell for Bash {
                 fi
 
                 command_not_found_handle() {{
-                    if {exe} hook-not-found -s bash -- "$1"; then
+                    if [ "$1" != "mise" ] && {exe} hook-not-found -s bash -- "$1"; then
                       _mise_hook
                       "$@"
                     elif [ -n "$(declare -f _command_not_found_handle)" ]; then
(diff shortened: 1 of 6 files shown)


bnorick:
Tests failed because I didn't update the test snap, I did that now. However, this made me realize that this'll need fixing for the other shells too, which I am not proficient in.

bnorick:
The remaining failure, related to `docker-compose` seems unrelated to this? Not sure what is going on there...

```
 mise docker-compose@2.32.3  install
  mise docker-compose@2.32.3  download docker-compose-linux-x86_64
  mise docker-compose@2.32.3  checksum docker-compose-linux-x86_64
  mise docker-compose@2.32.3  extract docker-compose-linux-x86_64
  mise docker-compose@2.32.3 ✓ installed
  mise $ /home/runner/.local/share/mise/installs/docker-compose/2.32.3/docker-cli-plugin-docker-compose --version
  Docker Compose version 083f676
  mise ERROR docker-compose: 
     0: expected output not found: Docker Compose version v2.32.3, got: Docker Compose version 083f676
 ```

File: src/shell/bash.rs [hunk 1]
```diff
@@ -71,7 +71,7 @@ impl Shell for Bash {
                 fi
 
                 command_not_found_handle() {{
-                    if {exe} hook-not-found -s bash -- "$1"; then
+                    if [ "$1" != "mise" ] && {exe} hook-not-found -s bash -- "$1"; then
```
jdx: this should match this logic https://github.com/jdx/mise/blob/801a880be383dbdd1502a857b7281b431489a1e5/src/cli/version.rs#L95 since "mise-foo" is a valid mise name

jdx:
I'd like at least support for zsh (which should be close to bash), fish would also be quite nice but the others I'm not so concerned about, I don't even think most of them have not_found handlers

bnorick:
I didn't check whether the `zsh` or `fish` implementations work and they were guided by an LLM. Any help here would be appreciated.

PR #4100, opened 2025-01-14
fix: use vars in [env] templates

hverlin:
https://github.com/jdx/mise/discussions/4099

### DIFF
--- a/e2e/env/test_env_template
+++ b/e2e/env/test_env_template
@@ -4,20 +4,20 @@ cat <<EOF >mise.toml
 [env]
 ENV_TMPL_EXEC = "{{ exec(command='echo foo') }}"
 EOF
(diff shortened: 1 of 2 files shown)


e2e/env/test_env_template
```diff
@@ -4,20 +4,20 @@ cat <<EOF >mise.toml
 [env]
 ENV_TMPL_EXEC = "{{ exec(command='echo foo') }}"
 EOF
-assert "mise env -s zsh | grep ENV_TMPL_EXEC" "export ENV_TMPL_EXEC=foo"
```
hverlin, line 7: I can revert, but it fails locally otherwise when I run:
```
mise run test:e2e -- test_env_template
```
(even on main)

PR #4101, opened 2025-01-14
feat: allow to specify Rust profile

### LINKED ISSUE #4085: Allow to specify Rust profile
### Discussed in https://github.com/jdx/mise/discussions/4083

<sup>Originally posted by **Birdie0** January 13, 2025</sup>
When Rust is being installed with mise it's not possible to specify [profile](https://rust-lang.github.io/rustup/concepts/profiles.html) which would be quite useful in cases when you need just compiler, allowing to save some space and time.
</div>

roele:
Fixes #4085

### DIFF
--- a/src/cmd.rs
+++ b/src/cmd.rs
@@ -248,6 +248,13 @@ impl<'a> CmdLineRunner<'a> {
         None
     }
 
+    pub fn opt_arg<S: AsRef<OsStr>>(mut self, arg: Option<S>) -> Self {
+        if let Some(arg) = arg {
+            self.cmd.arg(arg);
+        }
+        self
+    }
+
     pub fn arg<S: AsRef<OsStr>>(mut self, arg: S) -> Self {
         self.cmd.arg(arg.as_ref());
         self
--- a/src/plugins/core/rust.rs
+++ b/src/plugins/core/rust.rs
@@ -108,7 +108,7 @@ impl Backend for RustPlugin {
             .with_pr(&ctx.pr)
             .arg("toolchain")
             .arg("install")
-            .opt_arg(profile.as_ref().and_then(|_| Some("--profile")))
+            .opt_arg(profile.as_ref().map(|_| "--profile"))
             .opt_arg(profile)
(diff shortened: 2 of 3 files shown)


src/plugins/core/rust.rs
```diff
@@ -102,6 +110,8 @@ impl Backend for RustPlugin {
             .with_pr(&ctx.pr)
             .arg("toolchain")
             .arg("install")
+            .arg("--profile")
+            .arg(self.tv_profile(&tv))
```
jdx: I think if the user doesn't specify we should just not specify this arg at all. I haven't looked it up but I suspect there may be env vars and things that would also set this we should respect but specifying the arg may override

roele: valid concern, i assume the option will override the default set via `rustup set profile minimal`. could not find anything about environment variables though.

https://rust-lang.github.io/rustup/concepts/profiles.html

src/cmd.rs
```diff
@@ -248,6 +248,13 @@ impl<'a> CmdLineRunner<'a> {
         None
     }
 
+    pub fn opt_arg<S: AsRef<OsStr>>(mut self, arg: Option<S>) -> Self {
```
jdx, line 251: clever!

src/plugins/core/rust.rs
```diff
@@ -98,10 +102,17 @@ impl Backend for RustPlugin {
...
+        let profile = self.tv_profile(&tv);
+        let profile_option = match profile {
+            Some(_) => Some("--profile"),
+            None => None,
+        };
         CmdLineRunner::new(RUSTUP_BIN)
             .with_pr(&ctx.pr)
             .arg("toolchain")
             .arg("install")
+            .opt_arg(profile_option)
```
jdx:
```suggestion
            .opt_arg(profile.and_then(|_| "--profile")
```

jdx: might need an `.as_ref()` for this to work without moving `profile`

PR #4106, opened 2025-01-14
feat(registry): add aqua and cmdx

scop:
Makes it easier to contribute to aqua registry.

### DIFF
--- a/registry.toml
+++ b/registry.toml
@@ -83,6 +83,8 @@ apollo-router.backends = [
     "asdf:safx/asdf-apollo-router"
 ]
 apollo-rover.backends = ["ubi:apollographql/rover"]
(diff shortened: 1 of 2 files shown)


registry.toml
```diff
@@ -368,6 +370,8 @@ cmake.backends = [
     "vfox:mise-plugins/vfox-cmake"
 ]
 cmctl.backends = ["aqua:cert-manager/cmctl", "asdf:asdf-community/asdf-cmctl"]
+cmdx.backends = ["ubi:suzuki-shunsuke/cmdx"]
```
scop, line 373: Is available in aqua registry but can't use that here due to https://github.com/jdx/mise/discussions/4092

PR #4108, opened 2025-01-14
fix: incorrect config_root for project/.mise/config.toml

### LINKED ISSUE #4041: Incorrect config_root for project/.mise/config.toml
### Discussed in https://github.com/jdx/mise/discussions/4039

<sup>Originally posted by **Ernest0-Production** January 10, 2025</sup>
Based on the documentation https://mise.jdx.dev/templates.html#variables

> config_root: PathBuf – Locates the directory containing your mise.toml file, or in the case of something like ~/src/myproj/.config/mise.toml, it will point to ~/src/myproj.

I would expect that in the case of `~/src/myproj/.mise.config.toml`, the value from template variable `{{ config_root }}` would return me `~/src/myproj`, but it returns `~/src/myproj/.mise/`

**mise version:** 
2024.10.13 macos-arm64 (ee83488 2024-10-28</div>

roele:
Fixes #4041

### DIFF
--- a/src/config/config_file/mise_toml.rs
+++ b/src/config/config_file/mise_toml.rs
@@ -29,6 +29,8 @@ use crate::toolset::{ToolRequest, ToolRequestSet, ToolSource, ToolVersionOptions
 use crate::watch_files::WatchFile;
 use crate::{dirs, file};
 
+use super::config_root;
+
 #[derive(Default, Deserialize)]
 pub struct MiseToml {
     #[serde(rename = "_")]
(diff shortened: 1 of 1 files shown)


src/config/config_file/mise_toml.rs
```diff
@@ -514,6 +514,30 @@ impl ConfigFile for MiseToml {
...
+/// If the config directory is nested, this will return the top-most directory.
+/// For example
+/// ```text
+/// /src/myproj/mise.toml                   -> /src/myproj
+/// /src/myproj/mise/config.toml            -> /src/myproj
+/// /src/myproj/.mise/config.toml           -> /src/myproj
+/// /src/myproj/.config/mise/config.toml    -> /src/myproj
+/// /src/myproj/.config/mise/conf.d/a.toml  -> /src/myproj
+/// ```
+fn get_config_root(path: &Path) -> PathBuf {
```
jdx: I think we should instead use this https://github.com/jdx/mise/blob/1b0c5d716908a3a3ff7d677b3fdbb3f678dd0637/src/config/config_file/mod.rs#L246

PR #4138, opened 2025-01-17
feat(registry): add slangroom-exec

matteo-cristino:


jdx:
not notable enough, only has 1 star

PR #4144, opened 2025-01-17
fix: elixir installation failed

### LINKED ISSUE #4114: Elixir installation failed
### Discussed in https://github.com/jdx/mise/discussions/4103

<sup>Originally posted by **alec-c4** January 14, 2025</sup>
```sh
$ mise use -g erlang elixir --verbose                                                                                                                                                                                       
DEBUG ARGS: /opt/homebrew/bin/mise use -g erlang elixir --verbose
DEBUG EnvResults
DEBUG installing 1 leaf tools first
DEBUG install_some_versions: erlang@latest
DEBUG install_some_versions: elixir@latest
INFO  elixir@1.18.1-otp-27 install
INFO  elixir@1.18.1-otp-27 download v1.18.1-otp-27.zip
DEBUG GET Downloading https://builds.hex.pm/builds/elixir/v1.18.1-otp-27.zip to ~/.local/share/mise/downloads/elixir/1.18.1-otp-27/v1.18.1-otp-27.zip
DEBUG GET https://builds.hex.pm/builds/elixir/v1.18.1-otp-27.zip
DEBUG starting new connection: https://builds.hex.pm/
DEBUG GET https://builds.hex.pm/builds/elixir/v1.18.1-otp-27.zip 200 OK
INFO  elixir@1.18.1-otp-27 extract v1.18.1-otp-27.zip
DEBUG unzip /Users/alec/.local/share/mise/downloads/elixir/1.18.1-otp-27/v1.18.1-otp-27.zip -d /Users/alec/.local/share/mise/installs/elixir/1.18.1-otp-27
INFO  elixir@1.18.1-otp-27 elixir --version
DEBUG EnvResults
DEBUG EnvResults
DEBUG $ ~/.local/share/mise/installs/elixir/1.18.1-otp-27/bin/elixir --version
/Users/alec/.local/share/mise/installs/elixir/1.18.1-otp-27/bin/elixir: line 243: exec: erl: not found
ERROR ~/.local/share/mise/installs/elixir/1.18.1-otp-27/bin/elixir failed
Error:
   0: failed to install core:elixir@1.18.1-otp-27
   1: ~/.local/share/mise/installs/elixir/1.18.1-otp-27/bin/elixir exited with non-zero status: exit code 127

Location:
   src/cmd.rs:462

Version:
   2025.1.6 macos-arm64 (2025-01-12)

Run with RUST_BACKTRACE=full to include source snippets.```</div>

### PR DESCRIPTION
**Pull request by roele**
Fixes #4114 

The `dependency_toolset` function should not rely on dependencies in the Config to build the environment for the CmdLineRunner, especially when used with `install`.

Since this seems specific to Elixir so far and the method is used in other contexts, its probably best to keep the solution exclusive to elixir. If necessary we can move it up to the trait later. WDYT?

### DIFF
--- a/src/cli/use.rs
+++ b/src/cli/use.rs
@@ -101,6 +101,10 @@ impl Use {
         if self.tool.is_empty() && self.remove.is_empty() {
             self.tool = vec![self.tool_selector()?];
         }
+        crate::env::TOOL_ARGS
+            .write()
+            .unwrap()
+            .clone_from(&self.tool);
         let config = Config::try_get()?;
         let mut ts = ToolsetBuilder::new()
             .with_global_only(self.global)
--- a/src/toolset/tool_request_set.rs
+++ b/src/toolset/tool_request_set.rs
@@ -227,6 +227,22 @@ impl ToolRequestSetBuilder {
             }
             merge(trs, arg_ts);
         }
+
+        let tool_args = env::TOOL_ARGS.read().unwrap();
+        let mut arg_trs = ToolRequestSet::new();
+        for arg in tool_args.iter() {
+            if let Some(tvr) = &arg.tvr {
+                arg_trs.add_version(tvr.clone(), &ToolSource::Argument);
+            } else {
+                if !trs.tools.contains_key(&arg.ba) {
+                    // no active version, so use "latest"
+                    let tr = ToolRequest::new(arg.ba.clone(), "latest", ToolSource::Argument)?;
+                    arg_trs.add_version(tr, &ToolSource::Argument);
+                }
+            }
+        }
+        merge(trs, arg_trs);
+
         Ok(())
     }
 }
--- a/src/toolset/tool_request_set.rs
+++ b/src/toolset/tool_request_set.rs
@@ -233,12 +233,10 @@ impl ToolRequestSetBuilder {
         for arg in tool_args.iter() {
             if let Some(tvr) = &arg.tvr {
                 arg_trs.add_version(tvr.clone(), &ToolSource::Argument);
-            } else {
-                if !trs.tools.contains_key(&arg.ba) {
-                    // no active version, so use "latest"
-                    let tr = ToolRequest::new(arg.ba.clone(), "latest", ToolSource::Argument)?;
-                    arg_trs.add_version(tr, &ToolSource::Argument);
-                }
+            } else if !trs.tools.contains_key(&arg.ba) {
+                // no active version, so use "latest"
+                let tr = ToolRequest::new(arg.ba.clone(), "latest", ToolSource::Argument)?;
+                arg_trs.add_version(tr, &ToolSource::Argument);
             }
         }
         merge(trs, arg_trs);
--- a/src/cli/install.rs
+++ b/src/cli/install.rs
@@ -49,7 +49,10 @@ impl Install {
     pub fn run(self) -> Result<()> {
         let config = Config::try_get()?;
         match &self.tool {
-            Some(runtime) => self.install_runtimes(&config, runtime)?,
+            Some(runtime) => {
+                crate::env::TOOL_ARGS.write().unwrap().clone_from(&runtime);
+                self.install_runtimes(&config, runtime)?
+            }
             None => self.install_missing_runtimes(&config)?,
         };
         Ok(())
--- a/src/cli/install.rs
+++ b/src/cli/install.rs
@@ -50,7 +50,7 @@ impl Install {
         let config = Config::try_get()?;
         match &self.tool {
             Some(runtime) => {
-                crate::env::TOOL_ARGS.write().unwrap().clone_from(&runtime);
+                crate::env::TOOL_ARGS.write().unwrap().clone_from(runtime);
                 self.install_runtimes(&config, runtime)?
             }
             None => self.install_missing_runtimes(&config)?,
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by jdx** [comment]
doesn't this effectively make it so we'd always be using `latest` for dependencies and not the version the user configured?

**Comment by roele** [comment]
Yes that's ugly but get_dependencies only returns strings (eg "erlang") and with nothing in the config its the only option.

Your're right, the current implementation seems correct but does not take into account that dependencies might not be in the config (unless i'm missing something). An `install` does not add them and in case of `use`, which also does not seem to work, it might be that the config does not yet reflect the latest state during installation of multiple tools.

Maybe we need something more sophisticated, that tries the config first and falls back to `latest` versions.

**Comment by jdx** [comment]
> Your're right, the current implementation seems correct but does not take into account that dependencies might not be in the config (unless i'm missing something). An install does not add them and in case of use, which also does not seem to work, it might be that the config does not yet reflect the latest state during installation of multiple tools.

it sounds like you're misunderstanding the correct behavior. It should be possible for a user to install erlang with homebrew and elixir on top of that without also using mise to manage erlang.

**Comment by roele** [comment]
Ah I see... that complicates things. If `erlang` is installed via brew it will already be on the PATH so this issue will not surface. 

The culprit is the following which is passed to the `verify` function to check the version of elixir and should put erlang on the PATH.
https://github.com/jdx/mise/blob/1b5e7406ad969e8fc461d088beaf7886111bfbc3/src/backend/mod.rs#L601C1-L604C6

So i guess we need to check if the dependencies are part of the ToolSet (in InstallContext?) and add them, otherwise assume they're installed already (via brew or anyhow)?

**Comment by jdx** [comment]
yes, I think you're correct. The issue is that we always load the config from the config files and we're not including something that gets passed through command line arguments. Threading this through might be challenging.

**Comment by roele** [comment]
Played around a bit more with this but it seems a bit tricky and ugly if you ask me.

The `Config` is immutably shared across threads, therefore installed tools are not reflected and `use` will also not have `erlang` exposed in the environment.

Passing down the tools from the arguments can be done relatively easily via `InstallContext` but requires additional methods `dependency_env_with_ctx` and `dependency_toolset_with_ctx` as the context is not always available and the methods are used in quite a few places already.  This enables to resolve the tools that are defined as dependency and passed in as arguments.

One use-case that is not taken into account is when `erlang` is mananged via `mise` but not active. Installation of `elixir` will also fail in this case but i assume that is correct behaviour. It would also fail if `erlang` is not managed by `mise` and NOT on the PATH. Still it might be confusing for users.

**Comment by jdx** [comment]
I've thought a few things that `Config` should not be immutable and that it instead should be based on the arguments sent in always but this is difficult since occasionally we have 1-off `Config`s that are completely different than the main one.

I agree with your assessment, I wonder if we should have 1 global Config that is based on the arguments that we could reuse here but only use it in certain situations, like a `GLOBAL_CONFIG` OnceLock or something.

That's definitely easier said than done though. Maybe as a stop-gap we could have a mutex for tool arguments that we could populate when `mise install` starts which would then be used to optionally populate parts of the Config struct? Still not an _easy_ solution but maybe something along those lines?

**Comment by jdx** [comment]
looks like it's only used here when building it: https://github.com/jdx/mise/blob/5b2d79500ab89b2a4fe1a09a5abbcce1bde3cc55/src/toolset/tool_request_set.rs#L156

one refactor I have meant to do was to get rid of the almost identical logic in Toolset for this but haven't got around to that yet

**Comment by roele** [comment]
Using a global `TOOL_ARGS` seems to make things a lot easier. What i am not sure about yet is the existing code in `ToolRequestSet#load_runtime_args` for which the `self.args` never seems to be populated/set anywhere. For now i just kept things and added the additional handling for the TOOL_ARGS which seems to work fine so far.

File: src/cli/use.rs [hunk 1]
```diff
@@ -101,6 +101,10 @@ impl Use {
         if self.tool.is_empty() && self.remove.is_empty() {
             self.tool = vec![self.tool_selector()?];
         }
+        crate::env::TOOL_ARGS
```
jdx:
```suggestion
        env::TOOL_ARGS
```

minor nit

src/toolset/tool_request_set.rs
```diff
@@ -227,6 +227,20 @@ impl ToolRequestSetBuilder {
...
+        for arg in tool_args.iter() {
+            if let Some(tvr) = &arg.tvr {
+                arg_trs.add_version(tvr.clone(), &ToolSource::Argument);
+            } else if !trs.tools.contains_key(&arg.ba) {
+                // no active version, so use "latest"
+                let tr = ToolRequest::new(arg.ba.clone(), "latest", ToolSource::Argument)?;
+                arg_trs.add_version(tr, &ToolSource::Argument);
+            }
+        }
+        merge(trs, arg_trs);
```
jdx, line 242: I worry slightly this may have unintended consequences if we later create a config that isn't supposed to be "complete" but just a subset, but I don't know of any obvious places to check. I think it's good that we have this limited to install/use which should limit the blast radius.

Not a big enough issue that I think we shouldn't merge this but something to keep in mind if we see issues later.

jdx: the fact we don't have equivalent logic in the ToolsetBuilder might also be an issue, something to look out for later

roele: Yes, since the issue is limited to install/use at this point i think its better to only have this in ToolRequestSet for now. ToolsetBuilder is used much more broadly and the blast radius would be huge.

Issue #2488, opened 2024-08-22
standalone tasks


jdx:
tasks currently are assumed to be a part of a project or at least included in a mise.toml, but I think there are use-cases where you would want to have a task be just a single file you could execute outside of a mise project but have things like env vars and [tools](https://github.com/jdx/mise/issues/2487) defined just in the script

jdx:
these are all good, but I also think for one-liners we could also support `mise run ./path/to/build`

PR #4156, opened 2025-01-18
feat: split remote task

acesyde:
@jdx  what do you think about that ?

step one : refine the feth_tasks function <= we are here
step two : add cache system to avoid re-downloading tasks
step three : add more task providers


jdx:
code looks good, needs some docs though

jdx:
oh is all of this already documented?

acesyde:
In this pull request I have just refactored some code and I have added some non regression unit tests

After that I will open new pull requests for adding cache / git provider 

The next PRs will contain specific documentation according to the new features (cache etc...)

acesyde:
#2488

PR #4159, opened 2025-01-19
feat: set usage arguments and flags as environment variables for toml tasks

### LINKED ISSUE #4070: Set usage arguments and flag as environment variables before running the command
### Discussed in https://github.com/jdx/mise/discussions/4069

<sup>Originally posted by **gturi** January 12, 2025</sup>
I think it would be very useful to make mise populate the environment with the arguments and flags declared in the `usage` section before running the `run` section. In this way, tasks definition is made inside toml configuration, while their complex "implementation" is left inside the actual script.

```toml
[tasks.add-user]
description = "Add a user"
usage = '''
arg "<user>" default="unknown"
complete "user" run="mise run list-users-completion"
'''
run = 'node /path/to/test.js'

[tasks.list-users-completion]
hide = true
quiet = true
description = "List users"
run = 'echo "alice bob charlie"'
```

```node
// test.js
const { usage_user } = process.env;

console.log(`Adding user '${usage_user}'`);
```

Current output: `Adding user 'undefined'`

This enhancement also avoids the declaration of arguments both in `usage` and `run` section to make them work:

```toml
[tasks.add-user]
description = "Add a user"
usage = '''
arg "<user>"
'''
# the following definition feels a bit redundant 
# since I have already defined "user" argument in the usage section
run = 'echo {{arg(name="user")}}'
#run = 'echo $usage_user' # does not work
```
</div>

gturi:
Fixes #4070

@jdx I have tested it locally and everything seems to work fine, but before merging the PR I would like to have your opinion if everything seems ok. After that, I would like to add some e2e tests, but I do not know whether it is better to add them to e2e/tasks/test_task_run_toml, e2e/tasks/test_task_vars or to create a new test file just for this feature. Where do you think it is best to add them?

### DIFF
--- a/e2e/tasks/test_task_run_toml
+++ b/e2e/tasks/test_task_run_toml
@@ -44,3 +44,26 @@ arg "myarg" "myarg description" default="foo"
 run = 'echo {{arg(name="myarg")}}'
 EOF
 assert "mise run a" "foo"
+
+cat <<'EOF' >mise.toml
+[tasks.usage-env]
+usage = '''
+arg "myarg" "myarg description" default="foo"
+'''
+run = 'echo myarg=$usage_myarg'
+EOF
+
+assert "mise run usage-env" "myarg=foo"
+assert "mise run usage-env bar" "myarg=bar"
+
+cat <<'EOF' >mise.toml
+[tasks.usage-env]
+usage = '''
+flag "-m --myflag <myflag>" default="foo"
+'''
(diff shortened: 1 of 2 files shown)


jdx:
I think this is a solid change!

gturi:
Yes, unfortunately, it does not handle templates with environment variables right now... However, it could certainly be an interesting addition for the future

PR #4160, opened 2025-01-19
feat: add http cache

acesyde:
@jdx ,from your point of view where is the best place to store the cache ? I use the `MISE_CACHE_DIR` but according to the documentation [here](https://mise.jdx.dev/directories.html#cache-mise) i'm not really sure.


jdx:
I think we should probably make the cache optional doing something like the cache option for `exec()` in templates

acesyde:
Nice idea, i will implement it

acesyde:
Hello @jdx 

I think it's done if you want to check 👍

src/cli/run.rs
```diff
@@ -194,6 +194,10 @@ pub struct Run {
 
     #[clap(skip)]
     pub timed_outputs: Arc<Mutex<IndexMap<String, (SystemTime, String)>>>,
+
+    /// Don't cache the remote task file and always fetch it
+    #[clap(long, verbatim_doc_comment)]
+    pub no_cache: bool,
```
jdx, line 200: this is good, but we should offer it as a setting as well similar to what we do with task_timings:

https://github.com/jdx/mise/blob/77367caf17cca2edd97ebdfe4f07d5d607cafd8f/src/cli/run.rs#L870

it's pretty easy to do, just add the new setting to settings.toml in the project root and codegen will do all of the work

src/task/file_providers/http_file_provider.rs
```diff
@@ -1,17 +1,42 @@
...
-        Self { tmpdir }
+    pub fn new(cache_path: PathBuf, no_cache: bool) -> Self {
+        Self {
+            cache_path,
+            no_cache,
+        }
+    }
+}
+
+impl HttpTaskFileProvider {
```
jdx: I'd like to rename this and similar names from "FileProvider" to "RemoteTask" I think

```suggestion
impl RemoteTaskHttp {
```

src/task/file_providers/http_file_provider.rs
```diff
@@ -1,17 +1,42 @@
...
+    fn get_cache_key(&self, file: &str) -> String {
+        let mut hasher = Sha256::new();
+        hasher.update(file);
+        format!("{:x}", hasher.finalize())
```
jdx: I think we have a reusable function for this in hash.rs

src/task/file_providers/http_file_provider.rs
```diff
@@ -21,15 +46,37 @@ impl TaskFileProvider for HttpTaskFileProvider {
...
-            .path_segments()
-            .and_then(|segments| segments.last())
-            .unwrap();
-        let tmp_path = self.tmpdir.join(filename);
-        HTTP.download_file(file, &tmp_path, None)?;
-        file::make_executable(&tmp_path)?;
-        Ok(tmp_path)
+        match self.no_cache {
+            false => {
+                debug!("Cache mode enabled");
```
jdx: remove this one or make it trace level, the rest of the logs are fine though

src/task/file_providers/http_file_provider.rs
```diff
@@ -21,15 +46,37 @@ impl TaskFileProvider for HttpTaskFileProvider {
...
+                    debug!("Using cached file: {:?}", destination);
+                    return Ok(destination);
+                }
+
+                debug!("Downloading file: {}", file);
+                self.download_file(file, &destination)?;
+                Ok(destination)
+            }
+            true => {
+                debug!("Cache mode disabled");
```
jdx, line 65: remove or make trace level

src/task/file_providers/http_file_provider.rs
```diff
@@ -21,15 +46,37 @@ impl TaskFileProvider for HttpTaskFileProvider {
...
+                let url = url::Url::parse(file)?;
+                let filename = url
+                    .path_segments()
+                    .and_then(|segments| segments.last())
+                    .unwrap();
```
jdx: what would happen if someone just used a hostname like `https://myhost.com`? or if it ends in a trailing slash like `https://myhost.com/js/`? I think we'll need to handle that somehow without panicking or erroring

src/task/file_providers/http_file_provider.rs
```diff
@@ -64,13 +111,47 @@ mod tests {
                 .mock("GET", path)
                 .with_status(200)
                 .with_body("Random content")
+                .expect(2)
                 .create();
 
-            let provider = HttpTaskFileProvider::new(env::temp_dir());
+            let provider = HttpTaskFileProvider::new(env::temp_dir(), true);
```
jdx: this is optional, but I think it might be nice if we had a builder, maybe something like this:

```rs
let provider = RemoteTaskHttpBuilder::new(path).cache(true).build();
```

or perhaps we could make this work with dynamic dispatch:

```rs
let provider = RemoteTaskBuilder::new().http(path).cache(true).build();
```

though the latter is definitely tricky to do. This is totally up to you though, I'm fine with how things are currently. I've just found once I add 2 arguments to a create function it implies I'll probably need more later and builders make the code easier to read and work with. We could also add this later too.

src/task/file_providers/mod.rs
```diff
@@ -6,48 +7,54 @@ mod local_file_provider;
 pub use http_file_provider::HttpTaskFileProvider;
 pub use local_file_provider::LocalTaskFileProvider;
 
+use crate::dirs;
+
+static CACHE_FILE_PROVIDERS: Lazy<PathBuf> = Lazy::new(|| dirs::CACHE.join("tasks-file-provider"));
```
jdx:
```suggestion
static REMOTE_TASK_CACHE_DIR: Lazy<PathBuf> = Lazy::new(|| dirs::CACHE.join("tasks-file-provider"));
```

it wasn't clear to me from the name this is a directory path, it's named as if it contains a HttpTaskFileProvider and LocalTaskFileProvider struct

jdx (changes requested):
I think this is close, my main thing is that I think we should rename this since FileProvider sounds a bit generic and doesn't describe what these do very well. I suggested "RemoteTask" but I'm open to other ideas too.

src/task/file_providers/http_file_provider.rs
acesyde: I think I will keep that in mind for the next PRs 👍

src/cli/run.rs
acesyde: Yep, it's much better with settings 👍

src/task/file_providers/http_file_provider.rs
acesyde: Oh cool, less code to implement ^^

src/task/file_providers/http_file_provider.rs
acesyde: I have added some unit tests and validators in the `is_match` 👍

src/task/file_providers/mod.rs
acesyde: I renamed this variable and the generated directory

src/cli/run.rs
```diff
@@ -194,10 +194,6 @@ pub struct Run {
 
     #[clap(skip)]
     pub timed_outputs: Arc<Mutex<IndexMap<String, (SystemTime, String)>>>,
-
-    /// Don't cache the remote task file and always fetch it
-    #[clap(long, verbatim_doc_comment)]
-    pub no_cache: bool,
```
jdx: sorry I should've been more clear, I think we should have the cli flag _and_ a setting

acesyde: oh ok 😅

PR #4161, opened 2025-01-19
docs: neovim cookbook

EricDriussi:
From #3952, slightly improved code and explanation.

Not sure how to render the docs locally so hopefully it renders ok.


hverlin:
Maybe add a reference to this page from https://mise.jdx.dev/ide-integration.html#neovim?

PR #4233, opened 2025-01-26
feat: add git remote task provider 

acesyde:
## Description

Add a git remote task provider

This provider handle ssh and https generic format

Specific patterns like github or bitbucket are not translated because the ROI is too low and add useless complex code.

User must define a generic url with path


acesyde:
@jdx do you know how i can add some unit tests for this part 

image

Currently, we can't mock `git::` and `file::`

jdx:
I would use e2e tests for this

acesyde:
@jdx if you want to do a first review, all the code is done, I just have to do the e2e tests

acesyde:
@jdx All the code is finished and I'm trying to add some e2e tests but I'm having trouble with them :sweat_smile: if you have an idea how to implement them correcly

jdx:
hmm I don't know what it means:

```
Message:  called `Result::unwrap()` on an `Err` value: MietteDiagnostic { message: "git failed: Cmd([\"git\", \"-C\", \"/tmp/test_task_remote_git_https.EmQQUm/home/.cache/mise/remote-git-tasks-cache/791bacd406835a949d44cc2ac1106e0e4475d131ccc56b2540846d661bf5ff5c\", \"-c\", \"safe.directory=/tmp/test_task_remote_git_https.EmQQUm/home/.cache/mise/remote-git-tasks-cache/791bacd406835a949d44cc2ac1106e0e4475d131ccc56b2540846d661bf5ff5c\", \"fetch\", \"--prune\", \"--update-head-ok\", \"origin\", \"v2025.1.17:v2025.1.17\"]) error: cannot update ref 'refs/heads/v2025.1.17': trying to write non-commit object faa3b33c4b1d9a6efda1c7ade1c0b35e7146e1a8 to branch 'refs/heads/v2025.1.17'\nFrom [https://github.com/jdx/mise\n](https://github.com/jdx/mise/n) ! [new tag]           v2025.1.17 -> v2025.1.17  (unable to update local ref)\n * [new tag]           v2025.1.17 -> v2025.1.17\n", code: None, severity: None, help: None, url: None, labels: None }
```

it might be a bug in the git code that isn't correctly creating the repo or something. If I get some time next week I'll see if I can dive in but I don't think this is a test issue so much as an issue with the code.

acesyde:
Dont worry I found why, that why I want to add the clone options behavior to your git library, when it will be merged on main I will fix the issue in my pr 👍

acesyde:
@jdx should i use the crate::git or https://github.com/jdx/xx to clone repositories in this PR ?

acesyde:
@jdx i have fixed all the tests but to be able to pass ssh e2e in the github action i need to add a ssh key

This block of code cannot works because `RTX_SSH_KEY` doesn't exist for the test pipeline
```yaml
      - uses: shimataro/ssh-key-action@v2
        with:
          key: ${{ secrets.RTX_SSH_KEY }}
```

Can you do something for me ?

jdx:
you won't be able to use secrets in CI at all, is it possible to create one just for the test runs?

acesyde:
@jdx  I tried to create a temporary ssh key (latest commit) but it can't work properly because the key is not saved to any account :(

jdx:
I think long term you _should_ be able to use the `#MISE` comments. Not sure if it makes more sense to document this gap or just make it work.

PR #4250, opened 2025-01-28
Add args to the task prefix

borissmidt:
When running the same task with different arguments they are all printed the same.
So make it clear which task ran i propose to add the args to the prefix.

I had a stab at implementing the task dependency args but my rust skills/ knowledge of the project is a bit too limited for that.

### DIFF
--- a/src/task/mod.rs
+++ b/src/task/mod.rs
@@ -136,6 +136,7 @@ impl Debug for EitherStringOrIntOrBool {
     }
 }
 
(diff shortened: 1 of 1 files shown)


borissmidt:
But then how would you keep the tasks apart.
As an alternative could the prefix be truncated at 20 characters or something?

borissmidt:
I'll have a look.

PR #4178, opened 2025-01-22
docs: Add section about .python-version file

sconvent:
This adds a section to describe how `.python-version` files can be used, which was previously undocumented.
The documentation for other languages describe this feature similary.

PR #4252, opened 2025-01-28
docs: Update sections about idiomatic version files

hverlin:
- Apply suggestion from https://github.com/jdx/mise/pull/4178.
- Make each section a bit more consistent

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 /node_modules/
 package-lock.json
 .mise.lock
+/mise.local.toml
 
 *.log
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -3,6 +3,7 @@
 /node_modules/
 package-lock.json
 .mise.lock
+/mise.local.toml
```
hverlin, line 6: I can remove and use my local git exclude if you prefer. I think it might be a good thing to have it in the repo as an example, though :smiley:

PR #4260, opened 2025-01-29
docs: Update python documentation

hverlin:
- Update python documentation. 
- Indicate how one might use it with `uv` in the cookbook (at least according to my understanding and after experimenting a bit with it)
- Re-order the sections in the python page with the section about `uv`, `.python-version` and virtual env first
- Update the `mise` json schema to match the documentation (and observed behavior)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -274,7 +274,7 @@ in mise and nvm. Here are some of the supported idiomatic version files:
 | terraform | `.terraform-version`, `.packer-version`, `main.tf` |
 | yarn      | `.yarnrc`                                          |
 
-In mise these are enabled by default. You can disable them with [`mise settings idiomatic_version_file=false`](/configuration/settings.html#idiomatic_version_file) or for specific tools with [`mise settings idiomatic_version_file_disable_tools=['python']`](/configuration/settings.html#idiomatic_version_file_disable_tools).
+In mise these are enabled by default. You can disable them with [`mise settings idiomatic_version_file false](/configuration/settings.html#idiomatic_version_file) or for specific tools with [`mise settings idiomatic_version_file_disable_tools python`](/configuration/settings.html#idiomatic_version_file_disable_tools).
 
 There is a performance cost to having these when they're parsed as it's performed by the plugin in
 `bin/parse-version-file`. However, these are [cached](/cache-behavior) so it's not a huge deal.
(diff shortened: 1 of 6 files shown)


jdx:
this code seems completely wrong and I think it should just be deleted which would fix the issue https://github.com/jdx/mise/blob/a8ea4f31f0aa30ad15a742c54e554501196a6435/src/config/settings.rs#L123-L127

hverlin:
@jdx yes, after removing the code you mentioned, it's working :+1:  Added a test as well

jdx:
I added that code, I bet it was the result of using copilot lol

docs/configuration.md
```diff
@@ -274,7 +274,7 @@ in mise and nvm. Here are some of the supported idiomatic version files:
 | terraform | `.terraform-version`, `.packer-version`, `main.tf` |
 | yarn      | `.yarnrc`                                          |
 
-In mise these are enabled by default. You can disable them with [`mise settings idiomatic_version_file=false`](/configuration/settings.html#idiomatic_version_file) or for specific tools with [`mise settings idiomatic_version_file_disable_tools=['python']`](/configuration/settings.html#idiomatic_version_file_disable_tools).
+In mise these are enabled by default. You can disable them with [`mise settings idiomatic_version_file false](/configuration/settings.html#idiomatic_version_file) or for specific tools with [`mise settings idiomatic_version_file_disable_tools python`](/configuration/settings.html#idiomatic_version_file_disable_tools).
```
jdx, line 277: is this right? I would think you'd need to use `mise settings add|set`

hverlin: yeah, it works without `add/set` https://mise.jdx.dev/cli/settings.html#arguments. It's in the e2e test also now

docs/mise-cookbook/python.md
```diff
@@ -44,3 +44,112 @@ echo "Project: $PROJECT_NAME"
...
+Another option is to use `_.python.venv` in your `mise.toml` file to specify the path to the virtual environment created by `uv`.
+
+```toml [mise.toml]
+[env]
+_.python.venv = { path = ".venv" }
+```
+
+### Syncing python versions installed by mise and uv
+
+You can use [uv sync python --uv](/cli/sync/python.html#uv) to sync the python version installed by `mise` with the python version specified in the `.python-version` file in the `uv` project.
```
justbispo, line 99: I know this pull was already merged, but I don't think this bug needs an issue report.
The command should be `mise sync python --uv`, not `uv sync python --uv`

hverlin: Thank you for catching this. This should be `mise sync` indeed. I hope the link makes this obvious, for now. I will update it with my next PR

hverlin: Seems this got already fixed https://github.com/jdx/mise/commit/73d125ef16c2c591b5616f18ef6382ec8d2b835e

PR #4267, opened 2025-01-30
feat(registry): add falco to registry

ronnnnn:
Add [falco](https://github.com/ysugimoto/falco) to registry.


jdx:
asdf plugins are not accepted

PR #4287, opened 2025-02-02
docs: improve registry filtering performance

roele:
Return early on filtering/highlighting to improve filtering performance issues which were not showing in dev mode.

### DIFF
--- a/docs/components/registry.vue
+++ b/docs/components/registry.vue
@@ -14,7 +14,7 @@
       </tr>
     </thead>
     <tbody>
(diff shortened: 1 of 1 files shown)


jdx:
I'm not sure what to do about this, I think it's an existing issue, but it renders poorly on mobile

image

roele:
table is too wide, maybe we need to enable horizontal scroll on overflow or shorten long text with ellipses. i will have a look at this tomorrow...

PR #4289, opened 2025-02-02
feat: add gittool/gitversion

acesyde:
- add aqua and ubi for gitools/gitversion
- bump aqua-registry


jdx:
can you remove the aqua-registry change? that happens automatically as part of releases

PR #4324, opened 2025-02-06
docs(schema): fix description of task.dir default

risu729:
Resolves #4323. Since the default is explained by `default` property, just remove it from the `description`.


JanPokorny:
@risu729 Not sure if removing is the correct step, since that description is used as a tooltip in VSCode where the default is not visible otherwise:
image

risu729:
Oh sorry, I didn't notice that. Thanks for letting me know!

risu729:
I synced the description with the [docs](https://mise.jdx.dev/tasks/toml-tasks.html#VPSidebarNav:~:text=dir%20%3D%20%22%7B%7Bcwd%7D%7D%22%20%23%20run%20in%20user%27s%20cwd%2C%20default%20is%20the%20project%27s%20base%20directory).

PR #4325, opened 2025-02-06
refactor(registry): use aqua for slsa-verifier

scop:


jdx:
I think because aqua depends on slsa-verifier this is causing an endless loop

scop:
Hm, from https://aquaproj.github.io/docs/reference/security/cosign-slsa/

> You don't have to install Cosign and slsa-verifier, because aqua installs Cosign automatically and slsa-verifier is embedded to aqua as Go library.

scop:
Anyway `mise use -g aqua:slsa-framework/slsa-verifier` works for me.

jdx:
closing since it fails CI and I don't know how to even fix it

PR #4328, opened 2025-02-06
feat: task confirmation

### LINKED ISSUE #2050: task confirmation
just has a really cool feature to provide confirmation dialogs for a dangerous task:

```just
[confirm("Are you sure you want to delete everything?")]
delete-everything:
  rm -rf *
```

I'm thinking we could probably go beyond this by leveraging similar features in demand like choosing from different choices: https://github.com/jdx/demand

roele:
Resolves #2050 but requires a [fix](https://github.com/jdx/demand/issues/87) in demand.


roele:
Now with the proposed fixed implemented in `demand` the behaviour of parallel tasks writing to stdout while the confirmation dialog is shown is like follows.

https://github.com/user-attachments/assets/ee51e3ce-8b8a-4f49-b2df-d6b173185874

jdx:
needs docs

PR #4333, opened 2025-02-07
fix: lowercase desired shim names

KevSlashNull:
This fixes a bug on macOS where we are searching for a shim name with one or more characters uppercased, e.g. Python. But on macOS the APFS (default file system) is case insensitive, so `mise doctor` will always report an error in this case.

See https://github.com/jdx/mise/discussions/4332

### DIFF
--- a/src/shims.rs
+++ b/src/shims.rs
@@ -267,6 +267,9 @@ fn get_desired_shims(toolset: &Toolset) -> Result<HashSet<String>> {
                         ]
                     })
                     .collect()
+            } else if cfg!(macos) {
+                // some bins might be uppercased but on mac APFS is case insensitive
+                bins.into_iter().map(|b| b.to_lowercase()).collect()
             } else {
                 bins
             }


jdx:
do you have an example of a tool that has a bin with an uppercase name? is this a specific python flavor or something?

KevSlashNull:
@jdx, I assume it's something specific to the Python plugin but I don't really use Python and I also uninstalled the plugin during debugging and it still showed the doctor error.

I'll check on Monday when I'm back at my work laptop, and see where that shim is coming from. It doesn't happen on Linux seemingly.

KevSlashNull:
Thanks for merging and sorry for responding late, @jdx! For posterity, I found the source of the uppercase `Python` binary.

There exists a binary at `$HOME/.local/share/mise/installs/awscli/2.18.0/aws-cli.pkg/Payload/aws-cli/Python` for which mise tries to create a shim.

PR #4346, opened 2025-02-08
chore: build linux arm on arm

jdx:


jdx:
ah it seems cross doesn't support arm. No biggie, I don't think this would've improved performance or much else—just seemed like a smart thing to do

PR #4350, opened 2025-02-09
docs: update demo

hverlin:
- Update the demo to show that one can install multiple tools
- Add transcript
- Update readme (add link to the demo page with a video) + additional examples

image


README.md
```diff
@@ -57,14 +54,95 @@ echo '~/.local/bin/mise activate fish | source' >> ~/.config/fish/config.fish
...
+
+[tasks.deploy]
+description = "Deploy infrastructure after validation"
+depends = ["validate", "plan"]
+run = "terraform apply -auto-approve"
+```
+
+Run it with:
+
+```sh-session
```
acesyde, line 139:
```suggestion
```sh
```

README.md
```diff
@@ -57,14 +54,95 @@ echo '~/.local/bin/mise activate fish | source' >> ~/.config/fish/config.fish
...
+### Run tasks
+
+```toml
+# mise.toml
+[tasks.build]
+description = "build the project"
+run = "echo building..."
+```
+
+```sh-session
```
acesyde, line 94:
```suggestion
```sh
```

README.md
```diff
@@ -57,14 +54,95 @@ echo '~/.local/bin/mise activate fish | source' >> ~/.config/fish/config.fish
...
+
+### Manage environment variables
+
+```toml
+# mise.toml
+[env]
+SOME_VAR = "foo"
+```
+
+```sh-session
```
acesyde, line 77:
```suggestion
```sh
```

docs/tapes/demo.tape
```diff
@@ -0,0 +1,90 @@
+# Use mise run docs:demo to record this tape
+
+Output demo.gif
```
acesyde, line 3: if the demo.gif is not used you can probably removed it

docs/tapes/demo.gif
acesyde, line 1: seems to not be used

docs/tapes/demo.tape
hverlin: Yeah, I can remove it 👍
The other option I was thinking about is to leave the gif in the readme and go to the demo page if one clicks on it ?

acesyde: the mp4 can be paused / resume, it's a huge advantage imo

hverlin: Yeah, I agree. So my idea would be to have the video as `.mp4` on the demo page only - while we leave the `.gif` in the Readme. 

The `.gif` will work everywhere, while the video is not shown in some markdown renderer. If the user tries to click on the `.gif` to pause it, then it will send them to the [demo](https://mise.jdx.dev/demo.html) page where they can see the transcript + play/pause the video. I think this will also avoid having to host it through GitHub :thinking:

btpemercier: You're right

hverlin:
@jdx, Given the discussion in https://github.com/jdx/mise/pull/4414, maybe I should update the demo to show `mise exec` first as well?

jdx:
yeah I think so. `mise exec|x` is a much better way to test that mise is working for brand new users since it doesn't require config files or activation working properly so I think it's a better place to start to just make sure mise is working at all.

hverlin:
@jdx I have updated the demo to showcase `mise exec` now. Let me know if we should make any other changes and I will re-record it.

jdx:
it looks incredible! thanks so much!

PR #4359, opened 2025-02-10
docs: add hint about unsupported go.mod idiomatic version file

roele:


jdx:
I don't think it's necessary to document this

PR #4387, opened 2025-02-13
docs: update Fedora install for dnf5

rkben:
Fedora 41+ is using `dnf5` which has slightly different commands.


jdx:
well that's dumb

PR #4404, opened 2025-02-15
fix: properly iterate over depends_post entries to retrieve tasks

IxDay:
This is an attempt to fix: https://github.com/jdx/mise/discussions/4398 and https://github.com/jdx/mise/discussions/4222 
I am not familiar with Rust or the code base here. However, this change seems to fix the issue for my case. When running `mise task deps` I get the expected result from the setup in: https://github.com/jdx/mise/discussions/4398
When running the various tasks I get the proper output as well. Happy to get feedback and guidance here.

### DIFF
--- a/src/task/mod.rs
+++ b/src/task/mod.rs
@@ -291,9 +291,10 @@ impl Task {
             .flatten_ok()
             .filter_ok(|t| tasks_to_run.contains(t))
             .collect_vec();
-        let depends_post = tasks_to_run
+        let depends_post = self
+            .depends_post
             .iter()
-            .flat_map(|t| t.depends_post.iter().map(|td| match_tasks(&tasks, td)))
+            .map(|td| match_tasks(&tasks, td))
             .flatten_ok()
             .filter_ok(|t| t.name != self.name)
             .collect::<Result<Vec<_>>>()?;


src/task/mod.rs
```diff
@@ -291,9 +291,10 @@ impl Task {
             .flatten_ok()
             .filter_ok(|t| tasks_to_run.contains(t))
             .collect_vec();
-        let depends_post = tasks_to_run
+        let depends_post = self
+            .depends_post
             .iter()
-            .flat_map(|t| t.depends_post.iter().map(|td| match_tasks(&tasks, td)))
```
jdx, line 296: now I could be wrong about this and I would love it if we could have a simple fix. That said, if I recall this right it will be more complex.

if memory serves this method is supposed to return every dependency of self. The reason we don't iterate over depends_post the way wait_for and depends are done is because the way depends_post is implemented it's actually a reverse relationship—the task in depends_post actually depends on self.

so... I don't think this will work, but I also am not 100% on this so if you want to stab at it more and are able to get the test suite to work I would certainly welcome a simple fix like this.

jdx: I wonder if instead of having a channel that the dependency graph returns the next item with no leaves, perhaps it would be better to iterate through the list of tasks recursively: pushing depends and wait_fors in front of the task that defines those and depends_post after. That final 1-dimensional list would then be the work queue.

PR #4413, opened 2025-02-16
revert: "feat: set usage arguments and flags as environment variables for toml tasks"

### LINKED ISSUE #4379: Using flag in run array causes error "unexpected word"
### Discussed in https://github.com/jdx/mise/discussions/4370

<sup>Originally posted by **kehm** February 11, 2025</sup>
Hi. Assuming the error encountered in the below example is not intended?

Description:
The task below works fine in v2025.2.1 but returns "ERROR unexpected word: echo abc" in v2025.2.2 or higher.
This issue only seem to occur when using a flag on the second or later entry in the run array.

To reproduce:
[tasks.test]
run = [
  'echo abc',
  './script.sh {{flag(name="test", default="false")}}',
]

mise run test --test

Environment:
2025.2.2 macos-arm64 (2025-02-08)
zsh 5.9 (arm64-apple-darwin24.0)
experimental  true</div>

jdx:
Reverts jdx/mise#4159

Fixes #4379

### DIFF
--- a/e2e/tasks/test_task_run_toml
+++ b/e2e/tasks/test_task_run_toml
@@ -44,26 +44,3 @@ arg "myarg" "myarg description" default="foo"
 run = 'echo {{arg(name="myarg")}}'
 EOF
 assert "mise run a" "foo"
-
-cat <<'EOF' >mise.toml
-[tasks.usage-env]
-usage = '''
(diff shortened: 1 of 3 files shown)


jdx:
I added a test case to prevent this from regressing again @hverlin

hverlin:
Thank you!

PR #4414, opened 2025-02-16
docs: Update shims page

hverlin:


docs/dev-tools/shims.md
```diff
@@ -1,59 +1,61 @@
 # Shims
 
 ::: tip
-The [beginner's guide](https://dev.to/jdxcode/beginners-guide-to-rtx-ac4), and my [blog post](https://jdx.dev/posts/2024-04-13-shims-how-they-work-in-mise-en-place/) are helpful resources to dive deeper into shims.
-:::
-
-::: warning
-`mise activate --shims` does not support all the features of `mise activate`.<br>
-See [shims vs path](/dev-tools/shims.html#shims-vs-path) for more info.
+The [beginner's guide](https://dev.to/jdxcode/beginners-guide-to-rtx-ac4), and this [blog post](https://jdx.dev/posts/2024-04-13-shims-how-they-work-in-mise-en-place/) are helpful resources to dive deeper into shims.
```
hverlin: Do you think it makes sense to leave these links? I feel like the beginner guide might be confusing a bit - wondering if it should be updated an integrated somewhere in the docs?

docs/dev-tools/shims.md
```diff
@@ -73,23 +75,44 @@ echo 'eval "$(mise activate zsh)"' >> ~/.zshrc    # this sets up interactive ses
...
-- Most hooks won't trigger
+- [Env vars](/environments/) defined in mise are only available to mise tools
+- Most [hooks](/hooks.html) won't trigger
 - The unix `which` command points to the shim, obscuring the real executable
 
-In general, I recommend using PATH (`mise activate`) instead of shims for _interactive_ situations. The
-way activate works is every time the prompt is displayed, mise-en-place will determine what PATH and other
+In general, using PATH (`mise activate`) instead of shims for _interactive_ situations is recommended.
+
+The way `activate` works is every time the prompt is displayed, mise-en-place will determine what PATH and other
```
hverlin, line 132: I think this is not true anymore, right? mise hooks on `cd` now?

docs/dev-tools/shims.md
```diff
@@ -228,7 +244,7 @@ penalty. This is because when a shim is called, mise sets up the environment wit
 those PATH entries will be before the shim directory.
 
 In other words, which is better in terms of performance just depends on how you're calling mise. Really
-though I think most users won't notice a 5ms lag on their terminal so I suggest `mise activate`.
+though most users will not notice a few ms lag on their terminal caused by `mise activate`.
 
 ## Neither shims nor PATH
```
hverlin: Not really sure how to update this section. I can update it to mention `mise exec`/`mise run` and indicate that one can create aliases?

docs/dev-tools/shims.md
```diff
@@ -156,14 +178,11 @@ The [hooks](/hooks.html) `cd`, `enter`, `exit`, and `watch_files` only trigger w
...
-e.g:
+`which` is a command that a lot of users find great value in. Using shims effectively "break" `which` and cause it to show the location of the shim. A workaround is to use `mise which` will show the actual location. Some users prefer the "cleanliness" of running `which node` and getting back a real path with a version number inside of it. e.g:
 
 ```sh
 $ which node
-/Users/jdx/.mise/installs/node/20/bin/node
+~/.mise/installs/node/20/bin/node
 ```
 
 ## Hook on `cd`
```
hverlin: I am not sure if this is still accurate?

docs/dev-tools/shims.md
```diff
@@ -196,22 +215,19 @@ node some_script.js
 
 :::
 
-The only difference I can think of between these would be that using `hook-env` you will need to call
+The only difference between these would be that using `hook-env` you will need to call
```
hverlin: Should this section be renamed to `using shims in rc files?`

hverlin:
@jdx I started updating the page, but there are a few things I'm unsure about (see the inline comments). Please let me know your thoughts so I can finalize the PR

docs/dev-tools/shims.md
jdx: I don't feel like that beginner's article has been very useful for people as I had hoped. Maybe we can move it to the external resources page. It's also written for rtx so unless I get around to updating it (which I doubt given I don't think it's been helpful in the first place) I think we should reduce the number of times we mention it—but the external resources page makes sense to me.

docs/dev-tools/shims.md
jdx: this is actually shell dependent. I think that is true for bash/zsh/fish.

I think there is a bigger question here, which is that those I think of as "tier-1" shells because they're the only 3 that I actively maintain where the others are community maintained and I rarely use or test. That said, I don't know if I want to actually put that into the docs because that alone doesn't mean that they have worse support, "community maintained" might give the impression that it's abandoned but people _do_ maintain these and I do review the code.

The other thing is the tier-1 shells are similar to each other and have similar features, so they behave in similar ways like this one. Other shells may not offer a prompt display hook (powershell) and really I don't even know what the gaps are in the others.

I wonder if what we should do is build a table that shows exactly what features/behaviors we care about in the shell hooks. We could maybe have an icon for supported, not yet supported, and not possible in the given shell.

docs/dev-tools/shims.md
jdx: I think this is a fine change, of course which won't output a tilde but users will understand what you mean

docs/dev-tools/shims.md
jdx: yeah it's been a while since we've been in the 5ms area haha.

Taking a step back, the most important but also most difficult thing I want to convey is this:

using tools in mise requires one of the following approaches:

* `mise activate`
* `mise activate --shims`
* `mise x|run`

I have tried to explain this, but it is _definitely_ not understood. In fact I think this is why I wrote the "how I use mise" section. I often see people say "mise doesn't use shims" which as you know is not true and has pretty much never been true since shims were added almost 2 years ago now.

I'm wondering if the shims article is the wrong place for this, like maybe this should go into the getting started guide and we should give users a choice between these 3. Actually, now that I'm saying this I really like the idea of defaulting to `mise x|run` _instead_ of `mise activate`.

I think if we did that, users could get started quickly and we won't need to explain all the shell setup stuff that's different per environment and the relatively complex docs around updating `bashrc` which I'm sure is the hardest part. This would make setup way easier, teach users how to use mise without activation, and we could present activation as an optional step so they don't need to always use `mise x|run`.

The biggest downside I can think of here is that the env var stuff would only work if mise is run through `mise x|run`—but I don't think that's a dealbreaker.

Maybe we could have a new "Getting Started Advanced" or "Activation" doc that we would link from the current getting started which would basically be "here's how to setup mise without always needing `mise x|run`.

Unrelated to this, I was thinking today when I was using `mise run` in a script that called it a bunch. The overhead of `mise run` is significant because it has to resolve the whole toolset every time. I think there might be the ability to skip some of that logic if we can detect it has already been done in that environment which would make `mise run` when the environment is already active a lot faster. I'll open up a new issue for this.

docs/dev-tools/shims.md
hverlin: For now, I have clarified that `bash`, `zsh` on `fish` at least have a hook on cd

docs/dev-tools/shims.md
hverlin: Updated the getting started page as well as this section :+1:

PR #4421, opened 2025-02-16
fix: regression with env overriding

jdx:
regression caused by #4412


gawbul:
Ah, it looks like you have noticed this already 😂
I've been trying to debug all afternoon in #4420 😣 
I narrowed it down to the same change you made in `src/config/mod.rs`, but don't know Rust well enough (at all), and would never have picked up the other changes you made.
Good I don't need to raise a PR now 👍
Thank you 🙏

PR #4435, opened 2025-02-17
docs: Add a section on how to use environment variables

hverlin:
- update env. vars page to indicate under which conditions they work
- small updates to backend/plugins/registry pages
- clarify that `mise alias` command is only for version aliases (not backend aliases)

### DIFF
--- a/docs/components/registry.vue
+++ b/docs/components/registry.vue
@@ -50,7 +50,7 @@ export default {
   data() {
     return {
       filter:
-        new URLSearchParams(globalThis.location.search).get("filter") || "",
+        new URLSearchParams(globalThis?.location?.search).get("filter") || "",
       data: data,
     };
   },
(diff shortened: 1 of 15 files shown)


docs/components/registry.vue
```diff
@@ -60,6 +63,18 @@ export default {
       });
     },
   },
+  watch: {
```
hverlin, line 71: image

xtasks/fig/src/mise.ts
```diff
@@ -538,7 +538,7 @@ const completionSpec: Fig.Spec = {
     },
     {
       name: ["alias", "a"],
-      description: "Manage aliases",
+      description: "Manage version aliases.",
```
jdx, line 541: really this should be changed, I just haven't got around to updating the command

jdx (approved):

PR #4446, opened 2025-02-18
fix(cli/run): inherit stdio by --raw even when redactions are enabled

risu729:


src/cli/run.rs
```diff
@@ -629,11 +629,15 @@ impl Run {
...
         let redactions = config.redactions();
+        let raw = self.raw(Some(task));
         let mut cmd = CmdLineRunner::new(program.clone())
             .args(args)
             .envs(env)
             .redact(redactions.deref().clone())
-            .raw(self.raw(Some(task)));
+            .raw(raw);
+        if raw && !redactions.is_empty() {
+            warn_once!("--raw will prevent mise from being able to use redactions");
```
jdx: is this actually showing every time you run the CLI? I thought there was some logic that only showed once and saved a file to the state directory so it wouldn't display again, but I'm not actually seeing that.

I don't think we would want this to display every time the CLI is run so if I am correct this _is_ displaying every time please do one of 2 things (you're choice):
* Add a new macro like warn_once! that writes a file to the state directory to prevent it from showing a second time in a later execution of the CLI
* comment this out with a TODO that if we add that macro later we should enable this

risu729: Yes, it's printed every time.
I haven't touched macro, so I will leave it as TODO for now.

risu729: I don't know about the state directory, but should I re-warn when the config file is modified, or not needed?

jdx: ah! I was thinking of `hint!`, that should work here: https://github.com/jdx/mise/blob/e68b62484fbbf8a4cfc081221452179328b6de63/src/hint.rs#L9-L23

risu729: Thanks! confirmed it only showing the hint once.

PR #4476, opened 2025-02-21
fix: fix the behavior of .disable-self-update

ZeroAurora:
Closes #4468
A quick fix for this discussion


src/cli/self_update.rs
```diff
@@ -117,7 +117,12 @@ impl SelfUpdate {
             .ok()
             .and_then(|p| p.parent().map(|p| p.to_path_buf()))
             .and_then(|p| p.parent().map(|p| p.to_path_buf()))
-            .map(|p| p.join("lib").join(".disable-self-update").exists())
```
jdx, line 120: this is a breaking change, you need to look in both locations

ZeroAurora: completed

PR #4493, opened 2025-02-22
fix: mise outdated --json does not return json if all tools are up-to-date

### LINKED ISSUE #4471: mise outdated --json does not return json if all tools are up-to-date
### Discussed in https://github.com/jdx/mise/discussions/4470

<sup>Originally posted by **hverlin** February 20, 2025</sup>
For example, install `mise` and immediately run `mise outdated --json`
```sh
> mise outdated --json
All tools are up to date
```

I would expect an empty object?

Would it be ok to make a PR for it?
https://github.com/jdx/mise/blob/0824490c14d17cd93c7d68930b514eb11635c451/src/cli/outdated.rs#L56-L70</div>

roele:
Fixes #4471

### DIFF
--- a/src/cli/outdated.rs
+++ b/src/cli/outdated.rs
@@ -53,6 +53,19 @@ impl Outdated {
         ts.versions
             .retain(|_, tvl| tool_set.is_empty() || tool_set.contains(&tvl.backend));
         let outdated = ts.list_outdated_versions(self.bump);
+        self.display(outdated)?;
+        Ok(())
+    }
+
+    fn display(&self, outdated: Vec<OutdatedInfo>) -> Result<()> {
+        match self.json {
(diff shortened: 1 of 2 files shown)


src/cli/outdated.rs
```diff
@@ -82,6 +88,10 @@ impl Outdated {
 
     fn display_json(&self, outdated: Vec<OutdatedInfo>) -> Result<()> {
         let mut map = IndexMap::new();
+        if outdated.is_empty() {
+            miseprintln!("{}", serde_json::to_string_pretty(&map)?);
+            return Ok(());
+        }
```
jdx: if we left this out wouldn't it still do the same thing?

roele: Correct, good catch!

PR #4494, opened 2025-02-22
fix: bug when using mise use -g when MISE_ENV is filled

### LINKED ISSUE #4490: Bug when using mise use -g when MISE_ENV is filled
### Discussed in https://github.com/jdx/mise/discussions/4473

<sup>Originally posted by **jylenhof** February 20, 2025</sup>
Doing :
```
unset MISE_ENV
mise use -g uv@latest
mise ~/.config/mise/config.toml tools: uv@0.6.2
```

Now when using
```
export MISE_ENV=production
mise use -g uv@latest
mise WARN  uv is defined in ~/.config/mise/config.toml which overrides the global config (~/mise.production.toml)
mise ~/mise.production.toml tools: uv@0.6.2
```

When doing mise use -g the modified file should always be ~/.config/mise/toml, the global config file !</div>

roele:
Fixes #4490 

Using `MISE_ENV` circumvents the option overrides, so `global`, `path` and `env` options should come before `MISE_ENV` in the loading priority.

### DIFF
--- a/e2e/cli/test_use
+++ b/e2e/cli/test_use
@@ -75,6 +75,13 @@ assert "cat ~/.config/mise/config.toml" '[tools]
 gh = "2"'
 rm -f ~/.config/mise/config.toml
 
+export MISE_ENV=test
+mise use -g dummy@1
+assert "cat ~/.config/mise/config.toml" '[tools]
+dummy = "1"'
+rm -f ~/.config/mise/config.toml
+unset MISE_ENV
+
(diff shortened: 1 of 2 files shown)


jdx:
I think anytime we modify this code we should add a test. I don't think this would regress again, but this code has also been shown to be the result of so many bugs.

roele:
Agree. Will look into adding some e2e tests later today or tomorrow.

PR #4513, opened 2025-02-24
feat(registry): add cocogitto

reitzig:
Adding [cocogitto/cocogitto](https://github.com/cocogitto/cocogitto) (binary filename: `cog`) to the registry. 

Should (?) work:

```
❯ mise use aqua:cocogitto/cocogitto
mise aqua:cocogitto/cocogitto@6.2.0 ✓ installed                                                                                                                                                            mise ~/code/tools/my-project/.config/mise.toml tools: aqua:cocogitto/cocogitto@6.2.0
```

Hope this counts as obvious. ;)

### DIFF
--- a/registry.toml
+++ b/registry.toml
@@ -382,7 +382,7 @@ cockroach.backends = [
 ]
 cockroach.test = ["cockroach version", "v{{version}}"]
 cocoapods.backends = ["asdf:mise-plugins/mise-cocoapods"]
-cocogitto.aliases = ["cog"]
+cocogitto.test = ["cog --version", "cog {{version}}"]
 cocogitto.backends = ["aqua:cocogitto/cocogitto"]
 codefresh.backends = [
     "ubi:codefresh-io/cli[exe=codefresh]",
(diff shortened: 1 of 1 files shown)


registry.toml
```diff
@@ -382,6 +382,8 @@ cockroach.backends = [
 ]
 cockroach.test = ["cockroach version", "v{{version}}"]
 cocoapods.backends = ["asdf:mise-plugins/mise-cocoapods"]
+cocogitto.aliases = ["cog"]
```
jdx: doesn't seem notable enough to squat on this alias

jdx:
needs a test

registry.toml
reitzig: Okay. 🤷

For future reference, how do you measure "notable"? I see TLA aliases for projects with as few as 26 stars, so that can't be it?

reitzig:
I was looking for tests in the codebase and for hints in the contributing instructions, only to completely miss the innocent TOML lines right there. My bad! 😅

PR #4529, opened 2025-02-25
feat(registry): Add boilerplate

ZachGoldberg:


jdx:
asdf plugins are no longer accepted

PR #4539, opened 2025-02-26
fix: the graph calculation with post_depends.

boris-smidt-klarrio:
This bugfix fixes the graph calculation it seems that the 'post_depends' where incorrectly handled resulting in tasks where all dependending on eachother:
```toml
["task1"]
depends_post=[
    'task2',
    'task3',
    'task4',
]

["task2"]
run = "i'm task 2"

["task3"]
run = "i'm task 3"

["task4"]
run = "i'm task 4"
```

```
DEBUG Graph {
    Ty: "Directed",
    node_count: 3,
    edge_count: 6,
    edges: (2, 0), (1, 0), (1, 2), (2, 1), (0, 2), (0, 1),
```


jdx:
looks identical to https://github.com/jdx/mise/pull/4404

boris-smidt-klarrio:
Not sure why this isn't correct?
If i understand the code correct a depends_post means that it adds an extra depends to the task it references in its depends_post.

boris-smidt-klarrio:
Oke so the issue is that this will result in
```
/home/boris/projects/mise/target/debug/mise task deps one two three
one
two
└── one
three
└── one
```

So when you run two and three, it would run one even tough it isn't in the running tasks.

So what we actually want is:
```
/home/boris/projects/mise/target/debug/mise task deps one
one
two
└── one
three
└── one
```

but when you have the dependencies of two you want to get:
```
/home/boris/projects/mise/target/debug/mise task deps two
two
```

boris-smidt-klarrio:
I suspect the post_dependency is really a feature that increases the changes by a lot that you get 'cyclical' graphs.

boris-smidt-klarrio:
I've added an e2e test:
https://github.com/klarrio-oss/mise/blob/4daaf77c782296b5b224fc9434cb3d8e5173dbed/e2e/tasks/test_task_depends_post_inf_loop#L17-L17

boris-smidt-klarrio:
I've run the deps command locally and that one is still incorrect:
```
mise task deps one
# one

mise task deps two
# two

mise task deps one two
# one
# two
# └── one
```

So i suppose the deps needs to be updated that it shows
```
mise task deps one
# one
# two
# └── one
# three
# └── one
```

boris-smidt-klarrio:
@jdx Is there any way we can move this forward, it fixes the post_depends lockup bug.

boris-smidt-klarrio:
I hope you can find some time to have another look at this to atleast have a better working post_dependency instead of a 'fully broken' version of it.

PR #4540, opened 2025-02-26
feat: cleanup the mutex use.

boris-smidt-klarrio:
refactor: the condition on when the task runner finishes.
refactor: use the insert boolean instead of a !contains
feature: infinitive loop detection


boris-smidt-klarrio:
I can add a the case of pr #4539

boris-smidt-klarrio:
Thank you :+1:

PR #4576, opened 2025-03-04
docs: continuous-integration.md: fix gitlab caching example

nafg:
It seems MISE_DATA_DIR needs to be absolute, as per https://github.com/jdx/mise/discussions/4313

This change fixed it in my CI so perhaps the docs need to be updated.


hverlin:
👍 lgtm
Not sure why I did not hit this issue when I documented it 🤔 I am only using the [bootstrap](https://mise.jdx.dev/continuous-integration.html#example-with-the-bootstrap-script) solution on GitLab CI now, though

PR #4599, opened 2025-03-10
feat(registry): add curlie

reitzig:
Adding [rs/curlie][1] (binary filename: `curlie`) to the registry.

[1]: https://github.com/rs/curlie


reitzig:
It seems to me that neither of the failed checks relate to the proposed change; am I missing something?

PR #4605, opened 2025-03-11
fix(task): allow args to be used with tera tests

risu729:
This PR enables the use of `arg`, `option`, and `flag` in task templates within `if` conditions or similar syntaxes.

### Current Behavior
Currently, placeholders are replaced only after the Tera template is fully rendered. So, expressions like the following:

```toml
[tasks.test]
run = "echo {% if flag(name='test') == 'true' %}test is true{% endif %}"
```
would incorrectly render `mise test --test` as just `echo ` because `'MISE_TASK_ARG:test:MISE_TASK_ARG' != 'true'`.

### Changes
Since we need to parse the usage spec before rendering the full Tera template, this PR adds the second rendering after the usage spec is processed. This ensures that expressions with nest like this are evaluated correctly.

```toml
[tasks.test]
run = "echo {% if flag(name=env.FLAG_NAME) == 'true' %}TRUE{% endif %}"
```

Re-rendering the template might introduce some performance overhead, but I believe it's the simplest solution for this.
Of course, templates with more nests cannot be parsed correctly, but it is impossible if we are using usage to parse args in the template.

~~I still need to fix this PR to avoid code duplication and handle escapes, but I’d like to hear your thoughts on this approach.~~

---

Apologies for creating a PR directly. I was debugging locally and wanted to share the idea.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -697,7 +697,7 @@ version = "3.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fde0e0ec90c9dfb3b4b1a0891a7dcd0e2bffde2f7efed5fe7c9bb00e5bfb915e"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


risu729:
Depending on https://github.com/jdx/usage/pull/252, as `usage::parse::ParseValue` was not cloneable.

PR #4609, opened 2025-03-11
fix(uv): avoid deadlocks while initializing UV_VENV

### LINKED ISSUE #4600: `npm` backend hangs when used with `python.uv_venv_auto = true`
When `settings.python.uv_venv_auto` is set to `true` installing tools with the npm backend hangs indefinitely.

This was initially reported in https://github.com/jdx/mise/discussions/4286 and @jdx requested an issue with minimal reproduction steps.

```
❯ mise use -g --verbose npm:prettier
DEBUG ARGS: /opt/homebrew/bin/mise use -g --verbose npm:prettier
DEBUG install_some_versions: npm:prettier@latest
DEBUG $ npm view prettier dist-tags --json
```

### Steps to Reproduce

Without the `python.uv_venv_auto` setting the following succeeds:

```
mise uninstall --all
mise cache clear
mise use -g node
mise use -g npm:prettier
```

With the `python.uv_venv_auto` setting the following hangs:

```
# reset cache & set settings
mise uninstall --all
mise cache clear

mise config set settings.python.uv_venv_auto true
mise use -g --verbose npm:prettier
# hangs
```

risu729:
Fixes #4600.

`UV_VENV` initialization somtimes cause a deadlock in some cases because it requires loading `path` or `env` to get the toolset.

This PR avoids the deadlock by using `Mutex` and returning `None` while initializing itself.

### DIFF
--- a/src/uv.rs
+++ b/src/uv.rs
@@ -6,7 +6,7 @@ use crate::{dirs, file};
 use eyre::Result;
 use std::collections::HashMap;
 use std::path::PathBuf;
-use std::sync::LazyLock as Lazy;
+use std::sync::{LazyLock as Lazy, Mutex};
 
 #[derive(Debug)]
 pub struct Venv {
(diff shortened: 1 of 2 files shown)


src/uv.rs
```diff
@@ -6,27 +6,27 @@ use crate::{dirs, file};
...
+use std::sync::{LazyLock as Lazy, Mutex};
 
 #[derive(Debug)]
 pub struct Venv {
     pub venv_path: PathBuf,
     pub env: HashMap<String, String>,
 }
 
-pub static UV_VENV: Lazy<Option<Venv>> = Lazy::new(|| {
+pub static UV_VENV: Lazy<Mutex<Option<Venv>>> = Lazy::new(|| {
```
jdx, line 19: This is a good fix but would you mind adding a comment here explaining the rationale?

risu729: Sorry for missing the comment, I added it!

risu729:
I'm sorry for my mistake. I assumed my fix worked, but it actually does not. This PR does not resolve the deadlock.

I misunderstood the root cause, a deadlock occurring during initialization, so using `Mutex` inside of `LazyLock` does not fix the issue.

I will continue working on this.

PR #4635, opened 2025-03-14
feat(registry): add sonarscanner

thedavidharris:


thedavidharris:
@jdx anything in particular that should change with this? I see that I can add a test for this, but was unsure if it was rejected for other reasons (i.e. would rather the registry only take things in aqua, etc. going forward)

risu729:
I think new tools with asdf/vfox backend are basically not accepted.
It's mentioned in the [docs](https://mise.jdx.dev/plugins.html#:~:text=Tool%20plugins%20should%20be%20avoided%20for%20security%20reasons.%20New%20tools%20will%20not%20be%20accepted%20into%20mise%20built%20with%20asdf/vfox%20plugins%20unless%20they%20are%20very%20popular%20and%20aqua/ubi%20is%20not%20an%20option%20for%20some%20reason.), but might be hard to find.

PR #4637, opened 2025-03-14
[WORK IN PROGRESS] feat(core:zls) Add zls:core plugin

tamadamas:
The idea is to simplify the Zig developer life enabling `zls version` to follow `zig version`

Currently if I do like that, it fails
```
mise use -g zig@ref:master //good
mise use -g zls@ref:master //fail
```

```
mise use -g zig@0.14.0-dev.2577+271452d22 //mach latest -> good
mise use -g zls@0.14.0-dev.2577+271452d22 // fail
```

But we can fix it with official releases zigtools worker explained [here](https://github.com/zigtools/release-worker?tab=readme-ov-file#v1zlsselect-version)


tamadamas:
I have a problem because mise don't exec `core:zls` but instead uses `aqua registry`. 
I can't understand why? I've changed `aqua` to `core` in `registry.toml`.

I've tried to disable `aqua` but it fails.

Could someone help me?

```
MISE_DISABLE_BACKENDS=aqua cargo run use zls@ref:zig
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.74s
     Running `target/debug/mise use 'zls@ref:zig'`
Error:
   0: failed to install aqua:zigtools/zls@ref:zig
   1: HTTP status client error (404 Not Found) for url (https://api.github.com/repos/zigtools/zls/releases/tags/ref:zig)
   2: HTTP status client error (404 Not Found) for url (https://api.github.com/repos/zigtools/zls/releases/tags/vref:zig)
```

tamadamas:
Another solution I've thought about adding `zig` backend as `go` and `rust` but official tool doesn't seem supporting this yet. We need another cargo.
Or just do some fixing `core:zig` plugin like

```
mise use -g zig@ref:master --with-zls
```

jdx:
zls isn't something I would consider adding as a core tool

tamadamas:
@jdx so, could I create `Zig` backend and implement it there?

I don't want to contribute to `aqua` because it uses `go` and anyway `mise` is my tool of choice. Could you guide me through the right way to implement it?

jdx:
it should be in aqua

tamadamas:
aqua? Ok, I would check the way, thanks

PR #4643, opened 2025-03-15
feat(registry): use ubi for glab

scop:


registry.toml
```diff
@@ -753,7 +753,8 @@ gitsign.test = ["gitsign --version", "gitsign version v{{version}}"]
 gitu.backends = ["ubi:altsem/gitu", "cargo:gitu"]
 gitui.backends = ["aqua:extrawurst/gitui", "asdf:looztra/asdf-gitui"]
 gitversion.backends = ["aqua:gittools/gitversion", "ubi:gittools/gitversion"]
-glab.backends = ["asdf:mise-plugins/mise-glab"]
+glab.backends = ["ubi:gitlab-org/cli[forge=gitlab,exe=glab]", "asdf:mise-plugins/mise-glab"]
```
scop: In this form, requires https://github.com/jdx/mise/pull/4642

Without it, could do `"ubi:https://gitlab.com/gitlab-org/cli[exe=glab]"` but I'd hope we could make use of the forge option.

PR #4648, opened 2025-03-17
chore: Update to the latest version of ubi

autarch:


autarch:
This version has a bunch of fixes since 0.4. Also, I figured I should do this since 0.6.0 deprecates a method that `mise` was using.

autarch:
I'm not sure if "chore" is the right prefix.

jdx:
@autarch any reason this is still in draft state?

autarch:
> @autarch any reason this is still in draft state?

Oops, forgot to click the button.

jdx:
I had the same issue, try the latest version of zip now

autarch:
Ok, I updated this to `ubi` 0.6.1, which fixes the `zip` issue, and I rebased it off the latest `main`.

PR #4669, opened 2025-03-19
chore(registry): declare copier

looztra:
# Description

- chore(registry): declare copier with pipx/uvx backend and asdf plugin

## Note

- if there is a way to test this kind of PR locally, I'd be happy to do it (did not find something related to the registry PRs in the doc)


jdx:
pipx tools don't get included in the registry

jdx:
saying that though... it's a weird arbitrary situation and probably doesn't make sense to continue

PR #4704, opened 2025-03-25
feat: Nix backend.

vic:
This PR implements a nix-backend, allowing mise to install any of the [>120_000 packages](https://search.nixos.org/packages) from [nixpkgs](https://github.com/nixos/nixpkgs). 

Of course, the only requirement is to be running on a system where the `nix` executable is available.

This backend uses [NixHub API](https://www.nixhub.io/) in order to search for packages versions and determine the corresponding nixpkgs revision used to install it.

Tested locally and verified everything is installed correctly.

```
mise use nix:ruby@latest
```

If you prefer, I'd add a github action that exercises this feature by installing nix and use this backend to install something.


jdx:
Unfortunately this isn't something I'd be willing to maintain. This is something I would like to allow for with plugins though. I know @joshbode had worked on a nix plugin but I think there is more that could be done there.

ejrichards:
https://github.com/jdx/mise/discussions/3709

The nix as a backend would be great, as it would allow using the nix store / nixpkgs without having to write a flake, talked about this a bit here: https://github.com/jdx/mise/discussions/3538#discussioncomment-12363091

DrSensor:
I use just mise.toml or both mise.toml + flake.nix on my system depend if the project require system libraries or not (e.g libjpegturbo)

Would love it if mise has nix backend too so there is no duplicated toolchains that being cached on my machine.

jbadeau:
Try
https://github.com/jbadeau/mise-nix

based on the new backend plugin APIs

PR #4716, opened 2025-03-26
refactor(registry): use ubi for haxe

scop:


scop:
IDK, this may have been a bit hasty. Not a haxe user myself, just checked that it extracts to a bin/ dir and thought it's a single exe install only. I suggest reverting this one.

pavelzherdin:
Haxe was weird even before, with asdf. I don't know many other people who use it. I described one way to use it with mise [here](https://github.com/jdx/mise/discussions/4982). 

This adjustment to the registry will make it work on Linux at least:
```toml
# definitely need extract, since haxe bundles stdlib as a folder next to the binaries.
haxe.backends = ["ubi:HaxeFoundation/haxe[extract_all=true]", "asdf:mise-plugins/mise-haxe"]
haxe.test = ["haxe --version", "{{version}}"]
# Not sure about that part. Most likely not needed on Windows
haxe.depends = ["neko"]
# ...
# extract gives libneko.so.2, which is required by haxelib. But haxelib won't find it without some effort.
neko.backends = ["ubi:HaxeFoundation/neko[extract_all=true]", "asdf:asdf-community/asdf-neko"]
```

Not `haxelib`, though. That one doesn't have rpath, and requires libneko.so.2 somewhere. Hence, the LD_LIBRARY_PATH juggling. Ubuntu sets neko as haxe dependencies, and they bundle neko and libneko separately.

PR #4765, opened 2025-04-01
feat(ubi): add support for self-hosted GitHub/GitLab

### LINKED ISSUE #4749: Ubi support of GitLab
### Discussed in https://github.com/jdx/mise/discussions/4020

<sup>Originally posted by **q-b** January  9, 2025</sup>
Hi,

We would like to see `mise` support for GitLab as a forge. It has been added to `ubi` [v0.3.0](https://github.com/houseabsolute/ubi/releases/tag/v0.3.0) which is the version currently included in `mise`.

Our use case would be to retrieve some softwares from a private self-hosted repository (*not* gitlab dot com).

Thank you for help.</div>

roele:
Fixes #4749


docs/dev-tools/backends/ubi.md
```diff
@@ -69,6 +69,26 @@ then this will be ignored.
 "ubi:BurntSushi/ripgrep" = { version = "latest", matching = "musl" }
 ```
 
+### `forge`
```
jdx: maybe "provider"? not really happy with this name

src/env.rs
```diff
@@ -224,6 +224,42 @@ pub static GITHUB_TOKEN: Lazy<Option<String>> = Lazy::new(|| {
...
     token
 });
+pub static MISE_GITHUB_ENTERPRISE_TOKEN: Lazy<Option<String>> = Lazy::new(|| {
+    let token = var("MISE_GITHUB_ENTERPRISE_TOKEN")
+        .ok()
+        .and_then(|v| if v.is_empty() { None } else { Some(v) });
+    if let Some(token) = token.as_ref() {
+        set_var("MISE_GITHUB_ENTERPRISE_TOKEN", token);
+    } else {
+        remove_var("MISE_GITHUB_ENTERPRISE_TOKEN");
```
jdx: why are we removing?

src/env.rs
```diff
@@ -224,6 +224,42 @@ pub static GITHUB_TOKEN: Lazy<Option<String>> = Lazy::new(|| {
...
+pub static MISE_GITLAB_ENTERPRISE_TOKEN: Lazy<Option<String>> = Lazy::new(|| {
+    let token = var("MISE_GITLAB_ENTERPRISE_TOKEN")
+        .ok()
+        .and_then(|v| if v.is_empty() { None } else { Some(v) });
+    if let Some(token) = token.as_ref() {
+        set_var("MISE_GITLAB_ENTERPRISE_TOKEN", token);
+    } else {
+        remove_var("MISE_GITLAB_ENTERPRISE_TOKEN");
+    }
+    token
```
jdx: I think you could simplify this with a big match:

```suggestion
    match var("MISE_GITLAB_ENTERPRISE_TOKEN") {
      Ok("") | None => {
        remove_var()
        None
      },
      Ok(v) => {
        set_var();
        Some(v)
      }
    }
```

roele:
I just gave this a spin this morning with a test binary (`ubi:gitlab-org/cli`) and just adding the UBI forge support is not enough since `mise` fetches release versions to do the outdated check. So we also need an implementation of the GitLab API to fetch releases/tags and incorporate this into `_list_remote_versions` implementation.

src/env.rs
roele: not exactly sure why this is necessary, i copied this behaviour from the `MISE_GITHUB_TOKEN` which mentions the ubi use-case.

roele:
@jdx Maybe you can shed some light on the concept of options. There is a `BackendArg.opts()` and a `ToolRequest.options()`. They seem to be the same thing but are not synced.

If i add the following example to my `mise.toml`
```
"ubi:gitlab-org/cli" = { version = "1.54.0", exe = "glab", provider = "gitlab" }
```

i get the error
```
cd mise
mise ERROR [~/Downloads/mise/mise.toml] ubi:gitlab-org/cli@1.54.0: HTTP status client error (404 Not Found) for url (https://api.github.com/repos/gitlab-org/cli/releases)
mise ERROR error resolving versions
```
this is due to the fact that `BackendArgs.opts()` used when listing remote versions is empty, it would expect options in the backend string like `ubi:gitlab-org/cli[exe=glab,provider=gitlab]`


Currently to get this working without errors one might need to define the following which IMO is redundant.
```
"ubi:gitlab-org/cli[exe=glab,provider=gitlab]" = { version = "1.54.0", exe = "glab", provider = "gitlab" }
```

jdx:
yeah I think what we should do is probably have tool options include all of the backend opts

roele:
The problem i currently face is the other way around. Under some circumstances (uncached tool) the following code block is called, where `BackendArgs.opts()` is empty and i don't have access to the `ToolRequest` which contains the options for the provider and API URL.

https://github.com/jdx/mise/blob/370214def7429382066cb8ebcd517d506ebe4db4/src/backend/ubi.rs#L41

jdx:
I see. I don't think it would be possible to get the tool options at that point since mise doesn't necessarily have a specific tool when finding the versions. Maybe when we parse the toml we could include the tool options on the backend arg?

roele:
Turned out to be more work than anticipated but it should be working by now. Had to extend `http.rs` slightly since Github authentication was hardcoded. Only thing i could not test/verify properly is the token handling for the self-hosted URLs.

docs/dev-tools/backends/ubi.md
```diff
@@ -69,6 +69,26 @@ then this will be ignored.
...
 
+### `provider`
+
+Set the provider type to use for fetching assets and release information. Either `github` or `gitlab` (default is `github`).
+Ensure the `provider` is set to the correct type if you use `api_url` as the type probably cannot be derived correctly
+from the URL.
+
+```toml
+[tools]
+"ubi:gitlab-org/cli" = { version = "latest", exe = "glab", provider = "gitlab" }
```
jdx, line 80: do you happen to know of a tool other than the gitlab cli? I feel like it would make a clearer example if the CLI wasn't the same as the provider

src/toolset/tool_request.rs
```diff
@@ -107,11 +107,14 @@ impl ToolRequest {
         })
     }
     pub fn new_opts(
-        backend: BackendArg,
+        mut backend: BackendArg,
         s: &str,
         options: ToolVersionOptions,
         source: ToolSource,
     ) -> eyre::Result<Self> {
+        if backend.opts().is_empty() {
```
jdx: I feel like we should merge this using `options` on top of what was defined on the backend

src/toolset/tool_request.rs
```diff
@@ -107,11 +107,14 @@ impl ToolRequest {
         })
     }
     pub fn new_opts(
-        backend: BackendArg,
+        mut backend: BackendArg,
```
jdx: I haven't dove deep but this concerns me a bit since I know we cache these BackendArgs in a lot of places so this might result in weird behavior where we have BackendArgs with/without options

I'm not going to block on this, but figured it was worth mentioning

src/http.rs
```diff
@@ -143,13 +163,38 @@ impl Client {
         Ok((json, headers))
     }
 
+    pub fn json_headers_with_headers<T, U: IntoUrl>(
```
jdx, line 166: methods like this make me think we should be using builders but I don't think it's worth refactoring at this point

src/backend/ubi.rs
```diff
@@ -247,3 +304,37 @@ impl UbiBackend {
 fn name_is_url(n: &str) -> bool {
     n.starts_with("http")
 }
+
+fn set_token<'a>(mut builder: UbiBuilder<'a>, forge: &ForgeType) -> UbiBuilder<'a> {
+    match forge {
+        ForgeType::GitHub => {
```
jdx, line 310: was this ForgeType some existing enum?

roele: yes this comes from ubi (`use ubi::{ForgeType, UbiBuilder}`)

src/http.rs
roele: had similar thoughts (reminds me of naming patterns in Java 😂)

docs/dev-tools/backends/ubi.md
roele: unfortunately not, ~but i'll try to find another example...~ i'm unable to find another working example

src/toolset/tool_request.rs
roele: this way we can define a `ToolVersionOptions::merge(other: BTreeMap)` and can also get rid of the `mut backend`

PR #4788, opened 2025-04-04
feat(registry): add process-compose

evanleck:


evanleck:
I’m not sure why this failing, it’s a pretty straightforward addition to the registry. If anyone can point me in the right direction I’m happy to take a look.

PR #4804, opened 2025-04-09
Handle spaces in the windows path to mise

binyaminyblatt:
Moved the  path to mise.exe to a var for simpler script
Put the value in quotes so in cases that the path to it contains spaces it doesn't brake


binyaminyblatt:
Apologies I'm not a rust programmer I work in other languages so I'm just stumbling my way through making this work I'll get the test passed soon

jdx:
closing due to inactivity

PR #4249, opened 2025-01-28
fix: path option should take precedence over global configuration

roele:
Fixes #4244 and issue mentioned in #4211

PR #4817, opened 2025-04-10
fix: do not default to writing to mise.$MISE_ENV.toml

jdx:
I just hit a situation where I had set `MISE_ENV=macos` globally. This meant that `mise use` started writing to `mise.macos.toml` which I did not want.

I am not sure why I had this logic in the first place but it doesn't seem ideal. It seems to me if you specify `mise use -E macos` _that_ should write to `mise.macos.toml` but not if it's just an env var.

@roele I know you worked on this code a bit in https://github.com/jdx/mise/pull/4249 but I think it was more about making the `--path` flag work correctly. Any concerns about this change?

### DIFF
--- a/e2e/cli/test_use_env
+++ b/e2e/cli/test_use_env
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+
+mise use -e local dummy@1
+assert "cat mise.local.toml" '[tools]
+dummy = "1"'
+assert "rm mise.local.toml"
+
+MISE_ENV=local mise use dummy@1 # doesn't imply `-e local` for writing
+assert "cat mise.toml" '[tools]
+dummy = "1"'
--- a/src/cli/use.rs
+++ b/src/cli/use.rs
@@ -211,9 +211,6 @@ impl Use {
             } else {
                 cwd.join(format!("mise.{env}.toml"))
             }
-        } else if !env::MISE_ENV.is_empty() {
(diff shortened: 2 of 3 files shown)


roele:
Writing to `mise.{env}.toml` should be explicit via `--env` flag, so removing MISE_ENV makes sense. No objections from my side.

php-coder:
I've reported a similar issue (https://github.com/jdx/mise/discussions/4631) but with env-specific file is present. Does this PR fixes that case? Otherwise, perhaps, we should fix it as well.

php-coder:
BTW it seems like we also need to update `--help` output:

```console
$ mise use --help | head -15                                                                                                                                                                  11:40
Installs a tool and adds the version to mise.toml.

This will install the tool version if it is not already installed.
By default, this will use a `mise.toml` file in the current directory.

In the following order:
  - If `MISE_DEFAULT_CONFIG_FILENAME` is set, it will use that instead.
  - If `MISE_OVERRIDE_CONFIG_FILENAMES` is set, it will the first from that list.
  - If `MISE_ENV` is set, it will use a `mise.<env>.toml` instead.
  - Otherwise just "mise.toml"

Use the `--global` flag to use the global config file instead.

Usage: mise use [OPTIONS] [TOOL@VERSION]...
```

Note to the line `If `MISE_ENV` is set, it will use a `mise.<env>.toml` instead.`

PR #4847, opened 2025-04-15
fix: removed possible single-point-of-failure while running `mise upgrade` 

hitblast:
Key changes are as such:

`src/cli/upgrade.rs`
* Modified the upgrade process to update configuration files only for tools that were successfully installed.
* Changed the logic to uninstall old versions of tools only if their new versions were successfully upgraded.
* Adjusted the rebuild of shims and runtime symlinks to occur only for successful upgrades.

### DIFF
--- a/src/cli/upgrade.rs
+++ b/src/cli/upgrade.rs
@@ -178,9 +178,7 @@ impl Upgrade {
                 if let Err(e) =
                     cf.replace_versions(o.tool_request.ba(), vec![o.tool_request.clone()])
                 {
-                    had_errors = true;
-                    warn!("Failed to update config for {}: {}", o.name, e);
-                    continue;
+                    return Err(eyre!("Failed to update config for {}: {}", o.name, e));
                 }
 
                 if let Err(e) = cf.save() {
--- a/src/cli/upgrade.rs
+++ b/src/cli/upgrade.rs
@@ -136,6 +136,7 @@ impl Upgrade {
             }
             return Ok(());
         }
+
         let opts = InstallOptions {
             // TODO: can we remove this without breaking e2e/cli/test_upgrade? it may be causing tools to re-install
             force: true,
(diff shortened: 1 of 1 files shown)


src/cli/upgrade.rs
```diff
@@ -147,26 +148,73 @@ impl Upgrade {
...
-        if versions.iter().any(|v| v.short() == "python") {
+        if successful_versions.iter().any(|v| v.short() == "python") {
             PIPXBackend::reinstall_all().unwrap_or_else(|err| {
                 warn!("failed to reinstall pipx tools: {err}");
-            })
+            });
         }
+
+        if had_errors {
+            warn!("Some tools failed to upgrade. Check the logs for details.");
```
jdx: what logs?

src/cli/upgrade.rs
```diff
@@ -147,26 +148,73 @@ impl Upgrade {
...
 
-        if versions.iter().any(|v| v.short() == "python") {
+        if successful_versions.iter().any(|v| v.short() == "python") {
             PIPXBackend::reinstall_all().unwrap_or_else(|err| {
                 warn!("failed to reinstall pipx tools: {err}");
-            })
+            });
         }
+
+        if had_errors {
```
jdx, line 211: this should exit nonzero

src/cli/upgrade.rs
```diff
@@ -147,26 +148,73 @@ impl Upgrade {
...
+                let pr = mpr.add(&format!("uninstall {}@{}", o.name, tv));
+                if let Err(e) = self.uninstall_old_version(&o.tool_version, &pr) {
+                    warn!("Failed to uninstall old version of {}: {}", o.name, e);
+                }
+            }
         }
 
-        config::rebuild_shims_and_runtime_symlinks(&versions)?;
+        if let Err(e) = config::rebuild_shims_and_runtime_symlinks(&successful_versions) {
+            warn!("Failed to rebuild shims and runtime symlinks: {}", e);
```
jdx: I think we should output errors here with `:?` so it shows traceback

src/cli/upgrade.rs
```diff
@@ -147,26 +148,73 @@ impl Upgrade {
...
+            {
+                if let Err(e) =
+                    cf.replace_versions(o.tool_request.ba(), vec![o.tool_request.clone()])
+                {
+                    warn!("Failed to update config for {}: {}", o.name, e);
+                    continue;
+                }
+
+                if let Err(e) = cf.save() {
+                    warn!("Failed to save config for {}: {}", o.name, e);
```
jdx: this should return nonzero

hitblast: changed with latest commit

src/cli/upgrade.rs
hitblast: done!

src/cli/upgrade.rs
hitblast: pushed.

src/cli/upgrade.rs
hitblast: Removed the "logs part" and exited nonzero, thanks!

src/cli/upgrade.rs
```diff
@@ -147,26 +148,73 @@ impl Upgrade {
...
+                let pr = mpr.add(&format!("uninstall {}@{}", o.name, tv));
+                if let Err(e) = self.uninstall_old_version(&o.tool_version, &pr) {
+                    warn!("Failed to uninstall old version of {}: {}", o.name, e);
+                }
+            }
         }
 
-        config::rebuild_shims_and_runtime_symlinks(&versions)?;
+        if let Err(e) = config::rebuild_shims_and_runtime_symlinks(&successful_versions) {
+            warn!("Failed to rebuild shims and runtime symlinks: {:?}", e);
```
jdx: this should remain as a normal error I think

src/cli/upgrade.rs
```diff
@@ -147,26 +148,73 @@ impl Upgrade {
...
-            cf.replace_versions(o.tool_request.ba(), vec![o.tool_request.clone()])?;
-            cf.save()?;
+            if successful_versions
+                .iter()
+                .any(|v| v.ba() == o.tool_version.ba())
+            {
+                if let Err(e) =
+                    cf.replace_versions(o.tool_request.ba(), vec![o.tool_request.clone()])
+                {
+                    warn!("Failed to update config for {}: {}", o.name, e);
```
jdx: this should also return an error

src/cli/upgrade.rs
jdx: just leave this how it was before with `?`

hitblast: Okay, reverted.

src/cli/upgrade.rs
hitblast: Okay turns out GitHub glitched and I marked the wrong place. Resolving now.

hitblast: @jdx Resolved the issues related to the concurrent conversations. Thanks for the review!

PR #4850, opened 2025-04-15
fix: allow to install non-numeric elixir versions

roele:
As per discussion #4375


src/plugins/core/elixir.rs
```diff
@@ -11,9 +11,13 @@ use crate::ui::progress_report::SingleReport;
 use crate::{file, plugins};
 use eyre::Result;
 use itertools::Itertools;
+use once_cell::sync::Lazy;
+use regex::Regex;
 use versions::Versioning;
 use xx::regex;
 
+static NUMERIC_VERSION_REGEX: Lazy<&Regex> = Lazy::new(|| regex!(r"^[0-9]"));
```
jdx: the `regex!` macro already is a static lazy so you can just put the macro on line 48 directly

roele: Checking the source would have shown that 🤦🏼‍♂️ Thanks!

Issue #4842, opened 2025-04-14
`mise use` adding tool options from backend


jdx:
I thought this was fixed but apparently not

```
❯ mise use opentofu@1.6.2
mise Installed executable into /Users/jdickey/.local/share/mise/installs/opentofu/1.6.2/tofu
mise opentofu@1.6.2  ✓ installed
mise mise.toml tools: opentofu@1.6.2
❯ cat mise.toml
[tools]
opentofu = { version = "1.6.2", exe = "tofu" }
```

roele:
What exactly is the expected behaviour? 

Currently the default backend for `opentofu` is `aqua` which does not define/require any options.
```
mise use opentofu@1.6.2
mise WARN  aqua:opentofu/opentofu@1.6.2 can be verified with cosign but cosign is not installed
mise opentofu@1.6.2 ✓ installed

cat mise.toml
[tools]
opentofu = "1.6.2"
```

jdx:
ok I was able to repro on a different machine (mise 2025.4.4):

<img width="912" alt="Image" src="https://github.com/user-attachments/assets/bc9cba6b-1ed5-4f9a-b4f5-951bb819ad56" />

Looks like for some reason `mise use` is porting over the backend options from ubi?

doctor:

```
version: 2025.4.4 macos-arm64 (2025-04-15)
activated: no
shims_on_path: no

build_info:
  Target: aarch64-apple-darwin
  Features: DEFAULT, NATIVE_TLS, SELF_UPDATE
  Built: Tue, 15 Apr 2025 01:53:51 +0000
  Rust Version: rustc 1.86.0 (05f9846f8 2025-03-31) (Homebrew)
  Profile: release

shell:
  /usr/local/bin/fish
  fish, version 4.0b1

dirs:
  cache: ~/.mise/cache
  config: ~/.config/mise
  data: ~/.mise
  shims: ~/.mise/shims
  state: ~/.local/state/mise

config_files:
  ~/.config/mise/config.toml
  /private/var/folders/03/nk908szn7b35qksv9m08v7mw0000gn/T/tmp.l6Jf8tuGYR/mise.toml

ignored_config_files: (none)

backends:
  aqua
  asdf
  cargo
  core
  dotnet
  gem
  go
  npm
  pipx
  spm
  ubi
  vfox

plugins:
  1password                     https://github.com/NeoHsu/asdf-1password-cli.git#f5d5aab
[...]

PR #4857, opened 2025-04-16
refactor: introduce BackendOptions besides ToolVersionOptions

roele:
This is an early draft to split the `ToolVersionOption` which is currently used in the `BackendArg` and `ToolRequest` into a separate `BackendOption`.

When issueing a `mise use ubi:gitlab-org/cli[exe=glab,provider=gitlab]` with backend argument options these are currently copied over to the `ToolRequest`.

```
[tools]
"ubi:gitlab-org/cli" = { version = "latest", exe = "glab", provider = "gitlab" }
```

The new behaviour treats these as separate option variants keeping the backend options in the tool name.

```
[tools]
"ubi:gitlab-org/cli[exe=glab,provider=gitlab]" = "latest"
```

### DIFF
--- a/e2e/backend/test_ubi
+++ b/e2e/backend/test_ubi
@@ -7,12 +7,12 @@ assert_contains "$MISE_DATA_DIR/shims/jc --version" "jc version:  1.25.3"
 
 cat <<EOF >mise.toml
 [tools]
-"ubi:cilium/cilium-cli" = { version = "latest", exe = "cilium" }
+"ubi:cilium/cilium-cli[exe=cilium]" = "latest"
 EOF
 # re-uses tool options
 mise use ubi:cilium/cilium-cli
 assert "cat mise.toml" '[tools]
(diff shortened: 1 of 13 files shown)


jdx:
I think this is a good solution, but like I mentioned yesterday, I generally don't think I like how we're embedding data into a string here. I haven't come up with any better ideas though.

(very) long term, I actually wonder if we can completely get rid of the whole multiple runtimes feature. It makes the code way more complex and it really is only there to support python. If we did that, then we wouldn't need to distinguish between backend/tool options.

roele:
At this point this PR is merely a playground for me to explore and gauge the impact of any changes in another direction. I generally agree on the sentiment that encoding options in the backend identifier is not ideal.

I think in the short term we should leave it as is and observe #4842.

roele:
I am going to close this PR here. After digging more around ToolVersionOptions and fixing some issues with UBI/Java as part of #4886 i believe this is not the direction we should head towards.

PR #4863, opened 2025-04-16
feat(go): support build tags

bamorim:
## What

Allow installing tools like golang migrate that require people to pass -tags to work with specific databases.

## Reasonale

I've been working on a go project and we managed to install all our tools with mise with the exception of golang migrate. The reason being that installing it to work in our setup required passing `-tags postgres`. I recently saw how `ubi` allow some specific tags for specifying the binary to extract and I realized I could do something similar for go. 

Hopefully you like it 😄

### DIFF
--- a/docs/dev-tools/backends/go.md
+++ b/docs/dev-tools/backends/go.md
@@ -28,3 +28,18 @@ $ mise use -g go:github.com/DarthSim/hivemind
 $ hivemind --help
 Hivemind version 1.1.0
 ```
+
+## Tool Options
+
+The following [tool-options](/dev-tools/#tool-options) are available for the `go` backend—these
+go in `[tools]` in `mise.toml`.
+
+### `tags``
+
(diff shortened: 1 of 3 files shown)


jdx:
looks good but needs to go into the docs too

bamorim:
@jdx added some docs based on what other backends have. 😄

risu729:
It seems the CI is failing because [`awsebcli`](https://github.com/aws/aws-elastic-beanstalk-cli) released a new version [`3.22`](https://github.com/aws/aws-elastic-beanstalk-cli/releases/tag/3.22), and it has a module issue.
https://github.com/aws/aws-elastic-beanstalk-cli/issues/541

bamorim:
I can fix in this PR by setting it to the previous version maybe?

bamorim:
@jdx finally made CI pass xD

Thanks for the patience.

PR #4886, opened 2025-04-20
refactor: provide ToolRequest for some backend operations

roele:
This is ~draft for~ a slight refactor to address issues with tool options for operations such as `ls-remote`, `outdated` or `upgrade` (see #4020)

This alters some Backend trait methods to take a `Option<ToolRequest>` argument which can be used for example in `list_remote_versions` to get the ToolOptions. Currently this happens via `BackendArg` (`self.ba().opts()`) which is not the correct approach since the backend might not have any tool options for a certain version. This also led to code in `mise_toml.rs` which copies the tool options to the `BackendArg` which now could be removed.

Additionally this partially solves another issues with tool options recently reported. Java EA releases cannot be discovered via `ls-remote`. With this fix this would be possible via `mise ls-remote java[release_type=ea]@openjdk-25` (versions host interferes here so an additional method allows to disable the usage)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3731,6 +3731,7 @@ dependencies = [
  "nix",
  "num_cpus",
  "number_prefix",
+ "numeric-sort",
  "once_cell",
  "openssl",
  "os-release",
@@ -4016,6 +4017,12 @@ name = "number_prefix"
 version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "830b246a0e5f20af87141b25c173cd1b609bd7779a4617d6ec582abaf90870f3"
+
+[[package]]
+name = "numeric-sort"
+version = "0.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1f218700a47ff14f06bb676b97e63a6fb2fd25db3bf05d6f369971188e303f7f"
 
 [[package]]
 name = "object"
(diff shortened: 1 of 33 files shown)


jdx:
so I need some more time to think about this. This doesn't fit my mental model of how things like list_remote_versions should work. I'm particularly concerned about how this may impact things like `ls-remote` and caching.

I'll get back to you this week with what I think we should do.

roele:
I also tried to make sense for the reason why BackendArg is having ToolVersionOptions. Most of the time these are not populated (empty) because the backend is built when this information is not available or via a mise.toml which would have it but requires us to copy this information over. In the end there seems to be no reliable way to populate these for the backend.

Listing remote versions and remote version cache are certainly an issue. 

How can we list versions hosted on GitLab unless we use `mise ls-remote ubi:gitlab-org/cli[provider=gitlab]`? This affects other commands such as `outdated` and `upgrade` both tied to `ls-remote` to get latest versions.

Since the remote version cache key (`BackendArg.full()`) does not contain endpoint information the GitLab versions will be cached and `mise ls-remote ubi:gitlab-org/cli` (which should reach out to GitHub) would return GitLab versions.

Maybe there should be a method (e.g. `get_remote_version_cache_key`) to overwrite the remote cache key for a backend so we can provide different caches for different endpoints.

roele:
Closing this as discussed and raising a new one with a less intrusive solution.

PR #4888, opened 2025-04-20
fix(doctor): redact gitlab/enterprise tokens

risu729:
Fixes https://github.com/jdx/mise/discussions/4880.


roele:
Something i missed when adding support for GitLab 😖 Thanks for that PR!

PR #4891, opened 2025-04-20
fix: allow interactive upgrade to select nothing

risu729:
Fixes #4869.

This change allows users to proceed without selecting any options, but it doesn’t introduce a dedicated cancel operation.

### DIFF
--- a/src/cli/upgrade.rs
+++ b/src/cli/upgrade.rs
@@ -227,8 +227,7 @@ impl Upgrade {
         ui::ctrlc::show_cursor_after_ctrl_c();
         let mut ms = demand::MultiSelect::new("mise upgrade")
             .description("Select tools to upgrade")
-            .filterable(true)
-            .min(1);
+            .filterable(true);
         for out in outdated {
             ms = ms.option(DemandOption::new(out.clone()));
         }


risu729:
The test `mise test-tool bazel-watcher` is failing because the new release, [`0.26.0`](https://github.com/bazelbuild/bazel-watcher/releases/tag/v0.26.0), doesn't have the assets.
Since the release workflow failed, I think we just need to wait for it to be resolved.
https://github.com/bazelbuild/bazel-watcher/actions/runs/14561912040

nickjj:
Thanks for putting this up. Do you see this being a phase 1 release with dedicated cancel operations coming in the future or are those options off the table?

risu729:
First of all, I'm just a contributor, not a maintainer, so I can't speak for any future decisions made.

That said, I personally think that "selecting nothing" makes sense, while "cancelling the operation" should be handled via `Ctrl + C`, as it's more intuitive.

Just a note, such features should be added to [`demand`](https://github.com/jdx/demand?tab=readme-ov-file#multiselect), another Rust crate by jdx.

src/cli/upgrade.rs
```diff
@@ -227,8 +227,7 @@ impl Upgrade {
         ui::ctrlc::show_cursor_after_ctrl_c();
         let mut ms = demand::MultiSelect::new("mise upgrade")
             .description("Select tools to upgrade")
```
nickjj, line 229: What do you think about changing the wording to: `"Select tools to upgrade or press CTRL + c to skip all upgrades"`

My thought process is there's lots of new folks using command line tools and without a mention of CTRL + c I could see people getting stuck or confused. I think with the above wording or something similar any additional cancel actions wouldn't be needed.

risu729: Thanks for the suggestion.
I think it's a good idea, but I believe the description is not intended to describe the usage like that.

PR #4896, opened 2025-04-21
feat: added packaging as devcontainer features

atty303:
I’m making it possible to install mise as a Dev Container Feature.

Like other Features, it only supports Debian-based distributions and does not support Alpine, which is one of the official base images.
I’ve created a CI test and confirmed it works on my own fork. https://github.com/atty303/mise/actions/runs/14567218141

As for releasing, I haven’t set up a CI workflow to publish to the `jdx` GHCR.
However, I’ve confirmed that releases can be triggered from the release-devcontainer.yaml workflow in my fork. https://github.com/atty303/mise/actions/runs/14567216571
Since I can’t test this on the `jdx` repository, could the maintainer set up the release process?
You should be able to just replace the sample namespace `atty303` with `jdx`.


atty303:
I see. It would have been clearer if it were officially supported, but we'll handle it within the community.

PR #4900, opened 2025-04-22
fix: fix deadlocks caused by uv_venv_auto

### LINKED ISSUE #4600: `npm` backend hangs when used with `python.uv_venv_auto = true`

risu729:
Fixes #4600, ~~but still in WIP.~~


risu729:
**Note:**

The root cause of this issue lies in `uv_venv_auto` attempting to resolve the path to `uv`, which introduces a cyclic dependency:

```
toolset → ($npm/go dep env) → config → (uv path) → toolset
```

This cycle leads to a deadlock in `LazyLock` or `OnceCell` due to reentrant initialisation, where an item tries to access itself during its own initialisation.

Ideally, the correct fix would involve resolving the reentrancy issue directly.
However, since that seems complex, I'm trying to workaround by bypassing `LazyLock` or `OnceCell` during their initialization phase.
It means returning `None` or default values instead of trying to acquire the lock, thus avoiding the deadlock.

risu729:
@jdx I apologise for taking so long to fix my PR.
I confirmed this fixes the deadlock for `npm` or `go` backends with `uv_venv_auto`.

Issue #4919, opened 2025-04-23
mise settings add idiomatic_version_file_enable_tools stores duplicates in config


jdx:
### Discussed in https://github.com/jdx/mise/discussions/4918

<div type='discussions-op-text'>

<sup>Originally posted by **shaicoleman** April 23, 2025</sup>
```
$ mise --version
2025.4.7 linux-x64 (2025-04-23)
$ mise settings add idiomatic_version_file_enable_tools ruby
$ mise settings add idiomatic_version_file_enable_tools ruby
$ cat ~/.config/mise/config.toml
[settings]
idiomatic_version_file_enable_tools = ["ruby", "ruby"]
```</div>

PR #4925, opened 2025-04-24
fix: duplicate entries when running `mise settings add idiomatic_version_file_enable_tools`

hitblast:
fixes: #4919


hitblast:
Sorry for the clutter, for some reason rust-analyzer failed to check some of the errors beforehand.

src/cli/settings/set.rs
```diff
@@ -65,21 +68,30 @@ pub fn set(mut key: &str, value: &str, add: bool, local: bool) -> Result<()> {
...
-            Some(Some(array)) if add => {
-                let mut array = array.clone();
-                array.extend(value.as_array().unwrap().iter().cloned());
-                array.into()
+        let new_value: toml_edit::Value = if add {
+            if let Some(existing_arr) = settings.get(key).and_then(|item| item.as_array()).cloned()
+            {
+                let mut arr = existing_arr;
+                // Dedupe by comparing to the original &str
+                if !arr.iter().any(|it| it.as_str() == Some(value)) {
```
jdx, line 89: this needs to use SettingsMeta to only dedupe if it's a Set type

also needs a test

hitblast: Hmm, interesting. Since there isn't an explicit Set type, I wonder if you'd prefer something like this:

```rust
let is_set_type = meta
        .map(|m| matches!(m.type_, SettingsType::ListString | SettingsType::ListPath))
        .unwrap_or(false);
```

This would later be used to dedupe if needed;

```rust
let new_value: toml_edit::Value = if add {
    if let Some(existing_arr) = settings.get(key).and_then(|item| item.as_array()).cloned()
    {
        let mut arr = existing_arr;
        // Only dedupe if it's a set type
        if !is_set_type || !arr.iter().any(|it| it.as_str() == Some(value)) {
            arr.push(value);
        }
        toml_edit::Value::Array(arr)
    } else {
        parsed_value.clone()
    }
} else {
    parsed_value.clone()
};
```

hitblast: I've also added a test entry and pushed this code. Hope you'll have a look.

src/cli/settings/set.rs
```diff
@@ -27,7 +27,12 @@ impl SettingsSet {
 }
 
 pub fn set(mut key: &str, value: &str, add: bool, local: bool) -> Result<()> {
-    let value = if let Some(meta) = SETTINGS_META.get(key) {
+    let meta = SETTINGS_META.get(key);
+    let is_set_type = meta
+        .map(|m| matches!(m.type_, SettingsType::ListString | SettingsType::ListPath))
```
jdx: these aren't necessarily set types

hitblast: @jdx Umm, I'm not really sure what to look into in this case. Looking at the definitions, I can only see:

```rust

pub enum SettingsType {
    Bool,
    String,
    Integer,
    Duration,
    Path,
    Url,
    ListString,
    ListPath,
}

pub struct SettingsMeta {
    // pub key: String,
    pub type_: SettingsType,
    pub description: &'static str,
}
```

hitblast: Looking at the e2e test, turns out the dedupe mechanism for listed values work as expected: [(coverage 4)](https://github.com/jdx/mise/actions/runs/14666843267/job/41163710826?pr=4925#step:9:144)

```
$ mise settings add disable_hints a
  [mise settings add disable_hints a]
  $ mise settings add disable_hints b
  [mise settings add disable_hints b]
  $ mise settings get disable_hints
  [mise settings get disable_hints] output is equal to '["a", "b"]'
  $ mise settings add disable_hints a
  [mise settings add disable_hints a]
  $ mise settings get disable_hints
  [mise settings get disable_hints] output is equal to '["a", "b"]'
```

I'd really appreciate if you clarify here a little.

hitblast:
Since I'm not really sure how to implement the Set type directly into the code, I've tried an alternate approach to first cross-reference `DocumentMut` and only dedupe values if the setting in question is an array.

PR #4943, opened 2025-04-26
fix: Always use env::MISE_BIN when calling mise from itself (#4141)

hverlin:
- Update two places where `mise` assumed that `mise` is on `PATH` (in `mise dr --json` and `mise watch`).
- Update documentation to indicate how one might run tasks in series and mention that using `{{mise_bin}}` might be a good idea in this case.
- Add tests

Related discussion: https://github.com/jdx/mise/discussions/4141

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -1,6 +1,7 @@
 use crate::cli::args::ToolArg;
 use crate::cli::run::TaskOutput;
 use crate::config::{Config, Settings};
+use crate::env::MISE_BIN;
 use crate::exit::exit;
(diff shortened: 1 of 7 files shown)


src/cli/mod.rs
```diff
@@ -334,6 +335,7 @@ impl Cli {
         }
 
         debug!("ARGS: {}", &args.join(" "));
+        debug!("MISE_BIN: {}", MISE_BIN.to_string_lossy().to_string());
```
jdx: this could be trace, I think it's too noisy to need it all the time

PR #4944, opened 2025-04-26
fix(unuse): allow unusing any version if version not specified

### LINKED ISSUE #4933: `unuse` does not recurse up to global
I ran `mise unuse java` but because there was a config file in my local directory it seems to not remove java from the place it's actually installed to.

```
❯ mise ls java
Tool  Version  Source                      Requested
java  24.0.1   ~/.config/mise/config.toml  24
❯ mise cfg
Path                        Tools
~/.config/mise/config.toml  java, npm:prettier, opentofu, shellcheck, terraform, watchexec,...
~/src/mise/mise.toml        actionlint, bun, cargo-binstall, cargo:cargo-edit, cargo:cargo-ins...
❯ mise unuse java
❯ mise ls java
Tool  Version  Source                      Requested
java  24.0.1   ~/.config/mise/config.toml  24
```

I think there may have been a reason for this behavior but I can't recall.

jdx:
Fixes #4933

### DIFF
--- a/src/cli/unuse.rs
+++ b/src/cli/unuse.rs
@@ -21,10 +21,6 @@ pub struct Unuse {
     /// Do not also prune the installed version
     #[clap(long)]
     no_prune: bool,
-
-    /// Remove tool from global config
-    #[clap(short, long)]
-    global: bool,
 }
 
 impl Unuse {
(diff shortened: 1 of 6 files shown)


src/cli/unuse.rs
```diff
@@ -21,10 +21,6 @@ pub struct Unuse {
     /// Do not also prune the installed version
     #[clap(long)]
     no_prune: bool,
-
-    /// Remove tool from global config
-    #[clap(short, long)]
-    global: bool,
```
jdx, line 27: this wasn't used and doesn't make sense I don't think

PR #4960, opened 2025-04-28
fix: store tool version opts in .mise.backend

roele:
Less intrusive solution as replacement for #4886. This stores the tool options used when installing to the `full` representation in `.mise.backend` which allows the `BackendArg` to properly parse and initialize the backend.

That solves the following use case which currently fails because tool options are not available when tools are initialized from install state. 

```
mise use ubi:gitlab-org/cli[provider=gitlab,exe=glab]@1.55.0
sed -i '' 's/version = "1.55.0"/version = "latest"/' mise.toml
mise cache clear
mise outdated
```


efussi:
fwiw, I tested `mise` built from this PR's branch and based on a quick smoke test it also solves the 404 error I was getting when doing a `mise up` with tools installed from GH Enterprise. I had to clean the existing tool directories from `~/.local/share/mise/installs` first, though.

roele:
Thanks for verifying @efussi! Yes since this forces to store the full tool string in `.mise.backend` a "new" installation of the tool is required.

PR #4970, opened 2025-04-29
fix(java): filter out JetBrains releases with features

roele:
Java version data for JetBrains is available for a couple months but contains various packages per os/arch which leads to mise picking the wrong ones at random. Packages with features debug, fastdebug, jcef, freetype etc. should be filtered out.

See #4969


bric3:
@roele FYI, for Jetbrains Plugin development the `jcef` variant **should** be used. I'm not sure about the other variants, some might be useful depending on the platform.

Wouldn't it be possible to express this as an option like this ?

```
mise use java\[variant=jcef\]@jetbrains-21.0.7b968.13
```

Anyway such `variant` option/dimension could be useful in other places like the early access, e.g. to distinguish loom, valhalla, lilliput, leyden, babylon, etc.

roele:
@bric3 The current JDK version data has a `features` property which for JetBrains has `jcef` and/or `fastdebug` options. Usually we expose some of the features as part of the vendor string, Liberica for example has `liberica-lite-21.0.7+9`. For JetBrains this would become `jetbrains-jcef-21.0.7b968.13`. Unfortunately the features property in the current version data is not always properly set.

There are plans to switch to another version source for Java which is more tailored to `mise` and should be a tiny bit more consistent in this regard.

bric3:
Nice do you have an issue/discussion to follow?

roele:
No but i will push a draft PR in the next few days with required changes. Will need to coordinate with Jeff about the switch-over and new URL etc.

The change will be mostly compatible with a few changes that can/could be considered breaking changes.
- removed AdoptOpenJDK
- use of proper checksums from vendor
- JetBrains versioning will use suffix for builds (e.g. 21.0.7-b968.13 instead of 21.0.7b968.13)
- few missing patch versions which are no longer available

bric3:
@roele Thank you!

PR #4990, opened 2025-05-02
chore: enable clearing screen for confirm and dialog

roele:
Set new `clear_screen` flag on confirm and dialog. 

Confirmation dialogs displayed in tasks might be overwritten by other tasks output running in parallel which leads to broken dialog renditions. Clearing the screen helps mitigate this.


jdx:
I think we should solve this differently. I think we should pause the output while any dialog or confirm is displayed. Do you think that would be difficult?

roele:
Agree that this is not a proper solution for the issue. Something in the direction of pausing the logger and buffering log statements from other threads while paused would be required. Not sure if all output by tasks is written the same way currently.

PR #4991, opened 2025-05-02
feat(registry): use aqua for taplo

risu729:


evanleck:
FYI this change breaks Taplo usage as an LSP in e.g. Helix. There is a "full" build of Taplo available in aqua as well https://github.com/aquaproj/aqua-registry/tree/main/pkgs/tamasfe/taplo/full that likely has feature parity with the ubi version. FWIW the error I see in Helix using this version is:

> the LSP is not part of this build, please consult the documentation about enabling the functionality

risu729:
Sorry I didn't notice it. I've opened a new PR to fix that. #5017

PR #5002, opened 2025-05-04
feat: add mise_env tera variable for templates

auxesis:
Allows the user to query in a template what configuration environment a task is running in:

```bash
cat <<EOF >mise.toml
[tasks.print]
run = '{% if mise_env %}echo {{mise_env}}{% endif %}'
EOF
mise run print # output:
MISE_ENV=env1 mise run print # output: env1
mise --env env1,env2 run print # output: env1,env2
```

This can be useful when you want to `mise set` an environment variable for the current configuration environment:

```toml
[tasks.test]
run = "mise set --file mise.{{ mise_env | split(pat=',') | last }}.yaml KEY={{arg(name='value')}}"
```

### DIFF
--- a/src/tera.rs
+++ b/src/tera.rs
@@ -22,7 +22,9 @@ pub static BASE_CONTEXT: Lazy<Context> = Lazy::new(|| {
     context.insert("env", &*env::PRISTINE_ENV);
     context.insert("mise_bin", &*env::MISE_BIN);
     context.insert("mise_pid", &*env::MISE_PID);
-    context.insert("mise_env", &*env::MISE_ENV);
+    if !(&*env::MISE_ENV).is_empty() {
+        context.insert("mise_env", &*env::MISE_ENV);
+    }
     if let Ok(dir) = env::current_dir() {
         context.insert("cwd", &dir);
     }
@@ -384,7 +386,7 @@ mod tests {
 
     #[test]
     fn test_mise_env() {
-        assert_eq!(render("{{mise_env}}"), "[]");
+        assert_eq!(render("{% if mise_env %}{{mise_env}}{% endif %}"), "");
     }
 
     #[test]
--- a/src/tera.rs
+++ b/src/tera.rs
@@ -22,6 +22,7 @@ pub static BASE_CONTEXT: Lazy<Context> = Lazy::new(|| {
     context.insert("env", &*env::PRISTINE_ENV);
     context.insert("mise_bin", &*env::MISE_BIN);
     context.insert("mise_pid", &*env::MISE_PID);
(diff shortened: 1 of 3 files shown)


auxesis:
Additionally, I'm seeking feedback on the behaviour if an environment isn't specified. 

The current implementation always sets `mise_env` in Tera, even if the user hasn't specified an environment. 

When a user hasn't specified an environment, the "undefined variables are considered falsy" Tera convention doesn't work as expected, because `mise_env` will be an empty string. This leads to all sorts of hoop jumping for users in templates to check the length of `mise_env`. 

I think it may be better to not set `mise_env` if the environment is not specified, but I'm keen to hear what others think.

src/tera.rs
```diff
@@ -22,6 +22,7 @@ pub static BASE_CONTEXT: Lazy<Context> = Lazy::new(|| {
     context.insert("env", &*env::PRISTINE_ENV);
     context.insert("mise_bin", &*env::MISE_BIN);
     context.insert("mise_pid", &*env::MISE_PID);
+    context.insert("mise_env", &*env::MISE_ENV.join(","));
```
jdx: I think in this case we probably can keep the array type in the template

src/tera.rs
```diff
@@ -22,6 +22,9 @@ pub static BASE_CONTEXT: Lazy<Context> = Lazy::new(|| {
     context.insert("env", &*env::PRISTINE_ENV);
     context.insert("mise_bin", &*env::MISE_BIN);
     context.insert("mise_pid", &*env::MISE_PID);
+    if !(&*env::MISE_ENV).is_empty() {
```
jdx: I think you need to get this from settings or something because if it's specified as a CLI argument this will miss that (I think, I could be wrong)

auxesis: I double checked and it looks like it works:

```
$ cat <<EOF >mise.toml
[tasks.print]
run = '{% if mise_env %}echo {{mise_env}}{% endif %}'
EOF
$ MISE_ENV=production mise run print
[print] $ echo [production]
[production]
$ mise --env production run print
[print] $ echo [production]
[production]
```

I believe this is because the [`environment` function in `src/env.rs`](https://github.com/jdx/mise/blob/e35fa20f00ed884eebb1361cbf2b368237e29aca/src/env.rs#L464-L488) reads from both CLI arguments and environment varables.

auxesis: For the sake of completeness, I have added some additional e2e tests (38eb2a689cf2c3856256812c3e4c6d29f6be096d) that cover the different ways the configuration environment can be set.

jdx: thanks for checking

PR #4996, opened 2025-05-03
fix(env): fix dotenv files cascading (fix #4688)

noirbizarre:
Hi 👋🏼 

I posted an issue I had in #4688 and never had a response, so I gave a try at fixing it.

The test case is simple and covers 2 broken use-cases:
- dotenv file referencing a variable declared in the `env` section
- dotenv file referencing a variable declared in a previous dotenv file

In both cases, the referenced variables were replaced by an empty string.

Given `dotenvy` doesn't allow providing an initial substitution map ([I created an issue for that](https://github.com/allan2/dotenvy/issues/149)), I used the only workaround I found as of today: concatenate the initial environment with the dotenv file to load.

PR #5010, opened 2025-05-06
fix(dotenv): properly escape values in generated dotenv

noirbizarre:
Fix the poor escaping from #4996.

I was trying to avoid adding a new dependency, `shell-escape`, but it is already included, so I just rely on it instead of the poor uncomplete `replace()`

Fix #5008


david-hamilton-glean:
Just tested on 2025.5.2 and confirmed this fixes the issue.  Thanks @noirbizarre.

PR #5021, opened 2025-05-07
fix: unuse command does not support env, global and path options

roele:
Provides `--env`, `--global` and `--path` options for the `unuse` command as discussed in #5016


jdx:
like I mentioned in the thread, I'm not sure we actually need this and I think because of the prune logic stuff it might be a lot harder than it appears to support this. IMO this would only be necessary if you wanted to remove a tool from an overridden config file which doesn't seem very valuable to me.

jdx:
though idk, I suppose there might be use-cases where you explicitly only want to modify the global config and just want to make sure you don't modify the local one

roele:
you're probably right, did not think of the prune logic...

i sometimes remove tools from my global config which i don't use any longer as global default. that's where i see a `-g` flag could be useful.

jdx:
I guess the prune stuff should be ok in theory. If you had node@10 in global and node@12 in local, removed 10 from the global config and ran prune it should remove 10 and keep 12

src/cli/unuse.rs
```diff
@@ -1,14 +1,28 @@
...
+use crate::env::MISE_GLOBAL_CONFIG_FILE;
 use crate::file::display_path;
+use crate::{config, env};
 use eyre::Result;
 use itertools::Itertools;
+use path_absolutize::Absolutize;
 
 /// Removes installed tool versions from mise.toml
 ///
+/// By default, this will use a `mise.toml` file in the current directory.
```
jdx: this isn't accurate, it uses whichever file had the tool defined, correct?

roele: Good catch!

PR #5052, opened 2025-05-13
fix(registry): update glab ubi provider

StingRayZA:
Updates the `glab` ubi backend from `forge` to `provider`

Brings this in line with the documented api in https://mise.jdx.dev/dev-tools/backends/ubi.html#provider

I didn't find any other instances of `forge` in the registry


StingRayZA:
Hey @jdx 

Would you mind taking a look at this when you get a moment, please?

Thanks in advance!

PR #5087, opened 2025-05-15
chore: sign source tarball

digital-wonderland:
Sign the source tarball during each release as discussed [here](https://github.com/jdx/mise/discussions/4998).

Please have a look. It's - obviously? - untested but should work.

If it works I would like to do the same for [usage](https://github.com/jdx/usage).


jdx:
test run: https://github.com/jdx/mise/actions/runs/15055757913

PR #5089, opened 2025-05-15
feat: use new Java metadata source

roele:
This is draft to use a new Java version data source. Its output is currently tailored to `mise` but could provide more extensive data (architectures, properties etc.) if need be. 

It is using [roele/roast](https://github.com/roele/roast), a rewrite of [jdx/rtx-java-metadata](https://github.com/jdx/rtx-java-metadata) respectively [joschi/java-metadata](https://github.com/joschi/java-metadata).

Improvements:
* contains only `mise` relevant data/properties
* more consistent versioning
* uses proper vendor provided checksums

Possible breaking changes:
* removed AdoptOpenJDK (has been replaced by Temurin in 2021)
* few patch versions are missing as they are not provided by the vendors anymore

Question is if we should fork/move the repository for `mise` purpose to somewhere else and if we should push the data to a domain that is more related to mise. @jdx WDYT?

/cc @bric3

### DIFF
--- a/src/plugins/core/java.rs
+++ b/src/plugins/core/java.rs
@@ -101,9 +101,9 @@ impl JavaPlugin {
         pr.set_message(format!("download {filename}"));
         HTTP.download_file(&m.url, &tarball_path, Some(pr))?;
 
-        if !tv.checksums.contains_key(filename) {
+        if !tv.checksums.contains_key(filename) && m.checksum.is_some() {
             tv.checksums
-                .insert(filename.to_string(), format!("sha256:{}", m.sha256));
+                .insert(filename.to_string(), m.checksum.as_ref().unwrap().clone());
         }
         self.verify_checksum(ctx, tv, &tarball_path)?;
 
@@ -257,7 +257,7 @@ impl JavaPlugin {
 
     fn download_java_metadata(&self, release_type: &str) -> Result<Vec<JavaMetadata>> {
         let url = format!(
-            "https://rtx-java-metadata.jdx.dev/metadata/{}/{}/{}.json",
+            "https://mise-java.jdx.dev/jvm/{}/{}/{}.json",
             release_type,
             os(),
             arch()
@@ -266,8 +266,6 @@ impl JavaPlugin {
         let metadata = HTTP_FETCH
             .json::<Vec<JavaMetadata>, _>(url)?
             .into_iter()
-            // TODO: remove once we switch over to the new metadata which has these filtered out already
-            .filter(jetbrains_with_features)
             .filter(|m| {
                 m.file_type
                     .as_ref()
@@ -301,7 +299,7 @@ impl Backend for JavaPlugin {
                     .image_type
                     .as_ref()
                     .is_some_and(|image_type| image_type == "jdk");
-                let features = 10 - m.features.len();
+                let features = 10 - m.features.as_ref().map_or(0, |f| f.len());
                 let version = Versioning::new(v);
                 (
                     is_shorthand,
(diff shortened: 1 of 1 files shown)


jdx:
> Question is if we should fork/move the repository for mise purpose to somewhere else and if we should push the data to a domain that is more related to mise. @jdx WDYT?

I think this is a question for you not for me. I'm fine with it either way. I think if you wanted to maintain it that would make sense in case you wanted people to use it outside of mise, though I would probably move it to something like `java.roele.dev`.

That said, I think mise users would feel slightly more comfortable coming across `mise-java.jdx.dev` or `java.jdx.dev` but I don't think that's a huge deal. If you want me to host it I'm happy to help set that up of course.

src/plugins/core/java.rs
```diff
@@ -520,8 +516,9 @@ impl Display for JavaMetadata {
 }
 
 // only care about these features
-static JAVA_FEATURES: Lazy<HashSet<String>> =
-    Lazy::new(|| HashSet::from(["musl", "javafx", "lite", "large_heap"].map(|s| s.to_string())));
+static JAVA_FEATURES: Lazy<HashSet<String>> = Lazy::new(|| {
+    HashSet::from(["crac", "javafx", "jcef", "leyden", "lite", "musl"].map(|s| s.to_string()))
+});
```
bric3, line 503: **suggestion:** Shouldn't it be configurable ?

I.e. OpenJDK sometimes propose new build for other/new projects, that may or may not be built on the latest JDK.

E.g. https://jdk.java.net/loom/, https://jdk.java.net/valhalla/, maybe in the future babylon (already available on shipilev builds), etc.

roele:
@jdx We should definitely have something in the direction of `mise-java.jdx.dev`. Let me know once you have setup the R2 bucket and necessary credentials and i'll adapt GH workflow and PR accordingly.

jdx:
do you want to move the repo to jdx/mise-java ? I think first we need to both be contributors on the project so I can add the GHA secrets

roele:
Yes i guess it makes sense to move it. At this point i don't see a need for a non-mise tailored versions API. Added you as collaborator.

jdx:
hmm I can't move the repo or modify the admin settings, I think you need to transfer ownership then I think I can add you as an owner back since I have github pro (though I may be wrong about the pro part)

jdx:
ok so I was wrong about being able to add you as an owner, but I made you a collaborator at least. I got it configured with everything: https://mise-java.jdx.dev

roele:
Awesome, thanks!

I assume we gotta run both variants for a while after we merge this in until most have upgraded to the version containing this change. WDYT? 3-6 months?

Just a side note, the AWS PostgreSQL database only runs for ~1h during the scheduled update window to save costs. So manual triggers of update would require to manually start up the database beforehand. If you think you will require access i can set that up.

src/plugins/core/java.rs
```diff
@@ -257,7 +257,7 @@ impl JavaPlugin {
 
     fn download_java_metadata(&self, release_type: &str) -> Result<Vec<JavaMetadata>> {
         let url = format!(
-            "https://rtx-java-metadata.jdx.dev/metadata/{}/{}/{}.json",
```
jdx, line 260: I think this is the last vestige of rtx

jdx:
> I assume we gotta run both variants for a while after we merge this in until most have upgraded to the version containing this change. WDYT? 3-6 months?

of maintaining rtx-java-metadata? that sounds good

> Just a side note, the AWS PostgreSQL database only runs for ~1h during the scheduled update window to save costs. So manual triggers of update would require to manually start up the database beforehand. If you think you will require access i can set that up.

nah don't need access now, I'll let you know if that changes

src/plugins/core/java.rs
roele: yeah you're probably right, haven't come across any other remains of that era

roele:
> nah don't need access now, I'll let you know if that changes. If I used aws for mise I would offer to host it there, how much is it costing you?

its only a couple $ a month so don't worry about that

jdx:
lmk if that changes, people do donate to me for mise so it would make sense to put those funds towards that—or I'd also be happy to settle up with you annually or something too

src/plugins/core/java.rs
```diff
@@ -257,7 +257,7 @@ impl JavaPlugin {
...
+            "https://mise-java.jdx.dev/jvm/{}/{}/{}.json",
```
bric3, line 260: Have you thought about versioning if there are other changes needed to that metadata ?

bric3 (commented):
Sorry I didn't post my comment in time.

src/plugins/core/java.rs
roele: This will require adaptions in the [crawler](https://github.com/jdx/mise-java) first, so making this configurable wont do anything if the data is not available. Also releases such as Loom or Valhalla are considered `EA` (Early Access) and are therefore only available through the EA release type. You can explore the version data at https://mise-java.jdx.dev.

There are discussions to make the endpoint URL configurable as well to allow enterprises to have more control over supply chain security but this would mean that enterprises would have to provide the version data themselves.

src/plugins/core/java.rs
roele: I'd like to keep this as simple as possible. Java versioning is already a pain since there are so many vendors with each one having their own way of doing things.

bric3: Sorry I wasn't clear I meant it as endpoint version, rather than changing the host.

src/plugins/core/java.rs
bric3: OK I understand, but leyden isn't it an EA as well.

src/plugins/core/java.rs
roele: yes, i assume in the direction of https://mise-java.jdx.dev/jvm/v1/{}/{}/{}.json. that would imply to handle different api versions in mise which will further complicate things.

src/plugins/core/java.rs
roele: Yes Leyden is EA as well and OpenJDK provides this as version suffix while for Liberica this is exposed as a feature unfortunately (inconsistencies everywhere).

bric3: Arg. That's annoying indeed !

src/plugins/core/java.rs
bric3: Versioning can be a pain, let's close the topic. Thanks for reaching back on my questions !

zeitlinger:
@roele this feature caused a regression: https://github.com/jdx/mise/discussions/5123

PR #5116, opened 2025-05-18
fix(registry): use aqua for delta

risu729:


risu729:
I think the test failed because of rate limit

PR #5124, opened 2025-05-19
fix: SSH remote tasks do not support organizations in repository path

roele:
Fixes the issue discussed in #5101.

Loosens the constraints around `repo` named capture to allow (sub-)folders (GitLab supports nested folders) for SSH and HTTPS.

### DIFF
--- a/mise.lock
+++ b/mise.lock
@@ -74,10 +74,6 @@ jq-macos-arm64 = "sha256:0bbe619e663e0de2c550be2fe0d240d076799d6f8a652b70fa04aea
 version = "0.44.0"
 backend = "npm:markdownlint-cli"
 
-[tools."npm:pin-github-action"]
-version = "3.3.1"
-backend = "npm:pin-github-action"
-
 [tools."npm:prettier"]
 version = "3.5.3"
 backend = "npm:prettier"
--- a/src/task/task_file_providers/remote_task_git.rs
+++ b/src/task/task_file_providers/remote_task_git.rs
@@ -81,7 +81,7 @@ impl RemoteTaskGit {
     }
 
     fn detect_ssh(&self, file: &str) -> Result<GitRepoStructure, Box<dyn std::error::Error>> {
-        let re = Regex::new(r"^git::(?P<url>ssh://((?P<user>[^@]+)@)(?P<host>[^/]+)/(?P<repo>[^/]+)\.git)//(?P<path>[^?]+)(\?ref=(?P<branch>[^?]+))?$").unwrap();
+        let re = Regex::new(r"^git::(?P<url>ssh://((?P<user>[^@]+)@)(?P<host>[^/]+)/(?P<repo>.+)\.git)//(?P<path>[^?]+)(\?ref=(?P<branch>[^?]+))?$").unwrap();
 
         if !re.is_match(file) {
             return Err("Invalid SSH URL".into());
@@ -99,7 +99,7 @@ impl RemoteTaskGit {
     }
 
     fn detect_https(&self, file: &str) -> Result<GitRepoStructure, Box<dyn std::error::Error>> {
-        let re = Regex::new(r"^git::(?P<url>https://(?P<host>[^/]+)/(?P<repo>[^/]+(?:/[^/]+)?)\.git)//(?P<path>[^?]+)(\?ref=(?P<branch>[^?]+))?$").unwrap();
+        let re = Regex::new(r"^git::(?P<url>https://(?P<host>[^/]+)/(?P<repo>.+)\.git)//(?P<path>[^?]+)(\?ref=(?P<branch>[^?]+))?$").unwrap();
 
         if !re.is_match(file) {
             return Err("Invalid HTTPS URL".into());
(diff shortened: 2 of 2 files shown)


acesyde:
Hello @roele,

You have modified the non regression tests can you ensure the paths like this still work ?

git::ssh://git@github.com:myorg/my-repo.git

acesyde:
And you must update the documentation https://mise.jdx.dev/tasks/toml-tasks.html#git

btpemercier:
And from my point of view it's a breaking change, so you must maybe support the both rules and remove the previous pattern in a second row

roele:
@acesyde 

This feature is still `experimental` so breaking changes can be expected.

The tests will still pass when reverting the regression test to use `git::ssh://git@github.com:myorg/example.git//myfile?ref=v1.0.0` but the `host` is not resolved properly [^1]. The `host` value will be `github.com:myorg`. I'm not sure what exactly the intention was with `:`. Usually this should separate the port number from the host. In my opinion this should be `/` or `:/` (assuming empty port uses default 22).

Regarding the documentation i agree that the example should be updated accordingly.

[^1]: https://rregex.dev/?version=1.11&method=find&regex=%5Egit%3A%3A%28%3FP%3Curl%3Essh%3A%2F%2F%28%28%3FP%3Cuser%3E%5B%5E%40%5D%2B%29%40%29%28%3FP%3Chost%3E%5B%5E%2F%5D%2B%29%2F%28%3FP%3Crepo%3E.%2B%29%5C.git%29%2F%2F%28%3FP%3Cpath%3E%5B%5E%3F%5D%2B%29%28%5C%3Fref%3D%28%3FP%3Cbranch%3E%5B%5E%3F%5D%2B%29%29%3F%24&text=git%3A%3Assh%3A%2F%2Fgit%40github.com%3Amyorg%2Fexample.git%2F%2Fmyfile%3Fref%3Dv1.0.0%0A

acesyde:
of course you can introduce it but you can avoid it too, from my point of view every time i develop something even it's something in beta i avoid any breaking change if i can because a feature can be used by a lot of people.

that's why i have introduced a lot of tests in this part to avoid this kind of situation.

and why the : because i have used the terraform module formalism from this documentation https://developer.hashicorp.com/terraform/language/modules/sources

roele:
i try to keep backwards compatibility too generally but the whole point of the experimental flag is to be able to address flaws in a sustainable way. users of experimental features should be aware of the risks.

🤔  "scp-like" address syntax [^1] seems to be a Terraform feature and since we use `ssh://` syntax my take would be to handle it as such.

> If you use the ssh:// URL scheme then Terraform will assume that the colon marks the beginning of a port number, rather than the beginning of the path. This matches how Git itself interprets these different forms, aside from the Terraform-specific git:: selector prefix.

[^1]: https://developer.hashicorp.com/terraform/language/modules/sources#scp-like-address-syntax

acesyde:
I agree with you, but I discussed version pinning with @jdx here https://github.com/jdx/mise/discussions/4929 and some installation methods don't support version pinning. In this case, it's difficult, even experimental, to avoid a breaking change.

You can keep your change like this, but keep in mind that "mise" isn't a small project used by one or two people, but can be used by large-scale companies. Any change must be taken into account, as it can have a negative impact on several levels.

roele:
Yes i agree. 

So if we keep as is, we should probably adapt the example in the documentation and IMO also all the tests that still use the "scp-like" syntax to avoid confusion later.

acesyde:
@jdx 

I agree with you on that point. What I'm trying to highlight is that an experimental feature doesn’t necessarily need to be broken or disrupted when introducing a change—unless, of course, there’s no other viable option.

jdx:
The entire point of having things marked as experimental is because we know the design isn't final and that we're reserving the right to be able to change it at any point so you should _expect_ disruption if you decide to adopt them.

acesyde:
I understand the rationale behind the "experimental" label and the flexibility it implies. That said, I believe the term doesn't exclude the responsibility to consider the fact that the feature is already being used by many. Breaking things purely for the sake of convenience or to save a few minutes of work doesn’t always feel justified. But I suppose this comes down to a matter of philosophy.

roele:
Just to give you my view. 

Git supports the SSH protocol which allows for `ssh://[user@]server/project.git` or the scp-like `[user@]server:project.git` whereas the later seems to clash with defining a custom port. I could not find any info on how the scp-like syntax would support custom ports besides `scp` itself using a flag or the SSH config. Maybe i'm missing something?

So at this point for me the proper fix was to follow the `ssh://` notation use a `/` to denote the path and introduce a breaking change for the sake following the notation type used.

Do you see a better solution?

btpemercier:
Yes, I think the best approach is to adopt the ssh:// notation with the / separator. It not only solves the port issue but also ensures compatibility with other providers like Bitbucket, GitLab, or self-hosted instances.

acesyde:
Yes, I think the best approach is to adopt the ssh:// notation with the / separator. It not only solves the port issue but also ensures compatibility with other providers like Bitbucket, GitLab, or self-hosted instances.

PR #5130, opened 2025-05-19
support for 'usage' in shebang of tasks

NorthIsUp:
Currently mise will throw an error if usage is in the shebang of a standalone script task, this fixes that.


jdx:
needs a test, I don't think this will work as is

PR #5146, opened 2025-05-21
fix(aqua): use complete_windows_ext

### LINKED ISSUE #5141: aqua: mise does not use `complete_windows_ext` field
the current logic in mise for appending `.exe` for aqua urls needs to be replaced with [`complete_windows_ext`](https://aquaproj.github.io/docs/reference/registry-config/complete-windows-ext/)

https://github.com/jdx/mise/blob/9549a7ccb3aa52e132ceacef9d76297dde7a3293/src/aqua/aqua_registry.rs#L362-L371

jdx:
Fixes #5141

### DIFF
--- a/src/aqua/aqua_registry.rs
+++ b/src/aqua/aqua_registry.rs
@@ -13,7 +13,7 @@ use regex::Regex;
 use serde_derive::Deserialize;
 use std::cmp::PartialEq;
 use std::collections::HashMap;
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 use std::sync::LazyLock as Lazy;
 use url::Url;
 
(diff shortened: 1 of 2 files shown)


abernix:
👋 Haven't looked deeply at the root cause yet, but after the v2025.5.9 release, our installation of `helm`, with aqua, on Windows started failing with:

```
mise ERROR failed to install aqua:helm/helm@3.18.0
mise ERROR HTTP status client error (404 Not Found) for url (https://get.helm.sh/helm-3.18.0-windows-amd64.tar.gz.exe)
mise ERROR HTTP status client error (404 Not Found) for url (https://get.helm.sh/helm-v3.18.0-windows-amd64.tar.gz.exe)
```

This PR's changelog for v2025.5.9 stood out to me as a likely contender (though I haven't bisected to specifically this commit)

Now, I suspect this could be a problem with that aqua pkg (https://github.com/aquaproj/aqua-registry/commits/main/pkgs/helm), but thought it was worth mentioning.

We pinned to v2025.5.8 while we look into it.

jdx:
hmm I thinkl the logic here needs to not use .exe if the format is not defined or is "raw"

jdx:
hmm I think the logic might be skipped in aqua if filenames end in .tar.gz. I'll see if I can find the logic in aqua so we can replicate it

PR #5163, opened 2025-05-23
Update ruby.md: clarify how to enable ideomatic version file reading

amkisko:
Add information from command line to documentation as it becomes more critical before changing default behavior.


docs/lang/ruby.md
```diff
@@ -49,6 +49,12 @@ Create a `.ruby-version` file for the current version of ruby:
 ruby -v > .ruby-version
 ```
 
+Enable ideomatic version file reading for ruby:
```
jdx: typo

jdx (changes requested):

PR #5183, opened 2025-05-25
fix: remote task dependency does not work

roele:
Addresses an issue discussed in #5176 where remote tasks which are declared in `depends` are not properly resolved.

### DIFF
--- a/src/cli/run.rs
+++ b/src/cli/run.rs
@@ -238,7 +238,7 @@ impl Run {
             .clone()
     }
 
-    fn parallelize_tasks(mut self, mut tasks: Vec<Task>) -> Result<()> {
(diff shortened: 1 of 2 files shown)


acesyde:
Works with wait for but I think you found why it not works with depends ❤️

src/cli/run.rs
```diff
@@ -238,8 +238,8 @@ impl Run {
             .clone()
     }
 
-    fn parallelize_tasks(mut self, mut tasks: Vec<Task>) -> Result<()> {
-        time!("parallelize_tasks start");
+    fn parallelize_tasks(mut self, tasks: Vec<Task>) -> Result<()> {
+        time!("paralellize_tasks start");
```
jdx: typo

PR #5216, opened 2025-05-28
task-configuration.md: typo

mustafa0x:
fix `running tasks this will _will_ have`


mustafa0x:
completely unrelated: now that mise has `task_config`, can we add `set -euo pipefail` to it?

PR #5218, opened 2025-05-28
fix: uncaught panic when directory passed via -C doesn't exist

### LINKED ISSUE #5215: Uncaught panic when directory passed via `-C` doesn't exist
### Discussed in https://github.com/jdx/mise/discussions/5211

<sup>Originally posted by **corneliusroemer** May 27, 2025</sup>
`mise exec ... -C "nonexistentdir"` panics when the dir does not exist due to `.unwrap()` on an `Err` value. This error case ideally be handled and reported more succinctly to the end user:

```
$ mise exec "go@1.21.5" -C "p2" -- go version
=== Async Backtrace (panic occurred in tokio runtime) ===
  0: Location { name: Some("mise::config::Config::load::{{closure}}"), rest: ("src/config/mod.rs", 98, 5) }
  1: Location { name: Some("mise::cli::exec::Exec::run::{{closure}}"), rest: ("src/cli/exec.rs", 55, 5) }
  2: Location { name: Some("mise::cli::Cli::run::{{closure}}"), rest: ("src/cli/mod.rs", 322, 5) }

------- TASK DUMP TREE -------
╼ mise::cli::Cli::run::{{closure}} at src/cli/mod.rs:322:5
  └╼ mise::cli::exec::Exec::run::{{closure}} at src/cli/exec.rs:55:5
     └╼ mise::config::Config::load::{{closure}} at src/config/mod.rs:98:5
=== End Async Backtrace ===

Message:  called `Result::unwrap()` on an `Err` value:
   0: failed to set current directory to ~/p2
   1: No such file or directory (os error 2)

Location:
   src/env.rs:565

Location: src/config/settings.rs:103

```</div>

### PR DESCRIPTION
**Pull request by roele**
Fixes #5215

### DIFF
--- a/src/config/settings.rs
+++ b/src/config/settings.rs
@@ -1,8 +1,8 @@
 use crate::cli::Cli;
 use crate::config::ALL_TOML_CONFIG_FILES;
-use crate::duration;
 use crate::file::FindUp;
 use crate::{dirs, env, file};
+use crate::{duration, exit};
 #[allow(unused_imports)]
 use confique::env::parse::{list_by_colon, list_by_comma};
 use confique::{Config, Partial};
@@ -100,7 +100,13 @@ pub struct SettingsFile {
 
 impl Settings {
     pub fn get() -> Arc<Self> {
-        Self::try_get().unwrap()
+        match Self::try_get() {
+            Ok(settings) => settings,
+            Err(e) => {
+                error!("{e}");
+                exit(1);
+            }
+        }
     }
     pub fn try_get() -> Result<Arc<Self>> {
         if let Some(settings) = BASE_SETTINGS.read().unwrap().as_ref() {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/config/settings.rs [hunk 1]
```diff
@@ -100,7 +100,13 @@ pub struct SettingsFile {
 
 impl Settings {
     pub fn get() -> Arc<Self> {
-        Self::try_get().unwrap()
+        match Self::try_get() {
+            Ok(settings) => settings,
+            Err(e) => {
+                error!("{e}");
+                exit(1);
```
jdx, line 107: we shouldn't skip the main error handling logic. What we need to do is remove `Config::get()` and replace all uses with `Config::try_get()?`

roele: already had a feeling that this fix is not the way to go...

jdx: I'm trying to see if AI can do this

PR #5219, opened 2025-05-28
feat(registry): add aqua backend for maven

ZeroAurora:
https://github.com/aquaproj/aqua-registry/blob/main/pkgs/apache/maven/pkg.yaml
Tested locally using alias:
图片

mise checks latest versions from GitHub, so `@latest` works well. Verified not working with versions `<3.0.0 || >= 4.0.0` (need aqua-side changes)


ZeroAurora:
also added a test

PR #5240, opened 2025-05-30
fix(zig): exclude mach version from version list

mangkoran:
User still able to install mach versions, but the mach versions wouldn't be shown in `mise ls-remote zig`, which seems it was the [behavior before](https://github.com/jdx/mise-versions/blob/c7d45df8539d67e742c16de73a143a23491e3a5a/docs/zig) (only shows zig versions from zig index). Mach index seems also not updated recently so I don't think there's significant mise users who use the mach versions.

Due to user still able to install mach versions, I think we could safely assume this as not a breaking change.

Workaround: #5232

### DIFF
--- a/docs/lang/zig.md
+++ b/docs/lang/zig.md
@@ -12,14 +12,25 @@ The code for this is inside the mise repository at
 The following installs zig and makes it the global default:
 
 ```sh
-mise use -g zig@0.13     # install zig 0.13.x
-mise use -g zig@latest  # install latest zig release
-mise use -g zig@ref:master # install latest nightly from master
-mise use -g zig@ref:mach-latest # install latest nominated zig
-mise use -g zig@0.14.0-dev.2577+271452d22 # install dev version
+mise use -g zig@0.14           # install zig 0.14.x
+mise use -g zig@latest         # install latest zig release
+mise use -g zig@master         # install latest nightly from master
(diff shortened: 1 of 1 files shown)


mangkoran:
Tested in local

```sh
08:40:01 ❯ cargo run --bin mise -- cache clean
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.38s
     Running `target/debug/mise cache clean`
mise cache cleared

08:40:35 ❯ MISE_USE_VERSIONS_HOST=false cargo run --bin mise -- ls-remote zig
   Compiling mise v2025.5.16 (/home/mangkoran/repos/fork/mise)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 32.56s
     Running `target/debug/mise ls-remote zig`
master
0.1.1
0.2.0
0.3.0
0.4.0
0.5.0
0.6.0
0.7.0
0.7.1
0.8.0
0.8.1
0.9.0
0.9.1
0.10.0
0.10.1
0.11.0
0.12.0
0.12.1
0.13.0
0.14.0
0.14.1

08:41:10 ❯ cargo run --bin mise -- install zig@latest
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.42s
     Running `target/debug/mise install 'zig@latest'`
mise zig@0.14.1 ✓ installed

```

PR #5242, opened 2025-05-30
docs: add doc about github access token permissions

GiGurra:
Some installations, notably asdf:clang (try `mise use -g clang@latest` with a github token set) require additional token permissions (notably public repo access to be checked/enabled).

This PR suggests editing the doc to explain that. Please change the text as you see fit.

Before the change: 
* `mise use -g clang@latest` succeeds if no gh token is set
* `mise use -g clang@latest` fails if gh token is set (with 404 when trying to download `https://github.com/llvm/llvm-project/archive/llvmorg-<version>.tar.gz`)

After the change:
* `mise use -g clang@latest` succeeds if no gh token is set
* `mise use -g clang@latest` succeeds if gh token is set

How to set it:
image


GiGurra:
> should go on the plugin. the scope also isn't required.

Do you mean I should instead ask here https://github.com/mise-plugins/mise-llvm to have it documented there?

PR #5247, opened 2025-05-30
docs: add comment for installing mise using wget

mangkoran:
A minor note for installing mise using `wget` in https://mise.jdx.dev/getting-started.html. Could be useful inside some container environment where only wget is available.


mangkoran:
I also plan to add the note in https://github.com/jdx/mise-action/blob/13abe502c30c1559a5c37dff303831bab82c9402/README.md

jdx:
this is already on the linked page

PR #5271, opened 2025-06-03
fix(registry): use asdf-community repo for unforked plugins

risu729:
For these two plugins, it seems they're not forked to `mise-plugins`.
This PR reverts the registry to use the `asdf-community` repositories.

https://github.com/asdf-community/asdf-regal
https://github.com/asdf-community/asdf-scala-cli


jdx:
I just forked them

PR #5284, opened 2025-06-04
fix: make mise exec work again with unknown tool entries

stanhu:
With the switch from Rayon (parallel processing) to Tokio (async) in https://github.com/jdx/mise/pull/5172, `mise exec` no longer correctly resolves tool versions if `.tool-versions` contains an unknown entry.

This occurred because a single failure would cause a hard faliure in the `resolve()` method in `src/toolset/mod.rs`. Fix this by setting successful tool resolutions while returning the errors, which can be ignored by the caller.

Relates to https://github.com/jdx/mise/discussions/5282

### DIFF
--- a/src/toolset/mod.rs
+++ b/src/toolset/mod.rs
@@ -185,12 +185,11 @@ impl Toolset {
 
         self.versions = successful_versions;
 
-        match errors.is_empty() {
-            true => Ok(()),
-            false => {
-                let err = eyre::eyre!("error resolving versions");
-                Err(errors.into_iter().fold(err, |e, x| e.wrap_err(x)))
-            }
+        if errors.is_empty() {
+            Ok(())
+        } else {
+            let err = eyre::eyre!("error resolving versions");
+            Err(errors.into_iter().fold(err, |e, x| e.wrap_err(x)))
         }
     }
     #[async_backtrace::framed]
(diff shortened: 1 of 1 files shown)


stanhu:
@jdx Could you review?

stanhu:
@jdx Sorry to bother you again. Would you mind reviewing this?

src/toolset/mod.rs
```diff
@@ -158,13 +158,39 @@ impl Toolset {
...
-        Ok(())
+        let mut successful_versions = IndexMap::new();
+        let mut errors = Vec::new();
+        for (ba, tool_result) in safe_results {
+            match tool_result {
+                Ok(tvl) => {
+                    successful_versions.insert(ba, tvl);
+                }
+                Err(err) => {
+                    if Error::is_argument_err(&err) {
```
jdx, line 177: I don't really like checking for argument errors at this point. Could we instead exclude these tools if we know they won't be resolved?

src/toolset/mod.rs
```diff
@@ -158,13 +158,39 @@ impl Toolset {
...
+                        return Err(err);
+                    }
+                    warn!("failed to resolve tool {}: {err}", ba.short);
```
jdx, line 180: isn't this going to output the error as a warning and then also output it as an error?

PR #5290, opened 2025-06-06
fix: only show deprecation if not using 'tools-version'

timfallmk:
### Problem
The deprecation warning for `@system` was being shown for every usage, including in `.tool-versions` files where it's still valid for backward compatibility. This caused unnecessary warnings for users who have existing `.tool-versions` files with `@system` entries.

### Solution
Modified the `tv_pathname` method in `src/toolset/tool_version.rs` to check the source of the tool request. The deprecation warning is now suppressed when `@system` is used in `.tool-versions` files, but still shown for other configuration sources like `.mise.toml`.

### Changes
- Updated `ToolRequest::System` handling in `tv_pathname` method
- Added source checking to conditionally show deprecation warning
- Preserved backward compatibility for `.tool-versions` files

### Testing
- Verified that warnings are suppressed for `.tool-versions` 
- Verified that warnings still appear for `.mise.toml` usage

Fixes #5230

### DIFF
--- a/src/toolset/tool_version.rs
+++ b/src/toolset/tool_version.rs
@@ -156,7 +156,10 @@ impl ToolVersion {
             ToolRequest::Path { path: p, .. } => format!("path-{}", hash_to_str(p)),
             ToolRequest::System { .. } => {
                 // Only show deprecation warning if not from .tool-versions file
-                if !matches!(self.request.source(), crate::toolset::ToolSource::ToolVersions(_)) {
+                if !matches!(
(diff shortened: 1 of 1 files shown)


jdx:
bugbot run

timfallmk:
@jdx We use this internally and we've found sometimes it only responds to the capitalized version "BugBot run"

> bugbot run

timfallmk:
Gently bump @jdx . Thanks!

PR #5294, opened 2025-06-06
Update getting-started.md for missing shasum command

canepan:
On AmazonLinux2 (based on CentOS) `shasum` is not available, but there is `sha1sum`, `sha256sum`, etc. I believe sha256sum is required here


canepan (commented):
Updating commit title

PR #5320, opened 2025-06-10
feat: support matching_regex from the ubi backend

yjoer:
I had matching_regex landed on UBI in https://github.com/houseabsolute/ubi/pull/122. This change enables passing the option from the mise config to the backend.

Closes https://github.com/jdx/mise/discussions/5318


yjoer:
The builds are failing because `os_str_display` is stabilized in [Rust 1.87](https://github.com/rust-lang/rust/releases/tag/1.87.0) and an earlier version is used in the CI.

jdx:
1.87 is way too new to support—it's less than a month old. @autarch would it be possible to use alternate syntax to support somewhat older rust releases? this would be a problem for distros that have somewhat old rust versions?

jdx:
bugbot run

jdx:
bugbot run

jdx:
bugbot run verbose=true

jdx:
bugbot run verbose=true

autarch:
> 1.87 is way too new to support—it's less than a month old. @autarch would it be possible to use alternate syntax to support somewhat older rust releases? this would be a problem for distros that have somewhat old rust versions?

Sure, but I don't know exactly what thing I'm using that's 1.87-specific.

autarch:
Looks like this repo targets 1.85, so I'll do the same for the `ubi` crate.

yjoer:
The builds are passing now 🥳

Thanks @jdx and @houseabsolute

Cargo.lock
```diff
@@ -7192,27 +7190,14 @@ version = "2.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fabe6324e908f85a1c52063ce7aa26b68dcb7eb6dbc83a2d148403c9bc3eba50"
 dependencies = [
- "aes",
```
jdx, line 7195: this is strange but I assume it's fine

autarch: In `ubi` 0.7.0, I started adding `default-features = false` to most dependency declarations. This eliminated a bunch of transitive dependencies that weren't being used.

PR #4832, opened 2025-04-12
feat(registry): add docker cli

acesyde:
- feat: add docker cli in the registry

`cargo run --bin mise -- tool docker-cli`

```shell
Backend:            aqua:docker/cli
Description:        Docker CE CLI  
Installed Versions: 28.0.4         
Tool Options:       [none] 
```

PR #5341, opened 2025-06-13
fix(aqua): parse templates in version_filter

risu729:
`aqua:docker/cli` can be installed, but it caused warnings. 
https://github.com/jdx/mise/pull/4832

```
mise WARN  [aqua:docker/cli] aqua version filter error: Invalid operands for operator contains
```

This is because the registry uses `Version` in its `version_filter`.
https://github.com/aquaproj/aqua-registry/blob/8f3a2ae02b5b56a9583ddb1c9ee9739ceff19f9e/pkgs/docker/cli/rootless/registry.yaml

This PR resolves the issue by parsing aqua templates in `version_filter`.


risu729:
This was caused by the wrong operator precedence between `contains` and `or` used in `version_filter`, not the use of `Version`.
I fixed it in https://github.com/jdx/expr.rs/pull/16.

PR #5348, opened 2025-06-14
fix(registry): talosctl use aqua

mangkoran:
* chore(registry): talos,talosctl deprecate asdf

* test(registry): talos,talosctl add

* refactor(registry): talosctl alias

* fix(registry): arduino alias

Test
```
❯ talosctl version --client --short
Client:
Talos v1.10.4
```


mangkoran:
Please let me know if the asdf deprecation should not be included

mangkoran:
Actually `talos` and `talosctl` install the same executable: `talosctl`. Should we deprecate one of them?

jdx:
you could alias it

mangkoran:
Alright working on it

mangkoran:
While at it, I found `arduino` alias seems incorrect, as it aliased itself. I think it's supposed to alias `arduino-cli`.

Please let me know if this shouldn't be included in this PR.

PR #5352, opened 2025-06-14
docs: Document trivial task syntax

JayBazuzi:


hverlin:
👍 Otherwise, it's only mentioned in https://mise.jdx.dev/walkthrough.html#tasks

PR #5354, opened 2025-06-15
feat(java): add support for tar.xz in Java core plugin to support RedHat JDKs

roele:
With [b6bedc3](https://github.com/jdx/mise-java/commit/b6bedc3194a6382adcda382c8b2d879e7669a6c1) the new JDK vendor RedHat has been added. For Linux, only `tar.xz` archives are available, hence the extension to support this format.

### DIFF
--- a/src/plugins/core/java.rs
+++ b/src/plugins/core/java.rs
@@ -127,21 +127,17 @@ impl JavaPlugin {
     ) -> Result<()> {
         let filename = tarball_path.file_name().unwrap().to_string_lossy();
         pr.set_message(format!("extract {filename}"));
-        if m.file_type
-            .as_ref()
-            .is_some_and(|file_type| file_type == "zip")
-        {
-            file::unzip(tarball_path, &tv.download_path())?;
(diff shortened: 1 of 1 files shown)


src/plugins/core/java.rs
```diff
@@ -127,21 +127,25 @@ impl JavaPlugin {
...
+                let format = match ext {
+                    "tar.gz" => TarFormat::TarGz,
+                    "tar.xz" => TarFormat::TarXz,
+                    _ => unreachable!(),
+                };
+                file::untar(
+                    tarball_path,
+                    &tv.download_path(),
+                    &TarOptions {
+                        format,
```
jdx: could we just use auto? https://github.com/jdx/mise/blob/a63dd42a68386096063454af135ce526dd9aa659/src/file.rs#L626

roele: did not see that one, thanks for the hint!

jdx: it looks like it should support zip too

PR #5361, opened 2025-06-16
fix(shim): improve resolve_symlink for Windows

qianlongzt:
https://github.com/jdx/mise/discussions/5254
https://github.com/jdx/mise/discussions/5241

### DIFF
--- a/e2e-win/symlink.Tests.ps1
+++ b/e2e-win/symlink.Tests.ps1
@@ -1,0 +1,25 @@
+Describe 'windows runtime symlink' {
+
+    It 'version is correct, not latest' {
+        mise install yq@4.45.4
+
+        # correct output
+        # mise/2025.5.15/bin/mise ls yq
+        # Tool  Version  Source                                        Requested
+        # yq    4.45.4   D:\Users\qianlongzt\.config\mise\config.toml  latest
+
+        # wrong output
+        # mise/2025.5.16/bin/mise ls yq
+        # Tool  Version  Source                                        Requested
+        # yq    latest   D:\Users\qianlongzt\.config\mise\config.toml  latest
+        # yq    4.45.4
+
+        # https://github.com/jdx/mise/discussions/5254
+
+        
+        mise ls yq --json | jq '.[] | select(.source ) | .version' | Should -Be '"4.45.4"'
+
+        mise ls yq --json | jq '.[] | select(.version == "latest" ) | .version' | Should -Be $null
+
+    }
+}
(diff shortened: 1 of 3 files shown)


vemoo:
I also found this issue, but my fix was bit diferent:
https://github.com/vemoo/mise/commit/b311713761b2ed94f9f09b054e709c5e40e9a88e

I don't think it makes sense using or there because on linux it should always be a link and on windows it should always be a file.

vemoo:
Why is this a draft PR? Should I create a PR with my proposed changes?

qianlongzt:
Since the tests failed, I will try again today.

qianlongzt:
The previous error was caused by the incorrect use of link.metadata, which follows symlinks and thus led to incorrect judgment.

src/runtime_symlinks.rs
```diff
@@ -103,9 +103,11 @@ pub fn remove_missing_symlinks(backend: Arc<dyn Backend>) -> Result<()> {
     Ok(())
 }
 
+// for unix/linux ./1.1.1
+// for Windows  .\1.1.1
 pub fn is_runtime_symlink(path: &Path) -> bool {
     if let Ok(Some(link)) = file::resolve_symlink(path) {
-        return link.starts_with("./");
+        return link.starts_with(format!(".{}", std::path::MAIN_SEPARATOR_STR));
```
vemoo: This shouldn't be necessary, `Path::starts_with` compares components, not path separators, so on windows both `Path::new(".\\foo").starts_with("./")` and `Path::new("./foo").starts_with("./")` are `true`.

Also this allocates a new `String` each time.

e2e-win/symlink.Tests.ps1
```diff
@@ -0,0 +1,39 @@
...
+        # wrong output
+        # mise/2025.5.16/bin/mise ls yq
+        # Tool  Version  Source                                        Requested
+        # yq    latest   D:\Users\qianlongzt\.config\mise\config.toml  latest
+        # yq    4.45.4
+
+        # https://github.com/jdx/mise/discussions/5254
+
+        $output = mise ls --json yq
+        $output | jq '.[] | select(.source ) | .version' | Should -Be '"4.45.4"'
```
vemoo, line 36: I think this test will fail once a new version of `yq` gets released.

Maybe check against a regex? Or leave only the check that ensure the value is not `latest`?

vemoo: Nevermind I see that you then do `mise install yq@4.45.4`.

PR #5374, opened 2025-06-19
feat(registry): add carapace

jahands:
This is a nice shell completion tool: https://github.com/carapace-sh/carapace-bin

It has support for many tools: https://carapace-sh.github.io/carapace-bin/completers.html

And a completion format that's easier to use than typical zsh completions: https://carapace-sh.github.io/carapace-bin/spec.html

CleanShot_2025-06-19T06-37-48_CleanShot@2x


risu729:
https://github.com/jdx/mise/pull/5372 will fix the tests.

PR #5378, opened 2025-06-20
feat: Add sops.strict setting for non-strict decryption mode

pepicrft:
## Summary
- Add new `sops.strict` setting (default: `true`) to control sops decryption failure behavior
- When `sops.strict=false`, mise continues and skips encrypted content when decryption fails
- When `sops.strict=true`, mise fails when decryption fails (current behavior)

## Problem
Currently, mise fails to load environments when sops-encrypted files are present but the required age key is not available. This prevents users from working with repositories containing encrypted secrets unless they have access to the decryption keys.

## Solution
This feature adds a new setting `sops.strict` that allows users to opt into a non-strict mode where:
- Missing sops keys won't cause mise to fail
- Encrypted content is skipped and logged as a warning
- Unencrypted environment variables continue to work normally

## Test plan
- [x] Unit tests pass (`cargo test`)
- [x] Code compiles successfully (`cargo build`)  
- [x] Linting and formatting passes (`mise run lint`)
- [x] Added integration test for non-strict mode behavior
- [x] Verified backward compatibility (strict mode is default)

## Usage
```bash
# Enable non-strict mode
mise settings set sops.strict false

# Or via environment variable
export MISE_SOPS_STRICT=false
```

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/e2e/secrets/test_secrets_non_strict
+++ b/e2e/secrets/test_secrets_non_strict
@@ -1,0 +1,52 @@
+#!/usr/bin/env bash
+set -x
+
+# Test that with sops.strict=false, mise continues when age key is not available
+
+mise use sops age
+
+# Create an encrypted file but don't set age key
+age="$(mise x -- age-keygen 2>&1)"
+age_pub="$(echo "$age" | grep "# public key:" | awk '{print $4}')"
+
+# json test with strict=false
+echo '{ "SECRET": "mysecret" }' >.env.json
+mise x -- sops encrypt -i --age "$age_pub" .env.json
+
+# Set non-strict mode
+mise settings set sops.strict false
+
+# Clear age key to simulate missing key
+export MISE_SOPS_AGE_KEY=
+rm -f ~/.config/mise/age.txt
+
+# Should continue without failing and not include the secret
+assert "mise set _.file=.env.json"
+# The env should not contain the decrypted secret since key is missing
+assert_not_contains "mise env" "export SECRET=mysecret"
+
+# Test with strict=true (default behavior)
+mise settings set sops.strict true
+
+# This should fail
+assert_fail "mise env"
+
+# Test with yaml files too
+age_pub="$(mise x -- age-keygen -o ~/age.txt 2>&1 | awk '{print $3}')"
+echo 'SECRET: mysecret' >.env.yaml
+mise x -- sops encrypt -i --age "$age_pub" .env.yaml
+
+# Remove the key file
+rm -f ~/age.txt
+export MISE_SOPS_AGE_KEY=
+
+# Set non-strict mode
+mise settings set sops.strict false
+
+# Should work with non-strict mode
+assert "mise set _.file=.env.yaml"
+assert_not_contains "mise env" "export SECRET=mysecret"
+
+# Reset to strict mode and verify it fails
+mise settings set sops.strict true
+assert_fail "mise env"
--- a/src/sops.rs
+++ b/src/sops.rs
@@ -68,12 +68,24 @@ where
         env::set_var(age_env_key, age.trim());
     }
     let output = if Settings::get().sops.rops {
-        input
+        match input
             .parse::<RopsFile<EncryptedFile<AES256GCM, SHA512>, F>>()
-            .wrap_err("failed to parse sops file")?
-            .decrypt::<F>()
-            .wrap_err("failed to decrypt sops file")?
-            .to_string()
+            .wrap_err("failed to parse sops file")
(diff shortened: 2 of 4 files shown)


e2e/secrets/test_secrets_non_strict
```diff
@@ -0,0 +1,52 @@
...
+assert_not_contains "mise env" "export SECRET=mysecret"
+
+# Test with strict=true (default behavior)
+mise settings set sops.strict true
+
+# This should fail
+assert_fail "mise env"
+
+# Test with yaml files too
+age_pub="$(mise x -- age-keygen -o ~/age.txt 2>&1 | grep "# public key:" | awk '{print $4}')"
```
roele: this seems to fail for some reason

roele (commented):
👍🏼 LGTM besides the test failure

e2e/secrets/test_secrets_non_strict
roele: culprit seems to be the call with `mise x` which fails with...
```
++ mise x -- age-keygen -o /private/var/folders/_d/hzb38n8d6sb5lx_x43qnzc_m0000gn/T/test_secrets_non_strict.bnd0LD/home/age.txt
Error: 
   0: failed to decrypt sops file
   1: unable to decrypt file metadata
   2: unable to retrieve data key
   3: no data key retrieved from metadata, make sure at least one private integration key is present

Location:
   src/sops.rs:74
```

pepicrft: @roele the test is passing now :)

src/sops.rs
```diff
@@ -68,12 +68,24 @@ where
...
-            .to_string()
+            .wrap_err("failed to parse sops file")
+            .and_then(|file| file.decrypt::<F>().wrap_err("failed to decrypt sops file"))
+        {
+            Ok(decrypted) => decrypted.to_string(),
+            Err(e) => {
+                if Settings::get().sops.strict {
+                    return Err(e);
+                } else {
+                    warn!(
```
jdx: I think this warning will be very annoying, it'd be best to move this to debug level

src/sops.rs
```diff
@@ -68,12 +68,24 @@ where
...
+                        "sops decryption failed but continuing in non-strict mode: {}",
+                        e
+                    );
+                    return Ok(input.to_string()); // Return original content without decryption
```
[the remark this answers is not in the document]
jdx: this seems valid to me

e2e/secrets/.env.json
```diff
@@ -0,0 +1,15 @@
+{
+	"SECRET": "ENC[AES256_GCM,data:Arbiea+V/gw=,iv:FIZ6HOk53icxfE8v38Ggja4mWyKq797TZSYKRYuSNYc=,tag:fCCTrnCe4ELbKkGbZ1/RUw==,type:str]",
+	"sops": {
+		"age": [
+			{
+				"recipient": "age14k8sra8tyzz3e5fq9tft7zzpmnsq8mc3pt9zkhv3xlkqrdume4ssjsg4xe",
```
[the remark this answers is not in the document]
jdx: you can either ignore this or add maybe an unencrypted comment noting this is a test fixture

src/sops.rs
```diff
@@ -68,12 +68,24 @@ where
         env::set_var(age_env_key, age.trim());
     }
     let output = if Settings::get().sops.rops {
-        input
+        match input
             .parse::<RopsFile<EncryptedFile<AES256GCM, SHA512>, F>>()
```
jdx, line 78: you should detect if the file exists before attempting to read it. Creating errors (even if they're handled) takes a truly massive amount of time so it will noticeably slow down mise if the file isn't there even if strict is false

src/sops.rs
```diff
@@ -99,7 +111,21 @@ where
...
+            Ok(output) => output,
+            Err(e) => {
+                if Settings::get().sops.strict {
+                    return Err(e.into());
+                } else {
+                    warn!(
+                        "sops decryption failed but continuing in non-strict mode: {}",
+                        e
+                    );
+                    input.to_string() // Return original content without decryption
```
jdx: is the purpose here generally to handle the case where sops isn't installed? if so my comment about missing files applies here too, we should check if sops is available and avoid creating an error

jdx:
I think cursor is correct. I also don't think I understand the use-case where we return the original encrypted content, seems like we should return nothing instead.

pepicrft:
@jdx thanks a lot for the review. The comments should be addressed, and I also revisited the documentation of what the flag is for. Let me know what you think :).

jdx:
bugbot run

PR #5388, opened 2025-06-21
refactor(registry): add backends and tests

risu729:
Adds backends and tests to some tools in the registry.
This PR doesn't change the default backends, but adds fallback backends.

### DIFF
--- a/registry.toml
+++ b/registry.toml
@@ -558,6 +558,11 @@ djinni.backends = [
 ]
 # djinni.test = ["djinni-generator --version", "djinni generator version {{version}}"] # test fails on windows
 dmd.backends = ["asdf:mise-plugins/mise-dmd"]
+docker-cli.backends = ["aqua:docker/cli"]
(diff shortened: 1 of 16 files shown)


registry.toml
```diff
@@ -658,10 +664,11 @@ eza.backends = [
     {full = "aqua:eza-community/eza", platforms = [
         "linux"
     ]},
-    # "ubi:cargo-bins/cargo-quickinstall[exe=eza,tag_regex=^eza-]", # does not fetch tags right, needs to load everything
+    "ubi:eza-community/eza",
```
jdx: ubi doesn't offer macos

risu729: Thanks I didn't notice that.

risu729:
`biome@2.0.1` is broken (https://github.com/biomejs/biome/issues/6435), so we need to wait `mise-versions` will be updated to include `2.0.4`.

PR #5389, opened 2025-06-21
fix(registry): use aqua backend for slsa-verifier

risu729:


risu729:
I forgot about this https://github.com/jdx/mise/pull/4325.
I will try to fix but not sure I can.

PR #5401, opened 2025-06-22
ci: use cargo info to retrieve latest mise version

risu729:
I'm not sure if this will work, but it might work.
`--registry "crates-io"` avoids printing the local mise verison, and `--color never` avoids printing ANSI color codes.

```
cargo info --registry "crates-io" --color never --quiet mise | grep "^version:" | cut -d' ' -f2
```

### DIFF
--- a/xtasks/release-plz
+++ b/xtasks/release-plz
@@ -5,7 +5,7 @@ git config user.name mise-en-dev
 git config user.email release@mise.jdx.dev
 
 cur_version="$(cargo pkgid mise | cut -d# -f2)"
-latest_version="$(cargo search mise --limit 1 -q | grep '^mise = "' | cut -d'"' -f2)"
+latest_version="$(cargo info --registry "crates-io" --color never --quiet mise | grep "^version:" | cut -d' ' -f2)"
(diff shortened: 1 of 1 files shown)


risu729:
I got it. Usually the color is disabled when using pipes, but it is forcibly enabled with `CARGO_TERM_COLOR: always` in the workflow level.
https://github.com/jdx/mise/blob/d4bbeb6043585d37412ec7932e7873470b9d2759/.github/workflows/release-plz.yml#L22

risu729:
The original implementation before https://github.com/jdx/mise/commit/729aadc83e042b276e3ebd3ae378a7e647a54bc0 was:

```
curl -fSs https://crates.io/api/v1/crates/mise | jq -r '.versions[0].num'
```

However it fails even in my local environment with `curl: (22) The requested URL returned error: 403`, so I guess they changed something.

roele:
Odd the `curl` variant still works for me locally...

```
curl -fSs https://crates.io/api/v1/crates/mise | jq -r '.versions[0].num'
2025.6.5
```

jdx:
yeah the curl way works for me locally too but not on gha

Issue #2721, opened 2024-10-10
Faster installs for erlang


paradox460:
Currently, both the old ASDF erlang plugin and the built-in erlang plugin depend on kerl to build the erlang binary every single time.

This is acceptable on a dev machine, but when trying to use in CI, particularly on a slower project, where builds are infrequent, it can stretch a build to many times its length, as erlang has to be rebuilt every time.

https://github.com/erlef/setup-beam uses pre-built binaries that target various operating systems common in CI providers. It does this by sourcing them from various files on `builds.hex.pm` or on a mirror, as they are originally built by [bob](https://github.com/hexpm/bob).

The [logic for choosing which file to download is fairly simple, albeit long, due to the amount of things supported](https://github.com/erlef/setup-beam/blob/main/src/setup-beam.js#L751-L1002). But at its heart its just attempting to download `https://builds.hex.pm/builds/otp/<architecture>/<versionSpec>.tar.gz` for erlang versions.

There's been a [discussion about this for a long time on the old asdf-erlang plugin](https://github.com/asdf-vm/asdf-erlang/issues/165), but it hasn't seen much traction since 2022

cw789:
Also for MacOS there are now builds available: https://elixirforum.com/t/new-community-maintained-otp-builds-for-macos/67338

jdx:
I'd really love to see this as I can tell erlang+elixir are quite popular both with asdf and mise. I would greatly welcome a contribution if someone is willing to help. I may get around to it soon otherwise. [swift](https://github.com/jdx/mise/pull/3271) would be a good reference on how to do this.

PR #5402, opened 2025-06-23
Faster Erlang installs for Linux and Github Actions Ubuntu Runners

paradox460:
This PR attempts to enable installation of precompiled erlang binaries on Ubuntu systems, supplementing the current fast installs for MacOS and Windows (#2721).

# Details

Erlang builds are made available, precompiled, for a variety of Ubuntu systems via [bob](https://github.com/hexpm/bob).

These are generally available at `https://builds.hex.pm/builds/otp/${ARCH}/${OS_VER}/${REF}.tar.gz`

`ARCH` and `OS_VER` are typically a bit different than the other constants Mise uses in the erlang.rs core plugin.

The `REF` appears to be the same as the ones used for the other functions in the file, so it does not need to be changed.

From bob's documentation, the two arches available for precompiled builds are:

- `amd64`
- `arm64`

Additionally, the OS version encodes what build of Ubuntu is used as well:

- `ubuntu-20.04`
- `ubuntu-22.04`
- `ubuntu-24.04`

The architecture names are a fairly simple mapping, more or less just different names for concepts that already exist in the mise source code. However OS_VER is more complicated.

On Github Actions, this is exposed as the envar `ImageOS`. Determining the runtime from that would technically satisfy my ask, as this is primarily about CI. This is what I've implemented so far. If the envar is not present, then the build should be manual.

For determining the OS in an actual ubuntu context, one approach is that used by the [os_type](https://github.com/schultyy/os_type) crate.


paradox460:
I'm not actually sure if this does anything or works. My experience with rust is minimal, and so I just tried to follow conventions I saw in the file at hand.

src/plugins/core/erlang.rs
```diff
@@ -72,7 +72,58 @@ impl ErlangPlugin {
...
+            &tarball_path,
+            &tv.install_path(),
+            &TarOptions {
+                strip_components: 0,
+                pr: Some(&ctx.pr),
+                format: file::TarFormat::TarGz,
+            },
+        )?;
+
+        Ok(Some(tv));
```
roele: should be implicit return, so no semicolon

src/plugins/core/erlang.rs
```diff
@@ -72,7 +72,58 @@ impl ErlangPlugin {
...
+        let arch: String = match ARCH {
+            "x86_64" => "amd64",
+            "aarch64" => "arm64",
+            _ => {
+                debug!("Unsupported architecture: {}", ARCH);
+                return Ok(None);
+            }
+        };
+
+        let os_ver = match std::env::var("ImageOS") {
```
roele: i guess additional checks for the supported versions are required here, correct?

src/plugins/core/erlang.rs
```diff
@@ -72,7 +72,58 @@ impl ErlangPlugin {
...
+        ctx: &InstallContext,
+        mut tv: ToolVersion,
+    ) -> Result<Option<ToolVersion>> {
+        if Settings::get().erlang.compile == Some(true) {
+            return Ok(None);
+        }
+        let release_tag = format!("OTP-{}", tv.version);
+
+        let arch: String = match ARCH {
+            "x86_64" => "amd64",
```
roele: needs to be `"amd64".to_string()` and `"arm64".to_string()`

src/plugins/core/erlang.rs
roele: I would personally favour the use of the `os_crate` here so we can support Non-GitHub and Non-CI uses-cases.

src/plugins/core/erlang.rs
```diff
@@ -72,7 +72,67 @@ impl ErlangPlugin {
...
+        ctx.pr.set_message(format!("Extracting {filename}"));
+        file::untar(
+            &tarball_path,
+            &tv.install_path(),
+            &TarOptions {
+                strip_components: 0,
+                pr: Some(&ctx.pr),
+                format: file::TarFormat::TarGz,
+            },
+        )?;
```
roele, line 135:
```suggestion

        self.move_to_install_path(&tv)?;

        CmdLineRunner::new(&tv.install_path().join("Install"))
            .with_pr(&ctx.pr)
            .arg("-minimal")
            .arg(&tv.install_path())
            .execute()?;
```

src/plugins/core/erlang.rs
```diff
@@ -72,7 +72,67 @@ impl ErlangPlugin {
...
+
+        Ok(Some(tv))
+    }
+
```
roele, line 147:
```suggestion
    #[cfg(linux)]
    fn move_to_install_path(&self, tv: &ToolVersion) -> Result<()> {
        let base_dir = tv
            .download_path()
            .read_dir()?
            .find(|e| e.as_ref().unwrap().file_type().unwrap().is_dir())
            .unwrap()?
            .path();
        file::remove_all(tv.install_path())?;
        file::create_dir_all(tv.install_path())?;
        for entry in fs::read_dir(base_dir)? {
            let entry = entry?;
            let dest = tv.install_path().join(entry.file_name());
            trace!("moving {:?} to {:?}", entry.path(), &dest);
            file::rename(entry.path(), dest)?;
        }

        Ok(())
    }
    
```

roele (changes requested):
With these suggestions and some additional imports `std::fs;` and `use crate::cmd::CmdLineRunner;`  the installation works on an Ubuntu.

roele (approved):
There are some warnings for OS specific builds about dead code, will have a look at those later. Besides that looks okay to me now.

src/plugins/core/erlang.rs
```diff
@@ -72,7 +77,95 @@ impl ErlangPlugin {
...
+        ctx.pr.set_message(format!("Downloading {filename}"));
+        if !tarball_path.exists() {
+            HTTP.download_file(&url, &tarball_path, Some(&ctx.pr))
+                .await?;
+        }
+        ctx.pr.set_message(format!("Extracting {filename}"));
+        file::untar(
+            &tarball_path,
+            &tv.install_path(),
```
roele: forgot to suggest, this needs to be `&tv.download_path()`

PR #5406, opened 2025-06-23
fix(aqua): fix versions order

risu729:
My apologies, https://github.com/jdx/mise/pull/5371 broke the version order in `ls-remote`.

I misused `BTreeMap`, the key was `String`, not semver, so the versions were just compared lexicographically.

This PR fixes the regression by using `Vec<(String, String)>` instead of `BTreeMap`. I can also use `IndexMap`, but it seemed it's only used in `build.rs` so I guessed it's not preffered.

### DIFF
--- a/src/backend/aqua.rs
+++ b/src/backend/aqua.rs
@@ -58,9 +58,9 @@ impl Backend for AquaBackend {
     }
 
     async fn _list_remote_versions(&self, _config: &Arc<Config>) -> Result<Vec<String>> {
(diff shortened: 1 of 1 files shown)


risu729:
This broke the registry tests, because some tools changed their output of versions.
`registry.toml` tests depend on the latest version behaviours, so the tests fail if they're tested with older versions.

risu729:
The hyperfine test also failed because of this regression, as it uses `-N` flag which is not supported in an old version.

PR #5387, opened 2025-06-21
fix: `self-update` on MITM firewall

joehorsnell:
When using a traffic-inspecting/MITM firewall, such as Cloudflare WARP, the `self-update` command fails:

```
❯ ~/.local/bin/mise self-update --verbose
DEBUG ARGS: /Users/joe/.local/bin/mise self-update --verbose
DEBUG starting new connection: https://api.github.com/
Error:
   0: ReqwestError: error sending request for url (https://api.github.com/repos/jdx/mise/releases/latest)
   1: error sending request for url (https://api.github.com/repos/jdx/mise/releases/latest)
   2: client error (Connect)
   3: invalid peer certificate: UnknownIssuer

Location:
   src/cli/self_update.rs:78
```

This is because the request to GitHub releases is not using the local user's native certs, which include the necessary trusted root CA.

I'd appreciate some feedback @jdx:
- Is this the correct way to fix this issue? I mean, it certainly works locally, but I'm not sure what other implications there are from not using the vendored OpenSSL
- How to write a test for this?

Related/fixes:
- https://github.com/jdx/mise/discussions/4665
- https://github.com/jdx/mise/discussions/4931
- https://github.com/jdx/mise/discussions/5313

**Edit**: also related, #714 and #715.

PR #5459, opened 2025-06-28
fix: `self-update` on MITM firewall (attempt #2)

joehorsnell:
This is a follow-up to jdx/mise#5387, to fix the `self-update` command when using a traffic-inspecting/MITM firewall, such as Cloudflare WARP.

The previous PR broke the release build, so had to be reverted. I didn't realise that the [`reqwest`](https://crates.io/crates/reqwest) crate requirements are different on Linux (I use macOS), so OpenSSL is still needed.

I tried removing OpenSSL as a dependency completely, but it appears to not be straightforward, as many other crates also depend on `reqwest`.

The simplest thing I think is just to leave OpenSSL and add the `rustls-native-roots` feature.

I've tested `build-tarball.sh` locally on both macOS and Linux (using a VM) and verified that the resulting binaries can self-update correctly.

I don't have an easy way to test on Windows, but the `reqwest` docs suggest that it behaves the same as on macOS.

### DIFF
--- a/scripts/build-tarball.ps1
+++ b/scripts/build-tarball.ps1
@@ -6,7 +6,7 @@ $Version = ./scripts/get-version.ps1
 $BaseName = "mise-v$Version-$Env:OS-$Env:ARCH"
 
 # TODO: use "serious" feature
-cargo build --release --features openssl/vendored --target "$Target"
+cargo build --release --features rustls-native-roots,openssl/vendored --target "$Target"
 mkdir -p dist/mise/bin
 cp "target/$Target/release/mise.exe" dist/mise/bin/mise.exe
 cp README.md dist/mise/README.md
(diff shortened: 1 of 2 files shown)


jdx:
apologies, been super busy but I'll get around to it soon

joehorsnell:
No worries @jdx 👍

On a related note, do you see any sense in adding Mise tasks to run stuff in the `scripts` directory, eg the `build-tarball` scripts?

Yes, it would introduce a dependency on having Mise installed to run those scripts, but I think that already exists for some other tasks (like release?).

Mainly it would provide a natural place to define ENV vars for the various arguments (such as enabled features) that are shared across the Unix and Windows scripts.

Just a thought, can open a draft PR if you think it makes sense.

jdx:
sorry that took so long https://github.com/jdx/mise/actions/runs/16102935878

jdx:
> On a related note, do you see any sense in adding Mise tasks to run stuff in the scripts directory, eg the build-tarball scripts?

I don't really like messing with this code since it often results in bad releases

PR #5462, opened 2025-06-29
feat(registry): add PostgreSQL aqua backend and aliases

yashikota:


yashikota:
The CI job is currently failing with

```
0: failed to install aqua:bitwarden/clients@cli-v2025.6.0,
```

which appears to be unrelated to the changes in this PR.
Could someone take a look and address the issue? Thanks!

yashikota:
I’m also seeing the following error:

```
  mise ERROR postgres: 
     0: failed to install aqua:theseus-rs/postgresql-binaries@17.5
     1: HTTP status client error (404 Not Found) for url (https://api.github.com/repos/theseus-rs/postgresql-binaries/releases/tags/17.5)
     2: HTTP status client error (404 Not Found) for url (https://api.github.com/repos/theseus-rs/postgresql-binaries/releases/tags/v17.5)
```

The reason for the 404 is that the correct version tag is `17.5.0`, not `17.5`.
However, I’m not sure how to fix this in the current setup.
Could someone help me with this?

risu729:
> The CI job is currently failing with
> 
> ```
> 0: failed to install aqua:bitwarden/clients@cli-v2025.6.0,
> ```
> 
> which appears to be unrelated to the changes in this PR. Could someone take a look and address the issue? Thanks!

I opened an issue in the aqua-registry. https://github.com/jdx/mise/pull/5462#issuecomment-3016283125
I'll fix it, but we only update the aqua-registry in the release, so I think we need to disable the test for now.
I disabled it in https://github.com/jdx/mise/pull/5463, so this PR can be merged after it by updating the branch.

risu729:
Since `theseus-rs/postgresql-binaries` is a third-party release of the prebuilt binaries, I don't think we should include it in the default registry in terms of security concerns.

Regarding the versioning, they version the releases with `<postgres major>.<postgres minor>.<release>`, where `<release>` is the build version.
https://github.com/theseus-rs/postgresql-binaries?tab=readme-ov-file#versioning
So, we need to use their versions (e.g. `17.5.0`) instead of the postgres version (e.g. `17.5`).

However, [`mise-versions`](https://github.com/jdx/mise-versions/) contains the postgres versions, so the test is failing.
We can ignore the error with [`use_versions_host`](https://mise.jdx.dev/configuration/settings.html#use_versions_host) set to `false` for now, and the version list in `mise-versions` will be updated when this change is included in the new release.
However, it would be a breaking change to users using the older version of mise because `mise install postgres@17.5.0` fails.

PR #5463, opened 2025-06-29
fix(aqua): handle hard links in aqua packages

risu729:
Resolves #5419.

This PR fixes a bug that caused a `No such file or directory` error when extracting tar archives containing hard links.

### Cause

The issue is caused by [`tar::Entry::unpack`](https://docs.rs/tar/latest/tar/struct.Entry.html#method.unpack), which resolves hard link paths relative to the current working directory instead of the intended destination directory.
When `Entry::unpack` calls the internal `EntryFields::unpack` method, it passes `None` as the `target_base`, so the hard links are resolved to the current working directory.

This resulted in `std::fs::hard_link` being called with an incorrect source path (implementation [here](https://github.com/alexcrichton/tar-rs/blob/a1c3036af48fa02437909112239f0632e4cfcfae/src/entry.rs#L509)), leading to the following error:

```
No such file or directory (os error 2) when hard linking ./share/lima/templates/_images/ubuntu-24.04.yaml to /home/ryan/.local/share/mise/installs/lima/1.1.1/share/lima/templates/_images/ubuntu-lts.yaml
```

### Solution

The fix is to replace `Entry::unpack` with [`Entry::unpack_in`](https://docs.rs/tar/latest/tar/struct.Entry.html#method.unpack_in).

This method correctly uses the destination path as the base for unpacking contents, including hard links, ensuring that `std::fs::hard_link` receives the correct source path.

While `Archive::unpack` could also solve this issue, sticking with `Entry::unpack_in` on a per-entry basis allows us to retain the existing progress bar functionality.

However, `opts.strip_components` cannot be parsed directly, so I added logic to move the entries to upper directories.
This might slow down the process, but I couldn't find other solutions for this.

### Notes

- We might be able to use `std::env::set_current_dir` to change the current working directory, but that would be a global change and could cause issues in other parts of the code.

- The use of `Entry::unpack_in` mitigates the security issues that `absolutize_virtually` was introduced in #5279 to prevent, so it is no longer necessary.

### DIFF
--- a/registry.toml
+++ b/registry.toml
@@ -225,10 +225,8 @@ biome.backends = [
 ]
 biome.test = ["biome --version", "Version: {{version}}"]
 bitwarden.backends = ["aqua:bitwarden/clients", "asdf:vixus0/asdf-bitwarden"]
-bitwarden.test = [
-    "bw --version",
-    ""
-] # TODO: change version_filter to version_prefix in aqua registry
+# TODO: change version_filter to version_prefix in aqua registry
+# bitwarden.test = ["bw --version", ""] # latest version doesn't contain sha256
 bitwarden-secrets-manager.backends = [
     "ubi:bitwarden/sdk[tag_regex=^bws,exe=bws]",
     "asdf:asdf-community/asdf-bitwarden-secrets-manager"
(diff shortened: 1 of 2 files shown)


risu729:
Thanks! I'll have a look.

PR #5467, opened 2025-06-29
fix(registry): use aqua backends for all available tools

risu729:
This PR adds `aqua` backend before `ubi` backend for available tools.

I used this script to modify `registry.toml` automatically, but reviewed the change and added tests to the tools.

```bash
#!/bin/bash

# This script finds tools in registry.toml that use 'ubi' as their primary backend
# and have a corresponding package in the local aqua-registry. It then updates
# the registry.toml file in-place by prepending the correct 'aqua:' backend
# to the backends list for each matched tool.

yq --output-format=json . registry.toml | \
jq -c '.tools | to_entries[] | select(.value.backends[0]? | (type == "string" and startswith("ubi:"))) | {tool_key: .key, ubi_backend: .value.backends[0]}' | \
while IFS= read -r tool_json; do
    tool_key=$(echo "$tool_json" | jq -r '.tool_key')
    ubi_backend=$(echo "$tool_json" | jq -r '.ubi_backend')

    package_path=$(echo "$ubi_backend" | sed 's/^ubi://' | sed 's/\[.*\]//')
    aqua_registry_file="./aqua-registry/pkgs/${package_path}/registry.yaml"

    if [ -f "$aqua_registry_file" ]; then
        new_aqua_backend="aqua:${package_path}"
        echo "Found match for tool '${tool_key}'. Prepending backend '${new_aqua_backend}' to registry.toml." >&2
        sed -i "s|\"${ubi_backend}\"|\"${new_aqua_backend}\", \"${ubi_backend}\"|" registry.toml
    fi
done

taplo fmt
toml-sort -i settings.toml --spaces-indent-inline-array 4
toml-sort -i registry.toml --spaces-indent-inline-array 4

echo "Finished updating registry.toml." >&2
```

### DIFF
--- a/registry.toml
+++ b/registry.toml
@@ -225,10 +225,7 @@ biome.backends = [
 ]
 biome.test = ["biome --version", "Version: {{version}}"]
 bitwarden.backends = ["aqua:bitwarden/clients", "asdf:vixus0/asdf-bitwarden"]
-bitwarden.test = [
-    "bw --version",
-    ""
-] # TODO: change version_filter to version_prefix in aqua registry
+# bitwarden.test = ["bw --version", "{{version}}"] # latest version doesn't contain sha256
 bitwarden-secrets-manager.backends = [
     "ubi:bitwarden/sdk[tag_regex=^bws,exe=bws]",
     "asdf:asdf-community/asdf-bitwarden-secrets-manager"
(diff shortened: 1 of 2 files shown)


registry.toml
```diff
@@ -782,6 +826,8 @@ ghalint.backends = [
 ]
 ghalint.test = ["ghalint version", "{{version}}"]
 ghc.backends = ["asdf:mise-plugins/mise-ghcup"]
+# takes forever
+# ghc.test = ["ghc --version", "The Glorious Glasgow Haskell Compilation System, version {{version}}"]
```
risu729, line 829: Somehow this comments were placed below so moved to the relevant position.

PR #5474, opened 2025-06-30
fix(aqua): try input version before prepending v

risu729:


risu729:
This will break packages relying on the `v` prefix in `version_constraint`.

e.g.
`Version in ["v2.12.0", "v2.14.9", "v2.18.1", "v2.29.3"]` ([Byron/dua-cli](https://github.com/aquaproj/aqua-registry/blob/cdbc24f2374efda9d3be3091426dd2aaa55b18b1/pkgs/Byron/dua-cli/registry.yaml#L11))
`Version == "rc/v2.15.4.1"` ([commercialhaskell/stack](https://github.com/aquaproj/aqua-registry/blob/cdbc24f2374efda9d3be3091426dd2aaa55b18b1/pkgs/commercialhaskell/stack/registry.yaml#L635))

PR #5486, opened 2025-07-01
fix(aqua): handle hard links in aqua packages (attempt #2)

risu729:
Reverts jdx/mise#5485


risu729:
I confirmed `test_swift_slow` passes.

src/file.rs
```diff
@@ -675,75 +674,66 @@ pub fn untar(archive: &Path, dest: &Path, opts: &TarOptions) -> Result<()> {
...
-        trace!("extracting {}", display_path(&path));
-        entry.unpack(&path).wrap_err_with(err)?;
-        if let Some(pr) = &opts.pr {
-            pr.set_length(entry.raw_file_position());
+        // if let Some(pr) = &opts.pr {
+        //     pr.set_position(total);
+        // }
+    }
+    match opts.strip_components {
+        0 => {}
```
jdx, line 720: unique use of match here, but makes sense to me!

risu729: Thanks, but I just moved that part...

PR #5532, opened 2025-07-07
fix: mise panic in removed directory

roele:
Fixes #5525 by exiting before the unsafe `path_absolutize::update_cwd()` if the current directory does not exist.

Currently uses the standard `handle_err`  method which leads to following output. I have a feeling we should be more specific and show something like `The current working directory does not exist`. 
WDYT @jdx?

```
Error: 
   0: No such file or directory (os error 2)

Location:
   src/main.rs:108

```


src/main.rs
```diff
@@ -104,6 +104,9 @@ fn main() -> eyre::Result<()> {
 async fn main_() -> eyre::Result<()> {
     color_eyre::install()?;
     install_panic_hook();
+    if let Err(err) = std::env::current_dir() {
+        handle_err(err.into())?;
+    }
     unsafe {
         path_absolutize::update_cwd();
```
jdx: could we instead do the check here?

what I'm thinking is at least some subset of mise should ideally work even if the directory doesn't exist, at least things like `mise -v`

roele: Since `path_absolutize::update_cwd()` is the culprit, i assume we would need to move this somewhere inside `Cli::run` to have `version` or `-v` working.

jdx: huh? I'm just saying only call update_cwd() if there actually is a cwd

roele: ah yes sure, question for me is what the side-effects will be or where it will panic after

jdx:
yeah it'll probably break later, but we can fix that another time

PR #5534, opened 2025-07-07
feat: Add `self_update_available` to `mise doctor` output

joehorsnell:
Related to hverlin/mise-vscode#129.

### DIFF
--- a/src/cli/doctor/mod.rs
+++ b/src/cli/doctor/mod.rs
@@ -7,6 +7,7 @@ use crate::backend::backend_type::BackendType;
 use crate::build_time::built_info;
 use crate::cli::version;
 use crate::cli::version::VERSION;
+use crate::cli::self_update::SelfUpdate;
 use crate::config::{Config, IGNORED_CONFIG_FILES};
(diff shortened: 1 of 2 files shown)


src/cli/doctor/mod.rs
```diff
@@ -174,6 +179,7 @@ impl Doctor {
         #[cfg(unix)]
         info::inline_section("activated", yn(env::is_activated()))?;
         info::inline_section("shims_on_path", yn(shims_on_path()))?;
+        info::inline_section("self_update_available", yn(SelfUpdate::is_available()))?;
```
jdx, line 182: doesn't seem very valuable here... but I also think we should be consistent

joehorsnell: > doesn't seem very valuable here... but I also think we should be consistent

Yeah I know what you mean.

I can see how the output of `mise doctor` vs `mise doctor --json` _could_ be different, if required, but I think "The Principle of Least Surprise" would suggest they should be the same, just a different format.

Issue #5524, opened 2025-07-06
MISE_GITHUB_TOKEN and GITHUB_TOKEN behavior


jdx:
### Discussed in https://github.com/jdx/mise/discussions/5523

<div type='discussions-op-text'>

<sup>Originally posted by **JFenstermacher** July  6, 2025</sup>
I think the way GitHub token configurations are handled is incorrect. If I'm missing context, feel free to correct.

## Expected Behavior

When `MISE_GITHUB_TOKEN` is set, I would expect that `mise` would use the GitHub Token specified to download tools using it to prevent rate-limiting. It would only be used during tool installation, and not impact other behaviors.

## Actual Behavior

When `mise` runs, it tries to detect a valid GitHub Token configuration by checking the `MISE_GITHUB_TOKEN`, `GITHUB_TOKEN`, and `GITHUB_API_TOKEN` variables. If any are found, it then sets the all of the environment variables to the one found.

## Downstream Consequences

When `mise` runs in CI it relies on the shims by default, this means that every tool call is setting the `MISE_GITHUB_TOKEN` to `GITHUB_TOKEN`, and this leads to unintuitive behavior.

### Context

My day job has a GitHub token setup that allows tokens with different permissions to be requested based on OIDC JWT claims of a workflow. This allows us to have a terraform setup where plans can run with readonly permissions, and applies can run with mutation permissions based on an environment that has an approval.

We rely on the `GITHUB_TOKEN` configuration to be able to grant the permissions for the GitHub terraform provider. We must rely on the `GITHUB_TOKEN` environment variable, because if you pass the token into the plan explicitly (via a variable), then the plan will save that token in the state. Applies are always based on a written plan for us in CI, as to ensure we can inspect the plan prior to approval (this is best practice, and non-negotiable).

Additionally, we have an internal ASDF plugin I maintain. In order for `mise` to resolve our internal ASDF plugin I must provide `mise` a `GITHUB_TOKEN` with cross-repository read.

### The problem

I would expect that when I set `MISE_GITHUB_TOKEN`, then `mise` would use that token to handle installations. I set that env variable to our cross-repo read token.

Unexpectedly, in a later step terraform applies are failing when provided a `GITHUB_TOKEN` environment configuration that would allow for writing, because `mise` is overwriting `GITHUB_TOKEN` with the `MISE_GITHUB_TOKEN` which is the cross-repo read token.

### Shortest example

This would be the minimal example that would show off what I'm talking about:

```yaml
jobs:
[...]

Issue #2910, opened 2024-11-04
Mise does not work if GITHUB_TOKEN is not valid public Github token


dal13002:
**Describe the bug**
mise doesn't work by default when `GITHUB_TOKEN` is set but not a valid token to public github. We use Github Enterprise Server and set the `GITHUB_TOKEN` environment variable on our shell in order to interact with it. When talking to public GIthub (ie pull public repos from github.com) we do not want to use that environment variable but I believe due to this change that token is always send: https://github.com/jdx/mise/pull/2815 . This causes a problem because this is not a valid public Github token and we get 401 Unauthorized when using mise

**To Reproduce**
1. Set `GITHUB_TOKEN` to some non valid public github token (this will replicate setting a token for Github Enterprise Server)
2. Run `mise install bat --verbose`
3. We get an output similar to:
```
DEBUG mise::cli ARGS: mise install bat --verbose
DEBUG mise::config config: ~/.mise.toml
DEBUG mise::toolset install_versions: bat@latest
DEBUG mise::http GET https://api.github.com/repos/sharkdp/bat/releases/tags/0.24.0
DEBUG reqwest::connect starting new connection: https://api.github.com/
DEBUG mise::http GET https://api.github.com/repos/sharkdp/bat/releases/tags/0.24.0 401 Unauthorized
DEBUG mise::backend::ubi Failed to get release for ubi:sharkdp/bat@0.24.0, trying with 'v' prefix: HTTP status client error (401 Unauthorized) for url (https://api.github.com/repos/sharkdp/bat/releases/tags/0.24.0)
DEBUG ubi Parsed --project sharkdp/bat = sharkdp / bat
DEBUG ubi exe name = bat
DEBUG ubi install path = /home/<user>/.local/share/mise/installs/bat/0.24.0/bin/bat
DEBUG ubi adding GitHub token to GitHub requests
DEBUG reqwest::connect starting new connection: https://api.github.com/
DEBUG mise::backend::ubi Failed to install with ubi version 'v0.24.0': HTTP status client error (401 Unauthorized) for url (https://api.github.com/repos/sharkdp/bat/releases/tags/v0.24.0), trying with 'ubi:sharkdp/bat@0.24.0'
DEBUG ubi Parsed --project sharkdp/bat = sharkdp / bat
DEBUG ubi exe name = bat
DEBUG ubi install path = /home/<user>/.local/share/mise/installs/bat/0.24.0/bin/bat
[...]

Issue #3070, opened 2024-11-18
Empty MISE_GITHUB_TOKEN is not valid


Methy659:
**Describe the bug**
If you have a invalid GITHUB_TOKEN set, setting MISE_GITHUB_TOKEN to empty does not skip the rest of the checks. - same issue as: https://github.com/jdx/mise/issues/2910

The changes made in https://github.com/jdx/mise/pull/2915/files#diff-4da8e8ee5c65a27bd048a2f2e55de0a9c75e87618fc02e17b235cc5413e625a0R160 does not respect empty value for MISE_GITHUB_TOKEN.

**To Reproduce**
https://github.com/jdx/mise/issues/2910#issuecomment-2468863234

```Dockerfile
# Base image
FROM debian:bullseye-slim

# Install dependencies
RUN apt-get update && apt-get install -y \
    curl \
    bash \
    gnupg \
    && apt-get clean && rm -rf /var/lib/apt/lists/*

# Install mise
RUN curl https://mise.run | MISE_INSTALL_PATH=/usr/local/bin/mise sh && \
    echo 'eval "$(mise activate bash)"' >> /etc/profile && \
    mkdir -p .local/share/mise

# Set environment variables
ENV MISE_GITHUB_TOKEN=""
ENV GITHUB_TOKEN="invalid_token"

# Set cmd to attempt installing a package using mise
CMD bash -c "/usr/local/bin/mise --version && /usr/local/bin/mise install github-cli@2.4.0"

```

<img width="1497" alt="image" src="https://github.com/user-attachments/assets/ad60b4a4-840e-4842-9401-2ef93ca9f798">

**Expected behavior**
Treats MISE_GITHUB_TOKEN as a valid value even if it’s empty, stopping further lookups.

PR #5546, opened 2025-07-09
fix: do not overwrite github tokens environment variables

risu729:
Fixes https://github.com/jdx/mise/issues/5524.

https://github.com/jdx/mise/pull/3114 introduced the overwrite behaviour to resolve https://github.com/jdx/mise/issues/2910 and https://github.com/jdx/mise/issues/3070.
It was to ignore an invalid `GITHUB_TOKEN` if `MISE_GITHUB_TOKEN` is set to an empty string, but it was necessary only because we used to call `ubi` from the CLI.

For asdf plugins, this should be enough to overwrite `GITHUB_TOKEN`.
https://github.com/jdx/mise/blob/ce78f04bfb338d218df90301a093b01b1d67d13a/src/plugins/asdf_plugin.rs#L426

I'm not sure about vfox plugins, but I don't think this PR will change any behaviours because it seems `vfox.rs` not using the tokens.

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -205,62 +205,28 @@ pub static PATH_NON_PRISTINE: Lazy<Vec<PathBuf>> = Lazy::new(|| match var(&*PATH
 });
 pub static DIRENV_DIFF: Lazy<Option<String>> = Lazy::new(|| var("DIRENV_DIFF").ok());
 pub static GITHUB_TOKEN: Lazy<Option<String>> = Lazy::new(|| {
-    let token = var("MISE_GITHUB_TOKEN")
+    var("MISE_GITHUB_TOKEN")
         .or_else(|_| var("GITHUB_API_TOKEN"))
         .or_else(|_| var("GITHUB_TOKEN"))
         .ok()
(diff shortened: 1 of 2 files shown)


src/env.rs
```diff
@@ -205,62 +205,28 @@ pub static PATH_NON_PRISTINE: Lazy<Vec<PathBuf>> = Lazy::new(|| match var(&*PATH
 });
 pub static DIRENV_DIFF: Lazy<Option<String>> = Lazy::new(|| var("DIRENV_DIFF").ok());
 pub static GITHUB_TOKEN: Lazy<Option<String>> = Lazy::new(|| {
-    let token = var("MISE_GITHUB_TOKEN")
+    var("MISE_GITHUB_TOKEN")
```
jdx: we don't use them often but I feel this is an area where unit tests might be better than e2e. You'd need to refactor it into a function somehow

jdx: that said if it's too much of a pain don't worry about it

risu729: Thanks! I'll write the test.

risu729: I added the tests, but not really sure if they're good. Could you take a look?

PR #5556, opened 2025-07-10
feat(registry): add yarn aqua config

robsonpeixoto:


risu729:
I think it only supports yarn v1.
https://github.com/aquaproj/aqua-registry/issues/32299

PR #5562, opened 2025-07-11
fix(aqua): align version resolution logic in list_bin_paths

risu729:
Aligns the version resolution logic (from canonicalised version to GitHub tag) in `list_bin_paths` with the one in `install_version_`.

I think this doesn't change the behaviour for offline introduced in https://github.com/jdx/mise/pull/5195.

> TODO: instead of caching it would probably be better to create this as part of installation

Regarding this TODO, I agreed with it and tried to work on it, but was not sure how to.
I'd like to ask for some advice.

I thought of storing `bin_paths.msgpack.z` in the installation path, `tv.install_path()`, while the installation, and reading it when `list_bin_paths` is called.
Not to break the installations with old versions, or just in case `bin_paths.msgpack.z` is removed somehow, I think we need to fall back to the current implementation if it's not found.

However, then it would be the same as the current implementation, except that `list_bin_paths` is called in the installation and the cache is created.

### DIFF
--- a/src/backend/aqua.rs
+++ b/src/backend/aqua.rs
@@ -720,7 +720,7 @@ impl AquaBackend {
     fn srcs(
         &self,
         pkg: &AquaPackage,
-        install_path: &PathBuf,
+        install_path: &Path,
         v: &str,
     ) -> Result<Vec<(PathBuf, PathBuf)>> {
         let files: Vec<(PathBuf, PathBuf)> = pkg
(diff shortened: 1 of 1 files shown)


risu729:
My apologies, I should've marked it as a draft.

I thoght it was because it accesses GitHub API to list bin paths, but not really sure why it also slowed down the install itself.

PR #5573, opened 2025-07-11
test: Add e2e test for java + jib installation

mpsanchis:
Trying to have a failing PR that shows behaviour of discussion https://github.com/jdx/mise/discussions/5570

### DIFF
--- a/e2e/backend/test_java_direct_dependencies
+++ b/e2e/backend/test_java_direct_dependencies
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+#
+# Have java and jib in the dependencies
+cat >.mise.toml <<EOF
+[tools]
+jib = "0.13.0"
+java = "openjdk-17.0.2"
+
+[settings]
+jobs = 1
+
+EOF
+
+# Install the tools
+mise install
+
+# Assert that both tools are available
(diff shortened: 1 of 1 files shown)


e2e/backend/test_java_direct_dependencies
```diff
@@ -0,0 +1,19 @@
...
+[settings]
+jobs = 1
+
+EOF
+
+# Install the tools
+mise install
+
+# Assert that both tools are available
+assert_contains "which jib" "/mise/installs/jib/0.13.0/"
```
jdx: I would use `mise x`, this isn't going to work as mise is not activated in e2e tests. Look at the other e2e tests since they pretty much all use `mise x`

mpsanchis:
Hi @jdx I managed to get it fail the way I wanted: https://github.com/jdx/mise/actions/runs/16264199046/job/45916205726

It's a bit hacky (maybe you find a more elegant way), but proves the point.

In the meantime, I've opened this PR: https://github.com/mise-plugins/mise-jib/pull/1
It's not the solution to the base problem, but it helps me in my current situation, and I think it won't hurt others, because Jib can't work without Java in any case.

PR #5579, opened 2025-07-12
feat: custom backends through plugins

jdx:
Add support for vfox plugins that implement backend hooks, allowing
plugins to provide custom implementations for version listing, installation,
and environment setup. This enables more flexible plugin architectures
while maintaining backward compatibility with traditional vfox plugins.

- Add BackendExecEnvContext, BackendInstallContext, and BackendListVersionsContext
- Enhance VfoxBackend to support both traditional and backend-only plugins
- Add new backend hook types in vfox crate
- Include comprehensive test coverage for vfox backend functionality
- Update plugin system to handle vfox backend plugin types

### DIFF
--- a/src/backend/vfox.rs
+++ b/src/backend/vfox.rs
@@ -110,7 +110,7 @@ impl Backend for VfoxBackend {
                     return Ok(tv);
                 }
                 Err(e) => {
-                    return Err(e.wrap_err("Backend install method failed"));
+                    return Err(e).wrap_err("Backend install method failed");
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


risu729:
I'm a bit confused about the backend name.
I thought the repository name was used, but there's also `PLUGIN.name` in `metadata.lua`.
Is `metadata.lua` just intended to be used in the hooks, not by mise?

Also, if I were to make a Homebrew backend, for example, it would be better if I could use `brew:git` in `tools` instead of `vfox-homebrew:git`.
However, that seems difficult, as plugin names are currently determined by their directory names.

jdx:
bugbot run

jdx:
bugbot run

PR #5584, opened 2025-07-12
chore(release): mark a release as draft until assets are added

risu729:
This marks the release as a draft until the assets are successfully added.

However, since `release-r2.sh` runs before it, `https://mise.run` will still fail.
I think we need to move it to `post-release.sh` and run it after the assets creation.


jdx:
it's fine to add this but I'm not that worried about that step failing, I don't think its ever happened before

PR #5631, opened 2025-07-14
fix(static): support `strip_components` for zip files

risu729:
Fixes https://github.com/jdx/mise/discussions/5604#discussioncomment-13750422.

~~It's a bit weird, but only `file::untar` supports `strip_components` option.
It internally calls `file::unzip`, so some codes call `file::untar` even for `.zip`.~~

~~However, `inspect_tar_contents`, or `open_tar` in it, doesn't support `.zip`, so the auto detection doesn't work.~~

~~Can I refactor it to support `strip_components` by `file::unzip`?~~

### DIFF
--- a/src/backend/static_helpers.rs
+++ b/src/backend/static_helpers.rs
@@ -95,7 +95,10 @@ pub fn install_artifact(
     let ext = file_path.extension().and_then(|s| s.to_str()).unwrap_or("");
     let format = file::TarFormat::from_ext(ext);
     if format == file::TarFormat::Zip {
-        file::unzip(file_path, &install_path)?;
+        let zip_opts = file::ZipOptions {
+            strip_components: strip_components.unwrap_or(0),
+        };
+        file::unzip(file_path, &install_path, &zip_opts)?;
     } else if format == file::TarFormat::Raw {
         // Copy the file directly to the bin_path directory or install_path
         if let Some(bin_path_template) = opts.get("bin_path") {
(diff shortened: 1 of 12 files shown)


src/backend/static_helpers.rs
```diff
@@ -113,7 +111,7 @@ pub fn install_artifact(
     } else {
         // Auto-detect if we need strip_components=1 before extracting
         // Only do this if strip_components was not explicitly set by the user
-        if strip_components.is_none() {
+        if strip_components.is_none() && format != file::TarFormat::Zip {
```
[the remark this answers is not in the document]
risu729: `inspect_tar_contents` doesn't support zip format.

src/backend/static_helpers.rs
```diff
@@ -95,7 +95,10 @@ pub fn install_artifact(
     let ext = file_path.extension().and_then(|s| s.to_str()).unwrap_or("");
     let format = file::TarFormat::from_ext(ext);
     if format == file::TarFormat::Zip {
-        file::unzip(file_path, &install_path)?;
+        let zip_opts = file::ZipOptions {
```
jdx: can you add a todo for implementing the auto_detection for strip_components? (or implement it lol)

risu729: I (or Gemini) implemented it

risu729:
Let me know if it's better to separate the test files

PR #5632, opened 2025-07-14
feat: support extracting 7z archives for static backends

yjoer:
This update uses file::un7z, which is currently gated behind Windows. ~Another known limitation of this approach is that strip_components is not supported.~

### DIFF
--- a/src/file.rs
+++ b/src/file.rs
@@ -965,21 +965,16 @@ pub fn inspect_7z_contents(archive: &Path) -> Result<Vec<(String, bool)>> {
     Ok(top_level_components.into_iter().collect())
 }
 
+#[cfg(not(windows))]
+pub fn inspect_7z_contents(_archive: &Path) -> Result<Vec<(String, bool)>> {
+    unimplemented!("7z format not supported on this platform")
+}
+
 /// Determines if strip_components=1 should be applied based on archive structure
 pub fn should_strip_components(archive: &Path, format: TarFormat) -> Result<bool> {
     let top_level_entries = match format {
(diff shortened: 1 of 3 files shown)


jdx:
I think @risu729 is looking into this for zip

yjoer:
Uh-oh, I refactored in the opposite direction. Nevermind, I will refactor again when #5631 lands.

yjoer:
I might need another 7z archive with a nested directory structure to test split components.

jdx:
If you create one I can put it up so you can fetch it with the http backend

src/file.rs
```diff
@@ -917,10 +944,42 @@ pub fn inspect_zip_contents(archive: &Path) -> Result<Vec<(String, bool)>> {
...
         TarFormat::Zip => inspect_zip_contents(archive)?,
+        TarFormat::SevenZip => {
+            #[cfg(windows)]
+            {
+                inspect_7z_contents(archive)?
+            }
+
+            #[cfg(not(windows))]
+            {
+                bail!("7z format not supported on this platform");
```
jdx: use unimplemented!, no reason not to panic in this case. it'd also be cleaner to have separate functions, e.g.:

```rust
#[cfg(not(windows))]
pub fn inspect_7z_contents(archive: &Path) -> Result<Vec<(String, bool)>> {
  unimplemented!("")
}
```

yjoer: Applied in https://github.com/jdx/mise/pull/5632/commits/b67227618491d5d2538fd7b1f12992bca36552d3

PR #5635, opened 2025-07-14
Evaluate task dependency syntax clarity

jdx:
Add support for specifying environment variables within task dependencies using an object syntax.

This implements the requested feature to allow tasks to be run with different environment variable values when specified as dependencies. The object syntax `{task = "test", env = {PROP = "a"}}` was chosen over a string parsing approach like `depends = ["PROP=a test"]` for clarity, type safety, and extensibility, as per user preference and design principles.


jdx:
too many changes

PR #5638, opened 2025-07-14
feat(tasks): Add environment variable directives for mise tasks

jdx:
Add support for `_.file`, `_.path`, `_.source` directives and `false` value for removal in `[tasks.<name>.env]` configurations.

This feature allows loading environment variables from files directly within task definitions, making it easier to synchronize mise tasks with IDE configurations as requested in #4762.

BREAKING CHANGE: there is a minor breaking change. The internal structure of the `env` key in tasks has to change from an object to array. Unfortunately I don't see a way around this because directives can be specified multiple times so it has to be an array. This is *only* exposed via `mise task info --json` so you are only affected if using the `env` key in that command. If this is an issue for you, please let me know, perhaps we can brainstorm on a solution.


jdx:
bugbot run

jdx:
bugbot run

src/config/env_directive/mod.rs
```diff
@@ -294,7 +316,6 @@ impl EnvResults {
                                 if redact {
                                     r.redactions.push(k.clone());
                                 }
-                                r.env_remove.insert(k.clone());
```
jdx, line 297: no idea why this was set but seems like it was wrong

glasser:
This appears to cause a regression using `_.path` with `node`.
https://github.com/jdx/mise/discussions/5913

ssbarnea:
I think that this change introduced a regression that breaks even the documentation example from https://mise.jdx.dev/environments/  for loading multiple env files.

```toml
[env]
_.file = [
    # Load env from the json file relative to this config file
    '.env.json',
    # Load env from the dotenv file at an absolute path
    '/User/bob/.env',
    # Load env from the yaml file relative to this config file and redacts the values
    { path = ".secrets.yaml", redact = true }
]
```
```shell
> mise config
mise ERROR TOML parse error at line 19, column 5
   |
19 |     { path = ".env.secrets", redact = true }
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
invalid type: map, expected a string

```


I have no idea how to make it work now. I only want to redact only entries from one of the env files.

PR #5666, opened 2025-07-17
fix(nushell): fix `get -i` deprecation

JoaquinTrinanes:
Since https://github.com/nushell/nushell/pull/16007, the recommended flag to avoid erroring on missing fields is --optional.

To avoid compatibility issues, a new `cell-path` is built from the provided one, explicitly setting the `optional` prop to `true`.

This makes the change backwards-compatible.

More specifically, this PR fixes `nushell` showing a warning on every launch due to the syntax being deprecated. This deprecation isn't released (yet), but changing this ahead of time in a backward-compatible way should prevent future headaches :)

### DIFF
--- a/src/shell/nushell.rs
+++ b/src/shell/nushell.rs
@@ -64,8 +64,9 @@ impl Shell for Nushell {
           }}
 
           def --env add-hook [field: cell-path new_hook: any] {{
+            let field = $field | split cell-path | update optional true | into cell-path
             let old_config = $env.config? | default {{}}
-            let old_hooks = $old_config | get $field --ignore-errors | default []
+            let old_hooks = $old_config | get $field | default []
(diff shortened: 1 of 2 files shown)


jdx:
bugbot run

JoaquinTrinanes:
Just pushed the fixed revision ([diff](https://github.com/jdx/mise/compare/3627d55029ab4bcd68a7d111ae6a248b8390a1f8..4d3f210c08272c25953a19cb9b071e779a40eefd)).

Another alternative would be to have the `cell-paths` already be optional, like so:

```nushell
add-hook $.hooks?.pre_prompt $mise_hook # hooks?.pre_prompt would work as well, string is cast to cell-path automatically
add-hook $.hooks?.env_change.PWD $mise_hook
```

PR #5668, opened 2025-07-17
Revert "fix(aqua): improve warnings for packages without repo_owner and repo_name (#5644)"

jdx:
This reverts commit ad365393ff9700cfb37043bc98ecaca456fa2079.


jdx:
it's a weird bug I'm not sure is worth adding an e2e test around or not, I think we probably should since the "v" prefixing stuff has always caused issues. Maybe check to make sure all "action-validator" versions start with a digit?

risu729:
My apologies for the regression again.
I'll add a test, but I realized I don't understand about multiline strings in bash well, so I'll figure it out later.

PR #5672, opened 2025-07-17
fix: don't follow symlink to ignore symlinks from deletion

risu729:
Follow up of https://github.com/jdx/mise/pull/5662.
The copilot suggestion was wrong; `path.metadata()` follows symlinks, so use [`path.symlink_metadata()`](https://doc.rust-lang.org/beta/std/fs/fn.symlink_metadata.html) to return the metadata of the path itself.


jdx:
ai fooled me god

PR #5708, opened 2025-07-18
test(registry): enable disabled tests

risu729:
`maven` test fails because `mise-versions` doesn't include the latest `3.9.11`, but the maven only provides the latest download link.
Should we remove it from `mise-versions` to avoid errors?
https://dlcdn.apache.org/maven/maven-3/
https://github.com/jdx/mise-versions/commits/main/docs/maven

### DIFF
--- a/registry.toml
+++ b/registry.toml
@@ -1153,7 +1153,7 @@ grain.backends = [
     "ubi:grain-lang/grain[tag_regex=^grain-]",
     "asdf:mise-plugins/mise-grain"
 ]
-grain.test = ["grain --version", "{{version}}"]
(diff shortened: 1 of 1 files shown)


jdx:
any idea what is going on with maven?

risu729:
Sorry I wrote about it in the PR description.

`maven` test fails because `mise-versions` doesn't include the latest `3.9.11`, but maven only provides the latest download link.
Should we remove it from `mise-versions` to avoid errors?
https://dlcdn.apache.org/maven/maven-3/
https://github.com/jdx/mise-versions/commits/main/docs/maven

PR #5711, opened 2025-07-19
Change the primary choice for PHP to adwinying/php

adwinying:
As per @roele 's suggestion [in this discussion](https://github.com/jdx/mise/discussions/4720#discussioncomment-13812028), I've added my PHP repository as a primary source for PHP. 

This source just takes precompiled binaries from [static-php-cli](https://github.com/crazywhalecc/static-php-cli) and repackage it into Github releases, making it compatible with ubi. No dependencies needed, no compiling need, just good old PHP.

This should solve the other discussions regarding PHP:
- https://github.com/jdx/mise/discussions/5094
- https://github.com/jdx/mise/discussions/4560
- https://github.com/jdx/mise/discussions/4566

### DIFF
--- a/registry.toml
+++ b/registry.toml
@@ -2035,7 +2035,7 @@ periphery.backends = [
 ]
 perl.description = "self-contained, portable perl binaries"
 perl.backends = ["aqua:skaji/relocatable-perl", "asdf:ouest/asdf-perl"]
-php.backends = ["asdf:mise-plugins/asdf-php", "vfox:mise-plugins/vfox-php"]
+php.backends = ["ubi:adwinying/php", "asdf:mise-plugins/asdf-php", "vfox:mise-plugins/vfox-php"]
 pinact.description = "pinact is a CLI to edit GitHub Workflow and Composite action files and pin versions of Actions and Reusable Workflows. pinact can also update their versions and verify version annotations"
 pinact.backends = ["aqua:suzuki-shunsuke/pinact", "ubi:suzuki-shunsuke/pinact"]
 pinact.test = ["pinact version", "{{version}}"]


jdx:
thanks so much for working on this @adwinying. I know a lot of people will be excite about this (chief among them, @laluka). That said, I have concern about the supply chain security of this since we would have 2 participants in the supply chain. It's nothing against you personally but it doubles the chance of either one of you being compromised in some way.

It's not clear to me what changes you needed to make on top of @crazywhalecc's releases—I see he has github releases on the php assets. Are those changes you might be able to make a PR to in order to have them upstream? Or are these the sort of changes that mise/ubi might be able to handle on our end? Also see the new github backend which we have more control over than ubi so might also be an option to make more compatible with these tarballs.

jdx:
> The releases available in [static-php-cli](https://github.com/crazywhalecc/static-php-cli) are for the tool itself to compile PHP into a standalone binary

ah ok I missed that part.

I think what probably would be best then would be for this infra to be owned either by mise-plugins or by @crazywhalecc if we want to include this into the registry. My preference wouldn't be under me just because I don't know much about php so I'd be relying on PRs from the community entirely pretty much however I would still be happy to do it if you thought that was the best option.

adwinying:
Let's see if @crazywhalecc responds, else I think mise-plugins is the way to go.

By the way, if we were to go for the mise-plugins route, how would you envision this to work?
Would it mean a new plugin alongside the existing vfox and asdf variants?

jdx:
no I'm thinking it would just live in that account even though it isn't a plugin

crazywhalecc:
I don't know much what this repo does, but the pre-built PHP binaries in dl.static-php.dev is built by [static-php-cli-hosted](https://github.com/static-php/static-php-cli-hosted) GitHub Actions, and anyone can use them.

However, there is one point that you may need to be noted: in the bulk, common, and minimal releases, Linux is built using the musl-static method. That is, although it does not require any dependencies, it cannot load any external extensions (such as xdebug, etc.).

jdx:
Where this has worked really well was for python where @indygreg and @charliermarsh put together https://github.com/astral-sh/python-build-standalone which is used by different version-managers to host precompiled pythons in various flavors. I'd love to see something like that for PHP.

What you've proposed sounds like we would maintain our own GHA that use setup-static-php however that's not ideal since I'm not a php developer—nor is anyone on the advisory board so our ability to maintain it would be very limited. There would also need to be an important decision made about what exactly the default flavor would contain and I have no idea. That's not a good decision to leave up to the community either but would be best to have an expert in PHP able to make good decisions about that.

Am I understanding right the plan is to host precompiled php binaries in GHA alongside the deb/rpm packages? If so, that sounds like the right path forward to me.

Anyhow, I very much appreciate your work on this since whichever way we go here it's still huge for mise users. I just have to be realistic about what I'm able to maintain and how well I can do it.

crazywhalecc:
> Am I understanding right the plan is to host precompiled php binaries in GHA alongside the deb/rpm packages? If so, that sounds like the right path forward to me.

That's the best choice in the future, but it's still in very early stages.

It is possible that we may also provide a stable glibc-based static PHP binary in next major version of static-php. In short, if you want to use static-php now and avoid maintenance issues, just use the pre-built `bulk` binary mentioned above (or the package provided in the far future) directly.

> I just have to be realistic about what I'm able to maintain and how well I can do it.

The same goes for me🧐 But making it easier for anyone to use PHP and distribute PHP apps is also the original intention of this project. I'm open to help if you have any other questions.

risu729:
I think `type: http` package of Aqua would be suitable for this use-case, so I asked the author of Aqua in https://github.com/orgs/aquaproj/discussions/3204#discussioncomment-13825449.

However, he reminded me that we need to consider the environment variables.
I also mentioned it there, but it seems the asdf plugin sets `COMPOSER_HOME` environment variable to support composer's global binaries. https://github.com/asdf-community/asdf-php/pull/55
I'm also not familiar with PHP, so I'd like to ask about some thoughts on this.
The [vfox plugin](https://github.com/version-fox/vfox-php) doesn't seem to set `COMPOSER_HOME`.
Also, as I read the [docs](https://getcomposer.org/doc/03-cli.md#composer-home), it seems we don't need to separate `COMPOSER_HOME` for each PHP version.

If that's the case, I think we need to modify the asdf or vfox plugin to use the prebuilt binaries, as both `ubi` or `aqua` backends don't support setting environment variables.

jdx:
I imagine we'd do what we do for other tools and have a `MISE_PHP_COMPILE` setting

risu729:
> > Does this mean we'll add core:php? I think the settings cannot be used by plugins.
> 
> good point, it'd have to be a tool option but I think that's probably fine. Arguably it might have been better to use tool options for these configs even in core tools but I think the settings predate tool options. General support for modifying options via env vars would probably be useful at some point.

I see, I agree with it.

syhol:
for some reason `mise use ubi:adwinying/php` is giving me the the wrong binary on apple silicon.
```console
❯ file /Users/simon/.local/share/mise/installs/ubi-adwinying-php/8.4.10/php
/Users/simon/.local/share/mise/installs/ubi-adwinying-php/8.4.10/php: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), BuildID[sha1]=536b9c29fd70365f8cd18f1485a99ce5312e5e98, statically linked, no section header
```

Its weird because if I use ubi directly (`ubi -v --project adwinying/php`) or [download from the release](https://github.com/adwinying/php/releases/download/v8.4.8/php-8.4.8-macos-aarch64.tar.gz) directly, it works fine.

jdx:
what about github backend?

BrianGilbert:
> for some reason `mise use ubi:adwinying/php` is giving me the the wrong binary on apple silicon.
> 
> ```
> ❯ file /Users/simon/.local/share/mise/installs/ubi-adwinying-php/8.4.10/php
> /Users/simon/.local/share/mise/installs/ubi-adwinying-php/8.4.10/php: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), BuildID[sha1]=536b9c29fd70365f8cd18f1485a99ce5312e5e98, statically linked, no section header
> ```
> 
> Its weird because if I use ubi directly (`ubi -v --project adwinying/php`) or [download from the release](https://github.com/adwinying/php/releases/download/v8.4.8/php-8.4.8-macos-aarch64.tar.gz) directly, it works fine.

Same happened to me: https://github.com/adwinying/php/issues/1

BrianGilbert:
> You can also look into the static binaries provided by https://php.new that are purposes built to work with the most popular PHP framework, Laravel.

this only packages php 8.4, so not really an option I think

risu729:
Thank you for the `COMPOSER_HOME` explanation.

In my understanding, the best solution is to add the `MISE_PHP_COMPILE` option using static binaries to the vfox plugin.
Since `static-php-cli` provides several variations for the build, I think it would be better if we also support that in the option.

Adding Aqua packages works, but then we need to select the binaries installation and the previous method by `php` and `vfox:php`, which is not ideal.

PR #5725, opened 2025-07-21
fix(npm): use bin/ as bin_paths when installed with bun on windows

risu729:
Fixes https://github.com/jdx/mise/discussions/5709.

The cause was npm installs the bin in the root of the install path, but bun installs it in the `bin/` dir.

However, this will cause issues when changing the `npm.bun` options after the installation because `list_bin_paths` reads `npm.bun` when it's called, not when the tools are installed.
I remember you were not willing to save the bin paths while the installation, but I don't have any other ideas to resolve the issue.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -266,12 +266,12 @@ jobs:
           path: target/debug
       - run: ls target\debug
       - run: Add-Content $env:GITHUB_PATH "$env:GITHUB_WORKSPACE\target\debug"
-      - uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4
-        with:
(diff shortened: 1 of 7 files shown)


risu729:
> Thanks @risu729! The unit test shows that this will work correctly via `mise x`, but will this also create the shim executable in the expected location so that npm tools can be called without needing to use `mise x`?

I think so, as they use the same logic in my understanding.
However, since the test fails and I'm not sure for the cause, something should be wrong.

W1M0R:
At least your test clearly demonstrates the problem we are experiencing and that it seems to be related to bun. But I agree, I can't see why your fix isn't solving the issue:

image

jdx:
bugbot run

risu729:
I missed that the environment was not cleaned for each tests.

PR #5727, opened 2025-07-21
fix(registry): set matching_regex for glab on Windows to pick the correct asset

risu729:


registry.toml
```diff
@@ -1038,7 +1038,13 @@ gitui.backends = ["aqua:extrawurst/gitui", "asdf:looztra/asdf-gitui"]
...
 gitversion.backends = ["aqua:gittools/gitversion", "ubi:gittools/gitversion"]
 glab.backends = [
-    "ubi:gitlab-org/cli[provider=gitlab,exe=glab]",
+    {full = "ubi:gitlab-org/cli[provider=gitlab,exe=glab]", platforms = [
+        "linux",
+        "macos"
+    ]},
+    {full = "ubi:gitlab-org/cli[provider=gitlab,exe=glab,matching_regex=^glab\\.exe$]", platforms = [
+        "windows"
+    ]},
```
jdx, line 1047: have we actually done this before? I'm not sure it'll actually work

risu729: I think it's the first one.
I tested it locally by reordering them and confirmed it worked.

PR #5736, opened 2025-07-21
doc(config): fix alias section name

malept:
`[aliases]` doesn't actually exist according to [the schema](https://github.com/jdx/mise/blob/v2025.7.18/schema/mise.json), but `[alias]` does.


jdx:
I believe it is supported so we should update the schema

malept:
> I believe it is supported so we should update the schema

I'm running 2025.07.18, and I get this message when I add `[aliases]` to `~/.config/mise/config.toml`:

```
mise WARN  unknown field in ~/.config/mise/config.toml: aliases
```

PR #5745, opened 2025-07-22
feat(registry): use npm backend for yarn

mrazauskas:
Please consider including the npm backend for yarn.

The `mise-yarn` plugin does not work on Windows (https://github.com/mise-plugins/mise-yarn/issues/12). I just discovered that the binary installed from npm works on Windows (at least in CI).

The package on npm has no readme and strange name. But this is the same bundled binary. See `corepack`s config:

- the `mise-yarn` plugin loads yarn binary from https://repo.yarnpkg.com, https://github.com/nodejs/corepack/blob/0b492c9c97e4b3d5e9a139eeedb931860127b470/config.json#L137
- few lines below `@yarnpkg/cli-dist` package is included as an alternative, https://github.com/nodejs/corepack/blob/0b492c9c97e4b3d5e9a139eeedb931860127b470/config.json#L152


jdx:
I think it would be better to make a vfox plugin for yarn. I'm not sure how this might impact linux/macos users

registry.toml
```diff
@@ -2889,7 +2889,8 @@ yamlscript.backends = [
     "asdf:mise-plugins/mise-yamlscript"
 ]
 yamlscript.test = ["ys --version", "{{version}}"]
-yarn.backends = ["asdf:mise-plugins/mise-yarn"]
+yarn.backends = ["npm:@yarnpkg/cli-dist", "asdf:mise-plugins/mise-yarn"]
```
mrazauskas, line 2895: Several other entries have npm backend specified before asdf. For instance `pnpm`. And next to the npm backend a comment is included `# uses prebuilt binaries`. Shall I add it here too?

risu729: Sorry for the confusion. I just wrote it to indicate that the installation is relatively fast.

risu729:
It's because they return `404 Not Found` for `HEAD` requests.

```
$ xh --headers https://repo.yarnpkg.com/4.9.2/packages/yarnpkg-cli/bin/yarn.js
HTTP/2.0 200 OK
access-control-allow-origin: *
cf-ray: 964f4e5348e0aacf-SYD
content-encoding: gzip
content-type: application/javascript
date: Fri, 25 Jul 2025 23:06:07 GMT
server: cloudflare
vary: Accept-Encoding

$ xh HEAD --headers https://repo.yarnpkg.com/4.9.2/packages/yarnpkg-cli/bin/yarn.js
HTTP/2.0 404 Not Found
access-control-allow-origin: *
cf-ray: 964f4ea08d05aacf-SYD
content-length: 9
content-type: text/plain;charset=UTF-8
date: Fri, 25 Jul 2025 23:06:19 GMT
server: cloudflare
vary: Accept-Encoding
```

risu729:
Since `xh HEAD https://repo.yarnpkg.com/tags` also fails with 404, I think they are blocking `HEAD` requests.

Corepack started to use `https://repo.yarnpkg.com/{}/packages/yarnpkg-cli/bin/yarn.js` instead of `https://raw.githubusercontent.com/yarnpkg/berry/%40yarnpkg/cli/{}/packages/yarnpkg-cli/bin/yarn.js` in https://github.com/nodejs/corepack/pull/36.
Since they say `repo.yarnpkg.com` is a cache behind Cloudflare, I think we can use `https://raw.githubusercontent.com/yarnpkg/berry/%40yarnpkg/cli/{}/packages/yarnpkg-cli/bin/yarn.js` instead.

However, since using `repo.yarnpkg.com` is generally better, we cannot change the upstream `aqua-registry`.
I'm thinking of not sending `HEAD` requests if the URL candidates are only one, but not sure if it's possible yet.

mrazauskas:
Thanks for taking a look. Shall I open a separate issue, perhaps?

By the way, the [`mise-yarn` plugin](https://github.com/mise-plugins/mise-yarn/blob/5a1cb2785e8d8d9e9182152932de64085bfe2bbd/bin/install#L105) is also using `https://repo.yarnpkg.com`.

risu729:
I opened an issue https://github.com/yarnpkg/berry/issues/6863.
Also, I created a PR https://github.com/jdx/mise/pull/5800. It will allow `mise use aqua:yarnpkg/berry@4.9.2`.

risu729:
> I think it would be better to make a vfox plugin for yarn. I'm not sure how this might impact linux/macos users

We lack the `.yvmrc` support, but since it's deprecated since Feb 2022, I think it won't be a problem.

risu729:
I think this can be merged now, if we are okay with dropping support for `.yvmrc` and yarn v1.

meenie:
For anyone the Yarn 1.x gang and using `yarn = "1.22.22"` in their tools list, you need to change it to `asdf:yarn = "1.22.22"` for it to start working again.

PR #5788, opened 2025-07-25
feat(registry): add `hatch`

hasansezertasan:
Add [hatch](https://hatch.pypa.io/latest/) to the registry.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -80,7 +80,7 @@ jobs:
       - if: matrix.runs-on != 'ubuntu-latest'
         run: rustup target add ${{matrix.target}}
       - if: matrix.runs-on == 'ubuntu-latest'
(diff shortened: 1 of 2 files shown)


jdx:
why default to pipx?

hasansezertasan:
> why default to pipx?

My mistake, let me fix it 🤓.

risu729:
`extract_all` and `bin_path` will fix the ubi error, but since it has `python` in its `bin` directory, I think we should not use it.
It would cause a similar issue with https://github.com/jdx/mise/discussions/4675.

```
$ mise x ubi:pypa/hatch[extract_all=true,bin_path=bin] -- hatch --version
Hatch, version 1.14.1
$ mise x ubi:pypa/hatch[extract_all=true,bin_path=bin] -- which python
/home/risu/.local/share/mise/installs/ubi-pypa-hatch/hatch-v1.14.1/bin/python
```

PR #5800, opened 2025-07-27
fix(aqua): avoid unnecessary head requests in version resolution

risu729:
This change avoids unnecessary `HEAD` requests when resolving the Aqua package version.

Since we download the file after getting its URL, we generally don't need to check for its existence beforehand.
The only case that requires a check is when the URLs for both `v1.2.3` and `1.2.3` exist and are different.

As a side effect, this also resolves an issue https://github.com/jdx/mise/pull/5745#issuecomment-3116564192, which was caused by `repo.yarnpkg.com` rejecting `HEAD` requests.


jdx:
bugbot run

PR #5807, opened 2025-07-27
feat(registry): add trunk metalinter

daveio:
Add trunk.io metalinter support with aqua backend, description, and test.
The trunk metalinter provides comprehensive code quality checks via various sub-linters.


daveio:
The failure of `registry/test-tool-4` is correct. `trunk` can't be installed as `latest`. It needs a version specifier. I'm currently using `1.3.4`. If that's specified, it works fine. I don't know any way to tell `mise` about this, but if there is one, let me know and I'll fix it.

PR #5847, opened 2025-07-30
feat(tool-stubs): improve binary path detection in tool stub generator

jdx:
## Summary
- Adds smart binary detection logic that automatically finds the most likely executable when an exact match isn't found
- Detects and sets bin paths for platform-specific URLs
- Consolidates common bin paths to global level when all platforms use the same binary
- Improves error messages when binary detection fails

## Binary Detection Strategies

The generator now uses multiple strategies to find the correct binary:
1. Prefer executables in `bin/` or `sbin/` directories
2. Look for executables containing the tool name
3. Use single executable if only one exists
4. Prefer less nested paths

## Test plan
- [x] Run `mise run test:e2e test_generate_tool_stub test_generate_tool_stub_archive` - all tests pass
- [x] Manual testing with various tools shows correct binary detection

## Examples

Before (would fail without exact match):
```bash
$ mise g tool-stub test-gh --url 'https://github.com/cli/cli/releases/download/v2.45.0/gh_2.45.0_linux_amd64.tar.gz'
# Error: No executable found with exact filename 'test-gh'
```

After (automatically finds bin/gh):
```bash
$ mise g tool-stub test-gh --url 'https://github.com/cli/cli/releases/download/v2.45.0/gh_2.45.0_linux_amd64.tar.gz'
# Successfully generates stub with bin = "bin/gh"
```

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/docs/cli/generate/tool-stub.md
+++ b/docs/cli/generate/tool-stub.md
@@ -51,7 +51,7 @@ Examples: --platform-bin windows-x64:tool.exe --platform-bin linux-x64:bin/tool
 
 Binary path within the extracted archive
 
-If not specified and the archive is downloaded, will only auto-detect if an exact filename match is found
+If not specified and the archive is downloaded, will auto-detect the most likely binary
 
 ### `--skip-download`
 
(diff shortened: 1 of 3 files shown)


jdx:
bugbot run

PR #5854, opened 2025-07-31
docs: fix uv_venv_create_args reference for python

jasonraimondi:
Hey, disclaimer is I am a noob with `uv`, but when setting uv up following the mise + uv configurations, I kept having errors with `uv_create_args`, it appears to me that it is a typo and should actually be `uv_venv_create_args`.

Hopefully this is helpful for others. Feel free to merge or close if this is incorrect, no worries.

Thanks for this awesome tool. Been using since the rtx days.


jdx:
ah thanks @hverlin, somehow I also was thinking global settings for some reason

PR #5862, opened 2025-08-01
fix(node): Add NodeJS maintainer Antoine du Hamel's new GPG key

chadlwilson:
NodeJS 22.18.0 and 24.5.0 releases currently cannot be installed as looks like a maintainer regenerated their key.

See https://github.com/nodejs/node/pull/58877

Workaround for folks in the meantime: (please double check the key matches the PR https://github.com/nodejs/node/pull/58877/files rather than blindly trusting me :-) )

```shell
gpg --keyserver hkps://keys.openpgp.org --recv-keys 5BE8A3F6C8A5C01D106C0AD820B1A390B168D356
```

Believe we need to keep the old key as trusted as old releases won't be able to be installed.

When I re-ran `fetch-gpg-keys` I'm not sure why the other keys look so different now; possibly maintainers updated them for some reason; didn't dig through comparing them.


jdx:
it'd be nice if we had a way of only using older keys for older versions

chadlwilson:
> it'd be nice if we had a way of only using older keys for older versions

Well ideally the whole signing ecosystem should support that. Key is revoked at date x and nothing signed after that can be validated. But that requires secure timestamp server, proper revocations and trusted intermediary (or intermediaries) who can revoke/expire a key ... like the signing cert or modern attestation ecosystems?

jdx:
> Can it be this change does not work on macOS? The latest release of mise (2025.8.0 macos-arm64 (2025-08-01)) fails to validate gpg key when I try to use either node@22.18.0 or node@22.17.0. All worked fine before upgrading, but now it fails with older versions of Node.js too. (Also in CI jobs on macOS fail. Works with other OS.)

I can repro, working on a fix

jdx:
not sure yet

mrazauskas:
@jdx Just to confirm, all works now with the latest release. Thanks for a quick fix.

jdx:
it'd probably make sense for us to pull from that repo as well, though I think this is the first time we've actually needed to modify this since it was written (I could be wrong though)

PR #5863, opened 2025-08-01
fix(registry): add aqua backends to tools

risu729:


registry.toml
```diff
@@ -2235,7 +2278,13 @@ reckoner.backends = [
...
 redis.backends = ["asdf:mise-plugins/mise-redis"]
 redis-cli.backends = ["asdf:mise-plugins/mise-redis-cli"]
-redo.backends = ["ubi:barthr/redo", "asdf:chessmango/asdf-redo"]
+redo.description = "Redo is the ultimate tool to create reusable functions from your history in an interactive way"
+redo.backends = [
+    "aqua:barthr/redo",
+    "ubi:barthr/redo",
+    "asdf:chessmango/asdf-redo"
+]
+redo.test = ["which redo", "redo"]
```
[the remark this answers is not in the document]
risu729: This works, but yeah, since the install path contains the tool name, the test is meaningless.

registry.toml
```diff
@@ -685,10 +700,13 @@ dhall.backends = ["asdf:mise-plugins/mise-dhall"]
...
 diffoci.backends = ["aqua:reproducible-containers/diffoci"]
 diffoci.test = ["diffoci --version", "diffoci version v{{version}}"]
+difftastic.description = "a structural diff that understands syntax"
 difftastic.backends = [
-    "ubi:wilfred/difftastic[exe=difft]",
+    "aqua:Wilfred/difftastic",
+    "ubi:Wilfred/difftastic[exe=difft]",
     "asdf:volf52/asdf-difftastic"
 ]
+difftastic.test = ["difft --version", "Difftastic {{version}}"]
```
jdx, line 709: we're currently having issues where if I do a few PRs at a time we hit rate limits which is getting exacerbated by including more tests.
I don't have a great solution to this problem unfortunately. I don't think this means we shouldn't include tests though.

registry.toml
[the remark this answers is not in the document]
jdx: cursor isn't aware that it isn't checking the entire output

registry.toml
risu729: Maybe we can only test changed tools?
We cannot notice the broken tools anymore, but I think we can run scheduled tests to the main branch instead.

jdx: yeah I think that could work. Though I think I'd rather just test on the release branch since that's the one I check. I kind of want to get rid of tests on main since I don't really watch that branch.

PR #5877, opened 2025-08-01
feat(registry): add jjui

TyceHerrman:


TyceHerrman:
adds jjui to registry: https://github.com/idursun/jjui

PR #5885, opened 2025-08-02
registry: add git-lfs (aqua:git-lfs/git-lfs)

TyceHerrman:
Adding git-lfs.  "expected-output-pattern" in .test has system variables, but unsure how to account for those.


jdx:
just leave that part out

PR #5887, opened 2025-08-02
fix(registry): update kubecolor to use kubecolor/kubecolor

Darwiner:
* Change `kubecolor` to use to https://github.com/kubecolor/kubecolor instead of https://github.com/hidetatz/kubecolor as that project has been archived/deprecated.
* The aqua package entry exists in aqua at https://github.com/aquaproj/aqua-registry/blob/main/pkgs/kubecolor/kubecolor/registry.yaml and a `mise install aqua:kubecolor/kubecolor` runs successfully.

### DIFF
--- a/registry.toml
+++ b/registry.toml
@@ -1638,8 +1638,9 @@ kubebuilder.backends = [
 ]
 kubecm.description = "Manage your kubeconfig more easily"
 kubecm.backends = ["aqua:sunny0826/kubecm", "asdf:samhvw8/asdf-kubecm"]
-kubecolor.description = "colorizes kubectl output"
-kubecolor.backends = ["aqua:hidetatz/kubecolor", "asdf:dex4er/asdf-kubecolor"]
+kubecolor.description = "Colorize your kubectl output"
+kubecolor.backends = ["aqua:kubecolor/kubecolor", "asdf:dex4er/asdf-kubecolor"]
+kubecolor.test = ["kubecolor version --client=true", "Client Version: v"]
 kubeconform.description = "A FAST Kubernetes manifests validator, with support for Custom Resources"
 kubeconform.backends = [
     "aqua:yannh/kubeconform",


risu729:
There are some issues, so the test fails.
- `kubecolor` doesn't work at all if `kubectl` is not installed, so I think we should add it to `depends`.
  https://github.com/kubecolor/kubecolor/issues/220
- `kubecolor version` doesn't print kubecolor version if the coloured output is disabled, because then it just passes the command to `kubectl`.
  We need to use `--force-colors` to print it.
  https://github.com/kubecolor/kubecolor/blob/3cdd29b27bda4227a73e52c711682a698b0b2c37/command/runner.go#L88-L99

Could you update the registry to use this?
I tested it locally with `use_versions_host=0` and it worked.
```toml
kubecolor.description = "Colorize your kubectl output"
kubecolor.depends = ["kubectl"]
kubecolor.backends = ["aqua:kubecolor/kubecolor", "asdf:dex4er/asdf-kubecolor"]
kubecolor.test = [
  "kubecolor version --client=true --force-colors",
  "\u001b[96mKubecolor Version\u001b[0m: \u001b[93m{{version}}\u001b[0m",
]
```

However, `use_versions_host` is enabled in the test, so the test will fail.
This is because the current versions list is generated for `hidetatz/kubecolor`.
https://github.com/jdx/mise-versions/blob/main/docs/kubecolor

risu729:
Sorry, I forgot to mention, but I think we need to disable the test for now by commenting it out.

registry.toml
```diff
@@ -1641,8 +1641,10 @@ kubebuilder.backends = [
 ]
 kubecm.description = "Manage your kubeconfig more easily"
 kubecm.backends = ["aqua:sunny0826/kubecm", "asdf:samhvw8/asdf-kubecm"]
-kubecolor.description = "colorizes kubectl output"
-kubecolor.backends = ["aqua:hidetatz/kubecolor", "asdf:dex4er/asdf-kubecolor"]
+kubecolor.description = "Colorize your kubectl output"
+kubecolor.backends = ["aqua:kubecolor/kubecolor", "asdf:dex4er/asdf-kubecolor"]
+kubecolor.depends = ["kubectl"]
+kubecolor.test = ["kubecolor --kubecolor-version", "v{{version}}"]
```
risu729:
```suggestion
# kubecolor.test = ["kubecolor --kubecolor-version", "v{{version}}"] # disable until mise-versions will be updated
```

PR #5914, opened 2025-08-04
registry: add bob (aqua:MordechaiHadad/bob)

TyceHerrman:
Add neovim manager bob: https://github.com/MordechaiHadad/bob


registry.toml
```diff
@@ -288,6 +288,13 @@ bitwarden-secrets-manager.backends = [
 black.description = "The uncompromising Python code formatter"
 black.backends = ["aqua:psf/black"]
 black.test = ["black --version", "black, {{version}}"]
+bob.description = "A version manager for neovim"
+bob.backends = [
+    "aqua:MordechaiHadad/bob",
```
risu729:
```suggestion
    # "aqua:MordechaiHadad/bob",  # +x permission not set
```

Unlike Aqua, mise doesn't create a symlink so the files must have `+x` permission. See https://github.com/jdx/mise/discussions/5552#discussioncomment-13723271
`ubi` backend can handle it, so please comment it out for now.

PR #5925, opened 2025-08-05
registry: add pay-respects (aqua:iffse/pay-respects)

TyceHerrman:
Add pay-respects package: https://codeberg.org/iff/pay-respects


jdx:
not notable enough

PR #5972, opened 2025-08-10
fix: couldn't download node binary in Alpine, even if it exists in the mirror url

Hazer:
I was trying to use in Alpine the following build:

https://unofficial-builds.nodejs.org/download/release/v24.5.0/node-v24.5.0-linux-x64-musl.tar.gz

So I set:

```sh
export MISE_NODE_MIRROR_URL=https://unofficial-builds.nodejs.org/download/release/
export MISE_NODE_FLAVOR=musl
export MISE_NODE_COMPILE=false

# and/or

mise settings set node.mirror_url https://unofficial-builds.nodejs.org/download/release/
mise settings set node.flavor musl
mise settings set node.compile 0
```

But `mise install` would **always** fail trying to download `https://unofficial-builds.nodejs.org/download/release/v24.5.0/node-v24.5.0.tar.gz` that don't exist.

The main issue was that my `settings.node.compile false` was being overridden by the distro `all_compile` override, and it would then try to download the source code for compilation, and fail, took me a long while to figure that out. The real fix was just avoiding the `all_compile` override if  `node.compile` was set by the user manually. Added a `info` log to avoid future issues with not understand where this behavior came from.

The rest of the changes in `plugins/core/node.rs` are not necessary to solve the bug, but I would like to try pushing too:
 - I made a `fetch_binary` to reuse/deduplicate code for posix and windows, while at it made the logic about Ok explicit and about the NotFound case too; (_while the diff looks horrible, imho the final code is clearer to reason and safer now, for someone new to the codebase_)
 - it now makes explicitly clear when "precompiled node archive not found and compilation is disabled" happens.
 - Renamed `install_compiled` to `install_compiling` to make the intent of the method really clear, the first time I read the code I was confusing `install_compiled` and `install_precompiled`
 - Made the unarchiving logic reusable, I was thinking in moving to another place to let other packages use something like this, but it's just that for now;
 - ~~Made sure the errors where not encapsulated by using `Report::from` (I had a bug because of it while testing stuff that shouldn't ever happen, making an `Ok(())` be tested and treated as 404 somehow, but the previous flow allowed it when I made my broken test changes, so this could eventually cause a bug to some bad change in the future)~~ removed by autofix
 - Added 2 `info` logs that would've been crucial to understand this issue.
 - Added more `debug` logging around the areas that eventually led me to discover the issue, and areas I was under the impression could be potential troublemakers.
 
 
 The new install_precompiled and install_windows method bodies here for faster reading than the diff, just to try to defend my "unnecessary" changes easier 😆:
 ```rust
    async fn install_precompiled(
        &self,
        ctx: &InstallContext,
        tv: &mut ToolVersion,
        opts: &BuildOpts,
    ) -> Result<()> {
        match self
            .fetch_binary(ctx, tv, opts, || {
                self.extract_tarball(
                    &opts.binary_tarball_path,
                    &opts.install_path,
                    ctx,
                    1, // strip_components for binary tarball
                )
            })
            .await?
        {
            FetchOutcome::Downloaded => Ok(()),
            FetchOutcome::NotFound => {
                if Settings::get().node.compile != Some(false) {
                    self.install_compiling(ctx, tv, opts).await
                } else {
                    bail!("precompiled node archive not found and compilation is disabled")
                }
            }
        }
    }

    async fn install_windows(
        &self,
        ctx: &InstallContext,
        tv: &mut ToolVersion,
        opts: &BuildOpts,
    ) -> Result<()> {
        match self
            .fetch_binary(ctx, tv, opts, || self.extract_zip(opts, ctx))
            .await?
        {
            FetchOutcome::Downloaded => Ok(()),
            FetchOutcome::NotFound => bail!("precompiled node archive not found (404)"),
        }
    }
```

### DIFF
--- a/src/config/settings.rs
+++ b/src/config/settings.rs
@@ -78,8 +78,13 @@ static CLI_SETTINGS: Mutex<Option<SettingsPartial>> = Mutex::new(None);
 static DEFAULT_SETTINGS: Lazy<SettingsPartial> = Lazy::new(|| {
     let mut s = SettingsPartial::empty();
     s.python.default_packages_file = Some(env::HOME.join(".default-python-packages"));
-    if let Some("alpine" | "nixos") = env::LINUX_DISTRO.as_ref().map(|s| s.as_str()) {
+    let distro = env::LINUX_DISTRO.as_ref().map(|s| s.as_str());
+    if let Some("alpine" | "nixos") = distro {
         if !cfg!(test) {
+            info!(
+                "activating settings.all_compile as the {:?} distro default",
+                distro
+            );
             s.all_compile = Some(true);
         }
     }
@@ -176,7 +181,9 @@ impl Settings {
             settings.yes = true;
         }
         if settings.all_compile {
-            settings.node.compile = Some(true);
+            if settings.node.compile.is_none() {
+                settings.node.compile = Some(true);
+            }
             if settings.python.compile.is_none() {
                 settings.python.compile = Some(true);
             }
--- a/src/config/settings.rs
+++ b/src/config/settings.rs
@@ -78,13 +78,8 @@ static CLI_SETTINGS: Mutex<Option<SettingsPartial>> = Mutex::new(None);
 static DEFAULT_SETTINGS: Lazy<SettingsPartial> = Lazy::new(|| {
     let mut s = SettingsPartial::empty();
     s.python.default_packages_file = Some(env::HOME.join(".default-python-packages"));
-    let distro = env::LINUX_DISTRO.as_ref().map(|s| s.as_str());
-    if let Some("alpine" | "nixos") = distro {
+    if let Some("alpine" | "nixos") = env::LINUX_DISTRO.as_ref().map(|s| s.as_str()) {
         if !cfg!(test) {
-            info!(
-                "activating settings.all_compile as the {:?} distro default",
-                distro
-            );
             s.all_compile = Some(true);
         }
     }
--- a/src/plugins/core/node.rs
+++ b/src/plugins/core/node.rs
@@ -57,9 +57,7 @@ impl NodePlugin {
                 &opts.binary_tarball_path,
                 &opts.version,
             )
-            .await
-            .map_err(eyre::Report::from)
-        {
+            .await{
             Ok(()) => {
                 debug!("We successfully downloaded precompiled node archive");
             }
--- a/src/plugins/core/node.rs
+++ b/src/plugins/core/node.rs
@@ -57,7 +57,8 @@ impl NodePlugin {
                 &opts.binary_tarball_path,
                 &opts.version,
             )
-            .await{
+            .await
+        {
             Ok(()) => {
                 debug!("We successfully downloaded precompiled node archive");
             }
--- a/src/plugins/core/node.rs
+++ b/src/plugins/core/node.rs
@@ -27,6 +27,11 @@ pub struct NodePlugin {
     ba: Arc<BackendArg>,
 }
 
+enum FetchOutcome {
+    Downloaded,
+    NotFound,
+}
+
 impl NodePlugin {
     pub fn new() -> Self {
         Self {
@@ -34,13 +39,15 @@ impl NodePlugin {
         }
     }
 
-    async fn install_precompiled(
+    async fn fetch_binary(
         &self,
         ctx: &InstallContext,
         tv: &mut ToolVersion,
         opts: &BuildOpts,
-    ) -> Result<()> {
-        let settings = Settings::get();
+        extract: impl FnOnce() -> Result<()>,
+    ) -> Result<FetchOutcome> {
+        debug!("We will fetch a precompiled version");
+
         match self
             .fetch_tarball(
                 ctx,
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/config/settings.rs [hunk 1]
```diff
@@ -78,8 +78,13 @@ static CLI_SETTINGS: Mutex<Option<SettingsPartial>> = Mutex::new(None);
 static DEFAULT_SETTINGS: Lazy<SettingsPartial> = Lazy::new(|| {
     let mut s = SettingsPartial::empty();
     s.python.default_packages_file = Some(env::HOME.join(".default-python-packages"));
-    if let Some("alpine" | "nixos") = env::LINUX_DISTRO.as_ref().map(|s| s.as_str()) {
+    let distro = env::LINUX_DISTRO.as_ref().map(|s| s.as_str());
+    if let Some("alpine" | "nixos") = distro {
         if !cfg!(test) {
+            info!(
```
jdx: isn't this going to be shown every time mise runs?

jdx:
bugbot run

src/plugins/core/node.rs
```diff
@@ -454,10 +501,12 @@ impl Backend for NodePlugin {
         if cfg!(windows) {
             self.install_windows(ctx, &mut tv, &opts).await?;
         } else if settings.node.compile == Some(true) {
-            self.install_compiled(ctx, &mut tv, &opts).await?;
+            info!("node.compile true so we won't fetch a precompiled version");
```
jdx: we typically don't use `info!` level logs for this do we?

src/plugins/core/node.rs
```diff
@@ -52,54 +59,54 @@ impl NodePlugin {
...
+
+    fn extract_tarball(
+        &self,
+        tarball_path: &Path,
+        dest_path: &Path,
+        ctx: &InstallContext,
+        strip_components: usize,
+    ) -> Result<()> {
+        debug!("Removing destination before extraction: {:?}", dest_path);
+        file::remove_all(dest_path).map_err(|e| {
```
jdx: is this necessary?

src/plugins/core/node.rs
Hazer: I mean, if this log existed as info, I would have lost way less time, cause it was almost impossible to know why it was trying to fetch the wrong url without looking at the code at the time, I was setting node.compile false but it wasn't being respected and it wouldn't tell me it was forcing the source code route. But now that there's a fix, maybe it's really not needed, want me to remove?

src/config/settings.rs
Hazer: Thinking about that now, it may show multiple times in a single mise run 😞 
I forgot Settings::get() wasn't a singleton pattern.

I dislike how this workaround is silent but that's really not the way to do it, lol

Any suggestion or we just remove this info too?

src/plugins/core/node.rs
Hazer: The debug? Not really, I'll remove it.

src/plugins/core/node.rs
jdx: yes we don't want to add extra noise like this

src/plugins/core/node.rs
jdx: no removing the dest_path, is this the install directory? if so that'll already have been removed by mise

Hazer: I haven't changed this behavior:

- when on `install_precompiled` it will be `&opts.install_path` see [as is](https://github.com/jdx/mise/blob/cd8c65b4268d4aed9e754fabafcae5eea6c63f77/src/plugins/core/node.rs#L134-L137) 
- when on `install_compiling` it will be `opts.build_dir.parent().unwrap()` see [as is](https://github.com/jdx/mise/blob/cd8c65b4268d4aed9e754fabafcae5eea6c63f77/src/plugins/core/node.rs#L66-L69)

I'm not sure I should change this behavior, I don't see other removal calls

src/plugins/core/node.rs
```diff
@@ -27,20 +27,27 @@ pub struct NodePlugin {
...
+    async fn fetch_binary(
         &self,
         ctx: &InstallContext,
         tv: &mut ToolVersion,
         opts: &BuildOpts,
-    ) -> Result<()> {
-        let settings = Settings::get();
+        extract: impl FnOnce() -> Result<()>,
+    ) -> Result<FetchOutcome> {
+        debug!("We will fetch a precompiled version");
```
jdx: one last change, we should do something like `debug!("{self}: ...")` because otherwise it won't be easy to tell what these logs are referring to

Hazer: Done, but using `debug!("{:?}: ...", self)`

src/plugins/core/node.rs
Hazer: @jdx now I understood your doubts about it, my bad, seeing that the build broke and then https://github.com/jdx/mise/pull/5972#discussion_r2270661025 I moved the removal back out of the extract tarball to keep the same behavior as before.

I haven't tried locally running the source code install after my changes, totally at fault here.

src/plugins/core/node.rs
```diff
@@ -114,12 +116,60 @@ impl NodePlugin {
...
+        ctx: &InstallContext,
+        tv: &mut ToolVersion,
+        opts: &BuildOpts,
+    ) -> Result<()> {
+        match self
+            .fetch_binary(ctx, tv, opts, || {
+                file::remove_all(&opts.install_path).map_err(|e| {
+                    debug!("{:?}: Failed to remove {:?}: {e}", self, &opts.install_path);
+                    e
+                })?;
```
jdx:
```suggestion
```

I'm pretty sure we don't need this code. I also am not clear why the debug log was output

src/plugins/core/node.rs
```diff
@@ -131,15 +181,15 @@ impl NodePlugin {
         )
         .await?;
         ctx.pr.set_message(format!("extract {tarball_name}"));
-        file::remove_all(&opts.build_dir)?;
-        file::untar(
+        file::remove_all(&opts.build_dir).map_err(|e| {
+            debug!("{:?}: Failed to remove {:?}: {e}", self, &opts.build_dir);
```
jdx: why was this added?

PR #5973, opened 2025-08-11
registry: support usage on FreeBSD

risu729:


risu729:
I think it's a bug, I'll have a look.

PR #5981, opened 2025-08-11
docs: Update install instructions for nushell

Joniator:
[This PR](https://github.com/jdx/mise/pull/3592) should have fixed the path issues, but the problem described in [this discussion](https://github.com/jdx/mise/discussions/3555) was not fixed. 

The issue doesn't persist if the file is autoloaded, but the documentation currently describes the old way of importing the file with use. This updates the docs to write the autoload-way into the nu config

### DIFF
--- a/docs/installing-mise.md
+++ b/docs/installing-mise.md
@@ -295,14 +295,10 @@ echo 'mise activate pwsh | Out-String | Invoke-Expression' >> $HOME\Documents\Po
 
 Nu
 does [not support `eval`](https://www.nushell.sh/book/how_nushell_code_gets_run.html#eval-function)
-Install Mise by appending `env.nu` and `config.nu`:
+Install Mise by activating it into the autoload dir:
 
 ```nushell
-'
-let mise_path = $nu.default-config-dir | path join mise.nu
-^mise activate nu | save $mise_path --force
-' | save $nu.env-path --append
-"\nuse ($nu.default-config-dir | path join mise.nu)" | save $nu.config-path --append
+'^mise activate nu | save --force ($nu.data-dir | path join vendor autoload mise.nu)' | save $nu.config-path --append
 ```
 
 If you prefer to keep your dotfiles clean you can save it to a different directory then


jdx:
@JoaquinTrinanes @atty303 @samuelallan72 @finalchild @texastoland (people that have contributed to nushell)

can any of you have a glance at this and see if it seems reasonable? I don't know enough about nushell to properly review it myself

jdx:
thanks @atty303

docs/installing-mise.md
```diff
@@ -295,14 +295,10 @@ echo 'mise activate pwsh | Out-String | Invoke-Expression' >> $HOME\Documents\Po
...
-Install Mise by appending `env.nu` and `config.nu`:
+Install Mise by activating it into the autoload dir:
 
 ```nushell
-'
-let mise_path = $nu.default-config-dir | path join mise.nu
-^mise activate nu | save $mise_path --force
-' | save $nu.env-path --append
-"\nuse ($nu.default-config-dir | path join mise.nu)" | save $nu.config-path --append
+'^mise activate nu | save --force ($nu.data-dir | path join vendor autoload mise.nu)' | save $nu.config-path --append
```
samuelallan72, line 301: @jdx note that this fails if .local/share/nushell/vendor/autoload doesn't already exist (it probably doesn't by default; it wasn't created automatically for me).


```suggestion
mkdir ($nu.data-dir | path join vendor autoload mise.nu)
'^mise activate nu | save --force ($nu.data-dir | path join vendor autoload mise.nu)' | save $nu.config-path --append
```

samuelallan72:
@jdx @Joniator I tested this way according to the new instructions, but it doesn't work. The generated mise.nu exports a `main` function - this is designed to work with the `use mise.nu` format. Putting the mise.nu file in the autoload directory does indeed autoload the file, but the main function is registered in nushell as literally `main` instead of `mise`.

If we want to switch to this autoload method, we'll also need to make some changes to what is output by `mise activate nu`.

PR #6002, opened 2025-08-12
Fix setting env vars in nushell activation script

samuelallan72:
All env var setting in a nushell module must be within a single export-env block.

Addresses the remaining points of #3555

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,50 @@
 # Changelog
 
+## [2025.8.9](https://github.com/jdx/mise/compare/v2025.8.8..v2025.8.9) - 2025-08-13
+
+### 📦 Registry
+
+- add vfox-yarn as primary yarn backend by [@jdx](https://github.com/jdx) in [#5982](https://github.com/jdx/mise/pull/5982)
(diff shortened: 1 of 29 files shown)


samuelallan72:
@Joniator @elken please test this - it should address the issues you were seeing for https://github.com/jdx/mise/discussions/3555#discussioncomment-14081472 . I couldn't test easily, because I wasn't sure how to set up an environment to trigger adding extra env vars here.

Also, be sure to test by using the script in nushell as a module: `use mise.nu`.

Thanks!

src/shell/nushell.rs
```diff
@@ -131,7 +132,7 @@ impl Shell for Nushell {
     }
 
     fn prepend_env(&self, k: &str, v: &str) -> String {
-        format!("$env.{k} = ($env.{k} | prepend '{v}')\n")
```
samuelallan72, line 134: I changed this to match line 40 (use of a raw string in the SetEnv path of ActivatePrelude). It should help it be more robust against weirdly formed paths.

samuelallan72:
Thanks for checking @Joniator @elken ! :)

@jdx this is ready for your perusal. :)

PR #6011, opened 2025-08-13
feat(task): allow more #MISE comments patterns

risu729:
Resovles https://github.com/jdx/mise/discussions/5499.

mise currently supports:
- `# mise`
- `// mise`
- `#MISE`
- `//MISE`
- `::MISE`

This PR supports more patterns by making a space between `#` and `mise` optional for both cases.
- `#mise`
- `//mise`
- `::mise`
- `:: mise`
- `# MISE`
- `// MISE`
- `:: MISE`

---

Sorry, I found `# mise` is now deprecated. Was it because spaces caused problems?
https://github.com/jdx/mise/pull/2694

### DIFF
--- a/src/task/mod.rs
+++ b/src/task/mod.rs
@@ -163,7 +163,13 @@ impl Task {
         let mut task = Task::new(path, prefix, config_root)?;
         let info = file::read_to_string(path)?
             .lines()
-            .filter_map(|line| regex!(r"^(?:#|//|::) ?(?:MISE|mise) ([a-z_]+=.+)$").captures(line))
+            .filter_map(|line| {
+                regex!(r"^(?:#|//) mise ([a-z_]+=.+)$") // old deprecated syntax
+                    .captures(line)
+                    .or_else(|| {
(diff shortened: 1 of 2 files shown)


jdx:
maybe `// [MISE]` would be ok

jdx:
it's weird to me that formatters care about what comments look like

jdx:
as a short term solution you can probably wrap the mise config section with format disable comments

src/task/mod.rs
```diff
@@ -164,12 +164,14 @@ impl Task {
         let info = file::read_to_string(path)?
             .lines()
             .filter_map(|line| {
-                regex!(r"^(#|//) mise ([a-z_]+=.+)$")
+                regex!(r"^(?:#|//) mise ([a-z_]+=.+)$") // old deprecated syntax
```
jdx: I think this has been deprecated for a long time, we could probably show a warning at this point

jdx: maybe like I do with the idiomatic version warning where at a certain version 6 months from now we know to remove it altogether

PR #6013, opened 2025-08-13
registry: add cfssljson (aqua:cloudflare/cfssl/cfssljson)

disintegrator:
Adds `cfssljson`, which is a complementary program to `cfssl`, to the mise registry. Both of these programs are developed by Cloudflare here: https://github.com/cloudflare/cfssl

### DIFF
--- a/registry.toml
+++ b/registry.toml
@@ -460,6 +460,7 @@ cfssl.backends = [
 ]
 cfssljson.description = "cfssljson: Takes the JSON output from Cloudflare's cfssl and multirootca programs and writes certificates, keys, CSRs, and bundles to disk."
 cfssljson.backends = ["aqua:cloudflare/cfssl/cfssljson"]
+cfssljson.test = ["cfssljson -version", "Version: {{version}}"]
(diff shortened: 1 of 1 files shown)


disintegrator:
I'm not sure if this change will solve my problem because when I run `mise use aqua:cloudflare/cfssl/cfssljson` I get:

```
mise ERROR failed to install aqua:cloudflare/cfssl/cfssljson@1.6.5
mise ERROR unsupported env: linux/arm64
```

I'm not sure `mise use aqua:cloudflare/cfssl/cfssl` works but not when I switch to `cfssljson`.

disintegrator:
@jdx I'm not sure if this change will solve my problem because when I run `mise use aqua:cloudflare/cfssl/cfssljson` I get:

```
mise ERROR failed to install aqua:cloudflare/cfssl/cfssljson@1.6.5
mise ERROR unsupported env: linux/arm64
```

I'm not sure `mise use aqua:cloudflare/cfssl/cfssl` works but not when I switch to `cfssljson`.

PR #6023, opened 2025-08-14
registry: add prek (aqua:j178/prek)

HenryZhang-ZHY:
I've added prek to aqua-registry, so I will use it with mise.

[aquaproj/aqua-registry#39978](https://github.com/aquaproj/aqua-registry/pull/39978)


HenryZhang-ZHY:
Please take a look 🥺 @jdx

PR #6027, opened 2025-08-14
docs: fix backend_architecture docs

risu729:
- use correct package names (npm packages instead of Node.js packages, etc.)
- reorder aqua and ubi to show it's generally preferrable
- mark npm/pipx/cargo backends speed ⚠️ instead of ✅
- fix pipx backend dependency (python -> pipx)
- use yarn as an example for forcing backend because npm:prettier is the default for prettier


jdx:
bugbot run

PR #6044, opened 2025-08-16
chore: remove unused permissions in registry test

risu729:
It's added in https://github.com/jdx/mise/issues/5067, but the comment is moved to another workflow so this is no longer required.


risu729:
The `registry_comment` test fails because it's using `pull_request_target`. It should be resolved once this PR is merged.

PR #6082, opened 2025-08-20
chore: replace submodule with subtree

risu729:
ref: https://github.com/jdx/mise/discussions/6060#discussioncomment-14141845


jdx:
release-plz has some logic that needs to be updated too that updates aqua-registry during the release commit, we should squash all the changes from aqua-registry for the release branch

risu729:
As you see, the commit `Squashed 'aqua-registry/' changes from...` is not signed. I'm not sure but it might not be signed in the release PR too.

xtasks/release-plz
```diff
@@ -50,14 +48,14 @@ git add \
...
 	README.md \
-	aqua-registry \
 	default.nix \
 	packaging/rpm/mise.spec \
 	mise.usage.kdl \
 	completions \
 	man/
 git clean -df
 git checkout -B release
+git subtree pull --squash --prefix=aqua-registry https://github.com/aquaproj/aqua-registry main
```
jdx: I think this needs to be done after the commit but I could be wrong

risu729: Fixed. ~~I think the order doesn't matter~~ It does, uncommitted changes will cause an error.

risu729:
I'm not sure how many people are using [`aqua.registry_url`](https://aquaproj.github.io/docs/reference/config/#registries), but I see some problems with it.
At first, aqua supports multiple registries, and each tool needs to specify which registry to use.
https://aquaproj.github.io/docs/reference/config/#registries
Currently, I think the only reason to use this option is to use a fork of `aquaproj/aqua-registry`.

Also, we need to clear the cache (`.cache/mise/aqua-registry/`) to change this. I think we should save the `registry_url` with the cache.

Shall we remove this option, or should we support multiple registries as aqua?
I think it's useful if someone wants to install their private tool with aqua, but the github backend should be able to resolve most cases.

If we remove it, we can simply clone `aquaproj/aqua-registry` when the `baked_registry` option is disabled (to use the latest registry), or fetch each registry from GitHub.
e.g. https://raw.githubusercontent.com/aquaproj/aqua-registry/refs/heads/main/pkgs/BurntSushi/ripgrep/registry.yaml
I'm not really sure about the purpose of `mise-versions`, but if it was to avoid many API requests to GitHub, I think it's safe to remove the proxy because I don't expect many people to disable the option.

risu729:
Follow-up for the custom aqua registry discussion in this PR: opened draft PR #10179 to support multiple ordered custom aqua registries: https://github.com/jdx/mise/pull/10179

The implementation keeps the existing single-string `aqua.registry_url` syntax, adds list/comma-separated env support, scopes caches by registry URL/source, checks registries in configured order, and falls back to the baked registry when enabled. Aqua registry aliases remain registry-local; cross-registry/mise shorthand aliases should use `[tool_alias]`.

PR #6123, opened 2025-08-25
fix: Update ZLS registry.yaml

chirgjn:
https://zigtools.org/zls/releases/0.15.0/

> Release filenames have changed
> The operating system and cpu archtecture in the filename have been swapped to match the equivalent change that happened in Zig’s tarballs. armv7a also has been renamed to arm for the same reason.


risu729:
You need to send a PR to [`aquaproj/aqua-registry`](`aquaproj/aqua-registry`).
It's a clone and all changes will be overwritten on the mise release.

chirgjn:
Ref: aquaproj/aqua-registry#40606

PR #6137, opened 2025-08-27
feat(tasks): make auto outputs default

risu729:
This makes `{ auto = true }` the default for task `outputs`.
Since only setting `sources` doesn't affect anything, I think we can safely make it the default.

Even if `sources` is not set, `outputs` is set to `{ auto = true }`, but I don't think this is a problem.
The auto-tracked file won't be created if `sources` is empty, and it seems all related codes return early if `sources` is empty.

I preserved the `{ auto = true }` syntax for backwards compatibility, but I think we don't need to make it deprecated, as it's useful to make it explicit in the config.
We can remove this for now, but should I?
https://github.com/jdx/mise/blob/e5b71310016a8e9f2d73c9fca7c774f135818047/src/task/task_sources.rs#L62-L72

This will change `mise tasks info` output; the auto-tracked `outputs` path will be printed every time.


risu729:
If there're no use-cases for `outputs: []`, maybe it's simpler to make `{ auto = true }` the default behaviour for it.

jdx (approved):

jdx:
bugbot run

PR #6160, opened 2025-08-30
fix(env): allow complicated config for env._.path

risu729:


risu729:
Sorry I just haven't done this, I forgot to make it as a draft.

risu729:
Yes, I marked it as a draft.

schema/mise-task.json
```diff
@@ -550,11 +580,41 @@
                 },
                 {
                   "description": "bash scripts to load",
+                  "type": "array",
                   "items": {
-                    "description": "bash script to load",
-                    "type": "string"
-                  },
-                  "type": "array"
+                    "oneOf": [
```
jdx, line 530: could we add this as a `$def`?

risu729: Done, not using it for `env._.path` as `redact` shouldn't be shown in the suggestion.
It's allowed in the Rust code but shouldn't be allowed in the schema.

PR #6180, opened 2025-09-04
feat: include resolved sources in task templating context

the-wondersmith:
PR includes the resolved paths of a tasks's `sources` its templating context as `{{ task.sources }}`

### Example Usage

Given a "project" directory of:
```text
/tmp/mise
├── foo
│   ├── bar.txt
│   ├── baz
│   │   └── qux.example
│   └── quux
│       └── corge.grault
└── mise.toml
```

and a `mise.toml` of:

```toml
[vars]
target_source_file = "foo/quux/corge.grault"

[tasks.print-task-sources]
description = "Demo the inclusion of task sources in task templating context."
sources = ["**/*.txt", "**/baz/qux.example", "{{ vars.target_source_file }}"]
run = """
{%- for source in task.sources -%}
printf "source file: {{ source }}\n"
{%- endfor -%}
"""
```

`mise run print-task-sources` yields:

print task sources

Enabling verbose output shows source file resolution (as well as template string pass-through):

source file resolution

### DIFF
--- a/src/task/mod.rs
+++ b/src/task/mod.rs
@@ -43,6 +43,60 @@ pub use deps::Deps;
 use task_dep::TaskDep;
 use task_sources::TaskOutputs;
 
+#[inline(always)]
+fn resolve_sources<'de, D>(deserializer: D) -> Result<Vec<String>, D::Error>
+where
+    D: serde::de::Deserializer<'de>,
+{
+    let mut sources = <Vec<String> as serde::de::Deserialize>::deserialize(deserializer)?;
+
+    let mut resolved = Vec::with_capacity(sources.len());
+
+    for pattern in sources.drain(..) {
+        // pattern is considered a tera template string if it contains opening tags:
+        // - "{#" for comments
+        // - "{{" for expressions
+        // - "{%" for statements
+        if pattern.contains("{#") || pattern.contains("{{") || pattern.contains("{%") {
+            trace!("config::load::resolve_task_sources including tera template string in resolved task sources: {pattern}");
+                resolved.push(pattern);
+            continue
+        }
+
+        match glob::glob_with(
+            &pattern,
+            glob::MatchOptions {
+                case_sensitive: false,
+                require_literal_separator: false,
+                require_literal_leading_dot: false,
+            },
+        ) {
+            Err(error) => {
+                warn!("config::load::resolve_task_sources including '{pattern}' in resolved task sources, ignoring glob parsing error: {error:#?}");
+                resolved.push(pattern);
+            }
+            Ok(expanded) => {
+                for path in expanded {
+                    match path {
+                        Ok(path) => {
+                            let source = path.display();
+                            trace!("config::load::resolve_task_sources resolved source from pattern '{pattern}': {source}");
+                            resolved.push(source.to_string());
+                        },
+                         Err(error) => {
+                            let source = error.path().display();
+                            warn!("config::load::resolve_task_sources including '{source}' in resolved task sources despite: {:#?}", error.error());
+
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    Ok(resolved)
+}
+
 #[derive(Debug, Clone, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct Task {
@@ -78,7 +132,7 @@ pub struct Task {
     pub global: bool,
     #[serde(default)]
     pub raw: bool,
-    #[serde(default)]
+    #[serde(default, deserialize_with = "resolve_sources")]
     pub sources: Vec<String>,
     #[serde(default)]
     pub outputs: TaskOutputs,
@@ -492,6 +546,7 @@ impl Task {
         let ts = config.get_toolset().await?;
         let mut tera_ctx = ts.tera_ctx(config).await?.clone();
         tera_ctx.insert("config_root", &self.config_root);
+        tera_ctx.insert("task", &serde_json::json!({"sources": &self.sources }));
         Ok(tera_ctx)
     }
 


src/task/mod.rs
```diff
@@ -43,6 +43,68 @@ pub use deps::Deps;
...
+        // - "{%" for statements
+        if pattern.contains("{#") || pattern.contains("{{") || pattern.contains("{%") {
+            trace!(
+                "config::load::resolve_task_sources including tera template string in resolved task sources: {pattern}"
+            );
+            resolved.push(pattern);
+            continue;
+        }
+
+        match glob::glob_with(
```
jdx: this seems costly if there are a lot of sources

jdx: maybe we could make this a function instead?

the-wondersmith: > maybe we could make this a function instead?

Apologies for the confusion. Do you mean the... `match` block? Or "make source resolution its own function"?

the-wondersmith: FTR, I 100% agree that the implementation could be costly if there are a lot of sources, I just couldn't figure a better place for it. I did find the [`Run::sources_are_fresh`](https://github.com/jdx/mise/blob/main/src/cli/run.rs#L1127) method that *seems* to do something similar, but I had so much trouble figuring out where in the order of things it actually got called that I gave up and settled for this on the premise that it'd be better to get the idea out and then solicit input from someone more familiar with the `mise` codebase 😅.

tl;dr the feature would be hugely helpful to have I think, I'm just not 100% sure where it's best added and was hoping for guidance from the `mise` team.

jdx: you can make it a function here so the cost would only be paid when this feature is actually used https://github.com/jdx/mise/blob/5c2f066fec484f156dd24b878b6f882614a48adc/src/task/task_script_parser.rs#L114

the-wondersmith: Gotcha. Just to clarify - the suggestion / ask here is to move the actual *resolution* of source files to template rendering time, so that (instead of the example in the PR description) the cost of resolution is paid only if/when a user *wants* resolved source paths in their script, i.e.:

```toml
[vars]
target_source_file = "foo/quux/corge.grault"

[tasks.print-task-sources]
description = "Demo the inclusion of task sources in task templating context."
sources = ["**/*.txt", "**/baz/qux.example", "{{ vars.target_source_file }}"]
run = """
{%- for source in resolve(task.sources) -%}  # <-- resolution overhead is only paid when calling `resolve`
printf "source file: {{ source }}\n"
{%- endfor -%}
"""
```

Am I understanding that correctly? Hopefully I am, which means my only question is: do we want to leave the `task.sources` changes in place, or do we want to remove that as well and figure some different way of passing the list of source files to be resolved to the new at-template-render-time-only resolution function?

jdx: I think we need to remove it and figure out some other method. What I want to avoid is paying the cost of figuring out what all of the sources are when we don't need that information

jdx: unless what you're saying is we'd just store the unresolved sources? maybe we could store it in a hidden `._sources` as a way to pass it

the-wondersmith: I went and took a pass at moving source file resolution to being a tera function, and it's actually super reasonable. I've got it working on this branch, I just haven't committed it (as I wanted some feedback before I pushed up the changes).

I think the only unclear part is how the task's sources are passed to the function. At the moment, I left [the addition of `task.sources` to the tera context](https://github.com/jdx/mise/blob/4c1cd4eed4dbd7fd5f760a3cba9876e4ed6fb3da/src/task/mod.rs#L549) in place, and implemented exactly as you see it in the comment's example - `{%- for source in resolve(task.sources) -%}`.

If I'm interpreting correctly though, you probably agree that that seems a bit... *ugly*. Ideally, the function would just be something like `task_source_files()` (or whatever your preference is) and users would be able to call it as `{%- for source in task_source_files() -%}`. Is that more in line with what you were thinking?

Assuming it is, I just need some guidance on how you'd prefer to have the task sources made available to [`setup_tera_spec_for_parsing`](https://github.com/jdx/mise/blob/5c2f066fec484f156dd24b878b6f882614a48adc/src/task/task_script_parser.rs#L114). In short - `setup_...` is only called in two specific places ([here](https://github.com/jdx/mise/blob/5c2f066fec484f156dd24b878b6f882614a48adc/src/task/task_script_parser.rs#L401) and [here](https://github.com/jdx/mise/blob/5c2f066fec484f156dd24b878b6f882614a48adc/src/task/task_script_parser.rs#L439)). *Both* of the places it calls have access to the `Task` being run, so that's obviously ideal. The simplest, most straight-forward way I can think of to do it would be to just straight up pass a reference to `Task.sources` to `setup_tera_spec_for_parsing` (i.e. change `setup_tera_spec_for_parsing`'s signature from `fn setup_tera_for_spec_parsing(&self) -> TeraSpecParsingResult` to `fn setup_tera_for_spec_parsing(&self, sources: &[String]) -> TeraSpecParsingResult` and then just capture the passed sources in the closure registered with tera. That seems like setting things up to get out of hand in the future though, if/as more templating functions are added that want/need context from their associated task. It *feels* like it'd be better to do it as `fn setup_tera_for_spec_parsing(&self, task: &Task) -> TeraSpecParsingResult`.

Thoughts? If there aren't any objections to passing the task along to the setup method, I can get that done today or tomorrow, FWIW 🙂

jdx: yeah something like this?

```rust
fn setup_tera_for_spec_parsing(&self, task: &Task) -> TeraSpecParsingResult
```

I think that sounds good

the-wondersmith: > yeah something like this?
> 
> ```rust
> fn setup_tera_for_spec_parsing(&self, task: &Task) -> TeraSpecParsingResult
> ```
> 
> I think that sounds good

*Exactly* 😁. I'll get on it shortly 🫡

the-wondersmith: @jdx Do you happen to have an opinion on the name of the registered function? I *think* it'll be the first like... *task*-oriented template function in `mise` (i.e. template functions that rely on / implement reflection of the task definition in which they are called). Seems like a good idea to set a standard (start as we mean to proceed sorta deal).

the-wondersmith: @jdx Okie doke, got it all reimplemented as a tera template function `task_source_files`. Obviously happy to change the registered function name if/as needed 😅.

Give it a look over, whenever you have time? Eager to hear your thoughts.

jdx:
bugbot run

Cargo.toml
```diff
@@ -200,6 +200,7 @@ ctor = "0.4"
 insta = { version = "1", features = ["filters", "json"] }
 mockito = "1.6.1"
 pretty_assertions = "1"
+rstest = "0.26"
```
jdx: I'd rather just have more verbose tests than add a new crate for this

jdx (approved):
code seems fine, just remove the crate as mentioned

Cargo.toml
the-wondersmith: Removed the dependency and refactored the new test(s) 😅

docs/templates.md
```diff
@@ -177,6 +177,7 @@ Mise offers additional functions:
   64-character lowercase hex string.
 - `read_file(path) -> String` – Reads the contents of a file at the given path and returns
   it as a string.
+- `task_source_files() -> Vec<String>` – Returns the task's [`sources`](https://mise.jdx.dev/tasks/task-configuration.html#sources) as an array of resolved file paths. This function processes glob patterns and Tera template strings defined in the task's sources, expanding them into actual file paths. If a pattern doesn't match any files, it will be omitted from the result. Returns an empty array if no sources are configured or if no files match the patterns.
```
jdx: I think we need a new section for task functions

the-wondersmith: > I think we need a new section for task functions

@jdx can you expand on/clarify that?

jdx:
```
### Task Functions

These are functions available for tasks

- `task_source_files() -> Vec<String>`
```

we should also link to this from the task docs

the-wondersmith: > `{ ... }` we should also link to this from the task docs

@jdx ~~do you mean from the [TOML-based tasks](https://mise.jdx.dev/tasks/toml-tasks.html#arguments) page, or from the `sources` entry on the [Task Configuration](https://mise.jdx.dev/tasks/task-configuration.html#sources) page?~~

Figured if I had to ask, it was worth just doing both. Updates made and pushed 😁

jdx (approved):
looks good but needs a rebase

the-wondersmith:
@jdx rebased and pushed 😁

PR #6181, opened 2025-09-04
fix: add ansible depends on pipx

tony-sol:
ansible is installed via pipx, so pipx should be defined as ansible's dependency

### DIFF
--- a/registry.toml
+++ b/registry.toml
@@ -103,10 +103,12 @@ android-sdk.backends = ["asdf:mise-plugins/mise-android-sdk"]
 android-sdk.depends = ["java", "yq"]
 ansible.description = "ansible python package contains the core runtime and CLI tools, such as ansible and ansible-playbook and extra modules, plugins, and roles"
 ansible.backends = ["pipx:ansible[uvx=false,pipx_args=--include-deps]"]
+ansible.depends = ["pipx"]
 ansible.test = ['ansible --version', 'ansible']
 ansible-core.aliases = ["ansible-base"]
 ansible-core.description = "ansible-core python package contains the core runtime and CLI tools, such as ansible and ansible-playbook"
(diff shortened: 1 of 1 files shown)


risu729:
I tested it now, but it worked well for me. I tried setting `settings.pipx.uvx` to `true` and `false`, but it didn't matter.
Do you have `uv` or `pipx` in the `PATH` at any chance? It might be causing the issue, but I'm not sure.

Also, did this PR resolve the issue in your environment?

```
~/.ghr/github.com/risu729/dotfiles$ mise ls | rg "pipx|uv|ansible"
ansible                                 11.9.0 (missing)   ~/.config/mise/config.toml                    latest
pipx                                    1.7.1 (missing)    ~/.config/mise/config.toml                    latest
pipx:toml-sort                          0.24.2
mise WARN  missing: ansible@11.9.0
~/.ghr/github.com/risu729/dotfiles$ mise i
mise mise 2025.8.21 by @jdx –       ░░░░░░░░░░░░░░░░░░░░ 0/3 
creating virtual environment...
creating shared libraries...
upgrading shared libraries...
installing ansible from spec 'ansible==11.9.0'...
done! ✨ 🌟 ✨
```

tony-sol:
@risu729 probably i dont really get what 'depends' means

no, i dont have a pipx; i thought, that it should be installed automatically, when i install anything, which is depends on pipx :/

am i wrong?

risu729:
Oh, that's my bad. I tested with `mise install`, but I should've tested with `mise install ansible`.
Since the "dependency" in mise is just used to order the tool installations, it only works when both are installed at the same time.
It sounds like a bug because `mise use ansible` also doesn't work, but this cannot be solved by adding `depends` in `registry.toml`.

jdx:
Seems to be working as expected. depends should only affect installation order.

PR #6199, opened 2025-09-05
feat: Add ruby with devkit for Windows

JayBazuzi:


jdx:
you should create a vfox plugin for this, it's not something I'm willing to maintain

PR #6213, opened 2025-09-06
feat(cli): support scoped packages for all backend types

earlgray283:
## What

This PR enables backends other than npm to correctly parse package inputs with scopes and versions, such as '@antfu/ni@1.0.0'.

## Why

I was attempting to add a bun backend via a custom plugin, but discovered that the current parser only handles this format correctly for the npm backend. This change resolves that limitation, allowing other package managers to parse scoped packages properly.


jdx:
bugbot run

PR #6307, opened 2025-09-16
registry: add missing cargo backends

jayvdb:


registry.toml
```diff
@@ -830,7 +837,7 @@ delta.backends = [
 ]
 delta.test = ["delta --version", "delta {{version}}"]
 deno.description = "A modern runtime for JavaScript and TypeScript (Builtin plugin)"
-deno.backends = ["core:deno"]
+deno.backends = ["core:deno", "cargo:deno"]
```
risu729: I think we don't need fallbacks for `core` backends.

registry.toml
```diff
@@ -930,6 +938,7 @@ dotenv-linter.description = "Lightning-fast linter for .env files. Written in Ru
 dotenv-linter.backends = [
     "aqua:dotenv-linter/dotenv-linter",
     "ubi:dotenv-linter/dotenv-linter",
+    "cargo:dotenv-linter",
```
risu729: Any reasons for preferring `cargo` to `asdf` here?

registry.toml
```diff
@@ -2757,6 +2798,7 @@ rocq.aliases = ["coq"]
 ruby.description = "Ruby language"
 ruby.backends = ["core:ruby"]
 ruff.description = "An extremely fast Python linter and code formatter, written in Rust"
+# ruff intends to publish to crates.io soon.
```
risu729: Maybe we can add `cargo:ruff` and comment it out?

registry.toml
```diff
@@ -3196,6 +3248,7 @@ tilt.description = "Define your dev environment as code. For microservice apps o
 tilt.backends = ["aqua:tilt-dev/tilt", "asdf:eaceaser/asdf-tilt"]
 timoni.description = "Timoni is a package manager for Kubernetes, powered by CUE and inspired by Helm"
 timoni.backends = ["aqua:stefanprodan/timoni", "asdf:Smana/asdf-timoni"]
+# tiny is no longer published on crates.io
```
risu729: I believe this comment is not necessary. `tiny` is not a real tool.

registry.toml
jayvdb: I hadnt noticed any documented preferred order in the backends.
Happy to put cargo after asdf.

jdx:
bugbot run

PR #6317, opened 2025-09-18
feat(templates): join overloads for path vars/fns

Hazer:
# TBD

from
```toml
run = "{{ [config_root] | concat(with='/mise/setup-token.ts') | join_path | quote }}"
```

to shorter expressions while keeping compatibility:
```toml
run = "{{ config_root(join='/mise/setup-token.ts') | quote }}"
run = "{{ config_root | dirname(join='/something/else') }}"
```

### DIFF
--- a/src/tera.rs
+++ b/src/tera.rs
@@ -126,7 +126,7 @@ static TERA: Lazy<Tera> = Lazy::new(|| {
                 let p = Path::new(s).canonicalize()?;
                 if let Some(Value::String(join_path)) = args.get("join") {
                     let joined_path = p.join(join_path);
-                    return Ok(Value::String(joined_path.to_string_lossy().to_string()))
+                    return Ok(Value::String(joined_path.to_string_lossy().to_string()));
                 }
                 Ok(Value::String(p.to_string_lossy().to_string()))
             }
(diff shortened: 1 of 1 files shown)


Hazer:
@jdx Should I open an issue or discussion to ask if you're interested in this proposal or I can just continue this draft and we discuss it right here? Thanks

Hazer:
> I don't think it's useful enough to warrant. Can't you just do `run = "{{ config_root }}/mise/setup-token.ts"`?

@jdx Actually that’s exactly why this exists, this breaks, it won’t properly escape the spaces in the path, lol

So when any part of our config root absolute path had any spaces, like `/Users/myuser/my project/root`, run would fail to execute the file, because it would identify as a command only until the first space.

For this to work, I need to write like this:
`run = "'{{ config_root }}/mise/setup-token.ts'"` with extra single-quotes.

The actual command that I noticed the issue was exactly this:
`run = "{{ config_root }}/mise/setup-token.ts {{ flag(name='info') }}"`

The fix was:
`run = "'{{ config_root }}/mise/setup-token.ts' {{ flag(name='info') }}"`

But the fix wasn’t obvious to some of the team members, and some wanted some way to easily pipe/concat/join, to make it explicit and easy to lint too, but we also disliked the full list concat join pattern shown in the docs, we felt it was too much for a simple task, and as they felt about the extra single-quotes, it wasn’t that obvious why at first glance and this time way more verbose. And that’s how I came up with this middle ground proposal.

Hazer:
One other idea I had, was actually just having a pipe `shell_escaped`, so that I could write `{{ config_root | shell_escaped }}` and the path spaces would be slash escaped, but then I started to think about it and the behavior on Windows was really unclear, specially considering powershell and cmd, and I started to question if other chars are expected to be escaped too by using that name, so I scrapped that idea because I felt like it could become a source of bugs or non-ending requests for escape this or that.

PR #5629, opened 2025-07-14
feat: consolidate lockfile assets and add URL tracking

jdx:
This PR consolidates checksum and size fields into a unified assets section in lockfiles and introduces URL tracking for improved traceability.

Replace separate checksums maps with AssetInfo in ToolVersion and LockfileTool
Add migration logic to transform legacy lockfile formats into the new assets layout
Instrument multiple backends and plugins to record URLs, checksums, and sizes during downloads and verifications

PR #6348, opened 2025-09-20
fix: fix an issue where Swift could not be installed on arm64 Ubuntu

lish82:
Hi there! First and foremost, thank you for creating and maintaining this fantastic tool. I use it every day and really appreciate it.

I've read the [Contribution Guidelines](https://mise.jdx.dev/contributing.html) and believe this qualifies as a "something obvious" fix, so I've opened this PR. This is my first time contributing to this repository, so I'd appreciate it if you could point out any mistakes or areas for improvement.

## The Issue

I noticed that `mise install swift` fails on arm64 Ubuntu.

From my investigation, this issue seems to be caused by #5629, which changed the return value of `Settings::get().arch()` from `aarch64` to `arm64`.

## The Fix

I've adjusted the code to generate the correct download URL for Swift on arm64 Ubuntu, taking the new architecture name into account.

I couldn't find any corresponding tests for this functionality, so no tests have been modified.

---

Thanks for your time and consideration!

### DIFF
--- a/src/plugins/core/swift.rs
+++ b/src/plugins/core/swift.rs
@@ -208,8 +208,8 @@ fn platform_directory() -> String {
     } else if let Ok(os_release) = &*os_release::OS_RELEASE {
         let settings = Settings::get();
         let arch = settings.arch();
-        if os_release.id == "ubuntu" && arch == "aarch64" {
-            let retval = format!("{}{}-{}", os_release.id, os_release.version_id, arch);
+        if os_release.id == "ubuntu" && arch == "arm64" {
+            let retval = format!("{}{}-aarch64", os_release.id, os_release.version_id);
             retval.replace(".", "")
         } else {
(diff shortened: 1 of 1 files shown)


src/plugins/core/swift.rs
```diff
@@ -254,8 +254,12 @@ fn extension() -> &'static str {
...
 fn architecture(settings: &Settings) -> Option<&str> {
     let arch = settings.arch();
-    if cfg!(target_os = "linux") && arch != "x64" {
-        return Some(arch);
+    if cfg!(target_os = "linux") {
+        return match arch {
+            "x64" => None,
+            "arm64" => Some("aarch64"),
+            _ => Some(arch),
+        };
```
lish82, line 262: Since I'm not familiar with other Linux environments, I'm not sure if this implementation is suitable for anything beyond arm64 Ubuntu

PR #6368, opened 2025-09-22
registry: fix cmake

jayvdb:
See https://github.com/asdf-community/asdf-cmake/issues/4


risu729:
It seems like it's a third party release.
https://pypi.org/project/cmake/
https://github.com/scikit-build/cmake-python-distributions

PR #6376, opened 2025-09-22
fix(aqua): remove cosign.experimental

risu729:
`cosign.experimental` was removed in https://github.com/aquaproj/aqua/pull/2757. No registries are using it.
The implementation for experimental in https://github.com/jdx/mise/pull/6332 is never used, so I believe it's fine to remove it.


jdx:
bugbot run

PR #6377, opened 2025-09-22
feat(aqua): support cosign for the binary, not only for checksums

risu729:
Aqua allows cosign verification for binaries, not only for checksums.
https://aquaproj.github.io/docs/reference/registry-config/cosign

Currently, only these two packages are using it without `opts`, which doesn't work now.
https://github.com/aquaproj/aqua-registry/blob/9b2b783bba5c1637a5982f7c51ccd9d85af0e737/pkgs/k0sproject/k0s/registry.yaml#L69-L75
https://github.com/aquaproj/aqua-registry/blob/main/pkgs/technicalpickles/envsense/registry.yaml#L82-L93

`cosign.opts` is no longer supported since https://github.com/jdx/mise/pull/6332, but this PR doesn't do anything about it.


jdx:
bugbot run

risu729:
@jdx Any reasons for not using the [sigstore-rs](https://github.com/sigstore/sigstore-rs) crate?

PR #6396, opened 2025-09-24
registry: replace amplify-cli github backend with ubi

eggplants:
Currently, `mise use amplify-cli` installs `amplify-pkg-macos` / `amplify-pkg-linux` / `amplify-pkg-linux-arm64` instead of `amplify`.

Therefore, I modified it to be renamed to `amplify`.


jdx:
the test should be in registry.toml, not e2e

PR #6397, opened 2025-09-24
fix: use /c instead of -c on windows in postinstall hook

risu729:
Fixes https://github.com/jdx/mise/discussions/6388


jdx:
bugbot run

PR #6406, opened 2025-09-25
registry: revert opam/k3kcli backends to ubi

risu729:
Resolves https://github.com/jdx/mise/discussions/6402.

Neither works with the GitHub backend because the releases have archives, but we need to download non-archive files.
https://github.com/ocaml/opam/releases/tag/2.4.1
https://github.com/rancher/k3k/releases/tag/v0.3.4

We should fix it, but I don't have a good idea how to fix it.
Maybe we can only exclude small non-archive files (text files) when archive files exist, but I think this also causes some bugs.

### DIFF
--- a/registry.toml
+++ b/registry.toml
@@ -2354,8 +2354,9 @@ description = "Little helper to run CNCF's k3s in Docker"
 test = ["k3d --version", "k3d version v{{version}}"]
 
 [tools.k3kcli]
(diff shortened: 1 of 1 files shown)


jdx:
Seems like we could support non-archive binaires, no? I might have a look at this soon.

Is the tool supposed to be failing in this PR?

PR #6420, opened 2025-09-25
fix(json-schema): remove settings/additionalProperties

tpansino:
Resolves #6353

### DIFF
--- a/schema/mise.json
+++ b/schema/mise.json
@@ -313,6 +313,7 @@
     },
     "settings": {
       "type": "object",
+      "additionalProperties": false,
       "properties": {
         "activate_aggressive": {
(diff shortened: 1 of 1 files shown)


jdx:
that's not the right way to fix this

jdx:
according to gpt-5 (which I think is right):

schema/mise.json: Dropping the additionalProperties: false that sits beside the $ref does not make validation stricter; it just removes a keyword that draft‑2019‑09 ignores in that position, so extra keys are still accepted. To fail on unknown settings you need the constraint inside the target schema itself (e.g. add additionalProperties: false to #/$defs/settings, or wrap the $ref in an allOf that adds that keyword). GitHub PR #6420
Next step suggestion
Move additionalProperties: false into #/$defs/settings (or equivalent) so the validator enforces it, then re-run your schema tests.

jdx:
Yeah this looks correct to me

PR #6486, opened 2025-09-30
registry: prefer k3s from Aqua over ASDF plugin

TobiX:
I tried using the new `github` backend, but k3s has strange release artifacts :D

### DIFF
--- a/registry.toml
+++ b/registry.toml
@@ -2423,7 +2423,7 @@ description = "K3k, Kubernetes in Kubernetes, is a tool that empowers you to cre
 test = ["k3k --version", "k3kcli version v{{version}}"]
 
 [tools.k3s]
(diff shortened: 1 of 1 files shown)


registry.toml
```diff
@@ -2423,7 +2423,7 @@ description = "K3k, Kubernetes in Kubernetes, is a tool that empowers you to cre
 test = ["k3k --version", "k3kcli version v{{version}}"]
 
 [tools.k3s]
-backends = ["asdf:mise-plugins/mise-k3s"]
+backends = ["aqua:k3s-io/k3s", "asdf:mise-plugins/mise-k3s"]
```
jdx, line 2426: add a test

TobiX: Like this?

PR #6231, opened 2025-09-07
feat(lockfile): implement actual lockfile generation with platform metadata

jdx:
- Add tools_mut() and write() methods to Lockfile struct
- Implement complete lockfile generation pipeline in CLI command
- Add platform metadata fetching for tools using Backend trait methods
- Support creating/updating lockfiles with Multi-Version TOML format
- Add proper error handling for compilation issues
- Convert between toml::Value and toml_edit::Value for serialization
- Apply code formatting fixes

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude <noreply@anthropic.com>

PR #6496, opened 2025-09-30
fix(github): download assets via API to respect GITHUB_TOKEN

roele:
Currently GitHub backend is not able to download assets from private repositories since the `asset.browser_download_url` does not respect the `GITHUB_TOKEN` (Bearer token header respectively). We should use `asset.url` instead.

---

> [!NOTE]
> Switch downloads to API-backed asset URLs (token-aware), add asset name/url/url_api to lockfiles, and set correct headers for GitHub release assets.
> 
> - **Backend (GitHub/GitLab)**:
>   - Introduce `ReleaseAsset { name, url, url_api }` and propagate through resolution and install flows.
>   - Prefer existing lockfile asset info; otherwise resolve via API; store `name`, `url`, and `url_api` in `tv.lock_platforms`.
>   - Download logic: HEAD-check `url`; on failure, fall back to `url_api`; pass appropriate headers; set GitHub `accept: application/octet-stream` for `/releases/assets/`.
>   - Extend GitHub asset model with `GithubAsset.url`.
> - **Lockfile**:
>   - Extend `PlatformInfo` with `name` and `url_api` (serde read/write, conversions, tests updated).
>   - Populate new fields in various lockfile resolution helpers.
> - **Tests (e2e)**:
>   - Verify lockfile includes `name` and `url_api` alongside `url` and `checksum` for GitHub assets.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5d6e3477e598972d3855092337440b847dc10543. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/e2e/backend/test_github_url_tracking
+++ b/e2e/backend/test_github_url_tracking
@@ -29,7 +29,9 @@ assert_contains "cat mise.lock" 'backend = "github:jdx/mise-test-fixtures"'
 # Get the current platform key
 PLATFORM_KEY=$(mise x --cd . -- bash -c "echo \"\$(uname -s | tr '[:upper:]' '[:lower:]' | sed 's/darwin/macos/')-\$(uname -m | sed 's/x86_64/x64/' | sed 's/aarch64/arm64/')\"")
 assert_contains "cat mise.lock" "[tools.\"github:jdx/mise-test-fixtures\".platforms.$PLATFORM_KEY]"
+assert_contains "cat mise.lock" 'name = "hello-world-1.0.0.tar.gz"'
 assert_contains "cat mise.lock" 'url = "https://github.com/jdx/mise-test-fixtures/releases/download/v1.0.0/hello-world-1.0.0.tar.gz"'
+assert_contains "cat mise.lock" 'url_api = "https://api.github.com/repos/jdx/mise-test-fixtures/releases/assets/272317030"'
 
 echo "Lockfile after installation:"
 cat mise.lock
--- a/src/backend/github.rs
+++ b/src/backend/github.rs
@@ -23,6 +23,12 @@ pub struct UnifiedGitBackend {
     ba: Arc<BackendArg>,
 }
 
+struct ReleaseAsset {
+    name: String,
+    url: String,
+    url_api: String,
+}
+
 #[async_trait]
 impl Backend for UnifiedGitBackend {
     fn get_type(&self) -> BackendType {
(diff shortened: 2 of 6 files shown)


roele:
Guess since this adds more calls towards the infamous GitHub rate limit 😢

src/backend/github.rs
```diff
@@ -77,23 +82,24 @@ impl Backend for UnifiedGitBackend {
...
-        {
+        let asset = if let Some(existing_platform) = tv.lock_platforms.get(&platform_key) {
             debug!(
                 "Using existing URL from lockfile for platform {}: {}",
-                platform_key, existing_platform
+                platform_key,
+                existing_platform.url.clone().unwrap_or_default()
             );
-            existing_platform
+            // FIXME: we might need to add name to the lockfile as well
```
roele: @jdx Not quite sure how to proceed here. 
I guess we will need the asset name in the `PlatformInfo` (lockfile?) as well?

jdx: hmm yeah, related I was working on having a platform key for the lockfile https://github.com/jdx/mise/pull/6231

but maybe it would make sense to put the tarballs in there? not sure

e2e/backend/test_github_url_tracking
```diff
@@ -29,7 +29,8 @@ assert_contains "cat mise.lock" 'backend = "github:jdx/mise-test-fixtures"'
 # Get the current platform key
 PLATFORM_KEY=$(mise x --cd . -- bash -c "echo \"\$(uname -s | tr '[:upper:]' '[:lower:]' | sed 's/darwin/macos/')-\$(uname -m | sed 's/x86_64/x64/' | sed 's/aarch64/arm64/')\"")
 assert_contains "cat mise.lock" "[tools.\"github:jdx/mise-test-fixtures\".platforms.$PLATFORM_KEY]"
-assert_contains "cat mise.lock" 'url = "https://github.com/jdx/mise-test-fixtures/releases/download/v1.0.0/hello-world-1.0.0.tar.gz"'
+assert_contains "cat mise.lock" 'name = "hello-world-1.0.0.tar.gz"'
+assert_contains "cat mise.lock" 'url = "https://api.github.com/repos/jdx/mise-test-fixtures/releases/assets/272317030"'
```
jdx: hmm I definitely don't like not seeing the tarball url in here

roele: Yeah, the only URL containing the tarball is the `browser_download_url` which does not work for private repositories.

jdx: could we use browser_download_url if hostname == github.com?

roele: hostname is always `github.com` unless you self-host, so private repositories would still not work

jdx: ohh I see, sorry I was thinking of ghe

jdx: would it be a massive pain to make this fallback behavior? I really don't want to lose the tarball urls in the lockfile

roele: yes currently this is a breaking change since we need both url and name for it to work. we probably have to use a new `api_url` (beside the `name`) instead. question is when to use those? if we have a GITHUB_TOKEN set?

jdx: > if we have a GITHUB_TOKEN set?

I don't think this would work since we'd still want the tarball url for public repos even if GITHUB_TOKEN was set. I may wait for https://github.com/jdx/mise/pull/6231 when I can work on that more since I'll already need to make some breaking changes with lockfiles

roele: Okey reworked it a bit. 
Now we have a `name` and `url_api` derived from the release API call. `url` is the `browser_download_url` as before. We do a HEAD request to check if `url` (browser_download_url) works, if not we use the `url_api`. The `name` is used if it is set in the lockfile, if not we derive it from the `url` (like before).

jdx: great!

jdx:
bugbot run

src/backend/github.rs
```diff
@@ -325,7 +356,11 @@ impl UnifiedGitBackend {
                     )
                 })?;
 
-            return Ok(asset.direct_asset_url);
+            return Ok(ReleaseAsset {
+                name: asset.name,
+                url: asset.url,
+                url_api: asset.direct_asset_url,
+            });
```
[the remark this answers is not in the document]
roele: oh switched those up, good catch!

src/backend/github.rs
```diff
@@ -154,24 +163,34 @@ impl UnifiedGitBackend {
...
-        platform_info.url = Some(asset_url.to_string());
+        platform_info.name = Some(asset.name.clone());
+        platform_info.url = Some(asset.url.clone());
+        platform_info.url_api = Some(asset.url_api.clone());
+
+        // check if url is reachable, 404 might indicate a private repo or asset
+        let url = match HTTP.head(asset.url.clone()).await {
+            Ok(_) => asset.url.clone(),
+            Err(_) => asset.url_api.clone(),
+        };
```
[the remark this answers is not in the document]
roele: will move the header handling down to after url evaluation and move the `accept` header into the github::get_headers method.

jdx:
bugbot run

PR #6504, opened 2025-10-01
fix: set MIX_HOME and MIX_ARCHIVES when using the elixir plugin

numso:
fixes: https://github.com/jdx/mise/discussions/6478
previous implementation: https://github.com/mise-plugins/mise-elixir/blob/main/bin/exec-env
elixir docs on env vars: https://hexdocs.pm/mix/Mix.html#module-environment-variables

Elixir uses the `MIX_HOME` and `MIX_ARCHIVE` env vars to configure where it installs some global packages. This PR adds those to the elixir plugin.

This is my first time writing rust. I tried to copy the patterns I saw in the golang plugin. Please don't hesitate to give any feedback no matter how small! 

---

> [!NOTE]
> Elixir plugin now sets `MIX_HOME` and `MIX_ARCHIVES` in its exec env when unset, with e2e tests ensuring they aren’t overridden if preconfigured.
> 
> - **Elixir Plugin (`src/plugins/core/elixir.rs`)**:
>   - Implement `exec_env` to set `MIX_HOME` to ``<install>/.mix`` and `MIX_ARCHIVES` to ``<install>/.mix/archives`` when not present in pristine or config env.
>   - Add required imports (`BTreeMap`, `Toolset`, `env`).
> - **Tests (`e2e/core/test_erlang_slow`)**:
>   - Verify `MIX_HOME`/`MIX_ARCHIVES` absent before using Elixir, then present after `mise use elixir@1.17.3`.
>   - Ensure values are respected when set via `mise set` and not overridden when provided in a pristine environment.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 11c9b64c1ad68326a6cf15de57e8fc747dd112a1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/e2e/core/test_erlang_slow
+++ b/e2e/core/test_erlang_slow
@@ -10,3 +10,21 @@ assert_not_contains "mise env" "MIX_ARCHIVES"
 assert "mise use elixir@1.17.3"
 assert_contains "mise env" "MIX_HOME"
 assert_contains "mise env" "MIX_ARCHIVES"
+
+# Test that MIX_HOME is not overridden when set via mise set
+assert "mise set MIX_HOME=/mise/set/mix/home"
+assert_contains "mise env" "MIX_HOME=/mise/set/mix/home"
+assert "mise unset MIX_HOME"
+
+# Test that MIX_ARCHIVES is not overridden when set via mise set
+assert "mise set MIX_ARCHIVES=/mise/set/mix/archives"
+assert_contains "mise env" "MIX_ARCHIVES=/mise/set/mix/archives"
+assert "mise unset MIX_ARCHIVES"
+
+# Test that MIX_HOME is not overridden when set in pristine environment
+# When env var is already set, mise won't set it, so it won't appear in mise env output
+assert_not_contains "MIX_HOME=/custom/mix/home mise env" "MIX_HOME="
+
+# Test that MIX_ARCHIVES is not overridden when set in pristine environment
+# When env var is already set, mise won't set it, so it won't appear in mise env output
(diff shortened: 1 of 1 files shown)


src/plugins/core/elixir.rs
```diff
@@ -140,6 +141,28 @@ impl Backend for ElixirPlugin {
...
+        &self,
+        _config: &Arc<Config>,
+        _ts: &Toolset,
+        tv: &ToolVersion,
+    ) -> eyre::Result<BTreeMap<String, String>> {
+        let mut map = BTreeMap::new();
+        let mut set = |k: &str, v: PathBuf| {
+            map.insert(k.to_string(), v.to_string_lossy().to_string());
+        };
+        if !env::PRISTINE_ENV.contains_key("MIX_HOME") {
```
jdx: hmm I wonder if we should do something like `config.env().contains_key("MISE_ARCHIVES")`. I'm thinking someone may have a mise.toml like this:
```toml
[env]
MIX_ARCHIVES = 'foo'
[tools]
elixir = 'latest'
```

can you also modify the elixir e2e tests to check these scenarios?

numso: Done!

It looked like `config.env()` only contained env defined in a mise.toml and `env::PRISTINE_ENV` contained the actual environment. So I'm checking both. If that's not right let me know.

The only place I could find elixir e2e tests was in `test_erlang_slow` so I added some there. Is that what you were expecting or is there another place those belong?

PR #6472, opened 2025-09-29
feat(snapcraft): add snap package

phanect:
This pull request adds support for distributing `mise` as a Snap package.

- Added `snapcraft.yaml` to configure Snap packaging for `mise`
- Added GitHub Actions for testing and publishing the snap package to Snapcraft.io.
- Updated the release automation script (`xtasks/release-plz`) to automatically update the Snap package version in `snapcraft.yaml` during releases.
- Updated docs to add Snap installation instructions

You need the credentials to publish the snap package to snapcraft.io via GitHub Actions, 
Could you set the credentials by following these steps?

1. Create an Ubuntu account at https://snapcraft.io
2. [Install `snapcraft` CLI](https://documentation.ubuntu.com/snapcraft/stable/how-to/set-up-snapcraft/)
    - I guess you don't have to install LXD and Multipass.
3. Get snapcraft.io credentials via snapcraft CLI (See [the document](https://documentation.ubuntu.com/snapcraft/stable/how-to/publishing/) to learn more)
    ```sh
    snapcraft login # login to snapcraft.io with your Ubuntu account
    snapcraft register mise
    snapcraft export-login ./path/to/credentials.txt --snaps="mise" --expires="2025-12-31" # Change exported file path and expiration date
    ```
4. Open ./path/to/credentials.txt and set the content value to GitHub Actions' environment variable `SNAPCRAFT_STORE_CREDENTIALS`
5. Publish a Release on GitHub to publish the Snap package via GitHub Actions

---

> [!NOTE]
> Add Snapcraft packaging for mise with CI to build/publish and docs for Snap install.
> 
> - **Packaging**:
>   - Add `snapcraft.yaml` defining Snap (strict confinement, core24, amd64/arm64/armhf) and `apps.mise` with required plugs.
>   - Include `packaging/snapcraft/mise-self-update-instructions.toml` in snap via `update-instructions` part.
> - **CI/CD**:
>   - New workflow `/.github/workflows/snapcraft.yml` to build snap (beta channel), run a quick E2E test, upload artifact, and publish to Snap Store on release.
> - **Release Automation**:
>   - Update `xtasks/release-plz` to sync `version` in `snapcraft.yaml` during releases.
> - **Docs**:
>   - Add Snap installation instructions to `docs/getting-started.md` and `docs/installing-mise.md`.
> - **Repo hygiene**:
>   - Ignore Snap artifacts by adding `*.snap` to `.gitignore`.
> 
[...]

PR #6525, opened 2025-10-03
fix(snapcraft): add missing home plug for the home directory access permission

phanect:
This is a follow-up PR for #6472.
Sorry, there was a bug in the snap package.

As I mentioned in #6472, Snap packages can only access the explicitly allowed resources.
I found that I did not allow access to the home directory, and this PR adds `home` interface (permission) to `plugs`.
This `home` interface allows mise to access files under the home directory except for hidden files (i.e., dot files).

[The permissions to `$HOME/.cache/mise` and `$HOME/.local/share/mise` are already configured, and I added `$HOME/.config/mise` in this PR](https://github.com/jdx/mise/pull/6525/files#diff-e68b9afd67bb77dec8eda77796d3e0b784c7011203114996383eed6533be0cefR67-R72).
Does mise need file access permission to other hidden directories?
If mise needs access to other hidden directories, I want to add permission to those directories.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,3 +1,12 @@
+# To test this snap package on your local Linux machine,
+# Install snapcraft (`snap install snapcraft --classic`) and
+# run the following commands:
+# 
+# ```shell
+# snapcraft pack
+# snap install ./mise_20xx.xx.x_amd64.snap --dangerous
+# ```
+
 name: mise
 title: mise-en-place
 version: "2025.10.1"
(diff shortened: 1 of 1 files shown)


jdx:
are you sure it's not going to require classic confinement?

phanect:
If it consumes your time too much, I'm considering managing this snap package in my own repository and publishing this Snap package as my package first.
When it is mostly stable, I can transfer my ownership of mise Snap package to you.
Which is better for you, keeping the Snap package in this repository, or separating it to my repository for a while?

jdx:
I suspect sandboxing won't fit a lot of workflows. For example, if you ran `mise x -- sudo apt install`, that should function. There's really no limits on what a user may want to do with mise so I don't think sandboxing would make sense

jdx:
you did great! release code is never easy

PR #6533, opened 2025-10-04
feat: support latest suffix for Java, Python and Ruby flavoured versions

roele:
Adds support for `-latest` suffix in Java, Python and Ruby core plugins which have vendor builds. So `mise latest ruby@truffleruby` and `mise latest ruby@truffleruby-latest` are equivalent while the later is a bit more expressive in its intent.


roele:
Not really and i see the maintenance burden. I removed the overrides but we should probably keep the tests and updates of the docs that mention how to deal with `latest` version for vendor builds.

PR #6576, opened 2025-10-07
doc: document show tools for global tools

Konfekt:
See https://github.com/jdx/mise/discussions/6537


Konfekt:
It documents undocumented behavior that I found surprising. As it reads right now, one would expect only additional output if there is a `mise.toml` file in the working directory.

PR #6577, opened 2025-10-07
doc: Python 'system' deprecation and version setting precedence

Konfekt:
Clarify the deprecation of the 'system' value for Python in the documentation. See https://github.com/jdx/mise/discussions/6539

### DIFF
--- a/docs/lang/python.md
+++ b/docs/lang/python.md
@@ -71,6 +71,18 @@ If you have installed `uv` (for example, with `mise use -g uv@latest`), `mise` w
 
 Note that `uv` does not include `pip` by default (as `uv` provides `uv pip` instead). If you need the `pip` package, add the `uv_create_args = ['--seed']` option.
 
+The value `"system"` for `python` has been deprecated. To still use the system version, for example if other tools were linked to it by compile time (such as `vim`), note that
+
+- If no value (of such a Python version) is set as env-directive, then
+
+    - if `tools.python` is set, then that version is used.
+    - If not, or `settings.disable-tools = ["python"]`, then the latest Python version of `uv` is used.
+
+- If a value is set as env-directive, then `uv` uses the system python version if
+
+    - either no corresponding managed version was installed in `uv`,
+    - or `python.python-preference = "system"` is set in `uv`.
+
 See the [mise + uv Cookbook](/mise-cookbook/python.html#mise-uv) for more examples.
 
 ## Default Python packages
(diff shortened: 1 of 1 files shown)


docs/lang/python.md
```diff
@@ -71,6 +71,18 @@ If you have installed `uv` (for example, with `mise use -g uv@latest`), `mise` w
 
 Note that `uv` does not include `pip` by default (as `uv` provides `uv pip` instead). If you need the `pip` package, add the `uv_create_args = ['--seed']` option.
 
+The value `"system"` for `python` has been deprecated. To still use the system version, for example if other tools were linked to it at compile time (such as `vim`), note that
```
jdx, line 74: "system" never did this either

docs/lang/python.md
```diff
@@ -71,6 +71,18 @@ If you have installed `uv` (for example, with `mise use -g uv@latest`), `mise` w
...
+
+- If no value (of such a Python version) is set as env-directive, then
+
+  - if `tools.python` is set, then that version is used.
+  - If not, or `settings.disable-tools = ["python"]`, then the latest Python version of `uv` is used.
```
jdx, line 79: this is not the right place to document behavior for tools generally

docs/lang/python.md
```diff
@@ -71,6 +71,18 @@ If you have installed `uv` (for example, with `mise use -g uv@latest`), `mise` w
 
 Note that `uv` does not include `pip` by default (as `uv` provides `uv pip` instead). If you need the `pip` package, add the `uv_create_args = ['--seed']` option.
 
+The value `"system"` for `python` has been deprecated. To still use the system version, for example if other tools were linked to it at compile time (such as `vim`), note that
+
+- If no value (of such a Python version) is set as env-directive, then
```
jdx, line 76: idk what an "env-directive" is

Konfekt: I was just using the terminology coined in the [doc](https://mise.jdx.dev/environments/#env-directives)

docs/lang/python.md
Konfekt: Please feel free to move it to the suitable place as you've a much better overview

docs/lang/python.md
Konfekt: Then we should drop still ? I just assumed the `system` version to refer to the system's version, but have only started using Mise last week when it had already been abandoned

jdx:
I'm just going to close, this would take too much work to do properly

PR #6585, opened 2025-10-07
fix: make `silent` in toml task config works, and align the docs

gaojunran:


jdx:
bugbot run

risu729:
Was the document wrong?
Some people are trying to use that feature, so I think it'll be ideal if we implement it.
https://github.com/jdx/mise/discussions/5049

jdx:
I think we should implement it

PR #6593, opened 2025-10-07
refactor: remove duplicated display_name calculation

risu729:
`display_name` are set in `load_all_tasks_with_context`, which is the only function call `load_tasks_in_dir`, so this is redundant.

For the sort, we need to keep it there, not to sort all the tasks, but to sort tasks in the same dir.

### DIFF
--- a/tasks.md
+++ b/tasks.md
@@ -1,83 +1,83 @@
-## `aqua-tester`
+## ``
 
-- **Usage**: `aqua-tester`
+- **Usage**: ``
 
 Test Aqua configuration with fish shell
 
-## `build`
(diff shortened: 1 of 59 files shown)


tasks.md
```diff
@@ -1,83 +1,83 @@
-## `aqua-tester`
+## ``
```
jdx: Seems something broke?

risu729: Yeah, I forgot it was also used to generate task docs.

risu729:
I'll mark this as a draft for now, I think I need to refactor with task loading functions.

PR #6605, opened 2025-10-08
Collect env entries from all config files

davidolrik:


src/cli/run.rs
```diff
@@ -1099,16 +1101,57 @@ impl Run {
...
+                    entries.push((entry, source.clone()));
+                }
+            }
+        }
+
+        Ok(entries)
+    }
+
+    /// Build env directives from config file entries (now returns input directly)
+    fn build_config_env_directives(
```
jdx: this function isn't doing anything

src/cli/run.rs
```diff
@@ -1099,16 +1101,57 @@ impl Run {
...
+                // For task configs with a project root, include configs from the same project
+                cf.project_root().as_ref() == Some(task_root)
+            } else if let (Some(task_parent), Some(cf_parent)) =
+                (task_cf_path.parent(), source.parent())
+            {
+                // For configs without project root, check if they're in the same directory
+                task_parent == cf_parent || config::is_global_config(source)
+            } else {
+                config::is_global_config(source)
+            };
```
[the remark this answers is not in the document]
davidolrik: This makes it a bit more complicated to do correctly, as we would have to create a much more complex cache key...

src/cli/run.rs
```diff
@@ -1038,25 +1047,71 @@ impl Run {
...
 
         Ok((env, task_env))
     }
 
+    /// Compute a cache key for environment resolution
+    /// The key includes all factors that affect the resolved environment:
+    /// - Task config file path and modification time
+    /// - All config files with env entries and their modification times
+    /// - Toolset state (installed tools and versions)
+    fn compute_env_cache_key(
```
jdx, line 1070: the cache is just used in a single `mise run` execution so that it memoizes the results for each project and doesn't load the same project multiple times. Why is it checking the file times and resolving all of the tools here?

davidolrik: I misunderstood the purpose of the cache

PR #7775, opened 2026-01-20
[feat] Added support for mise's shell_aliases for nushell

mnalmahmud:


A2va:
Why was it closed?

mnalmahmud:
Hi.

Because the solution i proposed didn't work and i am not so good with rust.
I have figured out a not so good workaround:

in env.nu

if $nu.is-interactive {
    ^$mise activate nu | save -f ($nu.cache-dir | path join mise.nu)
    source ($nu.cache-dir | path join mise.nu)
    ^$mise shell-alias ls --no-header | lines
    | parse "{alias} {expansion}"
    | each { |r|
        $"alias ($r.alias | str trim) = ^($r.expansion | str trim)"
    } | str join (char newline)
    | save -f ($nu.cache-dir | path join mise-shell-aliases.nu)
    # source ($nu.cache-dir | path join mise-shell-aliases.nu)
    mise_hook
} else {
    ^$mise env --json | from json | load-env
    "" | save -f ($nu.cache-dir | path join mise-shell-aliases.nu)
}
source ($nu.cache-dir | path join mise-shell-aliases.nu)

Atleast the aliases are loaded from mise, although they are not updated on when we modify aliases in mise config.
In zsh, bash env vars and shell aliases both are updated immediately when mise config is updated.

PR #8247, opened 2026-02-18
feat: rename pipx backend to pypi

halms:
# WIP 🚧
After input from @jdx, I'll revisit this, probably rename `pipx` to `pypi` rather than `uv`.

Original note below:

This might be a bit controversial one, but I'll put it here to maybe have a discussion.

The `pipx` backend has defaulted to using `uv tool` for a while now and `uv` has more or less become the _default_ package manager for Python,

So I thought why not rename the backend to `uv` (while keeping `pipx` working for now).


lucasew:
Standardise on the name of the source/registry instead of the name of the tool really sounds like a better approach

my1e5:
Because we are running `uv tool install` - I feel like `uv` is the right direction for a rename. Calling it `pypi` gives the impression that this only installs from PyPI. But the scope of what can be installed by uv is very wide.

For example, I currently use tools installed from specific branches of private git repos:
```toml
[tools]
"pipx:git+ssh://git@github.com/my-org/my-repo.git" = "feature/my-branch"
```
In this context, calling it `pypi:git+ssh://git@github.com/my-org/my-repo.git` would feel strange. 

Also, depending on what uv environment variables you have set (e.g. `UV_INDEX`), you might be bypassing PyPI entirely and using your own package index. 

On `uv:psf/black` and `uv:gh:psf/black` - I think this is unnecessary complication. Cause there are different ways to authenticate (`https`, `ssh`) which you need to handle. I think just keep it straightforward and explicit. You pass in whatever you would normally pass in to `uv tool install`. As for `uv:pyx:...`, again you don't need to over complicate this. Access is already handled by your uv environment variables (`UV_INDEX`, `UV_INDEX_FOO_PASSWORD` etc.), which is what pyx and other package indexes instruct their users to set up.

PR #8890, opened 2026-04-04
feat: bump usage-lib v2 → v3 to render examples in task --help

baby-joel:
## Summary

`#USAGE example` directives parse correctly but are silently dropped from `--help` output. This is because usage-lib v2's CLI help templates don't include an Examples section.

usage-lib v3 added example rendering to its help templates. This bump picks that up.

## Before

```
$ mise run my-task --help
Usage: my-task [--name <name>]

Flags:
  --name <name>  Your name
```

## After

```
$ mise run my-task --help
Usage: my-task [--name <name>]

Flags:
  --name <name>  Your name

Examples:
  Basic usage:
    $ my-task --name world
```

## Changes

- `Cargo.toml`: `usage-lib` version `"2"` → `"3"`
- `task_script_parser.rs`: 4 callsites changed from struct literal init to `Default::default()` + field assignment (`Spec` became non-exhaustive in v3)

## Testing

558 unit tests pass, 0 failures. Manually verified with file tasks using `#USAGE example` directives.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -207,7 +207,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -218,7 +218,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -1006,7 +1006,7 @@ dependencies = [
  "bitflags",
  "cexpr",
  "clang-sys",
- "itertools 0.10.5",
+ "itertools 0.13.0",
  "log",
  "prettyplease",
  "proc-macro2",
@@ -1542,7 +1542,7 @@ version = "3.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "faf9468729b8cbcea668e36183cb69d317348c2e08e994829fb56ebfdfbaac34"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


zeke-ricon (commented):
**Review from @zeke-ricon**

Overall this is clean and mergeable. Notes:

**Looks good:**
- Cargo.toml bump is straightforward (usage-lib 2 → 3, features unchanged)
- All 4 Spec construction sites correctly adapted to v3 API — `default() + field assignment` is the right pattern
- Cargo.lock changes are mechanical (strum 0.28.0/0.28.0, roff 0.2.2 → 1.1.1 as transitive deps from usage-lib v3)
- New test covers the full path: `parse_script` → examples populated → `has_any_usage_spec` → `render_help` output. Good coverage of the feature being enabled

**One observation (non-blocking):**
The test exercises usage-lib's `Spec::parse_script` and `docs::cli::render_help` directly — this proves v3 works but doesn't go through mise's own `TaskScriptParser`. If `TaskScriptParser` ever strips or drops examples during its processing, this test wouldn't catch it. Consider a follow-up test that feeds a task with `#USAGE example` directives through `TaskScriptParser::parse_run_scripts` and asserts examples survive into the final Spec. Not blocking since the construction sites don't touch example fields.

Ship it once CI is green. Nice contribution.

ikma-ricon (changes requested):
Hey Baby Joel — reviewed the diff. Overall clean work. One substantive issue:

**`has_any_usage_spec` checks `spec.cmd.examples` but `parse_script` populates `spec.examples` — potential mismatch**

In usage-lib v3, `Spec` has both `spec.examples` (top-level) and `spec.cmd.examples` (on `SpecCommand`). The `parse_script` path puts `#USAGE example` directives into `spec.examples`. But `has_any_usage_spec` (line 916) checks:

```rust
|| !spec.cmd.examples.is_empty()
```

If a task script has **only** example directives (no flags/args), `has_any_usage_spec` returns false — the task falls through to generic help, and examples don't render.

Your test doesn't catch this because the script includes a flag (`--name`), so `has_any_args_defined` returns true regardless. I'd:
1. Add a test with **only** an example directive (no flags, no args) and assert `has_any_usage_spec` returns true
2. If that fails (I expect it will), fix `has_any_usage_spec` to also check `!spec.examples.is_empty()`

**Other notes:**
- The four `Spec::default()` + `spec.cmd = cmd` construction sites are consistent and correct (forced by `#[non_exhaustive]` in v3). No issues.
- The strum 0.27/0.28 duplication in Cargo.lock is expected — usage-lib v3 pulls strum 0.28 while the rest of the tree stays on 0.27. Fine.
- `render_help` assertions in the test are solid — good coverage of the rendering path.

— Ikma

PR #8902, opened 2026-04-04
fix: parse inline CLI options for ls-remote, overrides mise.toml

esteve:
This PR enables `ls-remote` to parse options from the CLI (e.g. `mise ls-remote "vfox-pixi:samtools[channels=bioconda]"`), overriding options from `mise.toml` if the latter exists, as introduced in #8875 . Also adds support for parsing TOML arrays (e.g. `mise ls-remote 'vfox-pixi:samtools[channels=["conda-forge", "bioconda"]]'`)

### DIFF
--- a/src/backend/conda.rs
+++ b/src/backend/conda.rs
@@ -632,8 +632,9 @@ impl Backend for CondaBackend {
     async fn list_remote_versions_with_info(
         &self,
         config: &Arc<Config>,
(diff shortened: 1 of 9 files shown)


esteve:
@jdx anything else that should be addressed? Would you prefer I add opts to  `_list_remote_versions` and update all the backends or is the current approach with `_list_remote_versions_with_opts` ok?

jdx:
going to pass for now, sorry

PR #8929, opened 2026-04-05
fix(task): make --timeout override task.timeout setting

jdx:
## Summary

- `--timeout` now overrides the `task.timeout` setting via `Settings::task_timeout_override()`, so it flows through the same codepath as the setting — including per-task timeout resolution
- Previously `--timeout` bypassed the setting and only wrapped the entire `parallelize_tasks` call directly

Fixes #8858

## Test plan

- [ ] `mise run --timeout 2s slow-task` should behave the same as `MISE_TASK_TIMEOUT=2s mise run slow-task`
- [ ] Without `--timeout`, behavior is unchanged

### DIFF
--- a/src/cli/run.rs
+++ b/src/cli/run.rs
@@ -7,7 +7,6 @@ use std::time::Duration;
 use super::args::ToolArg;
 use crate::cli::{Cli, unescape_task_args};
 use crate::config::{Config, Settings};
-use crate::duration;
 use crate::env;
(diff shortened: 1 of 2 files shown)


jdx:
Closing — the existing behavior is correct. `--timeout` already overrides the global `task.timeout` setting for the outer wrapper, and per-task timeouts are intentionally independent.

PR #9006, opened 2026-04-10
fix(env): preserve externally prepended paths when one is already in orig PATH

bdrewery:
## Summary

When both `mise activate` and `direnv hook` are active in zsh, paths
added via `PATH_add` in `.envrc` can end up at the **end** of PATH
instead of the beginning -- the opposite of what `PATH_add` is supposed
to do. Tools in a virtualenv or project-local scripts silently lose
priority over system binaries.

**Root cause:** `build_path_operations()` in `hook_env.rs` uses a
`seen_orig` boolean to mark the boundary between "prepended" (pre) and
"appended" (post_user) paths. It fires as soon as the first orig-PATH
entry appears in current PATH. But if an external tool prepended a path
that *already existed* in `__MISE_ORIG_PATH`, that entry appeared first
in the current PATH and triggered the flag immediately. Every genuinely
new prepend after it was misclassified as post_user and landed at the
end.

### 1. Bug fix (`hook_env.rs`)

Replace the single `seen_orig` boolean with boundary detection that
identifies the genuine original-position (tail) copy of each orig entry.
A path moved to the front by an external tool is correctly kept in the
pre zone; the boundary fires only when iteration reaches the tail copy.
This also handles the edge case where all orig entries are duplicated —
the boundary still fires correctly when the tail copies are reached.

### 2. Existing test fix (`test_path_interleaved_regression`)

This test was not exercising `hook-env`: it called `mise hook-env` after
overriding PATH, which removed the mise binary from PATH, so `eval` ran
on an empty string and all assertions were trivially checking the
manually-set PATH. Fixed by saving the mise binary path before the PATH
override (`MISE_BIN=$(which mise)`) and adding ordering assertions that
verify prepended paths stay before orig entries and interleaved paths are
not promoted ahead of them.

Proved the test was previously inert: introducing a deliberate ordering
bug in `hook_env.rs` passed the old test but failed the new one.

### 3. New tests

**`test_path_external_prepend_existing`** — directly reproduces the direnv
scenario: `__MISE_ORIG_PATH` contains `existing-bin`, then current PATH has
`existing-bin:prepended-bin:<orig>` simulating two `PATH_add` calls where one
path already exists in orig. Asserts that `prepended-bin` stays before
`system-bin` after `hook-env`, that the `check-priority` binary resolves from
the prepended location, and that `existing-bin` appears exactly twice in PATH
(prepend copy + original tail copy).

**`test_path_all_orig_duplicated`** — covers the edge case where an external
tool duplicates *every* orig path to the front (e.g., consecutive `PATH_add`
calls for all orig entries). Asserts that paths appended after the orig block
stay after it and are not promoted to the front.

## Reproduction

```zsh
# ~/.local/bin is already in PATH when mise activates
export PATH="${HOME}/.local/bin:${PATH}"

$ cat ~/project/.envrc
PATH_add "${PWD}/venv/bin"      # genuinely new
PATH_add "${HOME}/.local/bin"   # already in orig PATH, added last so ends up first

$ cd ~/project
$ echo $PATH | tr ':' '\n'
/home/user/.local/bin
/usr/local/bin
/usr/bin
/bin
/home/user/.local/bin
/home/user/project/venv/bin    # landed at the very end (should be near front)
```

Workarounds (both avoid the bug):
- Swap the `PATH_add` order so the already-in-orig entry is added first
- Remove the redundant `PATH_add` for a path already in orig PATH

## Test plan

- [x] `mise run test:e2e test_path_external_prepend_existing` -- passes (fails on old code)
- [x] `mise run test:e2e test_path_interleaved_regression` -- passes with real hook-env exercise
- [x] All 12 PATH-related e2e tests pass (`test_path*`)
- [x] `cargo test --all-features` passes
- [x] `TEST_ALL=1 mise run test:e2e -- env` passes
- [x] `mise run test:e2e test_path_all_orig_duplicated` -- boundary guard for all-orig-duplicated case

Co-authored-by: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/e2e/env/test_path_all_orig_duplicated
+++ b/e2e/env/test_path_all_orig_duplicated
@@ -1,0 +1,68 @@
+#!/usr/bin/env bash
+
+# Regression guard for the boundary-detection logic when every orig-PATH entry
+# has a stale tail copy AND a new path is appended after the tail copies.
+# This is a synthetic worst case, not a literal `direnv PATH_add` output
+# (real PATH_add prepends to the front, producing layouts like `b:a:a:b:SYS`).
+# It exercises the tail-anchored boundary detection in `build_path_operations()`
+# to ensure appended paths are not promoted to pre when all orig entries are
+# duplicated.
+
+MISE_BIN="$(which mise)"
+
+mkdir -p "$HOME/orig-bin-a"
+mkdir -p "$HOME/orig-bin-b"
+mkdir -p "$HOME/appended-bin"
+
+cat >"$HOME/orig-bin-a/check-priority" <<'EOF'
+#!/usr/bin/env bash
+echo "orig-a"
+EOF
+cat >"$HOME/appended-bin/check-priority" <<'EOF'
+#!/usr/bin/env bash
+echo "appended"
+EOF
+chmod +x "$HOME/orig-bin-a/check-priority"
+chmod +x "$HOME/appended-bin/check-priority"
+
+cat >mise.toml <<'EOF'
+EOF
+
+# Orig PATH has exactly two custom entries — both will be duplicated.
+export __MISE_ORIG_PATH="$HOME/orig-bin-a:$HOME/orig-bin-b"
+
+# External tool prepended copies of BOTH orig paths (all-duplicated scenario).
+# System paths are included so shell commands remain available after the override.
+# appended-bin was added after the orig block and must remain after it.
+export PATH="$HOME/orig-bin-a:$HOME/orig-bin-b:/usr/local/bin:/usr/bin:/bin:$HOME/orig-bin-a:$HOME/orig-bin-b:$HOME/appended-bin"
+
+eval "$("$MISE_BIN" hook-env -s bash)"
+
+echo "DEBUG: PATH=$PATH"
+
+ORIG_A_FIRST_POS=$(echo "$PATH" | tr ':' '\n' | grep -n "orig-bin-a" | head -1 | cut -d: -f1)
+APPENDED_POS=$(echo "$PATH" | tr ':' '\n' | grep -n "appended-bin" | head -1 | cut -d: -f1)
+
+echo "DEBUG: orig-bin-a first at $ORIG_A_FIRST_POS, appended-bin at $APPENDED_POS"
+
+if [[ -z $APPENDED_POS ]]; then
+	echo "FAIL: appended-bin not found in PATH"
+	exit 1
+fi
+
+if [[ -z $ORIG_A_FIRST_POS ]]; then
+	echo "FAIL: orig-bin-a not found in PATH"
+	exit 1
+fi
+
+# appended-bin must come AFTER the orig block, not be promoted to the front
+if [[ $APPENDED_POS -gt $ORIG_A_FIRST_POS ]]; then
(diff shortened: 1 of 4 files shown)


bdrewery:
Claude did all of the work here. I drove it to use TDD and not change any other existing behaviors.

jdx:
the docs are pretty clear that you shouldn't use mise and direnv together

bdrewery:
> the docs are pretty clear that you shouldn't use mise and direnv together

Does mise handle .env and .envrc?

bdrewery:
Good luck with adoption with that attitude. They are completely different tools. I was testing this tool this week and now won't bother with it again or adopt it into my projects. If you're going to dictate no direnv use then you should provide the same functionality or make it clear at runtime it is known and **intended to be broken**. What an asinine position. I don't even care if you take _my code_, I just wasted a lot of time with broken environment for simply installing this tool.
@jdx

jdx:
Screenshot 2026-04-10 at 11 52 44

bdrewery:
It's still an asinine position. When people are going to _do it anyway_ and someone _fixes a conflict_ that you refuse, you make a choice to only continue making user experience bad. Makes no sense. You're choosing to make user experience poor for some stubborn reason for fighting with the code in the past. If you're going to have such a hardline position you should implement .env/.envrc support to improve adoption.

jdx:
This PR is a good example of why we don't accept direnv compatibility changes.

The fix replaces a simple `seen_orig` boolean with count-based boundary detection that anchors on the "last occurrence" of each orig path. This works for the specific direnv `PATH_add` scenario you tested, but it introduces bugs for the far more common case where non-orig paths are legitimately interleaved between orig entries (e.g., paths added by `.zlogin`, Homebrew paths interleaved with system paths). Your change would promote those paths into `pre` that should stay in `post_user`, silently giving them higher priority than intended — breaking PATH ordering for users who don't use direnv at all.

We've been down this road before. Back in early 2023 I spent weeks trying to make mise fully compatible with direnv — patching `DIRENV_DIFF`, tracking individual PATH entries, trying different hook orderings. The pattern was always the same: fix one PATH edge case, create two new ones. At one point a direnv compat fix wiped out the entire PATH. PATH management is already the most fragile part of the codebase, and changes here affect every user, not just direnv users. This PR would have been no different — fixing your direnv scenario while introducing regressions for everyone else.

mise already provides the functionality that most people use direnv for — env vars, `env._.path`, virtualenv activation, file-based env loading. mise won't ever support `.envrc` directly because mise is declarative and `.envrc` is arbitrary shell scripts — they're fundamentally different models.

Re: "Claude did all of the work here" — that's part of the problem, not a selling point. An LLM can write code that passes the tests you give it without understanding the three years of PATH-ordering edge cases that informed the current implementation.

bdrewery:
https://mise.jdx.dev/getting-started.html does not have any warnings.
https://mise.jdx.dev/direnv.html has `While that's the official stance, the reality is mise and direnv usually will work together just fine despite this. It's only more advanced use-cases where problems arise`. That's claiming it should work _when not using an "advanced" thing_. So not using mise.toml, and having `PATH_add` in `.envrc` is considered "advanced"?
You're the maintainer. Reject this. Fine. But your docs and attitude need major improvements.

PR #9041, opened 2026-04-10
fix(docs): cap esbuild at 0.27.5

risu729:
## Summary
- Cap the docs build esbuild resolution at 0.27.5 instead of fully reverting to 0.21.5. This is the highest published version I found that keeps `bun run docs:build` green: npm has no 0.27.6, 0.27.7 fails, and 0.28.0 fails.
- Expand docs workflow path filters so Renovate dependency PRs and docs build inputs trigger docs CI: `package.json`, `bun.lock`, `mise.toml`, `settings.toml`, `registry/**`, `tasks.toml`, `tasks/docs/**`, `xtasks/docs/**`, `.github/actions/**`, and this workflow.

## Related
- Triggering failure observed on #9038: https://github.com/jdx/mise/actions/runs/24265801669/job/70860375765?pr=9038
- Regressing dependency PR: #9022 (`esbuild` 0.21.5 -> 0.28.0)
- Checked but not the trigger: #9023 (`vitepress-plugin-tabs` 0.8.0 still builds with esbuild 0.21.5)
- Upstream matching esbuild report: https://github.com/evanw/esbuild/issues/4436
- Upstream esbuild tracking issue for missing destructuring transform: https://github.com/evanw/esbuild/issues/3743
- Upstream compatibility-table change that marks the Safari destructuring bug: https://github.com/compat-table/compat-table/pull/2008
- Background Safari issue in Babel: https://github.com/babel/babel/issues/17773

I did not find matching Vite, VitePress, or vitepress-plugin-tabs issues by the failure text / esbuild Safari search terms.

## Verification
- Reproduced failure locally with esbuild 0.28.0: `bun run docs:build` fails with the same destructuring transform errors as #9038
- Confirmed esbuild 0.27.7 also fails locally with a destructuring transform error
- Confirmed esbuild 0.27.5 passes locally
- `bun install --frozen-lockfile`
- `bun run docs:build`
- `actionlint .github/workflows/docs.yml`


jdx:
Superseded by a new PR that upgrades esbuild to 0.28.0 instead of pinning to an older version. The fix is setting `build.target: "es2022"` in the VitePress Vite config so esbuild doesn't try to transform destructuring for old Safari. The CI path filter expansion from this PR is also included. Thanks for the thorough investigation @risu729!

PR #9050, opened 2026-04-11
fix: ignore MISE_TOOL_VERSION in nested postinstall runs

risu729:
Fixes https://github.com/jdx/mise/discussions/8041.

A tool-level `postinstall` hook can run a nested `mise run`. That nested process inherits hook context like `MISE_TOOL_VERSION`, and `ToolsetBuilder` was parsing it as a normal `MISE_<TOOL>_VERSION` request for a `tool` backend. This produced spurious `tool@<version>` registry warnings before running the task.

This updates `ToolsetBuilder` to ignore `MISE_TOOL_VERSION`, matching the existing `ToolRequestSet` behavior for hook-only version variables, and adds an e2e regression test that installs the local `dummy` tool with a `postinstall` task.

Verification:
- `cargo build`
- `./e2e/run_test e2e/cli/test_install_postinstall_task_env`
- `cargo test toolset`
- `cargo fmt --check`
- `git diff --check`


risu729:
Updated this PR with the fix in a second commit, without force-pushing.

Verification:
- `cargo build`
- `./e2e/run_test e2e/cli/test_install_postinstall_task_env`
- `cargo test toolset`
- `cargo fmt --check`
- `git diff --check`

PR #9066, opened 2026-04-13
fix: prevent trust prompt hang in process substitution contexts

brendanator:
## Summary

Fixes the unrecoverable terminal hang reported in https://github.com/jdx/mise/discussions/8940.

When mise runs inside `source <(mise completion zsh)` (process substitution), stdout is piped while stdin still reports as a TTY via `isatty()`. However, **stdin cannot actually deliver input** in this context — zsh consumes keystrokes before they reach the child process. The `demand` Dialog renders the trust prompt on stderr but blocks forever waiting for input, leaving the terminal completely stuck.

### Root cause

In zsh's `source <(command)`:
- `isatty(0)` → `true` (stdin appears to be a TTY)
- `isatty(1)` → `false` (stdout goes to the process substitution pipe)
- `isatty(2)` → `true` (stderr is the real terminal)

The existing check (`console::user_attended_stderr()`) passes because stderr IS a TTY. But the process cannot read from stdin — any keystrokes go to zsh, not to the child process. This makes the Dialog block forever.

### Fix

Add a `can_prompt()` helper that also requires **stdout to be a TTY**. When stdout is not a TTY, the process is inside a command substitution (`$(…)`) or process substitution (`<(…)`) — contexts where interactive prompts should not be shown.

### Reproduction

```bash
mkdir /tmp/mise-test && cd /tmp/mise-test
echo '[tools]' > mise.toml
mise trust --untrust .
zsh -l -i -c "echo hello"  # hangs if .zshrc has: source <(mise completion zsh)
```

### Testing

- All 629 existing unit tests pass
- Clippy clean
- Verified the fix with a PTY-based reproduction script that confirms:
  - `source <(mise completion zsh)` in untrusted dir: **was HUNG → now OK** (no prompt, no hang)
  - `eval "$(mise completion zsh)"` in untrusted dir: **still OK**
  - Direct `mise completion zsh` in a real TTY: **prompt still shows and is interactive**

### DIFF
--- a/src/ui/prompt.rs
+++ b/src/ui/prompt.rs
@@ -1,3 +1,4 @@
+use std::io::IsTerminal;
 use std::sync::Mutex;
 
 use demand::{Confirm, Dialog, DialogButton};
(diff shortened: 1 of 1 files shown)


jdx:
Thanks for the detailed investigation and repro — the root-cause analysis is accurate. Going to close this though, because I think the fix is aimed at the wrong layer.

`mise completion zsh` emits a static shell script and doesn't actually need any config or trust state to do its job. The intended usage is to write it to a file once (as shown in `mise completion --help`):

```
mise completion zsh > /usr/local/share/zsh/site-functions/_mise
```

Running `source <(mise completion zsh)` on every shell startup regenerates the script each time and, as a side effect, pulls in full config loading — which is what drags the trust prompt into a context where it can't work. That's also just slow for everyone who does it that way.

Widening the prompt guard to require stdout-is-TTY would also suppress legitimate prompts in cases like `mise install 2>&1 | tee log.txt`, where stdin is still the terminal and the user can answer fine. I'd rather not trade that away.

A better direction (needs investigation) would be to make `mise completion` skip config/trust loading entirely, since it doesn't need either. That fixes the hang at the source, speeds up the misuse case, and leaves prompt semantics alone for everyone else.

PR #9069, opened 2026-04-13
feat(task): add OpenTelemetry trace and log export for task executions

MatthiasGrandl:
Add native OTLP/HTTP JSON export for mise task runs. When `otel.endpoint` is configured, mise emits spans and logs for each task execution with proper monorepo grouping.

Streaming log export:
- Task stdout/stderr captured via CmdLineRunner callbacks
- Channel-based collector batches lines to /v1/logs (2s or 100 lines)
- Each log record correlated to its task span via trace_id/span_id

Span hierarchy:
- Root span covers entire `mise run` invocation
- Monorepo tasks grouped under parent spans by config_root
- Non-monorepo tasks are direct children of root

Settings:
- `otel.endpoint` — collector URL (e.g. http://localhost:4318)
- `otel.headers` — custom auth headers for the collector
- `otel.service_name` — service.name resource attribute (default: mise)

Example screenshot from signoz:

image

## Summary by CodeRabbit

* **New Features**
  * Added experimental OpenTelemetry support for `mise run`, including OTLP trace export and optional OTLP log export for task stdout/stderr.
  * Introduced separate opt-in settings for trace vs log export (`otel.enabled` and `otel.logs`) and propagate trace context to child processes.
* **Documentation**
  * Added a new OpenTelemetry guide and navigation entry, including OTLP environment variables, expected trace/log structure, and privacy/trust details.
* **Tests**
  * Added `test:oats` OpenTelemetry acceptance tests using a Docker Compose fixture.
* **Chores**
  * Added CI coverage for the new acceptance job and updated build/tooling/image setup.


MatthiasGrandl:
> ### Greptile Summary
> ...
> All findings are P2. The span-batching issue is a latent inefficiency, not a correctness bug. Core tracing logic, timing aggregation, log correlation, and nested-run propagation are correct and well-tested.
> ...

I deriberately didn't batch spans, because I doubt even the largest monorepos chain more than maybe 20 mise tasks and even then they are likely not happening in a manner of seconds.

risu729:
I'm not familiar with OTel, so I'm not too sure, but are there any reasons for not using https://github.com/open-telemetry/opentelemetry-rust? It seems a good 1st-party implementation.

MatthiasGrandl:
> I'm not familiar with OTel, so I'm not too sure, but are there any reasons for not using https://github.com/open-telemetry/opentelemetry-rust? It seems a good 1st-party implementation.

Just not a fan of pulling in dependencies for 2 http requests, but up to you.

MatthiasGrandl:
I tested this implementation thoroughly now and I am very happy with it.

jdx:
2 http requests? this PR is thousands of lines

MatthiasGrandl:
what I am trying to say is that using the official OTEL dependencies only gets rid of maybe 300 lines of already tested code, while pulling in shit like `prost` and `tonic` for no reason. The otel code in this PR is already cleanly separated from the rest of the codebase and should be very easy to maintain.

MatthiasGrandl:
so I tried implementing the same functionality with `opentelemetry` and `opentelemetry_sdk` crates and it saved around 37 LoC... and I am less confident in maintaining that implementation lol.

jdx:
ok, just to set expectations it'll take me a while to get around to reviewing this much code

jdx:
@MatthiasGrandl would you be willing to share the `opentelemetry` / `opentelemetry_sdk` port as a diff or branch? The 37-LoC delta is surprising — an idiomatic port should delete all of `types.rs` (~346 lines) and `span_collector.rs` (~41 lines) outright by using `BatchSpanProcessor` / `BatchLogRecordProcessor`, so naively I'd expect savings closer to 500+ LoC.

Seeing the attempt would help answer whether the 37-LoC number reflects "the SDK genuinely isn't worth it here" vs. "the port preserved the current no-batching / no-global-provider model and so ended up re-implementing the same structure with different types." Two very different tradeoffs.

Not blocking — the feature is clearly useful and the code is well-organized — just the one bit of data that would make the hand-rolled-vs-SDK call obvious.

MatthiasGrandl:
> @MatthiasGrandl would you be willing to share the `opentelemetry` / `opentelemetry_sdk` port as a diff or branch? The 37-LoC delta is surprising — an idiomatic port should delete all of `types.rs` (~346 lines) and `span_collector.rs` (~41 lines) outright by using `BatchSpanProcessor` / `BatchLogRecordProcessor`, so naively I'd expect savings closer to 500+ LoC.
> 
> Seeing the attempt would help answer whether the 37-LoC number reflects "the SDK genuinely isn't worth it here" vs. "the port preserved the current no-batching / no-global-provider model and so ended up re-implementing the same structure with different types." Two very different tradeoffs.
> 
> Not blocking — the feature is clearly useful and the code is well-organized — just the one bit of data that would make the hand-rolled-vs-SDK call obvious.
> 
> _This comment was generated by an AI coding assistant._

Yeah you can get ~500 loc saving by using the opentelemetry otlp dependency as well… i wanted to avoid that because it pulls in more transitive dependencies. But yeah I can add it as a second commit and you can then decide if you want to go that route.

MatthiasGrandl:
@jdx just pushed  a second commit with otel dependencies. note i had to use experiment feature flags for the `opentelemetry_sdk` crate to make the batch processor work without reqwest-blocking. i assume they are stable enough and i tested everything and works identical to my dependency-less implementation.

if you want to proceed with this approach i will gladly rebase and squash into a single commit.

MatthiasGrandl:
Ok I would say this is ready now.

MatthiasGrandl:
everything greptile keeps saying is complete noise at this point. ignoring.

zeitlinger:
I pulled a couple of the concerns apart and sent follow-ups rather than trying to mix them into this PR.

The main naming issue is that the current OTEL-facing types read too much like OpenTelemetry primitives rather than mise domain objects. In particular:
- `RunTrace` doesn’t say what the thing is in mise terms, and it also owns more than “a trace”
- `TraceContext` is especially confusing because OTEL already uses “trace context” for propagated upstream context like `traceparent` / `tracestate`
- `OtelLogCollector` is really forwarding task stdout/stderr, not acting like a general logging abstraction

I think the code would be easier to reason about if the names reflected their roles more explicitly, e.g. along the lines of:
- `TaskRunTelemetry` for the per-`mise run` lifecycle/orchestration object
- `TaskSpanTracker` for the mise-specific span bookkeeping object
- `TaskOutputForwarder` for the subprocess stdout/stderr to OTEL log bridge

Separately, I opened two stacked follow-up PRs against `otel`:

1. Propagation fidelity fix  
   https://github.com/MatthiasGrandl/mise/pull/2

   This preserves upstream OTEL propagation state instead of collapsing it down to IDs and reconstructing it manually:
- injects child context with `TraceContextPropagator`
- preserves `tracestate`
- preserves trace flags instead of hardcoding sampled state

2. OATs acceptance test  
   https://github.com/MatthiasGrandl/mise/pull/3

   This adds a minimal end-to-end acceptance test using `grafana/oats` to verify a real OTLP round trip for:
- the root `mise run` span
- a task span
- one exported stdout log line

I kept the rename suggestions out of those follow-ups so the implementation and test changes stay reviewable on their own.

MatthiasGrandl:
> I pulled a couple of the concerns apart and sent follow-ups rather than trying to mix them into this PR.
> 
> The main naming issue is that the current OTEL-facing types read too much like OpenTelemetry primitives rather than mise domain objects. In particular:
> 
>     * `RunTrace` doesn’t say what the thing is in mise terms, and it also owns more than “a trace”
> 
>     * `TraceContext` is especially confusing because OTEL already uses “trace context” for propagated upstream context like `traceparent` / `tracestate`
> 
>     * `OtelLogCollector` is really forwarding task stdout/stderr, not acting like a general logging abstraction
> 
> 
> I think the code would be easier to reason about if the names reflected their roles more explicitly, e.g. along the lines of:
> 
>     * `TaskRunTelemetry` for the per-`mise run` lifecycle/orchestration object
> 
>     * `TaskSpanTracker` for the mise-specific span bookkeeping object
> 
>     * `TaskOutputForwarder` for the subprocess stdout/stderr to OTEL log bridge
> 
> 
> Separately, I opened two stacked follow-up PRs against `otel`:
> 
>     1. Propagation fidelity fix
>        [preserve otel propagation context MatthiasGrandl/mise#2](https://github.com/MatthiasGrandl/mise/pull/2)
>        This preserves upstream OTEL propagation state instead of collapsing it down to IDs and reconstructing it manually:
> 
> 
>     * injects child context with `TraceContextPropagator`
> 
>     * preserves `tracestate`
> 
>     * preserves trace flags instead of hardcoding sampled state
> 
> 
>     2. OATs acceptance test
>        [add oats acceptance test for otel export MatthiasGrandl/mise#3](https://github.com/MatthiasGrandl/mise/pull/3)
>        This adds a minimal end-to-end acceptance test using `grafana/oats` to verify a real OTLP round trip for:
> 
> 
>     * the root `mise run` span
> 
>     * a task span
> 
>     * one exported stdout log line
> 
> 
> I kept the rename suggestions out of those follow-ups so the implementation and test changes stay reviewable on their own.

looks good on first glance. i merged your PRs and will test with my production OTEL setup tomorrow. don't disagree about the naming suggestions either.

MatthiasGrandl:
@zeitlinger tested on a large production pipeline including trace propagation. No regressions, works perfectly!

docs/tasks/opentelemetry.md
```diff
@@ -0,0 +1,174 @@
...
+| `mise.task.config_root` | Config root directory (for monorepo tasks)                         |
+| `mise.task.skipped`     | `"true"` when the task was skipped because sources were up to date |
+
+## Logs
+
+Task stdout and stderr are exported as logs and linked to the corresponding task span, so you can
+inspect output directly from the trace.
+
+- stdout is exported with severity `INFO`
+- stderr is exported with severity `WARN`
```
MaxwellBo: drive by review (as this is a PR that I'm interested in seeing merged!). Is there a reason this isn't `ERROR`?

MatthiasGrandl: Nope no reason. Seems like others log stderr as error so we probably should do the same. I don’t have a strong preference.

jdx:
dropping by again just to say I definitely have not forgotten about this and will review it but unfortunately it will take me a bit to get around to this still. I am thinking more broadly about potential log centralization with mise and this could play a part but I've got other priorities so haven't had the chance to really think through this yet.

MaxwellBo:
Also I just wanted to quickly flag https://opentelemetry.io/docs/specs/semconv/cli/cli-spans as maybe something you want to check against in terms of ensuring a baseline set of attributes are being sent, e.g. it doesn't appear that `process.exit.code` and `process.command_args` is being sent which are marked as required.

docs/tasks/opentelemetry.md
```diff
@@ -0,0 +1,174 @@
+# OpenTelemetry <Badge type="warning" text="experimental" />
+
+mise can export traces and logs for `mise run` to any OpenTelemetry-compatible backend such as
```
MaxwellBo: is there a reason we're choosing to instrument just `mise run`? 

The example in https://github.com/jdx/mise/discussions/6565 was for just `mise run`, but I imagine there are other commands that would be of interest to engineers managing a fleet of developer machines running `mise` (e.g. `install`, `doctor`, `upgrade`, etc.)

MatthiasGrandl: I personally was only interested in mise run. But feel free to take over the PR if you feel the scope should be expanded. I don’t feel like investing much more time here.

zeitlinger:
btw - I do want to use a release binary for oats - so that it will be faster to install

MatthiasGrandl:
> Also I just wanted to quickly flag https://opentelemetry.io/docs/specs/semconv/cli/cli-spans as maybe something you want to check against in terms of ensuring a baseline set of attributes are being sent, e.g. it doesn't appear that `process.exit.code` and `process.command_args` is being sent which are marked as required.

Now implemented

docs/tasks/opentelemetry.md
MatthiasGrandl: Changed!

jdx:
Thanks for working on this. Aside from the current conflicts/check failures, I think there are a few design/behavior issues worth tightening before merge:

1. **Task log export should be a separate explicit opt-in.** `otel.enabled` enabling traces is fine, but shipping task stdout/stderr is higher cost and a larger privacy/security boundary. I would not export logs just because OTEL is enabled. This should require its own setting, e.g. `otel.logs = true` / `MISE_OTEL_LOGS=1`, in addition to a logs endpoint. That also avoids changing task stdio behavior for users who only want traces.

2. **Signal-specific endpoint handling looks too broad.** `is_enabled()` turns the whole integration on if any of `OTEL_EXPORTER_OTLP_ENDPOINT`, `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`, or `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT` is present, but both tracer and logger providers are then built. I would expect traces/logs to be gated independently: traces only when trace export is configured, logs only when log export is explicitly enabled and configured.

3. **The root span timing does not match the PR description.** The PR says the root span covers the full `mise run`, but telemetry is initialized after task resolution, tool install, and auto deps, and the root span timing is folded from task start/end times. That excludes setup/deps/scheduler overhead. Either the span should truly cover the invocation, or the docs/PR text should describe it as covering task execution only.

4. **OTEL log export has an inherent stdio tradeoff that should stay explicit.** To export stdout/stderr logs, the PR attaches `CmdLineRunner` hooks. In `interleave`/`quiet`, those hooks keep stdout/stderr piped instead of inherited, so child processes no longer see a real TTY. That can change buffering, color, progress bars, prompts, and any `isatty()`-dependent behavior. This is probably the right implementation tradeoff for log export, but it reinforces why log export should be a separate explicit opt-in.

5. **Privacy docs should be more explicit.** This exports task args, `process.command_args`, stdout, and stderr to a collector. Even if args/output are already visible locally, shipping them to telemetry infrastructure is a different trust boundary. Please call out that secrets in args/output may leave the machine, and explain how this interacts with existing redaction behavior and `--raw`.

6. **Docs/code mismatch:** the docs say stderr is exported as `WARN`, but the code now exports stderr as `ERROR`.

I like the direction and the structure, especially correlating log records to task spans. I would want the log opt-in, signal gating, span semantics, and privacy/stdio docs tightened before merging.

MatthiasGrandl:
@jdx addressed your comments and rebased.

MatthiasGrandl:
sorry it seems i am too dense to fix the CI. running `mise run render` does a bunch of seemingly unrelated changes, that I don't want to commit. I think something in my dev setup is broken.

zeitlinger:
Suggestion: oats is in the aqua registry as `grafana/oats`, so it can be installed via `aqua:grafana/oats` — that lets `go` be dropped from `mise.toml` (nothing else here needs it). Stacked patch: https://github.com/MatthiasGrandl/mise/pull/4 (bumps to v0.7.0 since v0.6.1 ships no release binaries).

jdx:
@MatthiasGrandl I spent some time trying to pin down why this PR felt heavier than the feature seemed to warrant, and ended up sketching a traces-only version to see how much of the size is intrinsic: #11003 (draft).

It keeps your span model, attributes, and settings gates, but produces spans through the SDK lifecycle directly — start a real span when the task starts, end it when the task ends — instead of reserving IDs and emitting synthetic spans at the end. That removes the timing aggregation and the `StartedSpan` threading, and leaves `task_executor.rs`/`cmd.rs` untouched: ~230 lines of production code in one module, ~690 added lines total.

The part I'd most like your take on: since a task's span is *alive* while the task runs, its span context is available for log correlation without the reservation machinery. So I think log export could layer on top of this shape as a follow-up PR, keeping `otel.logs` and the stdio-piping trade-offs as their own review. Do you see anything this shape can't support — nested-run propagation, the semconv details you added, or anything you rely on in your production pipeline?

To be clear: the draft deliberately reuses your design work and is meant as a concrete anchor for the review conversation, not a competing submission. If the direction seems right to you, I'm happy either way — fold the approach into this PR, or land the small core first and rebase your logs work on top, whichever you prefer.

MatthiasGrandl:
> @MatthiasGrandl I spent some time trying to pin down why this PR felt heavier than the feature seemed to warrant, and ended up sketching a traces-only version to see how much of the size is intrinsic: #11003 (draft).
> 
> It keeps your span model, attributes, and settings gates, but produces spans through the SDK lifecycle directly — start a real span when the task starts, end it when the task ends — instead of reserving IDs and emitting synthetic spans at the end. That removes the timing aggregation and the `StartedSpan` threading, and leaves `task_executor.rs`/`cmd.rs` untouched: ~230 lines of production code in one module, ~690 added lines total.
> 
> The part I'd most like your take on: since a task's span is _alive_ while the task runs, its span context is available for log correlation without the reservation machinery. So I think log export could layer on top of this shape as a follow-up PR, keeping `otel.logs` and the stdio-piping trade-offs as their own review. Do you see anything this shape can't support — nested-run propagation, the semconv details you added, or anything you rely on in your production pipeline?
> 
> To be clear: the draft deliberately reuses your design work and is meant as a concrete anchor for the review conversation, not a competing submission. If the direction seems right to you, I'm happy either way — fold the approach into this PR, or land the small core first and rebase your logs work on top, whichever you prefer.
> 
> _This comment was generated by an AI coding assistant._

Seems fine, but I do think TRACEPARENT support should be there from the start.

jdx:
Fair point — agreed, and it's in now. #11003 has both directions of W3C trace context propagation:

- **extract**: the run joins an upstream trace from `TRACEPARENT`/`TRACESTATE` (CI or an outer `mise run`)
- **inject**: each task's env gets `TRACEPARENT`/`TRACESTATE` derived from its live span, excluded from `__MISE_DIFF` so a nested `hook-env` doesn't clobber them (same approach as yours)

Verified end-to-end against a local collector, including a nested `mise run` inside a task sharing the outer run's trace ID. It came out to ~60 lines plus tests since the executor only sees two plain env strings — the draft now totals ~840 lines vs ~2,560.

With that in, the proposed path would be: land the traces+propagation core, then your log export rebases on top as its own PR — the live task span's context is available for log correlation, so `otel.logs`, the output hooks, and the stdio piping trade-offs stay intact as their own review. Sound reasonable?

MatthiasGrandl:
> Fair point — agreed, and it's in now. #11003 has both directions of W3C trace context propagation:
> 
> * **extract**: the run joins an upstream trace from `TRACEPARENT`/`TRACESTATE` (CI or an outer `mise run`)
> * **inject**: each task's env gets `TRACEPARENT`/`TRACESTATE` derived from its live span, excluded from `__MISE_DIFF` so a nested `hook-env` doesn't clobber them (same approach as yours)
> 
> Verified end-to-end against a local collector, including a nested `mise run` inside a task sharing the outer run's trace ID. It came out to ~60 lines plus tests since the executor only sees two plain env strings — the draft now totals ~840 lines vs ~2,560.
> 
> With that in, the proposed path would be: land the traces+propagation core, then your log export rebases on top as its own PR — the live task span's context is available for log correlation, so `otel.logs`, the output hooks, and the stdio piping trade-offs stay intact as their own review. Sound reasonable?
> 
> _This comment was generated by an AI coding assistant._

Jep fine with me!

jdx:
@MatthiasGrandl thanks for all of this. I rebuilt it on current main as a stack, following the split we agreed on in July, with you (and @zeitlinger) as co-authors:

- #13556 failed-vs-cancelled task reporting (the non-OTel part of your traces commit)
- #13557 traces + W3C propagation, bumped to opentelemetry 0.33, which shares mise's reqwest 0.13 (no second reqwest, no tonic)
- #13558 new: setup phases (task resolution, tool install, deps, daemons) as child spans, so the root span covers the whole run
- #13559 your log export, including the nested-run claim handoff
- #13560 an e2e test against a small OTLP/JSON receiver in place of the OATs compose job, so it runs in the normal suite without Docker

Reviews and corrections on any of them are very welcome, especially if something in your production setup behaves differently.

*AI-assisted — Tool: Claude Code; model: anthropic/claude-opus-5-5; version: unavailable.*

jdx:
Closing in favour of the stack that carries this work forward on current main, with @MatthiasGrandl and @zeitlinger as co-authors: #13557 (traces), #13558 (setup spans), #13559 (log export), #13560 (e2e test). #13556 has already merged. Thanks for all of this.

*AI-assisted — Tool: Claude Code; model: anthropic/claude-opus-5-5; version: unavailable.*

PR #9072, opened 2026-04-13
feat(npm): use --min-release-age for npm 11.10.0+ supply chain protection

webkaz:
## Summary

For npm >= 11.10.0, replace the `--before <ISO timestamp>` flag with the purpose-built `--min-release-age=<days>` flag introduced in npm/cli#8965. Older npm continues to receive the existing `--before` args, so this change is fully backwards-compatible.

`--min-release-age` is preferred because:
- It mirrors bun's `--minimum-release-age` and pnpm's `--config.minimumReleaseAge` (both age-based)
- It is the flag npm now recommends for supply chain protection
- `--before` was originally designed for reproducible builds, not security

This is the implementation for #9042, which @jdx and @risu729 approved.

## Implementation
- `npm_supports_min_release_age_flag()` runs `npm --version` once per install (only when `install_before` is configured) and routes to `build_npm_min_release_age_args()` for npm 11.10.0+, falling back to the legacy `build_transitive_release_age_args(Npm, ...)` path otherwise
- Failures of either step are logged at debug level and silently fall back to `--before`, so users with broken npm setups don't see new errors
- Bun/Pnpm callers are unchanged; the build helper signature is untouched

## Test plan
- [x] Unit tests added (9 npm backend tests pass locally):
  - npm legacy `--before` path
  - `build_npm_min_release_age_args` for full days, partial day round-up, and `before == now` boundary
  - Version detection (>= 11.10.0, edge cases including `v` prefix, trailing newline, pre-release tags, garbage input)
- [x] CI

Closes #9042

---

*This PR was prepared with Claude Code.*

### DIFF
--- a/src/backend/npm.rs
+++ b/src/backend/npm.rs
@@ -227,7 +227,14 @@ impl Backend for NPMBackend {
             _ => {
                 let install_before_args = if let Some(before_date) = ctx.before_date {
                     let now = Timestamp::now();
-                    if self.npm_supports_min_release_age_flag(&ctx.config).await {
+                    // --min-release-age is day-granular, so fall back to --before for
+                    // sub-day windows (e.g. install_before = "1h") to preserve the
+                    // exact cutoff. div_ceil(86400) on <24h would round up to 1 day
+                    // and broaden the window, breaking backwards compatibility.
+                    let elapsed = Self::elapsed_seconds_ceil(before_date, now);
+                    let use_modern = elapsed >= 86400
+                        && self.npm_supports_min_release_age_flag(&ctx.config).await;
+                    if use_modern {
                         Self::build_npm_min_release_age_args(before_date, now)
                     } else {
                         Self::build_transitive_release_age_args(
@@ -329,9 +336,16 @@ impl NPMBackend {
 
     /// Returns true if the npm major.minor.patch version is >= 11.10.0,
     /// which is when the --min-release-age flag was added (npm/cli#8965).
+    /// Pre-release / build-metadata tags are rejected because in strict
+    /// semver `11.10.0-pre.1 < 11.10.0`, so a pre-release that shipped
+    /// before the flag landed would otherwise be misclassified as supporting
+    /// it and crash with an unknown-flag error.
     fn npm_version_supports_min_release_age(version: &str) -> bool {
         let trimmed = version.trim().trim_start_matches('v');
-        let mut parts = trimmed.split(['.', '-', '+']);
+        if trimmed.contains('-') || trimmed.contains('+') {
+            return false;
+        }
+        let mut parts = trimmed.split('.');
         let major: u64 = match parts.next().and_then(|p| p.parse().ok()) {
             Some(v) => v,
             None => return false,
@@ -350,7 +364,21 @@ impl NPMBackend {
     /// transparently fall back to the older --before flag. Failures are
     /// logged at debug level so users can troubleshoot why the fallback
     /// flag was chosen.
+    ///
+    /// The result is cached for the lifetime of the process via OnceLock so
+    /// repeated installs in one mise invocation only spawn `npm --version` once.
     async fn npm_supports_min_release_age_flag(&self, config: &Arc<Config>) -> bool {
+        static CACHE: std::sync::OnceLock<bool> = std::sync::OnceLock::new();
+        if let Some(&cached) = CACHE.get() {
+            return cached;
+        }
+        let result = self.detect_npm_supports_min_release_age_flag(config).await;
+        // OK if another thread won the race; both compute the same bool.
+        let _ = CACHE.set(result);
+        result
+    }
+
+    async fn detect_npm_supports_min_release_age_flag(&self, config: &Arc<Config>) -> bool {
         let env = match self.dependency_env(config).await {
             Ok(env) => env,
             Err(e) => {
@@ -551,9 +579,13 @@ mod tests {
         assert!(NPMBackend::npm_version_supports_min_release_age(
             "11.10.0\n"
         ));
-        // Pre-release of a supported version still satisfies the gate
-        assert!(NPMBackend::npm_version_supports_min_release_age(
+        // Pre-release tags are rejected (11.10.0-pre.1 < 11.10.0 in semver)
+        assert!(!NPMBackend::npm_version_supports_min_release_age(
             "11.10.0-pre.1"
+        ));
+        // Build metadata is also rejected for safety
+        assert!(!NPMBackend::npm_version_supports_min_release_age(
+            "11.10.0+build.1"
         ));
 
         assert!(!NPMBackend::npm_version_supports_min_release_age("11.9.9"));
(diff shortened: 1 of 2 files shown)


risu729:
I think you can use `semver` crate for the version comparison. 
Also, if we have `npm` tool in the ToolSet, detect version from there then we don't need to run `npm --version`. We still need it for node bundled npm, but we can remove unnecessary external command call.

risu729:
Sorry I meant https://github.com/jdx/mise/blob/main/src/semver.rs, not the crate.

webkaz:
All three review points have been addressed in the latest commit (de54717):

1. **Sub-day fallback** (Copilot): Implemented — when `elapsed_seconds_ceil < 86400`, we now fall back to `--before` to preserve the exact cutoff for sub-day `install_before` windows (e.g. `"1h"`).
2. **Pre-release rejection** (greptile): Implemented — version strings containing `-` or `+` now return `false`, preventing misclassification of pre-release builds where `--min-release-age` may not exist yet.
3. **OnceLock cache** (gemini-code-assist): Implemented — `npm --version` result is now cached via `std::sync::OnceLock` for the lifetime of the process, avoiding redundant subprocess spawns.

src/backend/npm.rs
```diff
@@ -225,13 +225,27 @@ impl Backend for NPMBackend {
...
+                    // sub-day windows (e.g. install_before = "1h") to preserve the
+                    // exact cutoff. div_ceil(86400) on <24h would round up to 1 day
+                    // and broaden the window, breaking backwards compatibility.
+                    let elapsed = Self::elapsed_seconds_ceil(before_date, now);
+                    let use_modern = elapsed >= 86400
+                        && self.npm_supports_min_release_age_flag(&ctx.config).await;
+                    if use_modern {
+                        Self::build_npm_min_release_age_args(before_date, now)
+                    } else {
+                        Self::build_transitive_release_age_args(
```
risu729: I believe we should move this logic into `build_transitive_release_age_args`.

src/backend/npm.rs
```diff
@@ -225,13 +225,27 @@ impl Backend for NPMBackend {
...
-                        before_date,
-                        Timestamp::now(),
-                    )
-                });
+                let install_before_args = if let Some(before_date) = ctx.before_date {
+                    let now = Timestamp::now();
+                    // --min-release-age is day-granular, so fall back to --before for
+                    // sub-day windows (e.g. install_before = "1h") to preserve the
+                    // exact cutoff. div_ceil(86400) on <24h would round up to 1 day
+                    // and broaden the window, breaking backwards compatibility.
```
risu729: This should be fine, but it should be documented.

webkaz:
@risu729 Thanks for the review — two great suggestions. Here's where I landed:

**1. Using `src/semver.rs` for version comparison**

I checked `src/semver.rs` — it provides `split_version_prefix` and `chunkify_version` but no comparison helpers. The `versions::Versioning` crate is available, but it's very permissive (falls back to `Mess` parsing for garbage input), which means `Versioning::new("garbage") >= Versioning::new("11.10.0")` could return an unexpected result instead of safely falling back to `--before`.

The current hand-rolled parser is 8 lines, rejects pre-release/build-metadata tags explicitly, and has full edge-case coverage in tests. I think it's the right trade-off for a single-purpose gate, but happy to switch if you or @jdx prefer the crate approach.

**2. Detecting npm version from the ToolSet instead of `npm --version`**

Good idea in principle — it would avoid the subprocess entirely when mise manages `npm` as a tool. The complication is that most users get npm bundled with `node` (not as a standalone mise tool entry), so the ToolSet lookup would miss the common case and we'd still need the `npm --version` fallback.

I've added `OnceLock` caching in the latest commit so `npm --version` runs at most once per mise invocation, which keeps the overhead negligible. A ToolSet-aware path could be a nice follow-up optimization if there's interest.

src/backend/npm.rs
```diff
@@ -309,6 +334,76 @@ impl NPMBackend {
             .expect("elapsed timestamp delta must fit into u64")
     }
 
+    /// Returns true if the npm major.minor.patch version is >= 11.10.0,
+    /// which is when the --min-release-age flag was added (npm/cli#8965).
+    /// Pre-release / build-metadata tags are rejected because in strict
+    /// semver `11.10.0-pre.1 < 11.10.0`, so a pre-release that shipped
```
risu729: There are no pre-releases for `11.10.0` afaik, so it should be safe to ignore this.

webkaz:
@risu729 Re-reading your second point more carefully — you're right that when `npm` is explicitly in the ToolSet (e.g. `mise use npm@11.10.0`), we can pull the resolved version from `dependency_toolset()` and skip the subprocess entirely for that case.

Implemented in f85b640: `detect_npm_supports_min_release_age_flag` now checks the dependency ToolSet first and only falls back to `npm --version` for node-bundled npm. Added `Suggested-by: risu729` to the commit.

src/backend/npm.rs
```diff
@@ -309,6 +334,76 @@ impl NPMBackend {
...
+    /// by invoking `npm --version`. Returns false on any failure so callers
+    /// transparently fall back to the older --before flag. Failures are
+    /// logged at debug level so users can troubleshoot why the fallback
+    /// flag was chosen.
+    ///
+    /// The result is cached for the lifetime of the process via OnceLock so
+    /// repeated installs in one mise invocation only spawn `npm --version` once.
+    async fn npm_supports_min_release_age_flag(&self, config: &Arc<Config>) -> bool {
+        static CACHE: std::sync::OnceLock<bool> = std::sync::OnceLock::new();
+        if let Some(&cached) = CACHE.get() {
```
risu729: I don't think `npm --version` is called several times as it's only used to build the install command.

risu729 (changes requested):

webkaz:
@risu729 Thanks for the detailed review — all four points addressed in 3ac30a0:

1. **Sub-day fallback moved into `build_transitive_release_age_args`** — all `--min-release-age` / `--before` routing is now in one place.
2. **Documented** — sub-day behavior is noted in the inline comment.
3. **Pre-release rejection reverted** — agreed, no known 11.10.0 pre-releases exist. Kept the simple split-based parser.
4. **OnceLock removed** — `npm --version` runs once per `install_version_` call which is sufficient. Note: when installing multiple npm packages in a single `mise install` (e.g. prettier + eslint), the subprocess runs per-package. If this proves to be a concern, the OnceLock can be added back — happy to do so if needed.

ToolSet-first version lookup is kept as you suggested.

risu729:
I mean, we should update this page as it now says `--before` is used.
https://mise.jdx.dev/dev-tools/backends/npm.html

webkaz:
@risu729 I've addressed all four points in the latest push. Could you re-review when you get a chance?

webkaz:
@risu729 Good catch — updated the npm backend docs in e801427 to reflect the new `--min-release-age` path and the sub-day `--before` fallback.

PR #9073, opened 2026-04-13
feat: add `arch` filter for tool version options

webkaz:
## Summary

Adds an `arch` filter to `[tools]` entries, complementing the existing `os` filter:

```toml
[tools]
"aqua:sharkdp/fd" = { version = "latest", arch = ["arm64"] }
"cargo:fd-find" = { version = "latest", arch = ["x64"] }
```

Use case from #8948: some prebuilt binaries only ship for `arm64` on macOS, so dotfiles managed across Intel and ARM Macs need to fall back to a different backend per architecture. Previously this required external templating (e.g., chezmoi); now it's native.

When both `os` and `arch` are specified, both must match. Accepted values mirror `crate::cli::version::ARCH` (`x64` / `arm64`); other `std::env::consts::ARCH` values pass through verbatim.

This is the implementation for #8948 ("Seems fine to me" — @jdx).

## Implementation
- New `arch: Option<Vec<String>>` field on `ToolVersionOptions`, included in the manual `Hash` impl
- New `arch()` accessor on `ToolRequest`, parallel to `os()`
- Filter logic split out into a private `matches_current_platform()` so unit tests can exercise the filter without touching the install-state singleton that `BackendArg::is_os_supported` consults
- `mise_toml.rs` deserializers (both `MiseTomlToolList` and `MiseTomlTool`) gain an `"arch" =>` arm; serializer writes `arch` next to `os`; `output_empty_opts` and `to_tool_request_set` updated to include `arch`
- `tool_stub.rs` exposes the same `arch` field on `ToolStubFile` so stubs can use it
- Documentation: new "Architecture-Specific Tools" section in `docs/dev-tools/index.md`, `os`/`arch` listed under Optional Fields in `docs/dev-tools/tool-stubs.md`
- Snapshots regenerated for the new field

## Test plan
- [x] Unit tests added: 5 new `tool_request` filter tests + snapshot updates for `mise_toml`. Full suite passes locally (633/633)
- [x] CI

Closes #8948

---

*This PR was prepared with Claude Code.*

### DIFF
--- a/docs/dev-tools/index.md
+++ b/docs/dev-tools/index.md
@@ -196,6 +196,31 @@ The `os` field accepts an array of operating system identifiers:
 
 If a tool specifies an `os` restriction and the current operating system is not in the list, mise will skip installing and using that tool.
 
+## Architecture-Specific Tools
+
+You can also restrict tools to specific CPU architectures using the `arch` field. This is useful when prebuilt binaries are only published for some architectures and you want to fall back to a different backend on others.
+
+```toml
+[tools]
+# Use the prebuilt binary on arm64 (Apple Silicon, ARM Linux)
+"aqua:sharkdp/fd" = { version = "latest", arch = ["arm64"] }
+
+# Compile from source on x64 where no prebuilt binary is available
+"cargo:fd-find" = { version = "latest", arch = ["x64"] }
+
+# Combine with `os` for finer-grained targeting
+"npm:my-tool" = { version = "latest", os = ["linux"], arch = ["x64"] }
+```
+
+The `arch` field accepts an array of architecture identifiers:
+
+- `"x64"` - 64-bit x86 (also known as `x86_64` / `amd64`)
+- `"arm64"` - 64-bit ARM (also known as `aarch64`)
+
+Other values from [`std::env::consts::ARCH`](https://doc.rust-lang.org/std/env/consts/constant.ARCH.html) (such as `"x86"`, `"arm"`, `"riscv64"`) are also accepted as-is for less common platforms.
+
+If a tool specifies an `arch` restriction and the current architecture is not in the list, mise will skip installing and using that tool. When both `os` and `arch` are specified, both must match.
+
 ## Tool Dependencies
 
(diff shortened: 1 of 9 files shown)


risu729:
I think it's better to support tools only supported on Linux and Windows, but only for x64. Some tools lack precompiled binaries only for Windows arm64.
I'm not sure if this is good syntax for mise, but Aqua supports `supported-envs`.
https://aquaproj.github.io/docs/reference/registry-config/supported-envs

It might be better to wait for jdx's decision. Current implementation could be fine.

webkaz:
All review points have been addressed in the latest commit (2953bd9):

1. **Scalar `arch`/`os` handling** (greptile, codex, gemini-code-assist): Fixed — `arch = "arm64"` (string without brackets) now stores `Some(vec![s])` instead of falling into `opts`. Same fix applied to `os` for consistency.
2. **Multi-version serialization** (codex, Copilot): Fixed — `replace_versions` multi-version path now writes `os`, `arch`, and `install_env` into inline tables, preventing filter loss on re-save.
3. **Safe array parsing** (Copilot, gemini-code-assist): Fixed — `unwrap()` replaced with `ok_or_else`/`collect::<Result>` pattern (matching `depends`), in both `MiseTomlToolList` and `MiseTomlTool` visitors.
4. **`is_empty()` update** (gemini-code-assist, Copilot): Fixed — now includes `os.is_none() && arch.is_none()` to prevent filter loss when only platform filters are set.

webkaz:
@risu729 Good point about the cross-product case (e.g. "Linux and Windows, but only x64"). The current `os` + `arch` design handles it as independent filters — both must match — which covers common use cases like "arm64 only" or "linux only", but can't express "x64 on linux + any arch on macos" in a single entry.

Aqua's `supported-envs` approach (explicit os/arch pairs like `linux/amd64`) is more expressive for those cases. That said, the separate-field approach was chosen to mirror the existing `os` filter syntax and keep backwards compatibility. If @jdx prefers the combined approach, happy to rework.

webkaz:
Closing — the os/arch compound syntax in #9088 and schema support in #9095 cover this use case. Thanks @RobertDeRose and @risu729!

PR #9081, opened 2026-04-14
fix(activate): export __MISE_EXE and resolve bare ARGV0 to absolute path

fru1tworld:
Refs https://github.com/jdx/mise/discussions/8984

`__MISE_EXE` was not exported in bash activate, so child shells couldn't access it and the `mise` function failed. Additionally on Linux, when ARGV0 is a bare name (e.g. `mise`) instead of an absolute path, PATH changes could break execution.

- add `export` to `__MISE_EXE=` in `activate.sh`
- resolve bare ARGV0 via `which::which`, falling back to `MISE_BIN`

Verification:
- `mise run lint-fix`
- `mise run test:unit` (629 passed)
- `mise run test:e2e test_activate_export`


fru1tworld:
Addressed the relative path edge case — `which` result is now joined with `current_dir()` when non-absolute.

PR #9089, opened 2026-04-14
feat(task): allow to set confirmation default

roele:
Currently the default selected option for `confirm` in tasks is `Yes` which might not be a sensible default for destructive actions. While keeping this behaviour we add an option to define `confirm` as map with `message` and `default=yes|no`.


jdx:
maybe it would also be useful to have some way to force the user to type something in like "I'm really sure" or an app name or something for confirmations?

doesn't replace this though, I think this is still useful on its own

PR #9105, opened 2026-04-15
fix(backend): honor install_before for pipx installs

risu729:
See https://github.com/renovatebot/renovate/issues/41654

## Summary
- forward `install_before` to transitive Python package resolution when `pipx:` tools are installed through uv
- pass uv's `--exclude-newer <timestamp>` flag on the uv install path only
- leave the `pipx` fallback path unchanged
- update docs for the uv-only transitive release-age behavior
- add focused unit coverage for uv `--exclude-newer` argument generation

## uv compatibility
`--exclude-newer` was added in https://github.com/astral-sh/uv/pull/2166 and is included in https://github.com/astral-sh/uv/releases/tag/0.1.15 (released in Mar 2024), so we do not need to check the uv version before using this flag.

## Tests
- `cargo fmt --check`
- `git diff --check`
- `cargo test -q uv_exclude_newer_args`


risu729:
Closing in favor of https://github.com/jdx/mise/pull/9190, which includes the uv `--exclude-newer` behavior from this PR and adds the pipx fallback behavior in the same small diff.

PR #9107, opened 2026-04-15
fix(conda): honor install_before for conda installs

risu729:
## Summary
- forward `install_before` to the rattler conda solver during fresh conda installs
- expose conda repodata timestamps in `VersionInfo` so `latest`/fuzzy version resolution can honor the cutoff before the install solve runs
- keep lockfile URL installs and non-install conda solve paths unchanged
- update the conda backend docs plus shared `install_before` docs
- replace the synthetic solver-task unit tests with focused coverage for the cutoff helper and conda repodata timestamp handling

## Why conda can support this
mise's conda backend already solves installs through rattler, so the cutoff maps directly to the solver instead of needing a separate package-manager flag:

- `rattler_solve::SolverTask` has an `exclude_newer` field documented as excluding packages newer than the configured cutoff: https://docs.rs/rattler_solve/latest/rattler_solve/struct.SolverTask.html#structfield.exclude_newer
- `rattler_solve::ExcludeNewer` is the public configuration for filtering packages newer than a cutoff: https://docs.rs/rattler_solve/latest/rattler_solve/struct.ExcludeNewer.html
- conda repodata records expose package creation time through `rattler_conda_types::PackageRecord::timestamp`, which mise now carries into version metadata: https://docs.rs/rattler_conda_types/latest/rattler_conda_types/struct.PackageRecord.html#structfield.timestamp
- Pixi, another rattler-based conda tool, documents the same `exclude-newer` concept for conda packages: https://pixi.prefix.dev/v0.51.0/reference/pixi_manifest/#exclude-newer-optional
- rattler-build also exposes `--exclude-newer` as a solver option for excluding packages newer than an RFC3339 timestamp: https://rattler-build.prefix.dev/latest/reference/cli/

Because rattler applies the cutoff while solving, this covers the requested conda package and its solved runtime dependencies during fresh installs. Locked installs still use the already-recorded URLs/checksums.

## Tests
- `cargo test --quiet conda::tests`
- `cargo fmt --check`
- `git diff --check`


risu729:
I dug into the root-version-resolution side of this after the solver-only fix.

`install_before` still does not affect `mise latest conda:...` on the current rattler-backed listing path. With the current branch built locally, `MISE_INSTALL_BEFORE=2024-01-01 target/debug/mise latest conda:ruff` still returned `0.15.11`, and the debug log showed:

```text
Backend conda:ruff does not provide release dates; --before filter may not work as expected
```

I also checked the raw conda repodata and it does contain `timestamp` fields for ruff records, but I could not find a way to get usable upload timestamps through the current `rattler_repodata_gateway::Gateway`-based version-listing path already used by this backend without switching to direct `api.anaconda.org` calls. Since we do not want to start using that direct API here, I’m closing this PR instead of leaving a partial implementation that only constrains the install-time solve.

PR #9109, opened 2026-04-15
feat(backend): add npm package-manager install options

risu729:
## Summary
- add npm backend tool options for manager-specific install arguments: `npm_args`, `pnpm_args`, `bun_args`, and `aube_args`
- resolve the active npm package manager at install time, reusing the upstream `auto`/`aube` flow and applying args inline in each manager branch, matching the `pipx` backend style
- include npm install-argument options in lockfile option resolution and install-time option filtering
- document the npm backend tool options and add e2e assertions that each package manager receives its configured args, including aube
- remove the unnecessary `install_env` / `env` alias support from the original version of this PR

## Tests
- `cargo fmt --check`
- `cargo test --quiet test_resolve_lockfile_options_includes_install_args_only`

Not run locally:
- `mise run test:e2e e2e/backend/test_npm_package_manager e2e/backend/test_npm_aube` - local task discovery reported `no tasks defined in ~/.ghr/github.com/risu729/mise`, so I did not run the e2e scripts directly.


jdx:
I think this is ok @risu729 but it has merge conflicts

risu729:
Sorry, I was busy. It should work now.

PR #9110, opened 2026-04-15
feat(backend): support aqua vars templates

risu729:
## Summary
- add support for aqua registry `vars` definitions in the aqua backend
- render string vars as `.Vars.<name>` in aqua templates, including function arguments and chained property access
- accept string vars from tool options as either `vars = { name = "value" }` or flat `name = "value"` keys
- apply aqua-style required/default validation, while rejecting non-string var values with a focused error
- include aqua string var options in lockfile artifact identity and bin-path cache keys so different var values do not reuse stale resolution data
- document aqua vars options and required/default behavior

## Background
- Fixes https://github.com/jdx/mise/discussions/7160.
- Aqua added registry `vars` in v2.31.0 via https://github.com/aquaproj/aqua/pull/3052.
- Aqua registry vars are exposed to templates as `.Vars.<name>` and can declare `required` and `default` fields; aqua applies defaults and errors when a required var has no value/default.
- Current registry entries that need this include:
  - `aqua:flutter/flutter`, which uses `.Vars.channel` with default `stable`
  - `aqua:scenarigo/scenarigo`, which uses required `.Vars.go_version`

## Difference From Aqua
- Upstream aqua stores package vars as `map[string]any` and passes that map directly into Go templates, so arrays/objects/non-string scalar values can exist there.
- mise's aqua registry renderer is string-oriented: template expansion ultimately resolves artifact names, URLs, and file paths as strings, and lockfile option identity is also string-valued.
- This PR intentionally supports only string aqua var values/defaults. Non-string values such as arrays or objects now fail with an explicit error like `aqua var \`vars.fixture_version\` must be a string, got array`.
- This keeps the implementation narrow for the registry entries this PR targets, instead of reimplementing Go template value semantics.

## Implementation Notes
- `crates/aqua-registry` models `vars`, validates missing required vars, and accepts string defaults only.
- The aqua backend maps mise tool options into string aqua var values before resolving assets, lockfile entries, installs, and bin paths.
- Nested `vars = { name = "value" }` wins over flat `name = "value"`, and both forms canonicalize to the same lockfile option identity.

## Tests
- `cargo fmt --check`
- `git diff --check -- . ':!mise.lock'`
- `cargo test -p aqua-registry`
- `cargo test backend::aqua::tests::`
- `mise run test:e2e e2e/backend/test_aqua_vars`

This PR was originally generated by Copilot and updated after manual review.


risu729:
@codex review

PR #9118, opened 2026-04-15
fix(task): forward task help args and add raw_args

jdx:
## Summary

- Forward task-side `--help`/`-h` after `--` to the task instead of letting the usage parser intercept it.
- Add `raw_args = true` for task definitions and file headers so proxy tasks can skip mise/usage argument parsing entirely.
- Treat `--` as the single boundary between mise parsing and task parsing: `mise run task -- -- --help` forwards `-- --help` to the task.
- Keep non-`run` subcommands intact when their command args contain a literal `run` after `--`, such as `mise exec -- npm run test --help`.

Refs https://github.com/jdx/mise/discussions/5460.

## Behavior

```toml
[tasks.manage]
raw_args = true
run = 'python manage.py'
```

```sh
mise run manage --help            # forwarded to manage.py
mise run manage migrate --fake    # all flags reach manage.py unchanged
mise run othertask -- --help      # forwarded ad hoc, no raw_args needed
mise run othertask -- -- --help   # forwarded as -- --help
```

`mise run task --help` without `--` and without `raw_args` still shows mise task help.

## Validation

- `cargo test --all-features cli::tests::test_escape_task_args_ignores_run_after_subcommand_separator`
- `cargo fmt --check`
- `git diff --check`
- `target/debug/mise run test:e2e e2e/tasks/test_task_raw_args`
- `target/debug/mise run test:e2e e2e/tasks/test_task_help`
- `target/debug/mise exec -- bash -c 'printf "%s\n" "$*"' _ npm run test --help`
- pre-commit `hk` checks during amend

*This PR description was updated by Codex.*

---

> [!NOTE]
> **Medium Risk**
> Changes core CLI argument preprocessing and task usage parsing, which can subtly affect how flags are routed between mise and tasks across multiple invocation forms.
> 
> **Overview**
> Task execution now supports **argument passthrough** via new `raw_args = true`, causing tasks to skip mise/usage argument parsing so all flags (including `--help`/`-h`) reach the underlying command.
> 
> Separately, CLI preprocessing was adjusted so a literal `--` is treated as the boundary between mise and task parsing: `-- --help` bypasses the usage parser (even when a usage spec exists), `-- -- --help` forwards `-- --help`, and non-`run` subcommands with their own `--` tails are left untouched. Docs, JSON schemas, and e2e/unit tests were updated to cover the new behavior and edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 846510fd44a8d19e6be8a913b57a91aa1c6530ad. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/tasks/task-configuration.md
+++ b/docs/tasks/task-configuration.md
@@ -308,6 +308,33 @@ this that no other tasks are running at the same time.
 In the future we could have a property like `single = true` or something that prevents multiple tasks
 from running at the same time. If that sounds useful, search/file a ticket.
 
+### `raw_args`
+
+- **Type**: `bool`
+- **Default**: `false`
+
+When `true`, mise does not parse arguments to the task at all — every argument
+is passed through verbatim to the underlying command, including `--help`/`-h`.
+Use this for tasks that act as a thin proxy for a tool which already has its
+own argument parser (e.g. `next build`, Django `manage.py`, Python scripts
+using `argparse`):
+
+```toml
+[tasks.manage]
+raw_args = true
+run = 'python manage.py'
+```
+
+```sh
+mise run manage --help          # forwarded to manage.py, not intercepted by mise
+mise run manage migrate --fake  # all flags reach manage.py unchanged
+```
+
+Without `raw_args`, mise intercepts `--help` and prints its own task help. As
+an ad-hoc alternative for individual invocations, you can also use
+`mise run task -- --help` — the `--` separator now bypasses mise's usage
+parser specifically for `--help`/`-h`.
+
 ### `interactive`
 
 - **Type**: `bool`
--- a/docs/tasks/task-configuration.md
+++ b/docs/tasks/task-configuration.md
@@ -333,9 +333,8 @@ mise run manage migrate --fake  # all flags reach manage.py unchanged
 Without `raw_args`, mise intercepts `--help` and prints its own task help. As
 an ad-hoc alternative for individual invocations, you can also use
 `mise run task -- --help` — the `--` separator now bypasses mise's usage
-parser specifically for `--help`/`-h`. A second, inner `--` escapes that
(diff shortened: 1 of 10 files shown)


jdx:
Rebased on main and pushed a fixup commit addressing review feedback:

**Addressed:**
- Extracted the bypass predicate into `Task::should_bypass_usage_parser()` so the rule lives in one place (Greptile P2 / Gemini medium).
- Added an inner-`--` escape hatch: `mise run task -- -- --help` stops the bypass scan at the second `--`, so the usage parser sees `--help` as a literal positional value instead of triggering the help passthrough.
- Added a file-task variant to `test_task_raw_args` so we exercise a real `\$@` proxy script (inline TOML scripts can't — args are appended to the command string, not passed as positional parameters). Greptile was right that the inline assertion passed for the wrong reason.

**Intentional / not changed:**
- The bypass for `-- <flag>` only triggers on `--help`/`-h`, not on arbitrary flags. Widening it to "any flag after `--`" would silently change semantics for existing users who rely on `-- --foo` going through the usage parser (covered by `test_task_options`, `test_task_flag_parsing`, etc.). The `raw_args = true` opt-in is the documented full-passthrough escape hatch.
- `has_any_args_defined` vs the explicit `args/flags/subcommands` check in `task_executor.rs`: this inconsistency is pre-existing, not introduced by this PR. Worth fixing in a follow-up but I'd rather keep this PR scoped.

**CI failure (`e2e-7`):** unrelated — `env/test_env_module_tools_no_spurious_warning` is failing because installing `github:jdx/fnox@1.20.0` hits `Verification failed: Invalid certificate issuer: expected sigstore, got 'Fulcio Intermediate l1'`. Reproduces locally on this branch and on main, so it's a pre-existing sigstore/Fulcio attestation issue, not something this PR touched.

PR #9120, opened 2026-04-15
fix: read 403 response body to show specific error instead of assuming rate limit

jasonwbarnett:
## Problem

The current 403 handler drops the GitHub response body entirely — `error_for_status_ref()` only includes the URL and status, not the body. So the "IP allow list" message GitHub sends is silently discarded and the user sees the generic warning instead.

To answer your question `do we not already display the message github is sending us?` — no, the body is not surfaced at all currently.

Full context: https://github.com/jdx/mise/discussions/9119

## Fix

**`src/http.rs`**: For 403 responses, read the body before bailing and include it in the error message. The body GitHub sends (e.g. the IP allow list message) is now visible to users.

**`src/main.rs`**: `show_github_rate_limit_err` now checks what's actually in the body:
- If body contains `"IP allow list"` → show the specific GitHub message plus a one-line actionable hint (unset GITHUB_TOKEN, since anonymous requests bypass IP allow lists for public repos)
- If body contains rate limit signals → already handled by `display_github_rate_limit_headers`
- No GITHUB_TOKEN → existing suggestion to set a token
- Anything else → the body is already in the error message, no extra noise

No bullet lists of possible causes. Each case shows exactly what GitHub told us.

## Before / After

**Before** (IP allow list case, x-ratelimit-remaining: 14987):
```
WARN  GitHub API returned a 403 Forbidden error. This likely means you have exceeded the rate limit.
```

**After**:
```
Error: HTTP status client error (403 Forbidden) for url (...): {"message": "Although you appear to have the correct authorization credentials, the `aquasecurity` organization has an IP allow list enabled, and your IP address is not permitted..."}
WARN  GitHub API returned 403: the target organization has an IP allow list and your IP is not permitted.
      Workaround: unset GITHUB_TOKEN — anonymous requests bypass IP allow lists for public repos.
```

**After** (actual rate limit):
```
WARN  GitHub rate limit exceeded. Resets at <time>
```
(handled by `display_github_rate_limit_headers`, body confirms it)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -169,15 +169,23 @@ fn handle_err(err: Report) -> eyre::Result<()> {
 fn show_github_rate_limit_err(err: &Report) {
     let msg = format!("{err:?}");
     if msg.contains("HTTP status client error (403 Forbidden) for url (https://api.github.com") {
-        warn!(
-            "GitHub API returned a 403 Forbidden error. This likely means you have exceeded the rate limit."
-        );
         if env::GITHUB_TOKEN.is_none() {
+            warn!(
+                "GitHub API returned a 403 Forbidden error with no GITHUB_TOKEN set."
+            );
             warn!(indoc!(
                 r#"GITHUB_TOKEN is not set. This means mise is making unauthenticated requests to GitHub which have a lower rate limit.
                    To increase the rate limit, set the GITHUB_TOKEN environment variable to a GitHub personal access token.
                    Create a token at https://github.com/settings/tokens and set it as GITHUB_TOKEN in your environment.
                    You do not need to give this token any scopes."#
+            ));
+        } else {
+            warn!(indoc!(
+                r#"GitHub API returned a 403 Forbidden error. This can have several causes:
+                   - Rate limit exceeded: check x-ratelimit-remaining in the response headers
+                   - IP allow list: the target GitHub organization has an IP allow list enabled and your IP is not permitted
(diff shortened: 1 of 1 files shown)


jdx:
this pr is wrong and we should not be listing all of the causes of 403s

jasonwbarnett:
@jdx thanks, I agree. I was iterating and wasn't planning on dumping the full list but being more introspective. 

Are you still against being more specific logging for 403 errors?

jdx:
I think so, 99% of the time this is due to rate limits so mentioning anything else is not helpful

jasonwbarnett:
@jdx I 100% agree that just listing potential causes is stupid and more confusing, for the reason you mention. I'm asking something more nuanced: What about actually introspecting response headers + body to give specific cause.

I'm motivated by this because I burned a half day tracking this down. So to save the future me or future someone else I'd love to contribute changes which validate it so rather than saying, "most likely" it says with confidence: "This is being rate limited" and even further if the headers show the low rates for anonymous it could give specific instructions to add GITHUB_TOKEN (or MISE_GITHUB_TOKEN). I'm just thinking more precise logging could help people.

Let me know your thoughts.

jdx:
do we not already display the message github is sending us?

jasonwbarnett:
@jdx no, it's just a generic error that says it's likely rate limiting (which to be fair, is true)

jdx:
ah ok, we should definitely fix that

PR #9131, opened 2026-04-15
feat(aqua): compress baked registry blobs with zstd

risu729:
## Summary
- compress each baked per-package registry YAML blob at build time with zstd
- embed compressed blobs via `include_bytes!` and keep alias/canonical lookup behavior unchanged
- lazily decompress selected blobs when fetching baked registry entries

## Binary size (serious profile, all features)
- before: `83,100,752` bytes
- after: `81,233,024` bytes
- delta: `-1,867,728` bytes (`-2.25%`)

## Notes
- this branch is based on `codex/aqua-baked-registry-yaml` and is expected to merge after #9043


risu729:
This compresses the size of binary 1,867,728 bytes; considering the original registry file is 3,003,656 bytes, this is good. However I think we don't need this as the binary is already 80MB and zstd compression costs in runtime.

PR #9135, opened 2026-04-16
fix(go): Remove install-time version override for subpath packages

c22:
The install method had a workaround that queried `go list -m -versions` on the full tool path (including subpath) and, if no versions were found, overrode tv.version to "latest". This was needed before #8968 because version resolution couldn't distinguish subpath packages from tagless modules.

After #8968, proxy-based resolution handles both cases correctly: subpaths get 404 (truncating to the root module), and tagless modules get their pseudo-version from @latest.

The old workaround now fights with the correct resolution, it queries the subpath (which always returns empty), overrides the version to "latest", and installs into a `.../latest/` directory.

On the next `mise up`, version resolution finds e.g. `1.7.0` but `.../1.7.0/` doesn't exist and causes a reinstall every time.

Apologies for missing this on the original PR!


c22:
> Stale test comment

Cleaned up

PR #9137, opened 2026-04-16
registry: add monodep

gracefullight:
Adds [monodep](https://github.com/gracefullight/mise-monodep-plugin) to the registry.

monodep deduplicates dependencies across polyglot monorepo workspaces managed by mise.
It reads `[monorepo].config_roots` from `mise.toml`, auto-detects the package manager
(bun/pnpm/npm for Node.js, uv for Python), and hardlinks shared packages
to a project-local `.monodep/store` to save disk space.

Commands: `install`, `sync`, `add`, `update`, `remove`, `doctor`, `why`, `plan`

Example usage: [first-fluke/fullstack-starter mise.toml](https://github.com/first-fluke/fullstack-starter/blob/main/mise.toml)

https://github.com/gracefullight/mise-monodep-plugin


gracefullight:
@jdx, could you let me know if there is any reason this PR should be closed?
I would appreciate it if you could point out anything I might have missed.

PR #9144, opened 2026-04-16
fix(self-update): ensure subcommand exists

salim-b:
even when the `self_update` Cargo feature is disabled.

Before, the `SelfUpdate` enum variant in `src/cli/mod.rs` inside the `Commands` enum was gated with `#[cfg(feature = "self_update")]`. This meant that when compiling without the `self_update` feature, the subcommand would be completely omitted from the CLI parser, instead of using the stub.

Furthermore, the `SelfUpdate` struct in `src/cli/self_update_stub.rs` did not derive the `clap::Parser` trait, nor did it have an asynchronous `run` method, which is required by the subcommand runner in `src/cli/mod.rs`.

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -241,7 +241,6 @@ pub enum Commands {
     Reshim(reshim::Reshim),
     Run(Box<run::Run>),
     Search(search::Search),
-    #[cfg(feature = "self_update")]
     SelfUpdate(self_update::SelfUpdate),
     Set(set::Set),
     Settings(settings::Settings),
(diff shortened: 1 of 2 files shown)


jdx:
Looks good overall, but I think the stub `SelfUpdate` struct should also mirror the fields from the real one (`-f/--force`, `-y/--yes`, `--version`). Otherwise on a no-`self_update` build, `mise self-update -y` will fail with a clap parse error for an unknown flag instead of hitting your nice "disabled at build time" bail. Ignoring the values in the stub's `run()` is fine — we just want the argument parsing to succeed so users get the helpful message.

PR #9247, opened 2026-04-18
feat(vfox): add backend batch install hook support

RobertDeRose:
## Summary
- add `BackendBatchInstall` support for `vfox-backend` plugins
- batch ready installs by backend while preserving the existing per-tool fallback path for plugins without the batch hook
- add focused e2e coverage for successful batched installs and partial per-item batch failures

## Details
This adds a new optional `backend_batch_install` hook to the vfox backend flow.
On the mise side, the install scheduler now groups ready tools by backend and dispatches batched installs when the backend supports them. For unsupported backends and plugins, installs continue through the existing per-tool path.
The vfox integration now:
- detects `backend_batch_install` from plugin metadata
- routes batch-capable `vfox-backend` plugins through the new hook
- preserves per-tool success/failure handling
- reuses the existing install wrapper semantics so lock handling, cleanup, and install state remain consistent

## Why
This is primarily for multi-tool backend plugins where one backend invocation can install several ready tools more efficiently than separate per-tool installs.
`mise-nix` is a good example:
- projects may define multiple `nix:` tools from nixpkgs or flakes
- without batching, mise calls the backend once per tool
- with batching, the backend can resolve/build multiple ready tools in a single `nix build` invocation
- that reduces repeated backend startup/work and makes multi-tool bootstrap and setup flows cheaper
This hook is also useful for other backend-style plugins that can naturally install several tools in one operation, such as:
- package-manager backends like npm/pnpm-style ecosystems
- custom organizational backends that fetch/build multiple tools from one registry or flake
- any `vfox-backend` plugin where one shared solver/download/build step is cheaper than repeating the same work per tool
The feature is intentionally opt-in:
- plugins without `BackendBatchInstall` continue using the existing per-tool path
- dependency ordering and per-tool success/failure behavior are preserved

## Documentation
- document `BackendBatchInstall` in backend plugin development docs
- document optional batch install support in the vfox backend docs

## Tests
- `cargo check`
- `mise run test:e2e e2e/plugins/test_backend_plugin_install`
- `mise run test:e2e e2e/plugins/test_backend_batch_install`
- `mise run test:e2e e2e/plugins/test_backend_batch_install_partial_failure`

### DIFF
--- a/crates/vfox/src/hooks/backend_batch_install.rs
+++ b/crates/vfox/src/hooks/backend_batch_install.rs
@@ -1,0 +1,156 @@
+use indexmap::IndexMap;
+use std::collections::HashMap;
+use std::path::PathBuf;
+
+use mlua::{FromLua, IntoLua, Lua, LuaSerdeExt, Value, prelude::LuaError};
+
+use crate::{Plugin, error::Result};
+
+#[derive(Debug, Clone, serde::Serialize)]
+pub struct BackendBatchInstallItem {
+    pub id: String,
+    pub tool: String,
+    pub version: String,
+    pub install_path: PathBuf,
+    pub download_path: PathBuf,
+    pub options: IndexMap<String, toml::Value>,
+}
+
+#[derive(Debug, Clone)]
+pub struct BackendBatchInstallContext {
+    pub tools: Vec<BackendBatchInstallItem>,
+}
+
+#[derive(Debug, Clone)]
+pub struct BackendBatchInstallResult {
+    pub version: Option<String>,
+    pub error: Option<String>,
+}
+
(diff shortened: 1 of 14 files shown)


jdx:
going to have to pass on this one, sorry. It greatly changes the way that tools work in mise in a way I'm not going to be able to investigate.

RobertDeRose:
> going to have to pass on this one, sorry. It greatly changes the way that tools work in mise in a way I'm not going to be able to investigate.

OK, can't say I'm not disappointed, it would have made tool backends like mise-nix much more efficient. But I understand. I did try very hard to ensure it doesn't change current paths unless the backend supports the optional hook.

PR #9269, opened 2026-04-20
feat(cli): add minimum release age flag to lock and ls-remote

risu729:
## Summary

Adds `--before` support to two more commands that need date-aware version discovery:

- `mise ls-remote --before <date>` filters remote version listings by release timestamp.
- `mise lock --before <date>` resolves fuzzy/latest tool requests against versions released before the cutoff before writing or previewing lockfile entries.
- Both commands honor all three cutoff sources: the CLI flag, per-tool `install_before`, and global `[settings] install_before`, with the CLI flag taking precedence.

This is intended to support potential mise integration work in Renovate, where tooling may need to list and lock versions using a release-date cutoff.

## Implementation Notes

`ls-remote` still fetches the normal cached full remote version metadata and applies the effective date filter after reading it. That keeps the cache independent of the requested cutoff and avoids caching a filtered result for one cutoff that could incorrectly affect a later query with a different cutoff.

`lock` builds the toolset with a `ResolveOptions.before_date` override when the flag is present, while preserving the existing per-tool and global `install_before` fallback behavior when it is not.

## Tests

- `mise run test:e2e e2e/cli/test_ls_remote e2e/cli/test_lock`
- `cargo fmt --check`


risu729:
Blocking: I need to make the scope of `mise lock` clear.

risu729:
CI note: `windows-e2e` failed for an unrelated external download timeout while running `e2e-win/zig.Tests.ps1`. The job timed out fetching `https://pkg.hexops.org/zig/zig-windows-x86_64-0.14.0-dev.2577+271452d22.zip`, so `test-ci` failed only because it aggregates the failed `windows-e2e` status.

All Linux E2E tranches, lint, unit/build jobs, docs, registry CI, and Greptile passed on the latest commit.

risu729:
CI note: the latest run is green except for `autofix`, which failed while running `mise run lint-fix` on schema validation for `schema/mise-registry-tool.json`:

`schema schema/mise-registry-tool.json is invalid`
`error: can't resolve reference #/definitions/toolId from id https://mise.en.dev/schema/mise-registry-tool.json`

This appears unrelated to this PR: the file was not changed here, and latest `upstream/main` moved the schema to `$defs` while that file still contains `#/definitions/...` refs. The normal `lint` job, all build/unit/e2e jobs including Windows, docs, registry CI, benchmark, Socket, and Greptile passed on commit `b7b4ca1ba600cf41ef1b8fd2c295580176846d63`.

docs/cli/ls-remote.md
```diff
@@ -25,6 +25,12 @@ same as the first argument after the "@"
 
 Show all installed plugins and versions
 
+### `--before <BEFORE>`
```
jdx: we should rename this --minimum-release-age to match the setting

PR #9276, opened 2026-04-21
fix(backend): stop fuzzy requests installing literal dirs

AsgardMuninn:
## Summary

Fix the root cause behind stale real install directories like `installs/<tool>/latest/` and explicit prefix requests resolving to literal on-disk directories when mise cannot determine a concrete version.

This PR also keeps the one-time migration for existing broken installs and teaches runtime symlink rebuild to replace stale real dirs with the correct symlink once a concrete install exists.

Importantly, this does **not** blanket-treat every short version like `1.0`, `4.13`, or `2026.1` as fuzzy. `~/src/mise-versions` contains many tools with legitimate 1-part or 2-part stable versions (`lua`, `ffmpeg`, `staticcheck`, `ccache`, `hlint`, `upx`, etc.), so exact short versions must continue to work.

## Root Cause

There were two separate problems:

1. Resolver fallback: truly fuzzy requests could collapse to the literal request string instead of a concrete version.

- `resolve_prefix` in `src/toolset/tool_version.rs` used `None => prefix`, so an explicit prefix request could resolve to the literal prefix when remote matching returned nothing.
- `latest` requests also had a bad fallback loop through `latest_installed_version` when `installs/<tool>/latest` existed as a real directory.

2. Runtime selection regression: after `dc840866` (`fix(env): use runtime symlink paths for fuzzy versions`), PATH/bin resolution began preferring request paths like `latest`, `24`, and `24.0`. That made previously latent bad directories actively win at runtime.

There was also a reinforcing loop in `latest_installed_version`: if `installs/<tool>/latest` existed as a real dir instead of a symlink, mise treated that as installed `latest` and kept reusing the alias literally.

## Changes

- `src/toolset/tool_version.rs`
  - Stop resolving `latest` and explicit `prefix:` requests to literal install dirs when no concrete version can be found.
  - Preserve exact/direct version requests, including legitimate short versions like `1.0`, `4.13`, or `2026.1`.

- `src/backend/mod.rs`
  - Change `latest_installed_version(None)` so a real `latest/` directory is no longer treated as installed `latest`.
  - If `latest/` is not a runtime symlink, fall back to the highest concrete installed version instead.

- `src/runtime_symlinks.rs`
  - Extend runtime symlink rebuild to replace stale real dirs whose basename differs from the concrete target (for example `latest -> ./2.0.0`, `24 -> ./24.3.1`).
  - This lets normal install/upgrade flows self-heal stale dirs once a concrete install exists.

- `src/migrate.rs`
  - Keep a one-time migration that rewrites existing stale fuzzy-version dirs into runtime symlinks and writes a marker file so it only runs once.
  - Removal note left in place so the migration can be dropped after the grace period.

## Tests

- Added `e2e/backend/test_fuzzy_versions_do_not_install_literal_dirs`
  - Verifies `mise install tool@latest`, `tool@prefix:24`, and `tool@prefix:24.0` fail instead of creating literal install dirs when no concrete version can be resolved.
- Kept `e2e/cli/test_upgrade_latest_stale`
  - Verifies a stale real `latest/` dir is repaired to `./2.0.0` during upgrade and runtime exec uses the new binary.
- Kept `e2e/tools/test_runtime_symlink_migration`
  - Verifies the one-time migration rewrites stale `1.0/` and `latest/` dirs and writes the marker file.
- Kept `e2e/tools/test_runtime_symlinks`
  - Verifies normal runtime symlink behavior still works.
- Added unit coverage for `latest_installed_version` ignoring a real `latest/` dir.

## Validation

- `cargo test --all-features test_latest_installed_version_ignores_real_latest_dir -- --nocapture`
- `cargo test --all-features runtime_path_does_not_return_file_based_runtime_symlink -- --nocapture`
- `mise run test:e2e e2e/backend/test_fuzzy_versions_do_not_install_literal_dirs e2e/cli/test_upgrade_latest_stale e2e/tools/test_runtime_symlink_migration e2e/tools/test_runtime_symlinks`

## Notes

This PR fixes the definite broken cases without regressing tools that genuinely use short exact versions:

- existing broken installs are migrated once
- stale dirs are self-healed during normal runtime symlink rebuilds
- new `latest` and explicit `prefix:` requests no longer create literal install directories when resolution fails
- exact short versions remain installable for tools whose real release scheme uses them

### DIFF
--- a/e2e/backend/test_fuzzy_versions_do_not_install_literal_dirs
+++ b/e2e/backend/test_fuzzy_versions_do_not_install_literal_dirs
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+
+PLUGIN_DIR="$MISE_DATA_DIR/plugins/fuzzytest"
+mkdir -p "$PLUGIN_DIR/bin"
+
+cat <<'LISTALL' >"$PLUGIN_DIR/bin/list-all"
+#!/usr/bin/env bash
+exit 0
+LISTALL
+chmod +x "$PLUGIN_DIR/bin/list-all"
+
+cat <<'INSTALL' >"$PLUGIN_DIR/bin/install"
+#!/usr/bin/env bash
+mkdir -p "$ASDF_INSTALL_PATH/bin"
+cat <<EOF >"$ASDF_INSTALL_PATH/bin/fuzzytest"
+#!/usr/bin/env bash
+echo "fuzzytest $ASDF_INSTALL_VERSION"
+EOF
+chmod +x "$ASDF_INSTALL_PATH/bin/fuzzytest"
+INSTALL
+chmod +x "$PLUGIN_DIR/bin/install"
+
+assert_fail "mise install fuzzytest@latest" "no versions found for fuzzytest"
+assert_fail "mise install fuzzytest@24" "no versions found for fuzzytest"
+assert_fail "mise install fuzzytest@24.0" "no versions found for fuzzytest"
+
+assert_fail "test -d $MISE_DATA_DIR/installs/fuzzytest/latest"
+assert_fail "test -d $MISE_DATA_DIR/installs/fuzzytest/24"
+assert_fail "test -d $MISE_DATA_DIR/installs/fuzzytest/24.0"
--- a/e2e/backend/test_fuzzy_versions_do_not_install_literal_dirs
+++ b/e2e/backend/test_fuzzy_versions_do_not_install_literal_dirs
@@ -21,8 +21,8 @@ INSTALL
 chmod +x "$PLUGIN_DIR/bin/install"
 
 assert_fail "mise install fuzzytest@latest" "no versions found for fuzzytest"
-assert_fail "mise install fuzzytest@24" "no versions found for fuzzytest"
-assert_fail "mise install fuzzytest@24.0" "no versions found for fuzzytest"
+assert_fail "mise install fuzzytest@prefix:24" "no versions found for fuzzytest"
+assert_fail "mise install fuzzytest@prefix:24.0" "no versions found for fuzzytest"
 
 assert_fail "test -d $MISE_DATA_DIR/installs/fuzzytest/latest"
 assert_fail "test -d $MISE_DATA_DIR/installs/fuzzytest/24"
--- a/e2e/cli/test_upgrade_latest_stale
+++ b/e2e/cli/test_upgrade_latest_stale
@@ -35,11 +35,10 @@ mise up
 # The resolved version dir must now exist — proving `install_version_` ran.
 assert "test -d $MISE_DATA_DIR/installs/dummy/2.0.0"
 
-# And the stale request dir must be repaired into the runtime symlink so execs
-# and future latest/version lookups resolve through the new concrete install.
-assert "readlink $MISE_DATA_DIR/installs/dummy/latest" "./2.0.0"
+# And runtime resolution must use the new concrete install, not the stale dir.
+assert "mise where dummy" "$MISE_DATA_DIR/installs/dummy/2.0.0"
 
-# The active binary must then report the new version.
+# The active binary must report the new version.
 assert_contains "mise x dummy -- dummy" "This is Dummy 2.0.0"
 
 rm -f mise.toml
--- a/e2e/cli/test_upgrade_latest_stale
+++ b/e2e/cli/test_upgrade_latest_stale
@@ -35,9 +35,11 @@ mise up
 # The resolved version dir must now exist — proving `install_version_` ran.
 assert "test -d $MISE_DATA_DIR/installs/dummy/2.0.0"
 
-# And the active binary must report the new version (mise's runtime path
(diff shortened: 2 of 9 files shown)


risu729:
You should add an e2e test that fails without this change, but passes with this change. I use `latest` for many tools, but it upgrades fine. I don't think this is the correct way to fix.
Also, doesn't it force-reinstall even the actual latest version is the same as the current one?

jdx:
I haven't looked deep but I suspect this will break shit. I'm surprised the e2e tests passed.

AsgardMuninn:
@risu729 @jdx — you were both right, the `force: true` approach was too broad. I've force-pushed a replacement commit that addresses the root cause in `is_version_installed` instead.

**TL;DR:** The early return compares `tv.request.install_path()` (which uses the **request** version like `"latest"`) against the filesystem, but never falls through to `tv.install_path()` (the **resolved** version). An existing guard already handled this for `ToolRequest::Prefix`; I've extended it to `ToolRequest::Version` so that channel pins like `@latest` check the resolved path when the request dir name doesn't match the resolved version.

This answers @risu729's "doesn't it force-reinstall even when the current version is the same as the latest?" — it no longer does. `force` stays `false` on the upgrade path; the narrow guard only triggers when the request basename (e.g. `"latest"`) genuinely differs from the resolved `tv.version` (e.g. `"1.21.0"`).

It also answers why you haven't hit this with most `@latest` tools — the bad state (a real `installs/<id>/latest/` directory instead of a symlink) only arises when the first install left `tv.version` set to the literal `"latest"`, which happens when resolve falls back to the alias in `resolve_prefix`. Backends like aqua / vfox / core / npm rarely trip that path; the Go backend is the most exposed because sub-module paths + goproxy behavior make transient resolve failures more likely.

E2e regression test included per @risu729's request: `e2e/cli/test_upgrade_latest_stale` sets up the bad state with a sentinel file and asserts that `mise up` actually installs the new version. Full root-cause analysis and tradeoffs in the updated PR description.

jdx:
oh is this why the "latest" as directory thing started happening? we definitely need to fix this

AsgardMuninn:
@jdx Confirmed — that's the mechanism. Three links in the chain:

1. `is_version_installed` short-circuit → fixed by this PR.
2. `runtime_symlinks::rebuild` (`src/runtime_symlinks.rs:58`) deliberately skips real dirs.
3. `latest_installed_version` (`src/backend/mod.rs:855-878`) returns the literal `"latest"` when the path isn't a symlink — that's the self-reinforcing loop: stale real dir → `"latest"` as a version → `resolve_version` re-pins `tv.version = "latest"` → runtime resolves back to the stale dir.

So the current PR makes `install_version_` run but the stale `installs/<id>/latest/` stays and wins at runtime — exactly what my e2e just caught.

Pushing a follow-up commit on this branch: after `install_version_` succeeds, when the request basename ≠ resolved `tv.version`, replace the stale real dir at the request path with a runtime symlink to the resolved dir — using the existing `make_symlink_or_file` helper, ~15 LOC, Windows-safe, tight guards (only `Version`/`Prefix` request, basename mismatch, non-symlink, inside our own installs dir). The existing `e2e/cli/test_upgrade_latest_stale` stays as regression proof (fails today, passes after).

Two calls for you:
- Local cleanup in `install_version` (surgical) or extend `runtime_symlinks::rebuild` to self-heal existing stuck states across all tools (broader, higher blast radius)?
- Want me to also harden `latest_installed_version` so it refuses to return the literal alias when the path isn't a symlink? That's the longer-lived root cause, but I'd do it as a separate PR unless you'd rather bundle.

jdx:
I want some time to look this over properly but should get back to you soon. Nice work identifying this.

jdx:
I dug into the regression path a bit more and found the behavior split across two changes:

1. The stale real directories themselves come from an old fallback in `resolve_prefix` / latest resolution where we fall back to the literal request when remote matching returns no result (`latest`, `24`, `24.0`, etc.). That can leave real install dirs named after the fuzzy request instead of the resolved version.
2. The reason it started breaking users recently is `dc840866` (`fix(env): use runtime symlink paths for fuzzy versions`). After that change, runtime PATH/bin resolution started preferring request paths like `latest`, `24`, and `24.0`, so once one of those existed as a real directory, it began winning at runtime instead of the concrete install dir.

To address both the new and existing bad states, I pushed a follow-up commit on this branch (`2f01b8051`):

- an install-time repair that rewrites a stale request-path dir into the correct runtime symlink after a successful resolved install
- a one-time startup migration in `migrate.rs` that rewrites existing stale fuzzy-version dirs into runtime symlinks and drops a marker file so it only runs once

That handles existing installs with low ongoing blast radius, which seems to match the concern here better than a perpetual broad cleanup. I also left a removal note on the migration so it can be deleted after the grace period.

Locally verified with:
- `mise run test:e2e e2e/cli/test_upgrade_latest_stale e2e/tools/test_runtime_symlink_migration e2e/tools/test_runtime_symlinks`

PR #9289, opened 2026-04-22
fix(backend): repair stale latest runtime directory

risu729:
## Summary

Fixes stale real request directories such as `installs/<tool>/latest/` without including #9276.

## Changes:
- add `e2e/cli/test_upgrade_latest_stale` as a non-executable e2e regression test
- before real installs, remove stale request paths for `ToolRequest::Version` aliases where the requested version differs from the resolved concrete version, so the existing install-skip checks no longer treat stale `latest/` as installed
- after successful concrete installs, run the same cleanup before runtime symlink rebuild
- limit cleanup to direct real files/directories under the backend managed installs directory
- skip runtime symlinks and do not apply the cleanup to `ToolRequest::Prefix`

`runtime_symlinks::rebuild` remains responsible for recreating `latest -> ./<resolved-version>` after the stale blocker is gone.

## Validation

Red test before fix commit:
- `mise run test:e2e e2e/cli/test_upgrade_latest_stale` failed at `test -d .../installs/dummy/2.0.0`, showing the stale `latest/` directory still short-circuited install on `jdx/mise:main`

Passing checks after fix commit:
- `mise run test:e2e e2e/cli/test_upgrade_latest_stale`
- `shellcheck e2e/cli/test_upgrade_latest_stale`
- `shfmt -d -s e2e/cli/test_upgrade_latest_stale`
- `mise run test:e2e e2e/cli/test_latest_explicit_latest`

### DIFF
--- a/e2e/cli/test_upgrade_latest_stale
+++ b/e2e/cli/test_upgrade_latest_stale
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+
+cat <<TOML >mise.toml
+[tools]
+dummy = "latest"
+TOML
+
+mise uninstall dummy --all
+
+mkdir -p "$MISE_DATA_DIR/installs/dummy/latest/bin"
+cat <<'SCRIPT' >"$MISE_DATA_DIR/installs/dummy/latest/bin/dummy"
+#!/usr/bin/env bash
+echo "This is stale dummy"
+SCRIPT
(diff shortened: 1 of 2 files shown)


risu729:
I think #9276 is sufficient for the fix. I don't think this PR is doing the correct thing.

PR #9329, opened 2026-04-23
feat: opt-in to pre-release versions for github and aqua backends

jakedgy:
Implements the opt-in proposed in discussion #9323 (approved by @jdx). Also wires support into the `aqua:` backend as a small follow-up, as requested.

## What changed

Adds a per-tool `prerelease = true` option to the `github:` and `aqua:` backends. When set:

- Releases flagged `prerelease: true` on GitHub appear in `mise ls-remote`
- `latest` resolves against the full list including pre-releases
- Fuzzy version queries (e.g. `1.2`) match pre-release tags under that prefix

```toml
[tools]
"github:myorg/mytool" = { version = "latest", prerelease = true }
"aqua:owner/tool"     = { version = "latest", prerelease = true }
```

Default behavior is unchanged (`prerelease = false`). Draft releases are always excluded.

## Where the filters lived

Two independent filters needed an escape hatch:

1. **GitHub API flag filter** in `src/github.rs::list_releases_` (`!r.prerelease`). Split so the cache holds unfiltered-except-drafts; `list_releases` preserves old behavior via read-time filter, new `list_releases_including_prereleases{,_from_url}` exposes the unfiltered view. Cache is shared — no extra API cost.
2. **`VERSION_REGEX` filter** in `fuzzy_match_filter` (trait default + `aqua` override). Extracted into a shared `fuzzy_match_versions(versions, query, filter_prereleases)` helper so both call paths can opt out.

Also, the `github:` backend's `latest_stable_version` uses the `/releases/latest` endpoint (stable-only by design) — when `prerelease = true` it now falls through to `latest_version_for_query("latest", None)` to resolve against the full list. Uses the `_for_query` variant deliberately to avoid recursing back through `latest_stable_version`.

## Tests

- `fuzzy_match_versions`: unit tests pinning both the default (filters prereleases) and opted-in (includes them) behavior across `latest` and partial queries
- `UnifiedGitBackend::fuzzy_match_filter`: unit tests verifying the override reads `self.ba.opts()` and toggles correctly
- `include_prereleases`: bool-parsing sanity check
- All 750+ existing unit tests still pass (one pre-existing failure — `test_inline_install_before_wins_over_config_entry` — reproduces on clean `main`, unrelated)

## What isn't here

**No e2e test yet** — `jdx/mise-test-fixtures` only has one stable release, so exercising the opt end-to-end needs a fixture with at least one release flagged `prerelease: true`. Happy to add that in a follow-up if you can flag an existing tag (or I can mock this differently if you'd prefer).

**Not touching GitLab/Forgejo** — the opt is documented as a no-op for those backends. Can extend later if there's demand.


docs/dev-tools/backends/github.md
```diff
@@ -260,6 +260,23 @@ For GitHub Enterprise or self-hosted GitHub instances, specify the API URL:
...
+
+```toml
+[tools]
+"github:myorg/mytool" = { version = "latest", prerelease = true }
+```
+
+When set:
+
+- Pre-release tags (e.g. `v1.0.0-rc1`, `v0.1.2-dev.86`) appear in `mise ls-remote`.
+- `latest` resolves to the newest version across stable **and** pre-releases, rather than taking the GitHub `/releases/latest` shortcut (which is stable-only by design).
```
risu729: This is not actually true. We can set an arbitrary release as "latest" using the API on GitHub.

src/backend/github.rs
```diff
@@ -257,6 +271,14 @@ impl Backend for UnifiedGitBackend {
         let api_url = self.get_api_url(&opts);
         let version_prefix = opts.get("version_prefix");
 
+        // When `prerelease = true`, skip the `/releases/latest` shortcut (which
+        // is stable-only by design) and resolve `latest` against the full list,
+        // which now includes pre-releases. Uses `latest_version_for_query` to
+        // avoid recursing back into this method.
+        if include_prereleases(&opts) {
+            return self.latest_version_for_query(config, "latest", None).await;
```
risu729:
```suggestion
            return Ok(None);
```

I refactored in https://github.com/jdx/mise/pull/9304, so just return `Ok(None)` and the caller will handle the fallback.

src/backend/mod.rs
```diff
@@ -1537,57 +1579,7 @@ pub trait Backend: Debug + Send + Sync {
     }
 
     fn fuzzy_match_filter(&self, versions: Vec<String>, query: &str) -> Vec<String> {
```
risu729: I think it's better to add `filter_prereleases` to this trait function. I might be wrong.

risu729 (changes requested):

src/github.rs
```diff
@@ -163,7 +186,7 @@ async fn list_releases_(api_url: &str, repo: &str) -> Result<Vec<GithubRelease>>
             headers = h;
         }
     }
-    releases.retain(|r| !r.draft && !r.prerelease);
+    releases.retain(|r| !r.draft);
 
     Ok(releases)
 }
```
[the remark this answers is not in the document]
jakedgy: Thanks for flagging this — we considered it and are going to accept it as a known one-time tradeoff. Reasoning:

**Duration bound:** default `fetch_remote_versions_cache = "1h"`, so the stale window is at most an hour on commands that miss the `prefer_offline` fast path (i.e. `ls-remote` / `install` — the two commands that surface this).

**Affected population is narrow:** requires that a user (a) ran `ls-remote`/`install` against the repo within the last hour on the old version, (b) upgraded mise, (c) added `prerelease = true` to their config, and (d) retried before the TTL expired.

**Workaround exists and is cheap:** `mise cache clear`, or deleting the specific `<repo>-releases.msgpack.z` file.

**Cost of fixing is non-zero:** bumping the cache key (e.g. `{key}-all`) orphans every existing on-disk cache file across the whole mise install base — a permanent cost in disk/latency to solve a ~1h transient for the small subset above.

Given the ratio, accepting the degradation seems like the right call. Documenting here so it's discoverable if anyone hits it.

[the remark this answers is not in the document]
jdx: users are not going to be able to work with this if you don't resolve this issue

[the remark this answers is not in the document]
jakedgy: Fair — you're right that surfacing "no prereleases" with no error is a real usability issue, not a small tradeoff. Fixed in 6bf294d21 by renaming the on-disk cache file from `{key}-releases.msgpack.z` to `{key}-all-releases.msgpack.z`. Old caches are ignored, post-upgrade runs start fresh, and the `list_releases` / `list_releases_including_prereleases` callers still share a single cache entry.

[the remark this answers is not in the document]
jdx: > and the list_releases / list_releases_including_prereleases callers still share a single cache entry.

this is still a problem, users that modify this config in mise.toml will not have updated versions so they will think it's broken

src/backend/github.rs
```diff
@@ -288,6 +310,15 @@ impl Backend for UnifiedGitBackend {
         }
     }
 
+    /// When `prerelease = true`, skip the default `VERSION_REGEX` filter so
+    /// that pre-release tags (e.g. `1.0.0-rc1`, `0.1.2-dev.86`) survive fuzzy
+    /// matching against `latest` and partial queries. Otherwise defer to the
+    /// default behavior.
+    fn fuzzy_match_filter(&self, versions: Vec<String>, query: &str) -> Vec<String> {
+        let filter = !include_prereleases(&self.ba.opts());
```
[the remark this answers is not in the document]
jdx: @greptileai I think this is still a problem

risu729:
I think there is still a cache correctness issue here around `prerelease` and the shared backend `remote_versions` cache.

`list_remote_versions_with_info()` caches the returned `Vec<VersionInfo>` via `get_remote_version_cache()`, keyed by `self.ba().full()` / `remote_versions.msgpack.z`. That cache key does not include config-derived tool opts like `prerelease`.

That means the result depends on which mode populated the cache first:

- first run with `prerelease = false`: `remote_versions` can be stable-only, and changing to `prerelease = true` will keep serving the stable-only cache until expiry/clear
- first run with `prerelease = true`: `remote_versions` can include prereleases, and changing back to false can still expose them in paths that read `list_remote_versions_with_info()` directly, e.g. `ls-remote`

I think the better fix is probably to make `remote_versions` option-independent by caching the superset, effectively fetching as though prereleases are included, then filter after reading the cache based on the current tool opts. To make that correct, `VersionInfo` may need to carry prerelease metadata instead of relying only on the version-string regex. Also, when `prerelease = true`, the versions-host path probably needs to be bypassed or extended to return prerelease-aware metadata, otherwise it can still serve a stable-only list before the backend code runs.

This is written by Codex, but I pointed out the bug to it so should be true.

jakedgy:
@risu729 thanks for catching this — you were right that the original change only solved half the problem. Pushed eba620653 to fix it.

A few notes on the approach and the principle behind it.

## The fix

I went with your superset suggestion. `VersionInfo` gains a `prerelease: bool` stamped from GitHub's release flag, and the shared remote-versions cache stores the full superset. The `prerelease` tool option is applied as a read-time filter rather than affecting cache content, so the cache becomes option-independent.

## Versions host: bypass always for github + aqua

I extended the bypass to *always* skip the host for these backends, not only when `prerelease = true`. The "skip only when opted in" approach left a real gap: flipping `prerelease = false → true` with a populated stable-only cache (filled by the host) would still serve the wrong list until TTL.

## The UX principle that drove the choice

The case I had in mind: a developer who needs a pre-release version of a tool while working on something specific. The natural way to express that is a project-local override — `mise.local.toml` with `prerelease = true` — and they'll `cd` in and out of that project regularly, possibly over weeks.

Under that workflow the right behavior is for flipping in/out to be **invisible** — no spinner, no API call, just the right list each time. Anything that requires a refetch on flip violates the principle of least surprise: *"why did mise hit GitHub? I didn't change anything, I just `cd`'d."*

The two main alternatives didn't satisfy that:

- **`is_superset` marker on the cache** — preserves the host shortcut for default users, but costs one refetch on the first flip into the project. Acceptable, but adds a one-time hiccup that needs explaining.
- **Two cache slots (per-opt)** — refetch on every TTL expiry in *each* mode independently, plus the possibility of the slots drifting (e.g. a new stable release visible in one mode but not the other around refresh time). Worse on both fronts.

The cost of always-bypass for default users is small. The on-disk releases cache (`{key}-all-releases.msgpack.z`) absorbs repeated upstream calls, and the per-backend cache absorbs the rest. The actual extra GitHub API hit is one call per repo per cache window on cold cache — well within unauth rate limits even for users with many tools, and a no-op for anyone with `GITHUB_TOKEN` set.

## Other things in this commit

- **Aqua's `get_version_tags`** (used during install/lock resolution) also drops its inline `prerelease` opt read and fetches the superset. This collapses the open P1 from Greptile naturally — a project-local override pinning a pre-release version now resolves correctly regardless of what the ambient mise config says.
- **Cache filenames bumped** (`remote_versions_v2.msgpack.z`, `version_tags_v2.msgpack.z`) so users upgrading don't carry forward pre-schema caches that the new read-time filter would mis-filter.
- **Aqua tag-source packages** (the `github_tag` version source) keep `prerelease = false` since git tags carry no flag — the existing regex-based filter in `fuzzy_match_versions` continues to handle those tags as it does today. Behavior on that path is unchanged from this PR's previous state.

## Verification

Manual integration test against `github:goreleaser/goreleaser` (which has rc tags):

| state | versions returned | cache file mtime |
|---|---|---|
| `prerelease = true` (cold) | 30 | written |
| flip to `prerelease = false` | 29 | unchanged |
| flip back to `prerelease = true` | 30 | unchanged |

Plus two new unit tests in `backend::tests` covering the read-time filter for backends that stamp the flag and those that don't.

src/backend/mod.rs
```diff
@@ -532,10 +662,15 @@ pub trait Backend: Debug + Send + Sync {
...
                     .is_some_and(|o| o.contains_key("version_list_url")));
         let versions_host_applies = match backend_type {
-            BackendType::Github
-            | BackendType::Gitlab
+            // github + aqua intentionally bypass the versions host. Both honor
+            // the `prerelease` tool option, and we cache the pre-release superset
+            // locally so a user flipping `prerelease = true` (e.g. via a project
+            // override) sees the right list immediately. The versions host serves
+            // a stable-only list, so reusing it would either re-introduce the
+            // option-dependent cache bug or force per-opt cache slots.
```
risu729: We shouldn't do this, the purpose of the versions host is to avoid GitHub rate limits, which applies for Aqua and GitHub backends mainly. We should include the prerelease metadata in the versions host ideally, but check if it doesn't break old mise versions. Then include `prerelease` metadata in `mise ls-remote --json` output. I think you also need to create a PR to `jdx/mise-versions` to update the DB schema.

risu729 (changes requested):

src/backend/mod.rs
jdx: ah I think you're right, I didn't think of this

jakedgy: @risu729 @jdx — agreed. Pushed aac50e05a taking the host-routes-everything approach you outlined.

## Summary

- Add `prerelease: bool` (with `serde(default = false)`) to the versions-host `VersionEntry` schema, propagated onto `VersionInfo` from `versions_host.rs`.
- Drop the `BackendType::Github | BackendType::Aqua => false` bypass so both backends route through the host like every other rate-limited backend.
- Add `prerelease` to `mise ls-remote --all --json` (it was already serialized in single-tool mode via `VersionInfo`; now also present in `--all --json` via `VersionOutputAll`).
- Backend-side `_list_remote_versions` (still used as fallback when the host is rate-limited / down / opted-out / `prefer_offline`) continues to populate `prerelease` from `github::list_releases_including_prereleases*`, so the offline/non-host path is consistent.

## Backwards compatibility

Old mise versions reading the host data: `toml-rs` ignores unknown fields by default (we don't set `deny_unknown_fields` on `VersionEntry`), so adding `prerelease = true/false` to entries in `mise-versions` is forward-compatible — old clients see exactly the same `created_at` + `release_url` they see today.

New mise reading old host data: `serde(default)` on `prerelease` falls back to `false`, so any host TOML that hasn't been re-rendered yet looks the same to the new client as today (no prereleases visible). Once the companion mise-versions PR ships and re-renders, prereleases start surfacing automatically — no client change required.

Net: this PR is safe to merge in either order with the mise-versions schema update.

## Companion PR

The host TOML producer side is in `jdx/mise-versions`. I'll open a PR there to:
1. Include pre-release entries (currently filtered out at fetch time).
2. Stamp each entry with the GitHub `prerelease` flag (and the equivalent for aqua sources).
3. Verify the rendered TOML stays parseable by the released mise versions before tagging.

The feature is functionally inert for github+aqua users until that ships and renders, but no regressions land in the meantime: with `prerelease = false` (the default) the user sees what they see today; with `prerelease = true` they see the stable-only list until the host catches up, which is the same as today minus the local-bypass shortcut.

PR #9335, opened 2026-04-23
fix(backend): skip versions host for custom remote options

risu729:
## Summary

- add list-remote option key hooks for backends whose remote version list depends on config
- skip `mise-versions.jdx.dev` when those options differ from registry/default backend options
- add an e2e regression for a registry GitHub tool with a custom `version_prefix`

## Why

The versions host is keyed only by tool short name, so it cannot represent config-specific backend options such as `version_prefix`, `api_url`, or the planned `prerelease` option (in https://github.com/jdx/mise/pull/9329). Registry tools with those overrides need to query the backend directly for `ls-remote`/version resolution.

## Validation

- `cargo fmt --check`
- `cargo check`
- `cargo build`
- `e2e/run_test cli/test_ls_cache`

Note: `mise run test:e2e cli/test_ls_cache` was attempted first, but this local checkout's task discovery returned `no tasks defined`; the repository e2e harness was run directly after building the local binary.


risu729:
Superceded by https://github.com/jdx/mise/pull/9568

PR #9392, opened 2026-04-25
fix(http): detect html without doctype

phateffect:
Bug Fix for Nexus PyPI responses
Older versions of Nexus PyPI return responses that don’t strictly start with <!DOCTYPE html> but still contain HTML.
This change ensures those responses are correctly handled as HTML.

### DIFF
--- a/src/http.rs
+++ b/src/http.rs
@@ -208,7 +208,7 @@ impl Client {
         let url = url.into_url().unwrap();
         let resp = self.get_async(url.clone()).await?;
         let html = resp.text().await?;
-        if !html.starts_with("<!DOCTYPE html>") {
+        if !looks_like_html(&html) {
             bail!("Got non-HTML text from {}", url);
         }
         Ok(html)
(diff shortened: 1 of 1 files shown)


jdx:
upgrade nexus?

phateffect:
> upgrade nexus?

I understand the perspective, but "upgrading Nexus" is often not an option for the end-user.

Major mirrors fail this check: Even one of the most widely used mirrors, Aliyun, doesn't pass this strict prefix check. If I `export MISE_PIPX_REGISTRY_URL=https://mirrors.aliyun.com/pypi/simple/{}/`, mise currently fails.
```bash
curl -s https://mirrors.aliyun.com/pypi/simple/flask/ | head -n 5
                 <-- Note the leading whitespace/newline here
<!DOCTYPE html> 
<html>
  <head>
    <meta name="pypi:repository-version" content="1.0">
```


Enterprise Constraints: In many corporate environments, developers have zero control over the Nexus/Artifactory versions managed by IT. If mise doesn't work out-of-the-box, users are forced to stick with asdf or uvx.

pip and other tools generally rely on the Content-Type header or a more resilient HTML parser rather than a strict DOCTYPE prefix.

If the current `looks_like_html` is not acceptable, would you be open to a fix that checks the Content-Type header instead?

PR #9398, opened 2026-04-25
registry: add dlv

scop:


registry/dlv.toml
```diff
@@ -0,0 +1,3 @@
+backends = ["aqua:go-delve/delve"]
+description = "Debugger for the Go programming language"
+test = { cmd = "dlv version", expected = "Version: {{version}}" }
```
scop, line 3: test-tool for this actually fails for me at the moment:
```
   0: Failed to install aqua:go-delve/delve@latest: package type `go_install` is not supported in the aqua backend. Use the go backend instead: go:github.com/go-delve/delve/cmd/dlv.
```
This looks like an issue with the mise aqua-registry implementation: the default type including the latest for go-delve/delve in it is github_release; go_install is just for one specific version override.

PR #9414, opened 2026-04-26
fix(http): retry transient HTTP failures with backoff and warn on rescue

jdx:
## Summary

- Turn on transient-failure retries by default (3 attempts) for all HTTP work in mise. Previously `http_retries` defaulted to 0, so a one-off 502 from GitHub releases (like the one in [run #24963676401](https://github.com/jdx/mise/actions/runs/24963676401/job/73095959056)) failed installs immediately even though the retry plumbing was already there.
- Classify errors: retry **5xx, 408, 429, network/connect/timeout, mid-stream body drops**; do **not** retry other 4xx (avoids wasting time on legitimate 404s when probing for tarball variants).
- Cover the chunk-streaming path in `download_file_with_headers` too — the original retry only wrapped the initial GET, so a connection drop mid-tarball still failed. Mirrored the same coverage in the `vfox` crate's `download()` and lua `download_file`.
- Emit a `warn!` whenever a retry rescues a request, with the original transient error included, so flaky infrastructure doesn't silently mask itself.
- Tighten the http→https fallback to fire only on connection-level errors, not on HTTP status errors — falling back to https after the server already returned a 4xx made no sense and was silently double-querying every non-2xx response.
- Hand-rolled backoff schedule (jittered: ~200ms / ~1s / ~4s / ~15s) instead of the previous `ExponentialBackoff::from_millis(10)` (10/100/1000ms — too short for real server outages). Worst-case added latency at default settings stays around ~20s.

`MISE_HTTP_RETRIES=0` still works as an opt-out.

## Test plan

- [x] New unit tests in `src/http.rs` using a tiny in-process TCP server cover: retry-rescues-after-502s, no-retry-on-404, retry-exhaustion-on-persistent-500, and `MISE_HTTP_RETRIES=0` regression check.
- [x] All 782 existing unit tests still pass (`mise run test:unit`).
- [x] `vfox` crate tests still pass (73 passed).
- [x] `mise run lint` clean.
- [ ] Manual smoke test: `MISE_DEBUG=1 mise install aqua:mvdan/sh@latest` (the originally failing tool).

---

> [!NOTE]
> **Medium Risk**
> Changes core HTTP request/download behavior by enabling retries by default and altering retry/backoff and http→https fallback logic, which can affect install/update flows and error surfaces under flaky networks.
> 
> **Overview**
> Enables **transient HTTP failure retries by default** (`http_retries` now defaults to 3) and documents the new behavior in `settings.toml` and `schema/mise.json`.
> 
> Replaces `tokio-retry` with a custom `retry_async`/backoff implementation in `src/http.rs`, including transient error classification (5xx/408/429 + network/connect/timeout/body drops), warn logging on each rescued attempt, and a clearer jittered schedule (~200ms/~1s/~4s/~15s). It also restructures downloads to retry the *entire* streamed body and tightens the `http→https` fallback to only trigger on connection-level failures.
> 
> Mirrors the same retry semantics in the `vfox` crate (including Lua HTTP download paths) and adds focused unit tests validating retry-on-5xx, no-retry-on-404, exhaustion behavior, and retries-disabled handling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ccb2d067b9f836a94beb6b26238883ff5f9c165a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5761,7 +5761,6 @@ dependencies = [
  "test-log",
  "thiserror 2.0.18",
  "tokio",
- "tokio-retry",
  "toml 1.1.2+spec-1.1.0",
  "toml_edit",
  "ubi",
@@ -9604,17 +9603,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bbae76ab933c85776efabc971569dd6119c580d8f5d448769dec1764bf796ef2"
 dependencies = [
  "native-tls",
- "tokio",
-]
-
-[[package]]
-name = "tokio-retry"
-version = "0.3.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "40f644c762e9d396831ae2f8935c954b0d758c4532e924bead0f666d0c1c8640"
-dependencies = [
- "pin-project-lite",
- "rand 0.10.1",
  "tokio",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -183,7 +183,6 @@ tera = "1"
 terminal_size = "0.4"
 thiserror = "2"
 tokio = { version = "1", features = ["full"] }
-tokio-retry = "0.3"
 toml = { version = "1.0", features = ["parse", "preserve_order"] }
 toml_edit = { version = "0.25", features = ["parse"] }
 ubi = { version = "0.9", default-features = false }
--- a/crates/vfox/src/http.rs
+++ b/crates/vfox/src/http.rs
@@ -1,5 +1,6 @@
-use reqwest::{Client, ClientBuilder};
+use reqwest::{Client, ClientBuilder, StatusCode};
 use std::sync::LazyLock;
+use std::time::Duration;
 
 pub static CLIENT: LazyLock<Client> = LazyLock::new(|| {
     ClientBuilder::new()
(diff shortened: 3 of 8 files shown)


jdx:
Thanks for the review. Addressed both issues:

**Issue 2 (vfox ignores `http_retries`)** — fixed in 043fc68. vfox now reads `MISE_HTTP_RETRIES` directly (since it can't depend on mise's `Settings` layer) and the backoff schedule matches the main crate's `~200ms / ~1s / ~4s / ~15s`.

**Issue 1 (5xx retry dead for non-download paths)** — I think this is a misread. `send_once` calls `resp.error_for_status_ref()?` at [src/http.rs:435](https://github.com/jdx/mise/blob/c17df5f/src/http.rs#L435), so a 502 surfaces as `Err(reqwest::Error)` *inside* the closure, not as `Ok(Response_5xx)`. The retry loop then sees `Err`, classifies it as transient via `status() == Some(502)`, and retries.

The two unit tests (`test_retry_succeeds_after_two_502s`, `test_retry_exhausted_on_persistent_500`) call `client.get_async()` — exactly the non-download path the analysis claims is broken — and both pass on macOS CI ([unit job](https://github.com/jdx/mise/actions/runs/24967007308/job/73103429292)). They assert `count == 3` and `count == 2` respectively, which would fail if no retry was happening.

That said, the redundant `error_for_status_ref()?` calls in `get_async_with_headers` and `head_async_with_headers` are a smell — they're no-ops because `send_once` already did the check. I'll leave the cleanup for a separate PR to keep this one focused.

PR #9426, opened 2026-04-27
feat(task): add task.surface_tools setting to surface task tools in project toolset

nkakouros:
## Summary

- Adds a new `task.surface_tools` boolean setting (default `false`). When enabled, tools declared in the `tools` table of task TOML files (via `task_config.includes`) are surfaced into the project toolset — made available in `PATH` and installed by `mise install`.
- Tools in `[tools]` of `mise.toml` always take precedence over task-declared tools (task tools are the lowest-priority layer).
- Reads task TOML files directly via `toml::from_str::<Tasks>` to avoid the circular dependency `task.render() → tera_ctx() → get_toolset()`.

## Implementation note

The logic for collecting task tools is duplicated across two builders:
- `ToolRequestSetBuilder::load_task_tools` in `src/toolset/tool_request_set.rs` — used by `Config::get_tool_request_set()` (feeds `mise ls`, tera context, etc.)
- `ToolsetBuilder::load_task_tools` in `src/toolset/builder.rs` — used directly by commands like `mise env`, `mise exec`, `mise run`, etc.

Both builders are needed to cover all call paths. The duplication is intentional and the implementations are kept in sync.

## Changed files

| File | Change |
|---|---|
| `settings.toml` | New `task.surface_tools` bool setting, default `false` |
| `schema/mise.json` | Regenerated to include new setting |
| `src/config/mod.rs` | `expand_task_include` made `pub(crate)` |
| `src/toolset/tool_request_set.rs` | `load_task_tools` on `ToolRequestSetBuilder` |
| `src/toolset/builder.rs` | `load_task_tools` on `ToolsetBuilder` |
| `e2e/tasks/test_task_surface_tools` | E2E test: 3 scenarios (basic, conflict resolution, setting off) |

## Test plan

- [ ] `mise run test:e2e test_task_surface_tools` — all 3 scenarios pass
- [ ] `mise run lint` — clean
- [ ] `mise run build` — clean
- [ ] No existing e2e tests regress

## Reviewer checklist

- [ ] `expand_task_include` visibility change is minimal and correct.
- [ ] `load_task_tools` is called before `load_config_files` in `build()`.
- [ ] `merge(task_trs, trs)` / `ts.merge(file_ts)` order is correct — task tools lose to everything that follows.
- [ ] No call to `config.tasks()`, `task.render()`, or `config.get_toolset()` inside `load_task_tools` (circular-dep guard).
- [ ] `git::` includes are skipped.
- [ ] Directories and non-`.toml` includes are silently skipped (no panic, no hard error).
- [ ] Parse/read errors warn but do not abort the build.
- [ ] Default-`false` setting — existing users see zero behaviour change.
- [ ] e2e scenario 2 (conflict resolution) is present and passes.

### DIFF
--- a/e2e/tasks/test_task_surface_tools
+++ b/e2e/tasks/test_task_surface_tools
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+# Test that task.surface_tools = true makes task tools available in the project toolset
+
+# --- Scenario 1: tool only declared in task, surface_tools = true ---
+cat >mise.toml <<'EOF'
+[settings]
+task.surface_tools = true
+[task_config]
+includes = ["tasks.toml"]
+EOF
+cat >tasks.toml <<'EOF'
+[greet]
+run = "dummy --version"
+tools.dummy = "2.0.0"
+EOF
+
+mise install
+assert_contains "mise ls" "dummy"
+assert_contains "mise ls" "2.0.0"
+assert_contains "mise env" "dummy/2.0.0"
+
+# --- Scenario 2: conflict — [tools] in mise.toml wins ---
+cat >mise.toml <<'EOF'
+[settings]
+task.surface_tools = true
+[tools]
+dummy = "1.0.0"
+[task_config]
+includes = ["tasks.toml"]
+EOF
+
+mise install
+assert_contains "mise ls" "dummy"
+assert_contains "mise env" "dummy/1.0.0"
+assert_not_contains "mise env" "dummy/2.0.0"
+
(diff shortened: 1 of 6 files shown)


nkakouros:
Sorry, this was not ready to be pushed here. But seeing that you closed it, I assume you disagree with the approach altogether? Do you think there is an acceptable approach to allow tools to be defined in a modular way, instead of in the main mise toml? For example, have Python-related tools in a Python-specific toml file, etc? Currently, only task includes are supported but these do not support a `[tool]` section.

PR #9429, opened 2026-04-27
feat(registry): add gest and doing

aaronmallen:
Adds two new tools to the registry

* [gest](https://gest.aaronmallen.dev)
* [doing](https://doing.aaronmallen.dev)


jdx:
not notable enough

PR #9432, opened 2026-04-27
refactor(COPR): centralize chroots config

bestagi:
Remove manual chroots input from workflow and move default chroots list to environment variable. This eliminates code duplication and simplifies the build configuration by using a single source of truth for target architectures across both the workflow and build script.

### DIFF
--- a/.github/workflows/copr-publish.yml
+++ b/.github/workflows/copr-publish.yml
@@ -5,11 +5,6 @@ on:
     types: [published]
   workflow_dispatch:
     inputs:
-      chroots:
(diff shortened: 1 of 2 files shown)


.github/workflows/copr-publish.yml
```diff
@@ -5,11 +5,6 @@ on:
     types: [published]
   workflow_dispatch:
     inputs:
-      chroots:
```
jdx, line 8: this is important

bestagi: fixed

PR #9436, opened 2026-04-27
fix(ls-remote): avoid degraded GitHub release metadata

jdx:
## Summary

Fixes the path where transient empty GitHub `/releases` responses could silently turn into degraded version metadata and then be cached or persisted downstream.

- retry empty GitHub release lists, then fail instead of caching `[]`
- make `aqua:` propagate that GitHub empty-release error instead of falling back to tags and emitting `prerelease = false` for every version
- keep UBI's normal releases-to-tags fallback, but make `--strict-metadata` fail before using that lower-metadata fallback
- add `mise ls-remote --json --strict-metadata` for metadata-producing automation that should not silently fall back to less-authoritative data
- update generated CLI docs, manpage, and Fig completion metadata

## Validation

- `cargo test -p mise github::tests::required_releases_reject_empty_results`
- `cargo check -p mise`
- `cargo build -p mise`
- `mise run render:usage`
- `mise run render:mangen`
- `mise run render:fig`
- `target/debug/mise ls-remote --strict-metadata node` fails because `--json` is required
- `MISE_CACHE_DIR=$(mktemp -d) MISE_USE_VERSIONS_HOST=0 target/debug/mise ls-remote github:jdx/mise --json` now errors on the live empty `/releases` response instead of printing `[]`
- `MISE_CACHE_DIR=$(mktemp -d) MISE_USE_VERSIONS_HOST=0 MISE_PRERELEASES=1 MISE_LIST_ALL_VERSIONS=1 target/debug/mise ls-remote aqua:nektos/act --json` returned a non-zero prerelease count when GitHub returned releases, and errors loudly when GitHub returned empty `/releases`
- `git diff --check`
- commit hook: `hk` pass, including `cargo check --all-features`

## Notes

`--strict-metadata` intentionally does not reject empty final result sets, placeholder-looking version strings, or versions without `created_at`. It only prevents known metadata-degrading fallback behavior.

The targeted e2e command `mise run test:e2e cli/test_ls_remote backend/test_github_alias_versions` got through `cli/test_ls_remote`, then failed in `backend/test_github_alias_versions` because the local unauthenticated GitHub API quota was exhausted (`403 Forbidden` for `https://api.github.com/repos/jdx/mise-test-fixtures/releases`).

### DIFF
--- a/docs/cli/ls-remote.md
+++ b/docs/cli/ls-remote.md
@@ -29,6 +29,14 @@ Show all installed plugins and versions
 
 Output in JSON format (includes version metadata like created_at timestamps when available)
 
+### `--strict-metadata`
+
+Fail instead of falling back when version metadata sources are unavailable
+
+This is intended for automation that writes canonical version metadata.
+It requires --json and prevents metadata source failures from silently
+degrading to less-authoritative fallback data.
+
 ### `--prerelease`
 
 Include pre-release versions in the output for backends that report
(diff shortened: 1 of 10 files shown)


jdx:
Closing this for now. We restored the mise-versions prerelease sync path instead and are not changing mise's normal GitHub release/tag fallback behavior in this PR.\n\n*This comment was generated by an AI coding assistant.*

PR #9442, opened 2026-04-27
fix(lockfile): update global lockfile on upgrade

jdx:
## Summary

Fixes automatic lockfile updates for global tool configs when upgrading a fuzzy version such as `latest`.

## Root Cause

`update_lockfiles` skipped global config files when grouping lockfile targets, so an existing `~/.config/mise/mise.lock` was not maintained after `mise upgrade`. In addition, fuzzy requests could still resolve through the old lockfile entry while the update was being prepared, preserving the stale version.

## Changes

- Include existing global lockfiles in the automatic lockfile update path.
- Overlay newly installed upgrade versions before writing lockfiles so stale locked resolutions are replaced.
- Add an e2e regression for `dummy = "latest"` in the global config upgrading from `1.0.0` to `2.0.0`.

## Validation

- `mise run test:e2e e2e/cli/test_upgrade`
- pre-commit hook: `cargo fmt --all -- --check`, `cargo check --all-features`, shell/lint/schema checks

---

> [!NOTE]
> **Medium Risk**
> Touches lockfile update logic used during `mise upgrade`, including global config handling and version overlay behavior; mistakes could lead to stale or incorrect lockfile pins across configs.
> 
> **Overview**
> Fixes automatic lockfile refresh when upgrading tools declared with fuzzy versions (e.g. `latest`), including **global** configs.
> 
> `update_lockfiles` now includes global `mise.toml` sources when determining which existing lockfiles to update, and overlays newly-installed `ToolVersion`s into the per-lockfile tool set so stale lockfile-resolved versions don’t persist during the update.
> 
> Adds an e2e regression that upgrades a globally-configured `dummy = "latest"` and asserts `~/.config/mise/mise.lock` is rewritten from `1.0.0` to `2.0.0` while the config stays `"latest"`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 71b1e21814f225a0220fed99d9dcf230d0a23269. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/e2e/cli/test_upgrade
+++ b/e2e/cli/test_upgrade
@@ -203,6 +203,7 @@ touch "$MISE_CONFIG_DIR/mise.lock"
 mise lock --global --platform linux-x64
 assert_contains "cat $MISE_CONFIG_DIR/mise.lock" "1.0.0"
 mise upgrade dummy@2.0.0
+assert_contains "cat $MISE_CONFIG_DIR/config.toml" '"latest"'
 assert_contains "cat $MISE_CONFIG_DIR/mise.lock" "2.0.0"
 assert_not_contains "cat $MISE_CONFIG_DIR/mise.lock" "1.0.0"
 rm -f "$MISE_CONFIG_DIR/config.toml" "$MISE_CONFIG_DIR/mise.lock"
--- a/e2e/cli/test_upgrade
+++ b/e2e/cli/test_upgrade
@@ -191,4 +191,19 @@ assert_contains "cat mise.toml" '"3"'
 assert_contains "cat mise.lock" "3.0.1"
 assert_not_contains "cat mise.lock" "2.0.0"
 rm -f mise.toml mise.lock
+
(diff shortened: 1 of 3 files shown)


jdx:
@greptileai this should only modify existing mise.lock files not create new ones

jdx:
Confirmed — both `update_lockfiles` and `auto_lock_new_versions` keep their `if !lockfile_path.exists() { continue; }` early-return, so the path that runs after install/upgrade only ever modifies an existing lockfile. The PR removes `is_global_config` guards (which gated *update*, not *creation*), so removing them doesn't change creation behavior. Verified locally that `mise install` and `mise upgrade` against a global-only config with no `~/.config/mise/mise.lock` leave the file absent. The new e2e test creates the file explicitly with `touch` + `mise lock --global` before exercising the upgrade path, matching the design intent.

PR #9451, opened 2026-04-28
fix(copr): remove stale pinned image digest and rebuild copr image on Dockerfile changes

bestagi:
## Problem

After #9421 was merged (which fixed `copr-cli` installation by switching from `pip3 install copr-cli` to `dnf install -y copr-cli`), the `copr-publish` workflow **continued to fail** with the exact same error:

```
ModuleNotFoundError: No module named 'rich'
```

Seen in: https://github.com/jdx/mise/actions/runs/25053108886/job/73385917506

### Root cause

The `copr-publish.yml` workflow uses a **hardcoded image digest**:

```yaml
container:
  image: ghcr.io/jdx/mise:copr@sha256:da63a0fcdf21a36be340dce61394b336113ea460b93aa9de2705bd7b35e5872e
```

This digest points to the **old image** built before the `packaging/copr/Dockerfile` was fixed. The `docker.yml` workflow only rebuilds images on tag pushes or `workflow_dispatch` — it does **not** trigger on commits to `main`. So even though the Dockerfile was fixed in #9421, the pinned digest in the workflow still referenced the broken pip-installed image.

### 1. `copr-publish.yml` — remove stale pinned digest

Replace:
```yaml
image: ghcr.io/jdx/mise:copr@sha256:da63a0fcdf21a36be340dce61394b336113ea460b93aa9de2705bd7b35e5872e
```
With:
```yaml
image: ghcr.io/jdx/mise:copr
```

This ensures the workflow always uses the latest built `:copr` image instead of a stale pinned version.

### 2. `docker.yml` — auto-rebuild copr image on Dockerfile changes

Add a `push` trigger scoped to `packaging/copr/Dockerfile` on `main` so the `:copr` image is automatically rebuilt and the tag updated whenever the Dockerfile changes. This prevents this class of stale-digest regression in the future.

```yaml
on:
  push:
    tags: ["v[0-9]*"]
    paths:
      - "packaging/copr/Dockerfile"
    branches:
      - main
```

## Testing

Once merged, re-run the `copr-publish` workflow — it will pull the latest `:copr` image (built from the fixed Dockerfile with `dnf install -y copr-cli`) and the `rich` module error will be gone.


jdx:
@greptileai I think this needs to be added to the renovate exclusions

jdx:
Thanks for digging into this! I think the scope can be trimmed — the two essential changes are dropping the `@sha256:` pin in `copr-publish.yml` and the renovate rule. Those alone fix the bug.

The new `docker-copr.yml` isn't needed: `copr-publish.yml` runs on `release: published`, and releases push a `v*` tag which already triggers `docker.yml` to rebuild `:copr`. Once the pin is gone, the next release publishes a fresh image and copr-publish pulls it. For ad-hoc rebuilds between releases, `workflow_dispatch` on `docker.yml` already works.

Could you drop the `docker-copr.yml` addition and keep just the two-line `copr-publish.yml` change plus the renovate rule? That's a tight, easy-to-review fix.

bestagi:
Thanks for the clear feedback, @jdx! You're absolutely right — the `docker-copr.yml` was redundant since `v*` tag pushes already trigger `docker.yml` to rebuild `:copr`, and `workflow_dispatch` covers ad-hoc cases.

I've just dropped `docker-copr.yml` in the latest commit ([797b5ab](https://github.com/bestagi/mise/commit/797b5ab7238848088c20f2673d0c441fc664fb70)). The PR now contains only the two essential changes:
- **`copr-publish.yml`**: floating `ghcr.io/jdx/mise:copr` tag (no stale digest)
- **`.github/renovate.json`**: `packageRule` to prevent Renovate from re-pinning the `:copr` tag

Should be a clean, easy-to-review diff now. Let me know if anything else needs adjusting!

PR #9468, opened 2026-04-29
fix(shims): compare PATH entries case-insensitively on macOS

jdx:
## Summary

- Fix infinite shim recursion on macOS when `$HOME` appears in PATH with different casing than the resolved `dirs::SHIMS` path.
- Introduce `file::paths_eq`, a platform-aware path comparator (case-insensitive on macOS/Windows, byte-equal on Linux, with separator normalisation on Windows), and route every "is this PATH entry the shims directory?" check through it.
- Reported in [discussion #9462](https://github.com/jdx/mise/discussions/9462).

## Background

A user on macOS hit a recursive `mise -> npm shim -> mise -> npm shim -> ...` loop while `mise up --bump` resolved npm package versions, sometimes crashing their session. Their PATH contained `/Users/Olfway/.local/share/mise/shims` but `dirs::SHIMS` resolved to `/Users/olfway/.local/share/mise/shims` via a lowercase `$HOME`. macOS volumes (HFS+/APFS) are case-insensitive by default, so both paths refer to the same directory — but mise compared PATH entries against the shims directory byte-equal.

The recursion site is `Backend::dependency_env`. When the npm backend ran `npm view <pkg> dist-tags --json` to resolve a "latest" version, it built a child env with the shims directory stripped from PATH so `npm` would resolve to the real binary. With the case mismatch the strip was a no-op, so `npm` re-resolved to the mise shim, which exec'd `mise … -- npm view …`, which spawned another `npm`, ad infinitum.

The same case-sensitivity bug existed at every other "compare against shims dir" call site, including the doctor's `shims_on_path` check (which incorrectly reported `no` for the affected user, masking the real problem).

## Changes

- New `file::paths_eq(a, b) -> bool` with `#[cfg]`-gated arms:
  - **Windows:** lowercase + `/` -> `\` normalisation (replaces a few inline copies of this logic).
  - **macOS:** lowercase compare.
  - **Linux:** byte-equal compare (case-sensitive filesystems are the norm).
- Routed through `paths_eq`:
  - `Backend::dependency_env` (the actual recursion-causing site)
  - `file::path_env_without_shims`, `file::strip_shims_from_path`, `file::which_no_shims`
  - `PathEnv::from_iter` shim partitioning
  - `doctor::shims_on_path`
  - Both Unix and Windows branches of `cli::exec::exec_program` program resolution
- Unit tests cover exact match, case-insensitivity (macOS/Windows), case-sensitivity (Linux), and `\`/`/` normalisation (Windows).

The Windows-specific normalisation block previously inlined in `Backend::dependency_env` and `cli::exec` is now collapsed into a single cross-platform call.

## Test plan

- [x] `cargo build --bin mise` clean
- [x] `cargo test --bin mise file::` — 94 tests pass, including new `paths_eq` tests
- [x] `mise run lint-fix` clean
- [ ] Manually verify on a macOS box with mixed-case `$HOME` that `mise up --bump` against npm packages no longer recurses (covered by the user's reproduction in the linked discussion)
- [ ] CI

---

> [!NOTE]
> **Medium Risk**
> Touches multiple call sites that manipulate PATH and executable resolution, so subtle path-comparison behavior changes could affect tool lookup order across platforms despite the changes being narrowly scoped and well-tested.
> 
> **Overview**
> Prevents infinite shim recursion on macOS (and hardens Windows handling) by introducing `file::paths_eq` and switching all “is this PATH entry the shims dir?” checks to use case-insensitive, component-based path comparison (including `~/` expansion via `replace_path`).
> 
> This updates shim stripping and PATH reordering in `Backend::dependency_env`, `cli exec` program resolution (Unix+Windows), `PathEnv` shim partitioning, and `doctor`’s `shims_on_path`, and adds focused unit tests covering exact match, macOS/Windows case-insensitivity, separator/trailing-slash normalization, and Linux case-sensitivity.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a341252cb778333ee7fab0d8c793e662a212ea63. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -1708,32 +1708,18 @@ pub trait Backend: Debug + Send + Sync {
         // dependency tool (e.g., go) is configured but not installed. Without this,
         // the shim for the dependency would call `mise exec` which would call the
         // shim again infinitely.
+        //
+        // `paths_eq` handles case-insensitive matching on macOS/Windows: e.g. if
+        // `$HOME` is mixed-case in PATH (`/Users/Foo`) but lowercase in the
+        // resolved shims path, byte-equal comparison would miss it and the shim
+        // would survive in the child env.
         if let Some(path_val) = env.get(&*env::PATH_KEY) {
             let paths: Vec<_> = env::split_paths(path_val).collect();
             let original_len = paths.len();
-            #[cfg(not(windows))]
             let filtered: Vec<_> = paths
                 .into_iter()
-                .filter(|p| p.as_path() != *dirs::SHIMS)
+                .filter(|p| !file::paths_eq(&file::replace_path(p), &dirs::SHIMS))
                 .collect();
-            #[cfg(windows)]
-            let filtered: Vec<_> = {
-                // Pre-compute once; case-insensitive + separator-normalised to handle
-                // path variations such as ~/.local/share/mise\shims vs
-                // C:\Users\user\.local\share\mise\shims
-                let shims_normalized = dirs::SHIMS
-                    .to_string_lossy()
-                    .to_lowercase()
-                    .replace('/', "\\");
-                paths
-                    .into_iter()
-                    .filter(|p| {
-                        let expanded = file::replace_path(p);
-                        expanded.to_string_lossy().to_lowercase().replace('/', "\\")
-                            != shims_normalized
-                    })
-                    .collect()
-            };
             if filtered.len() != original_len {
                 let joined = env::join_paths(&filtered)?;
                 env.insert(
--- a/src/cli/doctor/mod.rs
+++ b/src/cli/doctor/mod.rs
@@ -571,7 +571,10 @@ impl Doctor {
 }
 
 fn shims_on_path() -> bool {
-    env::PATH.contains(&dirs::SHIMS.to_path_buf())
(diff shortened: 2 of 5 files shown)


jdx:
Addressed in a341252cb:

- **Greptile P2 (path_env.rs)**: added `replace_path` before `paths_eq` for consistency with every other call site.
- **Gemini high (paths_eq)**: switched to component-based comparison. This folds trailing slashes, redundant separators, and (on Windows) `/` vs `\` since `Path::components` already treats both as separators — so the explicit `replace('/', "\\")` is no longer needed. Added a `test_paths_eq_trailing_separator` test.
- **Gemini medium (perf in `dependency_env`)**: deliberately not addressed. PATH typically has tens of entries; `paths_eq` is a handful of small string allocations per call, and `dependency_env` runs once per backend version check, not in a tight inner loop. Pre-normalising the target would require an awkward two-arg helper that propagates platform-specific normalised state through callers, which isn't worth it for a sub-microsecond win on a non-hot path. Happy to revisit if it ever shows up in a profile.

PR #9496, opened 2026-04-30
feat(task): support exclusion patterns in task sources

jlarmstrongiv:
Hi 👋 I don’t normally use Claude to code, but I saw some co-authored commits in main, so I thought I would post it anyway to continue the discussion https://github.com/jdx/mise/discussions/7809.  I’m unfamiliar with rust, so I may not be able to push the PR across the finish line.

---

Tasks now accept exclusion patterns so files like tests no longer invalidate caches when only the build is meant to rebuild. Two complementary forms compose as a union:

- entries in `sources` prefixed with `!` are excluded
- a new `sources_exclude` field accepts dedicated exclude patterns

Exclusions apply uniformly to source freshness checks, the `task_source_files` template function, and `mise watch` (via watchexec `--ignore`). The task's own config file is always treated as a source even if matched by an exclude, so a stray `!mise.toml` cannot silently disable invalidation. A literal `!` prefix in a path can be escaped as `\!`.

### DIFF
--- a/docs/tasks/task-configuration.md
+++ b/docs/tasks/task-configuration.md
@@ -377,6 +377,27 @@ has changed since the last build.
 The [`task_source_files`](../templates.md#task-source-files) function can be used to iterate over a task's
 `sources` within its template context.
 
+#### Excluding sources
+
+Entries in `sources` prefixed with `!` are excluded, matching the convention
+used by gitignore, watchexec, and rsync. Exclusions affect the freshness
+check, the `task_source_files` template function, and which files
+`mise watch` watches for changes.
+
+```mise-toml
+[tasks.build]
+sources = ["src/**/*.ts", "!src/**/*.test.ts", "!src/**/*.spec.ts", "tsconfig.json"]
+run = "npm run build"
+```
+
+Entries are evaluated in order, and the latest matching entry wins. A later
+non-negated entry can re-include a file an earlier `!` excluded — for example,
+`["src/**/*.ts", "!src/**/*.test.ts", "src/keep.test.ts"]` excludes all
+`*.test.ts` files except `src/keep.test.ts`.
+
+To include a literal path that begins with `!`, escape the prefix as `\!`
+(e.g. `"\\!important.txt"` in TOML).
+
 #### Dependency invalidation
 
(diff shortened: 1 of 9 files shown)


jdx:
Thanks for picking this up! One ask before we land it: please drop `sources_exclude` and keep only the `!`-prefix form in `sources`.

Two ways to express the same thing doubles the surface area (docs, schema, validation, merge semantics in task templates, JSON output) without adding expressiveness — `!`-prefix alone matches the convention from gitignore / watchexec / rsync that users already know, and composes fine on its own.

Everything else (the `!mise.toml` carve-out, watchexec wiring, validation of stripped patterns) looks like the right shape — just want the API surface to be one thing.

jlarmstrongiv:
Thank you @jdx !  The main reasons I included both `!` and `sources_exclude` are:
- `!` is a valid file path character
- It’s very common to provide both options, e.g. [tinyglobby](https://github.com/SuperchupuDev/tinyglobby#usage), [fast-glob](https://github.com/mrmlnc/fast-glob#ignore), [globby](https://github.com/sindresorhus/globby)

Please confirm if you would like to keep or remove `sources_exclude`.  Thanks again!

jdx:
remove it

jlarmstrongiv:
Thank you @jdx !  I updated the PR and dropped `sources_exclude`.  All tests are now passing.

jdx:
We already depend on the [`ignore`](https://docs.rs/ignore) crate (currently used in `src/config/mod.rs` for directory walking), and it ships [`ignore::gitignore::Gitignore`](https://docs.rs/ignore/latest/ignore/gitignore/struct.Gitignore.html) which handles exactly this case natively — `!` negation, `\!` literal-escape, and order-dependent re-inclusion (a later `!pat` re-includes files a prior pattern excluded).

Using it here would let us drop `split_sources` and `build_exclude_set`, and we'd get gitignore re-inclusion semantics for free — which the union-of-excludes approach in this PR can't express.

The `!mise.toml`-must-not-disable-invalidation case still works: just always treat the task's config path as a source after the gitignore match, same as today.

jlarmstrongiv:
@jdx good news!  The PR has been refactored to use the [ignore](https://docs.rs/ignore) crate and still works with absolute paths

PR #9506, opened 2026-05-01
fix(github): retry ip allow list errors without auth

risu729:
## Summary
- keep GitHub API requests authenticated by default so installation-token/user-token rate limits still apply
- when an authenticated GitHub REST API request returns a 403 body containing "IP allow list", retry the same request without Authorization
- leave vfox GitHub header behavior unchanged

Fixes https://github.com/jdx/mise/discussions/9119

## Notes
mise-versions appears to use a user-token pool today, but its JSON enrichment path can still burn the workflow-level MISE_GITHUB_TOKEN because mise resolves MISE_GITHUB_TOKEN before the per-tool GITHUB_API_TOKEN override. This PR avoids making that worse by only falling back to anonymous requests after the specific IP allow-list 403.

## Tests
- cargo fmt --check
- cargo test github::tests
- cargo test http::tests
- cargo test -p vfox lua_mod::http::tests::test_add_default_headers

### DIFF
--- a/src/http.rs
+++ b/src/http.rs
@@ -8,7 +8,8 @@ use base64::Engine;
 use base64::prelude::BASE64_STANDARD;
 use eyre::{Report, Result, bail, ensure};
 use regex::Regex;
-use reqwest::header::{CONTENT_TYPE, HeaderMap, HeaderValue};
+use reqwest::StatusCode;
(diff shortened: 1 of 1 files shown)


jasonwbarnett:
Let's go! 🔥 

Thanks @risu729 ❤️

PR #9523, opened 2026-05-01
fix(backend): mark -nightly/-canary/-experimental as prereleases

jdx:
## Summary

When `minimum_release_age` is set, latest version resolution skips the backend's `latest_stable_version` fast path (e.g. npm dist-tags) and falls back to the shared version-list path, which relies on `VERSION_REGEX` to flag pre-releases. The regex did not catch `-nightly`, `-canary`, `-experimental`, `-insider`, or `-edge` channel tags, so npm packages like `@google/gemini-cli` resolved `latest` to a `-nightly` tarball instead of the stable maintainer-tagged release.

This adds those channel tags to `VERSION_REGEX` so they are filtered out of `latest` resolution and `ls-remote` (without `--prerelease`) on every backend that opts into pattern-based prerelease detection.

Fixes #9503

## Test plan

- [x] `cargo test test_version_regex_filters_prerelease`
- [x] `cargo test` (full unit suite, 796 passing)
- [ ] Manually verified `mise install npm:@google/gemini-cli@latest` with `MISE_MINIMUM_RELEASE_AGE=0d` no longer picks the `-nightly` tarball

---

> [!NOTE]
> **Medium Risk**
> Changes prerelease detection used in version listing/resolution, which may alter which versions are considered installable as “stable” (especially for npm packages) and could inadvertently filter edge-case version strings.
> 
> **Overview**
> Improves prerelease filtering during version resolution so npm channel builds like `-nightly`, `-canary`, `-experimental`, `-insider`, and `-edge` are no longer treated as stable when `VERSION_REGEX`-based filtering is used.
> 
> The npm backend now explicitly marks versions as prereleases based on strict semver rule 9 (hyphen suffix, ignoring `+build` metadata), and adds focused unit tests covering both the new regex channels and semver prerelease parsing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 92467940a64cc0a926f711a7a244cccdee4aa29c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/backend/npm.rs
+++ b/src/backend/npm.rs
@@ -158,6 +158,7 @@ impl Backend for NPMBackend {
                         VersionInfo {
                             version: version.to_string(),
                             created_at,
+                            prerelease: is_semver_prerelease(version),
                             ..Default::default()
                         }
                     })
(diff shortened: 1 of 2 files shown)


jdx:
Addressed review feedback in e8e9b000d:

- **cursor-bot (correct catch):** the `18.0.0-experimental-abc1234` fixture was already matched by the pre-existing `([abc])[0-9]+` alternative via `c1234`, so the assertion would pass even if `-experimental` was removed from the regex. Switched to `18.0.0-experimental.1` which exercises only the new alternative. Did the same audit for the other tags (`13.0.0-canary` → no incidental match; `1.99.0-insider` / `1.99.0-edge` → no incidental match).
- **greptile / gemini:** added `-insider` and `-edge` test cases.
- **gemini "use `[-\.]nightly` etc. for consistency":** skipped — these are npm-style channel tags that universally use `-`. Broadening to `.` would risk false positives for backends using literals like `.canary` in stable version strings, and there's no current evidence any tool actually publishes the dot form.

PR #9540, opened 2026-05-02
fix(plugin): detect plugin types after cloning

risu729:
## Summary

- remove URL-name plugin type heuristics for unprefixed plugin config and installs
- detect asdf/vfox/vfox-backend plugins from cloned plugin files (`metadata.lua`, `hooks/backend_install.lua`, `bin/list-all`)
- update install state and clear backend cache after plugin clone so commands can continue with the detected backend
- refresh plugin-backed `latest` and `ls-remote` backends after auto-install
- add e2e coverage with mise-plugins asdf/vfox plugins and the allowed `jbadeau/mise-nix` backend plugin

Fixes discussion #9333.

## Validation

- `/home/risu/.cargo/bin/cargo fmt --check`
- `/home/risu/.cargo/bin/cargo test --bin mise plugins::tests::test_plugin_type -- --nocapture`
- `/home/risu/.cargo/bin/cargo test --bin mise backend_arg -- --nocapture`
- `MISE_TRUSTED_CONFIG_PATHS=/tmp/mise-plugin-type-detection/mise.toml ./target/debug/mise run test:e2e e2e/plugins/test_plugin_type_lazy e2e/plugins/test_backend_plugin_install`
- `git diff --check`

### DIFF
--- a/src/plugins/mod.rs
+++ b/src/plugins/mod.rs
@@ -189,7 +189,7 @@ impl PluginType {
         }
     }
 
-    pub fn from_plugin_config<'a>(key: &'a str, url: &str) -> (Self, &'a str) {
+    pub fn from_plugin_config(key: &str) -> (Self, &str) {
         if let Some(name) = key.strip_prefix("vfox:") {
             (Self::Vfox, name)
         } else if let Some(name) = key.strip_prefix("vfox-backend:") {
@@ -197,26 +197,23 @@ impl PluginType {
         } else if let Some(name) = key.strip_prefix("asdf:") {
             (Self::Asdf, name)
         } else {
-            (Self::from_plugin_url(url), key)
+            let path = dirs::PLUGINS.join(key.to_kebab_case());
+            (Self::from_plugin_path(&path).unwrap_or(Self::Asdf), key)
         }
     }
 
-    pub fn from_plugin_url(url: &str) -> Self {
-        if url.starts_with("vfox-backend:") {
-            Self::VfoxBackend
-        } else if url.starts_with("vfox:") {
-            Self::Vfox
-        } else if url.starts_with("asdf:") {
-            Self::Asdf
-        } else if Self::looks_like_vfox_url(url) {
-            Self::Vfox
+    pub fn from_plugin_path(path: &Path) -> Option<Self> {
+        if path.join("metadata.lua").exists() {
+            if path.join("hooks").join("backend_install.lua").exists() {
+                Some(Self::VfoxBackend)
+            } else {
+                Some(Self::Vfox)
+            }
+        } else if path.join("bin").join("list-all").exists() {
+            Some(Self::Asdf)
         } else {
-            Self::Asdf
+            None
         }
-    }
-
-    fn looks_like_vfox_url(url: &str) -> bool {
-        url.contains("vfox-")
     }
 
     pub fn plugin(&self, short: String) -> PluginEnum {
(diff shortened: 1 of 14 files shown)


src/plugins/mod.rs
```diff
@@ -189,6 +189,36 @@ impl PluginType {
...
+            (Self::from_plugin_url(url), key)
+        }
+    }
+
+    pub fn from_plugin_url(url: &str) -> Self {
+        if url.starts_with("vfox-backend:") {
+            Self::VfoxBackend
+        } else if url.starts_with("vfox:") {
+            Self::Vfox
+        } else if url.starts_with("asdf:") {
```
jdx: would the url start with this?

src/plugins/mod.rs
```diff
@@ -189,6 +189,36 @@ impl PluginType {
...
+            Self::Asdf
+        } else if Self::looks_like_vfox_url(url) {
+            Self::Vfox
+        } else {
+            Self::Asdf
+        }
+    }
+
+    fn looks_like_vfox_url(url: &str) -> bool {
+        url.contains("vfox-")
```
jdx: this really is not a good heuristic. The heuristic should be whether it contains metadata.lua. I think the direction I wanted to go was for whether or not the plugin was asdf or not to simply not matter until it's cloned, then after it's cloned mise would know which type it is.

You didn't add this heuristic so this might be fine, but it might be worth exploring why this happened and make sure that things labeled as asdf still work fine even if they are vfox plugins. Not all vfox plugins are going to start with vfox-.

risu729: I see. I'll try to refactor this PR in that way.

risu729:
Self-review after `98163cd1e`:

- Suspicious: unprefixed plugins are still registered as `asdf` before they are cloned. Evaluation: this is intentional as the temporary clone path; after `ensure_installed` completes, the plugin directory is inspected and install state is updated from `metadata.lua` / `hooks/backend_install.lua` / `bin/list-all`.
- Suspicious: a command may keep using the backend object created before clone. Evaluation: `latest` and `ls-remote` now re-fetch the backend after auto-install, and plugin install invalidates the backend cache for the plugin short name. `mise install` re-resolves each tool after config plugins are installed, so it also picks up the detected type.
- Suspicious: backend plugin cache entries use `plugin:tool` keys, while cache invalidation removes the plugin short name. Evaluation: configured `plugin:tool` entries already resolve as `VfoxBackend` before clone, so the cached backend type is already the desired one; the standard lazy vfox case is keyed by the short tool name and is invalidated.
- Suspicious: explicit `asdf:` config pointing at a vfox repo will become vfox after clone. Evaluation: this follows the requested metadata-based behavior; the prefix chooses the clone path before install, but cloned plugin files are authoritative afterward.
- Suspicious: `PluginEnum::install` itself does not update install state. Evaluation: current production call sites install plugins through `ensure_installed`, where the state update happens. No direct call sites were found.
- Suspicious: the backend plugin e2e now uses `mise latest` instead of installing a Nix package. Evaluation: this avoids requiring Nix on e2e runners while still exercising auto-install and lazy backend-plugin detection with the allowed `jbadeau/mise-nix` plugin.

No follow-up patch from this self-review.

src/plugins/mod.rs
risu729: I think not, fixed.

n-rodriguez:
@risu729 thank you!

PR #9542, opened 2026-05-02
registry: add tool yr (backend:github:VirusTotal/yara-x)

adam-moss:
Add `yr` (yara-x), the pattern matching swiss army knife from VirusTotal.


jdx:
Remove the alias

PR #9600, opened 2026-05-04
registry: change aqua and vfox order in 1password.toml

gisripa:
Change the order of registry to have aqua as first preference.

>Not accepted:
New vfox and asdf tools are not accepted for supply-chain security reasons — use aqua (preferred) or github instead. The ubi backend is deprecated and is not accepted for new registry entries. Users can still install via any backend themselves with explicit syntax (mise use vfox:owner/repo, mise use cargo:name, etc.) — they just don't get a registry shorthand for it.

### DIFF
--- a/registry/1password.toml
+++ b/registry/1password.toml
@@ -1,4 +1,4 @@
 aliases = ["1password-cli", "op"]
-backends = ["vfox:mise-plugins/vfox-1password", "aqua:1password/cli"]
+backends = ["aqua:1password/cli", "vfox:mise-plugins/vfox-1password"]
 description = "Password manager developed by AgileBits Inc"
 test = { cmd = "op --version", expected = "{{version}}" }


risu729:
See https://github.com/jdx/mise/pull/6116. The asdf plugin is replaced by the vfox plugin later.
If the issue is already fully resolved we can use the aqua backend. 
Also, the docs you referenced is for new tools, it's better to use the aqua backend but for some tools we need to keep vfox backends.

gisripa:
@risu729  I keep seeing this error recently after mise upgrade which made me trigger this PR. 

```
mise WARN  Error listing bin paths for vfox:mise-plugins/vfox-1password@2.32.0: registry not available: no aqua-registry found for mise-plugins/vfox-1password
```

It was completely failing and I had to remove registry shorthand and explicitly do this. `"aqua:1password/cli" = "2.32.0"` and atleast `mise install` started working only with this `WARN`

Any guidance why thats happening ?

risu729:
Could you share `mise doctor`, config, and a command to reproduce the error message? I don't have any ideas why.

gisripa:
I ran the `mise doctor` It emitted a warning and I followed it. now the WARN message is gone, so I guess that fixes it ?

```
2. tool '1password' installed with backend 'aqua:1password/cli'
   but registry now recommends 'vfox:mise-plugins/vfox-1password'.
   To migrate: mise uninstall --all 1password && mise install 1password
```

risu729:
Hmm, if the warning is gone, maybe that's fine.

PR #9604, opened 2026-05-05
fix(hooks): align hook run syntax with tasks

risu729:
## Summary
- Extend hook `run` from scalar inline commands to task-style run entries, including script strings, `{ task, args, env }`, and `{ tasks = [...] }`.
- Add `run_windows` as a Windows-specific hook run override.
- Keep string hook values as `run` shorthand, and keep `{ task = "name" }` as shorthand for `run = [{ task = "name" }]`.
- Keep `script + shell` as current-shell syntax for `enter`, `leave`, and `cd`, and allow `script` arrays for those hooks.
- Keep install-hook `script` as the legacy alias for `run`; when paired with `shell`, the shell value is ignored and the script runs through the default inline shell.
- Update docs and JSON schema for `run`, `run_windows`, task-style run entries, and current-shell `script` behavior.

Fixes #9636.
Fixes #8293.

## Hook Run Forms
Scalar `run` hooks continue to work:

```toml
[hooks]
postinstall = { run = "echo installed", shell = "bash -c" }
```

Hook `run` also accepts task-style entries:

```toml
[hooks]
postinstall = { run = [
  "echo first shell",
  "echo second shell",
  { task = "setup", args = ["--fast"], env = { CI = "1" } },
  { tasks = ["lint", "test"] },
] }
```

`run.shell` follows task `shell` semantics: it is an inline shell command such as `bash -c`, `zsh -c`, or `pwsh -Command`. Without `shell`, hook script entries use the default inline shell settings.

`run_windows` can be used by itself for a Windows-only hook or alongside `run` as the Windows-specific override:

```toml
[hooks]
preinstall = { run_windows = "Write-Output installing" }
```

Current-shell hooks continue to use `script + shell`; here `shell` is a shell-name selector, not an inline shell command:

```toml
[[hooks.enter]]
shell = "fish"
script = [
  "set -gx PROJECT_MODE dev",
  "source completions.fish",
]
```

## TaskExecutor Rationale
Hook `run` now accepts real task run entries, so hooks need to execute a synthetic `Task` with its `RunEntry` list instead of flattening those entries into generated `mise run` argv. Keeping the entries as Rust objects preserves task-entry `args`, high-priority `env`, task groups, dependency scheduling, and monorepo-relative task resolution.

The `TaskExecutor` change is the narrow in-process entry point for that existing run-entry executor and scheduler. Hook scripts are rendered before calling it, so preinstall hooks can keep using the preinstall-safe tera context and avoid resolving `tools = true` env directives before tools are installed.

The only remaining subprocess fallback is for task entries on hooks loaded from a previous shell session after the current `Config` no longer contains that root, such as leave hooks after changing out of a project. In that case there is no in-process task graph for the old root to resolve against, so preserving the old behavior still requires a `mise --cd <old-root> run ...` subprocess.

## Tests
- `cargo fmt --all`
- `cargo fmt --all --check`
- `jq empty schema/mise.json`
- `shellcheck -x e2e/config/test_hooks_run`
- `shfmt -d --apply-ignore e2e/config/test_hooks_run`
- `git diff --check`
- `CARGO_BUILD_JOBS=1 cargo check -q`
- `mise run test:e2e e2e/config/test_hooks_run`
- `mise run test:e2e e2e/config/test_hooks_task_ref`
- `mise run test:e2e e2e/config/test_hooks e2e/config/test_hooks_config_root e2e/config/test_hooks_error_handling e2e/config/test_hooks_global e2e/config/test_hooks_installed_tools e2e/config/test_hooks_postinstall_env e2e/config/test_hooks_shell_env e2e/config/test_hooks_tool_env e2e/config/test_hooks_trust_enter`

### DIFF
--- a/src/hooks.rs
+++ b/src/hooks.rs
@@ -540,7 +540,7 @@ async fn execute(
         })
         .collect_vec();
     let mut task = Task {
-        name: format!("hook:{}", hook.hook),
+        name: hook_task_name(config, root, hook.hook),
         display_name: format!("hook:{}", hook.hook),
         config_source: root.join("mise.toml"),
         config_root: Some(root.to_path_buf()),
@@ -557,6 +557,11 @@ async fn execute(
     } else {
         render_run_entries_for_hook(config, ts, hook.hook, root, &env, &mut task.run).await?;
     }
+    let rendered_run_scripts = task
+        .run_script_strings()
+        .into_iter()
+        .map(|script| (script, vec![]))
+        .collect_vec();
     let output_handler = OutputHandler::new(OutputHandlerConfig {
         output: Some(TaskOutput::Quiet),
         silent: false,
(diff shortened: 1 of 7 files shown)


risu729:
CI finished on the latest commit (`0a698f4`).

Unrelated failures:
- `e2e-0`, `e2e-3`, and `e2e-4` all fail while installing `aqua:astral-sh/uv@latest`, before reaching this PR’s hook parsing path. The error is GitHub artifact attestation verification: expected `astral-sh/uv/.github/workflows/release.yml`, found certificate/provenance `None`. The same failure appears in `shell/test_xonsh`, `cli/test_activate_multiple_xonsh`, `sync/test_sync_python_uv`, and `cli/test_deactivate_xonsh`.
- `test-ci` fails only because those e2e shards failed.
- `benchmark` fails on `mise x -- echo` due to a noisy hyperfine result (`25.9 ms ± 7.8 ms` vs `19.8 ms ± 0.9 ms`, with an outlier warning). This PR only changes hook config deserialization, docs, schema, and focused e2e coverage; it does not touch the `x` command path.

Greptile passed on the latest commit, and I applied/resolved the one valid AI review suggestion from Gemini (doc comment update).

src/hooks.rs
```diff
@@ -69,6 +69,22 @@ pub enum HookDef {
...
+pub enum HookScript {
+    String(String),
+    Array(Vec<String>),
+}
+
+impl HookScript {
+    pub fn into_script(self) -> String {
+        match self {
+            HookScript::String(script) => script,
+            HookScript::Array(script) => script.join("\n"),
```
jdx: not sure we should do it this way since it departs from how tasks do this, in tasks they're individual shell sessions

risu729:
Follow-up CI triage: the latest run for head `0a698f4` is green now. The earlier xonsh/sync failures were from the temporary upstream `uv@latest` attestation problem, which has since been fixed upstream, so I did not create the uv pin workaround PR.

src/hooks.rs
risu729: That's right. Are there any reasons hooks didn't support this before? Can we share the logic between tasks, including task references?

risu729: Hmm, it doesn't make sense `shell=fish` for tasks. I guess we should flatten the array instead?

risu729:
CI note for head `501939d`: `autofix` failed in `mise run lint-fix` while validating `schema/mise-registry-tool.json`:

`can't resolve reference #/definitions/toolId from id https://mise.en.dev/schema/mise-registry-tool.json`

This PR does not touch `schema/mise-registry-tool.json`; the same CI log reports `schema/mise.json is valid`. I also applied Greptile's valid `run_windows` edge-case feedback in `501939d` by allowing hook run tables with either `run` or `run_windows`. The `script` table without `shell` remains the intentional deprecated spawned-run alias described in the PR.

risu729:
CI note for the current head (`a8d0938`): `lint` failed before the branch-specific checks completed because `taplo` cannot resolve the registry schema reference `https://mise.en.dev/schema/mise-registry-tool.json#/definitions/os` while validating existing `registry/*.toml` files. The hook schema file from this PR was validated successfully (`schema/mise.json is valid`), and the hook e2e shell file passed both `shellcheck` and `shfmt` in the same job. `test-ci` failed as a dependent aggregate because `lint` was not successful.

risu729:
CI finished for head `ebc7a0202` with one unrelated failure:

- `e2e-0` was cancelled while in the `Test in Docker` step. The log shows normal e2e output through `config/test_env_non_ascii` at `2026-05-11T12:10:30Z`, then no test failure before GitHub cancelled the job at `2026-05-11T12:39:38Z` with `The operation was canceled`.
- `test-ci` failed only as the aggregate because `e2e-0` was cancelled.

All other checks completed successfully, including lint, benchmark, build-ubuntu/windows, unit-macos/windows, e2e-1 through e2e-7, windows-e2e, registry, docs, autofix, and Socket checks. I don’t see evidence that this cancellation is related to the hook/task changes in this PR.

PR #9607, opened 2026-05-05
fix(pipx): filter yanked pypi releases

risu729:
## Summary
- ignore PyPI releases that have no non-yanked files when listing pipx versions
- preserve exact pinned behavior by keeping the filter in fuzzy/latest version listing
- add unit coverage for fully yanked, mixed, empty, and string-yanked PyPI releases

## How it works
- For PyPI packages, the pipx backend uses the PyPI JSON API, whose `releases` map contains each version and the files published for that version.
- Each file can carry `yanked` metadata. The backend filters out a release only when there are no usable files: either the file list is empty or every file for that release is yanked.
- Mixed releases are kept if at least one file is not yanked, and `created_at` is computed from non-yanked files so date filtering reflects usable artifacts.
- The JSON `latest_stable_version` fast path now derives latest from the same filtered version list instead of returning `info.version`, because `info.version` can still point at a yanked release.
- The `yanked` field is deserialized defensively because PyPI-compatible JSON can represent yanked metadata as either a bool or a string reason.

## Why this is required
- Yanked PyPI releases remain visible in JSON metadata, but pip avoids them unless the user asks for an exact pinned version. mise should follow that behavior for fuzzy/latest resolution.
- Filtering only during version listing keeps exact pinned installs available while preventing `latest` or prefix resolution from choosing a yanked release by default.

## Tests
- `cargo fmt --check`
- `git diff --check`
- `cargo test backend::pipx::tests`

### DIFF
--- a/src/backend/pipx.rs
+++ b/src/backend/pipx.rs
@@ -21,6 +21,8 @@ use indexmap::IndexMap;
 use itertools::Itertools;
 use jiff::Timestamp;
 use regex::Regex;
+use serde::Deserialize;
+use serde_json::Value;
 use std::collections::BTreeMap;
 use std::ffi::OsString;
 use std::path::{Path, PathBuf};
@@ -148,7 +150,7 @@ impl Backend for PIPXBackend {
                                 debug!("Fetching JSON for {}", package);
                                 let url = registry_url.replace("{}", &package);
                                 let pkg: PypiPackage = HTTP_FETCH.json(url).await?;
-                                Ok(Some(pkg.info.version))
+                                Ok(Self::latest_stable_from_pypi_package(pkg))
                             } else {
                                 debug!("Fetching HTML for {}", package);
                                 let url = registry_url.replace("{}", &package);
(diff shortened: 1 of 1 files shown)


risu729:
CI note: the `e2e-3` job failed in unrelated tests: `cli/test_activate_multiple_xonsh` and `sync/test_sync_python_uv`. This PR only changes `src/backend/pipx.rs`, and the focused local `cargo test backend::pipx::tests` suite passes, so I am treating this as unrelated to the pipx yanked-release change.

risu729:
CI note: final failing checks appear unrelated to this pipx backend change:

- `benchmark`: benchmark completed, then failed while `thollander/actions-comment-pull-request` tried to create a PR comment: `Resource not accessible by integration`
- `e2e-0`: `shell/test_xonsh`
- `e2e-3`: `cli/test_activate_multiple_xonsh`, `sync/test_sync_python_uv`
- `e2e-4`: `cli/test_deactivate_xonsh`
- `test-ci`: aggregate failure from the failed e2e tranches above

This PR changes `src/backend/pipx.rs`; the focused local checks passed (`cargo test backend::pipx::tests`, `cargo fmt --check`, and `git diff --check`).

PR #9610, opened 2026-05-05
fix(aqua): support registry file links

risu729:
## Summary
- Parse aqua registry `files[].link` and `files[].hard` entries.
- Create aqua file aliases during install: relative symlinks on Unix by default, hard links when `hard: true`, and hard links for explicit file aliases on Windows.
- Keep PATH exposure aligned with aqua `files` entries, including linked aliases when `symlink_bins` is enabled.
- Add coverage for link templating, link/hard path mapping, relative symlink targets, hard-link error handling, and an e2e install from a local aqua registry.

## Aqua feature background
Aqua added these fields for packages where the executable name used to launch the binary matters:

- `files[].link` was added in https://github.com/aquaproj/aqua/pull/3134 to let a registry entry create an alias link next to the extracted binary. The motivating issue, https://github.com/aquaproj/aqua/issues/3130, uses `granted` / `assumego` as the example: one binary changes behavior based on `$0` / `argv[0]`, so invoking it through a different filename selects a different mode.
- `files[].hard` was added in https://github.com/aquaproj/aqua/pull/3780 for cases where a symlink is not enough and the installed executable should appear under the requested name as a hard link. The linked issue, https://github.com/aquaproj/aqua/issues/3775, covers `pnpm`, where newer pnpm versions observe the executable path/name and can fail when launched as the platform-specific release asset name instead of `pnpm`.

Relevant aqua docs and implementation:

- Files docs: https://github.com/aquaproj/aqua/blob/main/website/docs/reference/registry-config/files.md
- File schema: https://github.com/aquaproj/aqua/blob/main/json-schema/registry.json
- File config struct: https://github.com/aquaproj/aqua/blob/main/pkg/config/registry/config.go
- Link creation implementation: https://github.com/aquaproj/aqua/blob/main/pkg/installpackage/check_file.go

## Behavior changes
Before this PR, mise parsed `files[].name` and `files[].src`, but ignored `files[].link` and `files[].hard`. That meant packages relying on aqua file aliases could install the source binary but miss the alias that users actually execute, or execute a binary under the wrong filename for tools that inspect `$0` / `argv[0]`.

This PR changes native aqua installs as follows:

- `AquaFile` now models `link` and `hard`.
- `files[].link` supports the same aqua template context used for `files[].src`, including `AssetWithoutExt` and `FileName`.
- When `link` is present, mise resolves the source path from `src` when provided, or from `name` when `src` is omitted, then creates the linked destination next to the source file.
- On Unix, explicit links are relative symlinks unless `hard: true` is set.
- On Windows, explicit file links are hard links, matching aqua's behavior for this field.
- When `hard: true`, mise creates a hard link and replaces an existing file/symlink destination. It refuses to replace a destination directory.
- `complete_windows_ext` is applied to both the source and linked destination path during file resolution.
- `.mise-bins` creation now uses the resolved destination paths, so linked aliases are exposed when `symlink_bins` is enabled.

## Tests
- `cargo test -p aqua-registry test_aqua_file_link_template`
- `cargo test -p mise test_srcs`
- `cargo test -p mise test_relative_path`
- `cargo test -p mise test_create_file_link_rejects_hard_link_directory_destination`
- `mise run test:e2e e2e/backend/test_aqua_link_hard`
- `git diff --check`


risu729:
CI completed for head `2789142fa`.

Unrelated failures observed:
- `benchmark`: failed installing `aqua:sharkdp/hyperfine@latest` because the GitHub API returned `403 Forbidden` after the rate limit was exceeded.
- `e2e-1`: `shell/test_xonsh` failed.
- `e2e-4`: `cli/test_activate_multiple_xonsh` and `sync/test_sync_python_uv` failed. The new `backend/test_aqua_link_hard` test ran in this shard and passed, including `hello-world`, `hello-link`, and `hello-hard` assertions.
- `e2e-5`: `cli/test_deactivate_xonsh` failed.
- `windows-e2e`: `backend_aqua.executes tree-sitter via aqua backend on Windows` and `backend_github.installs and executes docker-compose via github backend` failed because GitHub API calls returned `403 Forbidden` after rate limiting.
- `test-ci`: aggregate failure from the failed e2e jobs.

All other checks passed, including build, lint, unit, registry, release, autofix, Socket, and Greptile. Greptile review threads from the first review were addressed and resolved; the follow-up Greptile check passed.

risu729:
For the follow-up Greptile P2 observations, I left both as-is:

- `relative_path` may preserve `.` in diverging tails, but the resulting symlink target is still correct and the helper avoids filesystem canonicalization.
- Conflicting `hard` settings for an identical `(src, dst)` pair are ambiguous registry input; the implementation keeps the first mapping, and the supported use case is one source fanning out to distinct destinations.

There are no unresolved Greptile review threads.

risu729:
Follow-up CI triage: the xonsh/sync `uv@latest` failures were the same temporary upstream uv attestation issue, now fixed by the maintainer. A fresh run is in progress; the previous GitHub `403 Forbidden` failures were API rate-limit noise, not this PR. I am not opening the uv pin workaround PR.

risu729:
CI completed for head `eab26b605`.

Unrelated failure observed:
- `windows-e2e`: `zig.executes zig 2024.11.0-mach` failed because downloading `https://pkg.hexops.org/zig/zig-windows-x86_64-0.14.0-dev.2577+271452d22.zip` timed out after three 30s attempts. This is external network/download flakiness and unrelated to the aqua link/hard-link changes.
- `test-ci`: failed only as the aggregate result from the failed `windows-e2e` job.

All other checks passed or were intentionally skipped, including build, lint, Linux e2e shards, registry, benchmark, autofix, release, Socket, and Greptile. Greptile added no new review comments on this head, and the existing AI review threads are resolved.

PR #9630, opened 2026-05-05
fix(ci): ignore noisy hyperfine regressions

risu729:
## Summary
- export hyperfine JSON alongside the existing markdown report
- fail only when a slowdown stays above the 10% threshold after accounting for measurement error
- keep noisy >10% measurements in the PR comment as inconclusive instead of failing CI

## Why
Recent hyperfine failures were tripping on point estimates with very large uncertainty, such as `mise env` at `1.15 ± 0.45` and `mise x -- echo` at `1.69 ± 0.81`, across unrelated PRs.

## Test
- `actionlint .github/workflows/hyperfine.yml`
- `git diff --check`
- extracted benchmark shell block and ran `bash -n` after replacing the GitHub expression
- sanity-checked the threshold logic against noisy and clear-regression samples

### DIFF
--- a/.github/workflows/hyperfine.yml
+++ b/.github/workflows/hyperfine.yml
@@ -80,33 +80,77 @@ jobs:
           echo "## Hyperfine Performance" >> comment.md
           for cmd in "${CMDS[@]}"; do
             if [ -n "${MISE_ALT:-}" ]; then
-              mise x hyperfine -- hyperfine -N -w 10 -r 500 --export-markdown out.md --reference "$MISE_ALT $cmd" "mise $cmd"
+              reference="$MISE_ALT $cmd"
             else
-              mise x hyperfine -- hyperfine -N -w 10 -r 500 --export-markdown out.md --reference "mise-${{ steps.versions.outputs.release }} $cmd" "mise $cmd"
+              reference="mise-${{ steps.versions.outputs.release }} $cmd"
             fi
+            mise x hyperfine -- hyperfine -N -w 10 -r 500 --export-markdown out.md --export-json out.json --reference "$reference" "$reference" "mise $cmd"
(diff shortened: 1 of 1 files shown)


risu729:
I'm keeping the relative-noise gate from `1b84f4e` in place. A lower-bound-only rule would still fail one of the observed flaky runs: `mise x -- echo` in run `25393624788` reported `1.43 ± 0.32`, so `1.43 - 0.32 = 1.11`, just over the `1.10` threshold, even though the same PR later passed.

The gate does not suppress the example from the Greptile summary: `ratio=1.50, error=0.15` would still fail because `0.15 <= 0.5 * (1.50 - 1.0)`. It only classifies results as inconclusive when the uncertainty is large relative to the observed slowdown.

risu729:
CI note: `test-tool-1` failed for an unrelated tool-install test tranche, not because of this PR's `.github/workflows/hyperfine.yml` change. The failed job reports `No API secret provided, skipping token fetch`, then ends with `Tools still failing after retry` for many registry tools including `amass`, `apko`, `cosign`, `taplo`, `yt-dlp`, and `zizmor`.

The corrected `benchmark` job is green and now compares exactly `mise-2026.5.0 ...` vs `mise ...` for each command.

risu729:
CI note: the current `lint` failure, and therefore the aggregate `test-ci` failure, happened before the fork branch ref was restored. The job failed during `actions/checkout` while fetching `refs/heads/fix/hyperfine-noisy-regressions`, before any lint/test command ran.

I restored the branch with a normal non-force push, but I do not have permission to rerun upstream Actions jobs (`gh run rerun --failed` is denied without repository admin rights). The benchmark check for this PR passed after the hyperfine workflow change.

PR #9640, opened 2026-05-06
feat: add --inactive option to outdated and upgrade commands for inactive tools

roele:
Adds a new `--inactive` flag to `outdated` and `upgrade` command to allow to show/upgrade tools which are currently not active or  defined in a configuration.

### DIFF
--- a/docs/cli/outdated.md
+++ b/docs/cli/outdated.md
@@ -31,6 +31,12 @@ show other 20.x versions, not 21.x or 22.x versions.
 
 Using this flag, if there are 21.x or newer versions it will display those instead of 20.x.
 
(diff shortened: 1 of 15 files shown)


docs/cli/outdated.md
```diff
@@ -42,6 +42,12 @@ will skip tools defined in the global config (~/.config/mise/config.toml).
 
 Don't show table header
 
+### `--no-source`
```
jdx: I think it should be --inactive or --all

roele: Yes was also unsure about the naming, I'd probably lean towards `--inactive`

PR #9645, opened 2026-05-06
refactor(aqua): store aqua var defaults as strings

risu729:
## Summary
- Store aqua var defaults as `Option<String>` instead of preserving arbitrary YAML values.
- Rely on Serde's normal `Option<String>` YAML deserialization; there is no custom default validator or tagged-value handling.
- Simplify runtime var resolution now that defaults are stored as strings.

## Notes
- Upstream aqua's schema/type allows arbitrary var default values.
- mise intentionally stores defaults as strings for implementation simplicity.
- With serde_yaml, scalar defaults deserialize into strings, explicit `null` deserializes as no default, and sequence/mapping defaults fail during typed YAML parsing.
- The vendored registry currently only uses string defaults or required vars without defaults.

## Test Plan
- [x] `cargo fmt --package aqua-registry --check`
- [x] `cargo check -p aqua-registry`
- [x] `cargo test -p aqua-registry`

*This PR body was generated by an AI coding assistant.*

### DIFF
--- a/crates/aqua-registry/build.rs
+++ b/crates/aqua-registry/build.rs
@@ -50,6 +50,7 @@ fn generate_baked_registry(out_dir: &str) -> Result<()> {
                 registry_file.display()
             )
         })?;
+    validate_var_defaults(packages)?;
     let registries = package_registries(packages)?;
     if registries.is_empty() {
         return Err(eyre!(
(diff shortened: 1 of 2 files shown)


risu729:
CI failure looks unrelated to this AquaVar change. `windows-e2e` failed in `e2e-win/zig.Tests.ps1` because downloading `https://pkg.hexops.org/zig/zig-windows-x86_64-0.14.0-dev.2577+271452d22.zip` timed out after three attempts. `test-ci` then failed only because it aggregates the failed `windows-e2e` result.

risu729:
Greptile noted that `test_vars_scalar_defaults_deserialize_as_strings` should fail because serde_yaml would not coerce YAML bool/number scalars to `String`. I checked this locally in this workspace: `cargo test -p aqua-registry` passes, including `test_vars_scalar_defaults_deserialize_as_strings`; serde_yaml 0.9 deserializes those scalar defaults into strings for the `Option<String>` field.

The build/codegen validator was intentionally removed per review direction. The PR body now says the actual behavior: scalar defaults deserialize into strings, explicit `null` becomes no default, and sequence/mapping defaults fail during typed YAML parsing.

PR #9582, opened 2026-05-03
refactor(schema): reuse task props with unevaluatedProperties

risu729:
## Summary
- Restore task and task template schemas to `allOf` + `unevaluatedProperties` now that Tombi supports the evaluated-property behavior needed here.
- Keep `task_props` in the generated included-task schema so task properties are defined once instead of duplicated into task and template objects.
- Use `unevaluatedProperties: false` consistently for closed object schemas across the other schema files too.
- Update the Tombi schema e2e test to assert the pinned Tombi version and document why the pin is new enough.

## Tombi support
- Tombi v0.9.19 includes `Fix issue 1719 unevaluatedProperties handling`: https://github.com/tombi-toml/tombi/releases/tag/v0.9.19
- The supporting Tombi PR is https://github.com/tombi-toml/tombi/pull/1725, which fixed evaluated-property collection for referenced schemas inside applicators so `allOf` + `unevaluatedProperties` does not report false diagnostics.
- The test remains pinned to Tombi v0.9.22, which is newer than v0.9.19, so no version bump was necessary.

## Test plan
- `mise run render:schema`
- `hk check schema/mise.json schema/mise-task.json schema/miserc.json schema/mise-settings.json schema/mise-registry-tool.json schema/mise.plugin.json xtasks/render/schema.ts`
- `taplo check settings.toml registry/*.toml test/fixtures/mise.plugin.toml`
- `mise run test:e2e config/test_schema_tombi`

This PR was generated by an AI coding assistant.

PR #9646, opened 2026-05-06
refactor(schema): use $defs in schema files

risu729:
## Why
#9582 moved the generated schemas onto the JSON Schema draft 2020-12 shape we need for `allOf` + `unevaluatedProperties`. Once we rely on that behavior, the remaining schema files should use the draft 2020-12 definition namespace consistently.

Keeping `definitions` around works in some validators as legacy compatibility, but it leaves these files half-migrated: new reusable schema pieces are expected to live under `$defs`, while older local refs still point at `#/definitions/...`. This PR finishes that migration for the settings and registry-tool schemas so follow-up schema refactors can add shared definitions without mixing two local-ref conventions.

## What changed
- Migrates `schema/mise-settings.json` and `schema/mise-registry-tool.json` from `definitions` to `$defs`.
- Updates local refs from `#/definitions/...` to `#/$defs/...`.
- Keeps the accepted schema values unchanged; this is a namespace/ref cleanup for draft 2020-12 consistency.

## Validation
- `jq empty schema/mise-settings.json schema/mise-registry-tool.json`
- `git diff --check`

### DIFF
--- a/schema/mise-registry-tool.json
+++ b/schema/mise-registry-tool.json
@@ -24,13 +24,13 @@
         {
           "type": "array",
           "items": {
-            "$ref": "#/definitions/optionValue"
+            "$ref": "#/$defs/optionValue"
           }
         },
         {
           "type": "object",
           "additionalProperties": {
-            "$ref": "#/definitions/optionValue"
+            "$ref": "#/$defs/optionValue"
           }
         }
       ]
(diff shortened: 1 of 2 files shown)


risu729:
The current CI failures look unrelated to this schema refactor. `windows-e2e` failed in `e2e-win\zig.Tests.ps1:7` while installing `zig@2024.11.0-mach`; the download of `zig-windows-x86_64-0.14.0-dev.2577+271452d22.zip` from `pkg.hexops.org` timed out after three 30s attempts and ended with Windows `os error 10060`.

The `test-ci` failure is only the aggregate job reporting that `windows-e2e` failed or was skipped.

PR #9648, opened 2026-05-06
refactor(schema): reuse registry common types

risu729:
## Why
After #9582, the registry schema can lean on shared definitions without tripping Tombi's older `unevaluatedProperties` false positives. That means repeated primitive schema fragments no longer need to stay inline just to avoid validator issues.

The registry schema repeats the same tool-id, OS, and backend option value shapes in several places. Pulling them into shared definitions reduces drift risk: aliases, overrides, test tools, backend platforms, and backend options now validate against the same source of truth.

## What changed
- Extracts common registry schema definitions for tool IDs, OS names, and recursive backend option values.
- Reuses those definitions across aliases, overrides, test tools, platform lists, and backend options.
- Keeps the registry schema behavior the same while making future edits smaller and less error-prone.

## Validation
- `jq empty schema/mise-registry-tool.json`
- `git diff --check`

### DIFF
--- a/schema/mise-registry-tool.json
+++ b/schema/mise-registry-tool.json
@@ -9,39 +9,30 @@
       "type": "string",
       "pattern": "^(aqua|asdf|cargo|conda|core|dotnet|gem|github|gitlab|go|http|npm|pipx|spm|ubi|vfox):([^\\[\\]]|\\[[^\\]=]*\\])+$"
     },
-    "optionValue": {
-      "description": "Backend option value.",
-      "anyOf": [
+    "backendOptionValue": {
+      "oneOf": [
         {
           "type": "string"
         },
         {
-          "type": "number"
-        },
-        {
           "type": "boolean"
-        },
-        {
-          "type": "array",
-          "items": {
-            "$ref": "#/definitions/optionValue"
-          }
-        },
-        {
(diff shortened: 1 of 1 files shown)


risu729:
I am not applying a code change for the remaining Greptile concerns on this PR:

- Draft-7 compatibility: these schema files explicitly declare JSON Schema Draft 2020-12, and #9582 intentionally switches to `unevaluatedProperties` for `allOf` + `$ref` composition now that the target Tombi version supports it. Keeping Draft-7 semantics would conflict with the declared dialect and the parent PR.
- `task_props` sync: `schema/mise-task.json` is generated from `schema/mise.json` by `xtasks/render/schema.ts` via `crawlReferencedDefs`, so this is not a manually maintained second copy.

risu729:
The current CI failures look unrelated to this schema refactor. `windows-e2e` failed in `e2e-win\zig.Tests.ps1:7` while installing `zig@2024.11.0-mach`; the download of `zig-windows-x86_64-0.14.0-dev.2577+271452d22.zip` from `pkg.hexops.org` timed out after three 30s attempts and ended with Windows `os error 10060`.

The `test-ci` failure is only the aggregate job reporting that `windows-e2e` failed or was skipped.

PR #9649, opened 2026-05-06
refactor(schema): extract tool options definition

risu729:
## Why
#9582 restored the schema pattern we wanted for task objects: shared property definitions composed with `allOf`, then closed with `unevaluatedProperties`. With Tombi pinned to a version that handles evaluated properties correctly, we can use the same pattern for tool option objects.

The top-level `[tools]` schema and task `tools` schema both describe the same `{ version, os }` object shape, but with different extra-property rules. Duplicating the common part makes it easy for those two locations to drift. This PR extracts the shared `version`/`os` piece and lets each call site keep its own closure behavior.

## What changed
- Adds a shared `$defs.tool_options` schema for the common `version` and `os` fields.
- Reuses it from both top-level tool options and task tool options.
- Preserves the important difference between call sites: root tool objects still allow backend-specific values, while task tool objects still only allow string extras.
- Extends the Tombi regression test so warnings fail and the shared-schema behavior is covered.

## Validation
- `git diff --check`
- Direct Tombi validation with `tombi@0.9.22` and `--error-on-warnings` for valid root/task tool options and invalid task tool extra options.
- Full e2e should run in CI; my earlier local targeted e2e run was stopped after the build step stalled behind another active cargo build on this machine.

### DIFF
--- a/e2e/config/test_schema_tombi
+++ b/e2e/config/test_schema_tombi
@@ -14,6 +14,7 @@ mise use -g tombi@$TOMBI_VERSION
 SCHEMA_PATH="$ROOT/schema/mise.json"
 TASK_SCHEMA_PATH="$ROOT/schema/mise-task.json"
 TOMBI="mise x tombi@$TOMBI_VERSION -- tombi"
+TOMBI_LINT="$TOMBI lint --offline --no-cache --error-on-warnings --quiet"
 assert_contains "$TOMBI --version" "tombi $TOMBI_VERSION"
 
 # Set up tombi config pointing to local schema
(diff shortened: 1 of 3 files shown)


risu729:
Greptile note on the root `tool` extra-property types looks stale against the actual stack/current main. `upstream/main:schema/mise.json` already allows `string | number | boolean | array | object` for root tool extra properties; this PR preserves that set by moving the same schema under `unevaluatedProperties` around the new `$defs.tool_options` ref. No code change needed.

This comment was generated by an AI coding assistant.

risu729:
CI note: `windows-e2e` failed for an unrelated network/download issue. The failing test was `zig.executes zig 2024.11.0-mach`; both retry attempts timed out downloading `zig-windows-x86_64-0.14.0-dev.2577+271452d22.zip` from `https://pkg.hexops.org/...` with Windows socket error 10060. This does not appear related to the schema-only changes in this PR.

This comment was generated by an AI coding assistant.

risu729:
Follow-up on CI: the final `test-ci` failure is the aggregate job reporting `windows-e2e failed or was skipped`. The underlying `windows-e2e` failure is the unrelated Zig download timeout noted above; all Linux e2e shards and lint passed.

This comment was generated by an AI coding assistant.

PR #9651, opened 2026-05-06
refactor(schema): reuse env directive property schemas

risu729:
## Why
This is the same follow-up enabled by #9582 and extended by #9649: now that Tombi correctly tracks evaluated properties through referenced schemas, repeated env directive property schemas can be shared instead of copied into every object variant.

The env schema has many repeated pieces: primitive env values, path/path-array handling, `tools`, `redact`, `required`, and age encryption fields. Keeping those inline makes the schema large and makes it easy for one directive form to drift from another. Reusing definitions keeps env object validation consistent across the main config schema and the generated task schema.

## What changed
- Adds shared env schema definitions for primitive env values, path/path arrays, `tools`, `redact`, `required`, and age value/format fields.
- Replaces repeated inline env and env-directive property schemas with refs to those definitions.
- Expands Tombi coverage for env value objects, env file/source/path directives, age format validation, and directive `required` validation.
- Inherits the #9649 test harness change that treats Tombi warnings as failures.

## Validation
- `git diff --check`
- Direct Tombi validation with `tombi@0.9.22` and `--error-on-warnings` for valid env value/directive cases and invalid age/directive cases.
- Full e2e should run in CI; my earlier local targeted e2e run was stopped after the build step stalled behind another active cargo build on this machine.

### DIFF
--- a/e2e/config/test_schema_tombi
+++ b/e2e/config/test_schema_tombi
@@ -98,6 +98,7 @@ TOML
 cd "$HOME/workdir"
 assert_succeed "$TOMBI_LINT mise.toml"
 assert_succeed "$TOMBI_LINT mise-age.toml"
+assert_succeed "$TOMBI_LINT mise-env-directives.toml"
 assert_succeed "$TOMBI_LINT mise-os.toml"
 assert_succeed "$TOMBI_LINT mise-task-os.toml"
 
--- a/e2e/config/test_schema_tombi
+++ b/e2e/config/test_schema_tombi
@@ -27,7 +27,7 @@ strict = true
 
 [[schemas]]
 path = "file://$SCHEMA_PATH"
-include = ["mise.toml", "mise-age.toml", "mise-os.toml"]
+include = ["mise.toml", "mise-age.toml", "mise-env-directives.toml", "mise-os.toml"]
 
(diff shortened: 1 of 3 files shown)


risu729:
Applied Greptile’s valid note by adding `assert_succeed "$TOMBI_LINT mise-env-directives.toml"` in commit 7b044c894, so the new positive env directive fixture is exercised.

This comment was generated by an AI coding assistant.

risu729:
CI note: `windows-e2e` failed for the same unrelated network/download issue seen on #9649. The failing test was `zig.executes zig 2024.11.0-mach`; both retry attempts timed out downloading `zig-windows-x86_64-0.14.0-dev.2577+271452d22.zip` from `https://pkg.hexops.org/...` with Windows socket error 10060. This does not appear related to the schema-only changes in this PR.

This comment was generated by an AI coding assistant.

risu729:
Follow-up on CI: the final `test-ci` failure is the aggregate job reporting `windows-e2e failed or was skipped`. The underlying `windows-e2e` failure is the unrelated Zig download timeout noted above; all Linux e2e shards and lint passed.

This comment was generated by an AI coding assistant.

PR #9657, opened 2026-05-06
fix(task): stable MISE_PROJECT_ROOT for monorepo tasks, add MISE_MONOREPO_ROOT

jdx:
## Summary

- For monorepo tasks, `MISE_PROJECT_ROOT` was cwd-dependent: from inside the subproject it pointed at the subproject's directory, but from the monorepo root or an intermediate directory it pointed at the monorepo root. The value was sourced from `config.project_root` (which is the closest project to cwd) before falling back to the task's own `config_root`.
- Prefer the task's own `config_root` so `MISE_PROJECT_ROOT` is always the directory of the mise.toml that defined the task — stable regardless of where the task was invoked from. For non-monorepo configs `task.config_root == config.project_root`, so behavior is unchanged.
- Add `MISE_MONOREPO_ROOT` (the directory of the config with `experimental_monorepo_root = true`) so subproject tasks can reference the monorepo root without a cwd-dependent fallback.

Refs #9639.

## Test plan

- [x] New e2e test `e2e/tasks/test_task_monorepo_project_root` exercises a subproject task invoked from the monorepo root, an intermediate dir, and the subproject dir, and asserts `MISE_PROJECT_ROOT` and `MISE_MONOREPO_ROOT` both stay stable.
- [x] Existing monorepo task tests still pass (`test_task_monorepo_config_context`, `test_task_monorepo_run_project_tasks`, `test_task_monorepo_run_project_local_tasks`, `test_task_monorepo_global_tasks`).

---

> [!NOTE]
> **Medium Risk**
> Changes task execution environment variable semantics for `MISE_PROJECT_ROOT` (especially in monorepos) and adds a new `MISE_MONOREPO_ROOT`, which could affect scripts relying on prior cwd-dependent behavior despite added e2e coverage.
> 
> **Overview**
> Makes `MISE_PROJECT_ROOT` resolve to the task’s defining `mise.toml` directory (via `task.config_root`) so monorepo subproject tasks get a stable project root regardless of invocation cwd, while preserving prior behavior for *global* and *remote* tasks.
> 
> Adds `MISE_MONOREPO_ROOT` to the task environment (derived from the config with `experimental_monorepo_root = true`), updates task docs accordingly, and introduces a new e2e test covering monorepo/subproject/root/global task cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ea7f632ecae44ef64850104c8e2d7a748c4a47ed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/tasks/index.md
+++ b/docs/tasks/index.md
@@ -53,7 +53,8 @@ The following environment variables are passed to the task:
 
 - `MISE_ORIGINAL_CWD`: The original working directory from where the task was run.
 - `MISE_CONFIG_ROOT`: The directory containing the `mise.toml` file where the task was defined or if the config path is something like `~/src/myproj/.config/mise.toml`, it will be `~/src/myproj`.
-- `MISE_PROJECT_ROOT`: The root of the project.
+- `MISE_PROJECT_ROOT`: The root of the project that defines the task. For monorepo subproject tasks this is the subproject's directory and is stable regardless of the directory the task is invoked from.
+- `MISE_MONOREPO_ROOT`: The root of the monorepo (the directory containing the config with `experimental_monorepo_root = true`). Only set inside a monorepo.
 - `MISE_TASK_NAME`: The name of the task being run.
 - `MISE_TASK_DIR`: The directory containing the task script.
 - `MISE_TASK_FILE`: The full path to the task script.
--- a/e2e/tasks/test_task_monorepo_project_root
+++ b/e2e/tasks/test_task_monorepo_project_root
@@ -64,3 +64,22 @@ output_global_task_from_sub=$(cd projects/frontend && mise run global-show-root)
 echo "global task from subproject: $output_global_task_from_sub"
 # From inside the subproject the local project root is the subproject's mise.toml dir.
 assert_contains "echo $output_global_task_from_sub" "PROJECT_ROOT=$frontend_root"
+
+# Global *file* task (script under ~/.config/mise/tasks/) must also use the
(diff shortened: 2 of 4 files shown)


jdx:
Greptile is right — preferring \`task.config_root\` unconditionally regressed \`MISE_PROJECT_ROOT\` for tasks defined in the global config (\`~/.config/mise/config.toml\`), where \`task.config_root\` is the global config dir rather than the project the user is running the task from.

Fix in 08e09f6: only flip the precedence when the task is from a project config. For global/system-config tasks the original \`config.project_root → task.config_root\` order is preserved. Extended the e2e test to invoke a global task from both the monorepo root and a subproject and assert \`MISE_PROJECT_ROOT\` is the local project root in each case.

jdx:
Greptile, Cursor Bugbot and Gemini all converged on the same gap: \`is_global_config(&task.config_source)\` only matches global TOML config paths, so global *file* tasks under \`~/.config/mise/tasks/\` (whose \`config_source\` is the script's own path) fell through to the new branch and would have received the global config dir as \`MISE_PROJECT_ROOT\`.

Switched the guard to \`task.global || task.is_remote()\` in ea7f632, which \`mark_tasks_as_global\` already sets for both inline TOML and file-based global tasks (and \`is_remote()\` covers \`git::\`/\`http(s)://\` includes that Gemini flagged). Extended the e2e test with a script-based global task that's invoked from both the monorepo root and the subproject; both assert the local project root.

PR #9667, opened 2026-05-07
fix(http): prefer current config over install manifest opts

risu729:
Fixes the HTTP backend option layering reported in https://github.com/jdx/mise/discussions/8880.

## Summary
- Track install-manifest tool opts as their own lower-priority source instead of treating them like inline backend args.
- Keep true inline bracket opts as the highest-priority override.
- Add an HTTP `ls-remote` regression where a stale manifest `version_json_path` must not override the current config.

## Tests
- `cargo test test_resolve_tool_opts_prefers_config_over_install_manifest_opts --all-features`
- `cargo test get_tool_opts_with_overrides --all-features`
- `mise run test:e2e e2e/backend/test_http_ls_remote_current_opts`
- `mise run test:e2e e2e/backend/test_github_ls_remote_inline_opts`


risu729:
CI note: `autofix` and `lint` failed while validating schemas, not from this HTTP/tool option change. The failing logs show `schema/mise-registry-tool.json is invalid` with `error: can't resolve reference #/definitions/toolId from id https://mise.en.dev/schema/mise-registry-tool.json`, plus taplo registry validation errors for unresolved `#/definitions/os`.

`test-ci` failed as the aggregate result after `lint` failed. The other relevant checks passed, including build, unit, Windows e2e, Linux e2e tranches 0-7, benchmark, Socket Security, and Greptile.

risu729:
CI update for the latest head (`426fe21bb`): `lint` was cancelled after failing the existing schema/taplo validation path. The lint log shows `schema/mise-registry-tool.json is invalid` with `error: can't resolve reference #/definitions/toolId from id https://mise.en.dev/schema/mise-registry-tool.json`, plus taplo registry validation errors for `Invalid reference: https://mise.en.dev/schema/mise-registry-tool.json#/definitions/os`.

`test-ci` failed only as the aggregate result because `lint` failed (`lint failed or was skipped`). All other checks passed or were intentionally skipped, including autofix, benchmark, build, unit, nightly, Linux e2e tranches 0-7, Windows e2e, registry checks, release, Socket Security, and the latest Greptile review.

This appears unrelated to this PR; the patch does not touch `schema/` or `registry/`.

risu729:
Follow-up implementation direction from the tool-option categorization review:

- Do not use `remote_version_listing_tool_option_keys()` directly as the install-manifest persistence filter. That method is for versions-host/cache source decisions, and some keys in it also affect install/download source or installed layout.
- Keep the source precedence fix: install-manifest opts should be weaker than current config/inline opts, but stronger than registry defaults.
- Replace the current global listing-key filter with a backend-aware install-manifest persistence contract, likely something like `install_manifest_excluded_tool_option_keys()` or a persisted-key allowlist.
- Persist opts that describe the installed artifact or exposed binaries, such as `url`, `api_url` where needed for self-hosted downloads, `provider`, `channel`, `asset_pattern`, `checksum`, `size`, `format`, `strip_components`, `bin`, `rename_exe`, `bin_path`, `filter_bins`, aqua `vars`, dotnet `runtime`, rust components/targets/profile, etc.
- Do not persist transient resolution policy: `minimum_release_age`, deprecated `install_before`, `prerelease`, and pure version-list parser keys like HTTP `version_list_url`, `version_regex`, `version_json_path`, and `version_expr`.
- Be careful with mixed-role keys. `s3.url`, `s3.region`, `s3.endpoint`, `conda.channel`, `ubi.provider`, `ubi.api_url`, `spm.provider`, `spm.api_url`, and `java.release_type` affect remote listing and install/source reconstruction, so they should generally be handled by precedence rather than blanket exclusion.
- Revisit aqua `symlink_bins`: it is not artifact identity, so excluding it from lockfile can be defensible, but it is expose/PATH identity and should probably be retained in install metadata. There may also be a follow-up bug if toggling `symlink_bins=true` after install returns `.mise-bins` without creating it until reinstall.

I am leaving this as a note for a later follow-up rather than expanding this PR further.

PR #9668, opened 2026-05-07
refactor(deps): use std::sync::LazyLock instead of once_cell::Lazy

risu729:
Replace `once_cell::sync::Lazy` with `std::sync::LazyLock` where initialization is infallible:

- `crates/vfox/src/config.rs` (Linux `env_type` static)
- `crates/vfox/src/registry.rs` (`SDKS`)
- `crates/vfox/src/runtime.rs`
- `src/task/task_file_providers/remote_task_git.rs`

This keeps the direct `once_cell` dependency for `OnceCell` / `get_or_try_init` call sites. `std::sync::OnceLock::get_or_try_init` remains unstable behind `once_cell_try` ([rust-lang/rust#109737](https://github.com/rust-lang/rust/issues/109737)), so those are unchanged.


risu729:
CI has finished. The remaining failures look unrelated to this dependency-removal change:

- `autofix` fails in `mise run lint-fix` while validating `schema/mise-registry-tool.json`: `#/definitions/toolId` cannot be resolved.
- `lint` emits the same schema/taplo invalid-reference errors for `schema/mise-registry-tool.json` (`toolId`/`os`) and then ends as cancelled.
- `test-ci` is the aggregate job and fails because `lint` did not complete successfully.

Greptile completed successfully after the review follow-ups.

risu729:
We need https://github.com/rust-lang/rust/issues/109737 to be stable.

PR #9717, opened 2026-05-08
fix(forgejo): include prereleases when opted in

risu729:
## Summary
- keep Forgejo prerelease releases in the cached remote release list while continuing to exclude drafts
- propagate Forgejo release `prerelease` metadata into `VersionInfo` so the shared `prerelease = true` / `MISE_PRERELEASES=1` filtering path works for `ls-remote`, `latest`, and fuzzy matching
- document the Forgejo `prerelease` tool option and update shared prerelease docs

## Forgejo docs
- https://forgejo.org/docs/latest/user/releases/

## Testing
- `cargo fmt --check`
- `cargo check --bins`
- `cargo test forgejo`
- `mise run render:schema` (ran successfully; generated schema diff was formatting-only because the changed settings text is not emitted into the JSON schema)

### DIFF
--- a/src/forgejo.rs
+++ b/src/forgejo.rs
@@ -346,6 +346,7 @@ mod tests {
         assert!(is_published_release(&release("1.0.0", false, false)));
         assert!(is_published_release(&release("1.1.0-rc1", false, true)));
         assert!(!is_published_release(&release("1.2.0", true, false)));
+        assert!(!is_published_release(&release("1.2.0-rc1", true, true)));
     }
 
     #[test]
--- a/src/forgejo.rs
+++ b/src/forgejo.rs
@@ -46,7 +46,7 @@ async fn get_releases_cache(key: &str) -> RwLockReadGuard<'_, CacheGroup<Vec<For
         .await
         .entry(key.to_string())
         .or_insert_with(|| {
-            CacheManagerBuilder::new(cache_dir().join(format!("{key}-releases.msgpack.z")))
+            CacheManagerBuilder::new(cache_dir().join(format!("{key}-all-releases.msgpack.z")))
                 .with_fresh_duration(Settings::get().fetch_remote_versions_cache())
                 .build()
         });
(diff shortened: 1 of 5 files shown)


risu729 (commented):
Self-review completed after PR creation. Findings: none.

Review checklist:
- Forgejo release listing now caches the non-draft release superset and preserves upstream `prerelease` metadata.
- Default behavior still excludes prereleases through the shared `filter_cached_prereleases` path.
- `prerelease = true`, `MISE_PRERELEASES=1`, and `mise ls-remote --prerelease` can now include Forgejo prereleases because `VersionInfo.prerelease` is populated.
- `latest` with prereleases enabled skips the Forgejo `/releases/latest` shortcut via the existing `latest_stable_version` guard and resolves from the full list.
- The cache filename changed from `*-releases` to `*-all-releases` so existing stable-only Forgejo cache entries cannot hide prereleases until expiration.
- Docs now describe the Forgejo option, update the shared prerelease setting text, and remove the stale GitHub-doc statement that Forgejo is unaffected.

Local verification run:
- `cargo fmt --check`
- `cargo check --bins`
- `cargo test forgejo`
- `mise run render:schema` (schema output was formatting-only, so no schema files are committed)

Residual risk: I did not add or run a live Forgejo e2e because `e2e/backend/test_forgejo` is intentionally disabled for intermittent API flakiness.

risu729:
Applied Greptile’s valid coverage suggestion in `3d6ee5988` by adding the `draft && prerelease` assertion to the Forgejo release predicate test. The review thread is resolved.

src/forgejo.rs
```diff
@@ -46,7 +46,7 @@ async fn get_releases_cache(key: &str) -> RwLockReadGuard<'_, CacheGroup<Vec<For
         .await
         .entry(key.to_string())
         .or_insert_with(|| {
-            CacheManagerBuilder::new(cache_dir().join(format!("{key}-releases.msgpack.z")))
+            CacheManagerBuilder::new(cache_dir().join(format!("{key}-all-releases.msgpack.z")))
```
jdx: the cache gets reset whenever the version changes doesn't it?

risu729: Yes, sorry I missed that. I followed GitHub/Aqua patterns blindly.

PR #9722, opened 2026-05-08
feat(shim): add experimental_lazy_shims for aqua-backed tools

yumafuu:
Related discussion: https://github.com/jdx/mise/discussions/9674

## Summary

Adds an experimental `experimental_lazy_shims` setting that generates shims for configured-but-not-installed aqua-backed tools using aqua registry metadata.

When such a shim (or a `command_not_found` event in `mise activate` mode) is invoked, mise auto-installs the tool and executes the requested binary.

The feature is gated behind:

```toml
experimental = true
experimental_lazy_shims = true
```

and currently limited to the aqua backend.

## Why aqua first

This PR starts with aqua because aqua registry metadata can provide executable names before installation, allowing mise to pre-generate deterministic shims without downloading the tool first.

A new `Backend::lazy_shim_bin_names` trait method is added as the extension point. The default implementation returns `Ok(None)`, and only `AquaBackend` overrides it in this PR.

## Unified lazy install path

The lazy auto-install logic lives in `Toolset::install_missing_bin`, which is used by both:

1. shim invocation (`handle_shim` → `which_shim`)
2. `mise activate`'s `hook-not-found`

This keeps shim-based and non-shim environments on the same code path.

## Falls back to current behavior when

* the aqua package uses `files[*].link`
* the package is unsupported on the current platform
* executable names cannot be derived safely
* multiple configured tools advertise the same executable name
* an installed tool already owns the same shim name

## Tests

Added e2e coverage for:

* lazy shim generation + feature flag gating
* activate-mode `hook-not-found` behavior

All existing tests pass and clippy is clean.

## Known limitations

* `pkg.supported_envs` is not checked yet
* `files[*].link` packages are skipped entirely
* `lazy_shim_bin_names` is currently sequential
* only e2e coverage exists for lazy shim resolution

### DIFF
--- a/settings.toml
+++ b/settings.toml
@@ -578,6 +578,31 @@ right, try disabling it if you can.
 env = "MISE_EXPERIMENTAL"
 type = "Bool"
 
+[experimental_lazy_shims]
+description = "[experimental] Generate shims for aqua-backed tools before they are installed and auto-install on first invocation."
+docs = """
+[experimental] When set together with `experimental = true`, mise will generate shims for
+aqua-backed tools that are configured but not yet installed, when the executable names
+can be determined unambiguously from the aqua registry metadata.
+
+When such a shim is executed, the underlying tool is installed automatically and then
+the requested binary is run. The same code path is used by `mise activate`'s
+`command_not_found_handler`, so users without shims on PATH are also covered.
+
+Currently this is limited to the aqua backend. The lazy shim is *not* generated when:
+
+- The aqua registry package has no `files` entry and no derivable name
+- Any of the package's files use the `link` field (per-platform template — too risky
+  to resolve ahead of install)
+- The package is unsupported on the current platform (`no_asset` / `error_message`)
+- Two or more configured aqua tools advertise the same executable name
+
+In any of these cases the existing behavior applies (no lazy shim, install required
+before the binary can be invoked).
+"""
+env = "MISE_EXPERIMENTAL_LAZY_SHIMS"
+type = "Bool"
+
 [fetch_remote_versions_cache]
 default = "1h"
 description = "How long to cache remote versions for tools."
--- a/src/backend/aqua.rs
+++ b/src/backend/aqua.rs
@@ -545,6 +545,75 @@ impl Backend for AquaBackend {
         Ok(paths)
     }
 
+    /// Determine shim names from aqua registry metadata, *without* requiring
+    /// the package to be installed. Used by `experimental_lazy_shims`.
+    ///
+    /// Returns `Ok(None)` whenever names cannot be determined unambiguously
+    /// from registry data so the caller falls back to current behavior.
+    async fn lazy_shim_bin_names(
+        &self,
+        _config: &Arc<Config>,
+        tv: &ToolVersion,
+    ) -> Result<Option<Vec<String>>> {
+        let pkg = match self.package_with_options(tv, &[&tv.version]).await {
+            Ok(p) => p,
+            Err(e) => {
+                trace!(
+                    "lazy_shim_bin_names: failed to load aqua package {}: {:#}",
+                    self.id, e
+                );
+                return Ok(None);
+            }
+        };
+
+        if pkg.no_asset || pkg.error_message.is_some() {
+            return Ok(None);
+        }
+
+        // Skip when any file uses `link` — the install-time symlink-target
+        // resolution is non-trivial (template-evaluated per platform) and we
+        // don't want to risk producing the wrong shim name. Falling back is
+        // safe: the user gets current behavior.
+        if pkg.files.iter().any(|f| f.link.is_some()) {
+            return Ok(None);
+        }
+
+        let names: Vec<String> = if pkg.files.is_empty() {
+            // Single-binary fallback: aqua exposes the package as repo_name (or
+            // the last segment of `name` when set). Mirrors `srcs_for_platform`.
+            let fallback = pkg
+                .name
+                .as_deref()
+                .and_then(|n| n.split('/').next_back())
+                .filter(|n| !n.is_empty())
+                .map(str::to_string)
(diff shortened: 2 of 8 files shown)


src/backend/aqua.rs
```diff
@@ -545,6 +545,92 @@ impl Backend for AquaBackend {
...
+                trace!(
+                    "lazy_shim_bin_names: failed to load aqua package {}: {:#}",
+                    self.id, e
+                );
+                return Ok(None);
+            }
+        };
+
+        if pkg.no_asset || pkg.error_message.is_some() {
+            return Ok(None);
```
risu729, line 570: `package_with_options` applies platform/version overrides, but this only filters `no_asset`/`error_message`. Packages can also be excluded via `supported_envs`; `install_version_` rejects those later in `validate(&pkg)`. As written, an unsupported package with static `files` will still get a lazy shim and `hook-not-found` will try to install it, producing an install error instead of falling through to the current not-found behavior. We should check `is_platform_supported(&pkg.supported_envs, os(), arch())` here too (and if the matching override depends on the GitHub tag form, resolve the local version to that tag as the install path does before making the decision).

src/backend/aqua.rs
```diff
@@ -545,6 +545,92 @@ impl Backend for AquaBackend {
...
+        }
+
+        // Skip when any file uses `link` — the install-time symlink-target
+        // resolution is non-trivial (template-evaluated per platform) and we
+        // don't want to risk producing the wrong shim name. Falling back is
+        // safe: the user gets current behavior.
+        if pkg.files.iter().any(|f| f.link.is_some()) {
```
risu729, line 577: This drops packages where `link` is still fully deterministic. `file_link_for_version` already evaluates `f.link(...)` and uses it as the destination basename, so a package like `oven-sh/bun` has static `bun`/`bunx` names but no lazy shims because one file has `link: bunx`. Could this derive names from `srcs_for_platform(...).dst.file_name()` (or only fall back when `link` fails to render) instead of rejecting the whole package?

src/shims.rs
```diff
@@ -490,19 +492,94 @@ async fn get_desired_shims(
...
+            for shim in names {
+                // Skip if another (installed) tool already claims this shim.
+                if let Some(owner) = shim_owner.get(&shim)
+                    && owner != &tool_id
+                {
+                    trace!("skipping lazy shim {shim} for {tool_id} (owned by installed {owner})");
+                    continue;
+                }
+                // Skip if multiple configured aqua tools would advertise this name.
+                if let Some(claimants) = lazy_candidates.get(&shim) {
```
risu729, line 555: Skipping all lazy candidates for a shared shim name prevents `mise reshim` from creating the dispatch shim at all. Installed tools with the same executable do not behave that way; `get_desired_shims` only needs one shim name and `Toolset::which` resolves the owner by tool order/overrides at execution time. For lazy-only conflicts, this should still emit the shim and let the missing-bin resolver apply the same ordering, otherwise two configured aqua tools that both expose `foo` make `foo` unavailable.

src/toolset/toolset_install.rs
```diff
@@ -518,6 +519,53 @@ impl Toolset {
...
+        // auto-install entirely — same determinism rule as `get_desired_shims`.
+        //
+        // Name matching mirrors the platform transformations done at shim
+        // generation time (`shim_names_for_platform`):
+        //   - macOS: APFS is case-insensitive and shim files are lowercased,
+        //     so a shim invocation may reach us with a case that differs from
+        //     `pkg.files[*].name` in the registry.
+        //   - Windows: shim names get `.exe`/`.cmd`/etc. appended; the bin
+        //     name from the shim entry point may carry that extension.
+        if plugins.is_empty() {
```
risu729, line 535: This gate means the lazy lookup is skipped as soon as any installed backend already provides `bin_name`. In shim execution that installed provider wins even earlier via `ts.which(...)`, so a local, configured-but-missing aqua tool that should take precedence can never lazy-install if some lower-priority installed tool has the same executable. That changes the current one-shim/ordered-resolution behavior for duplicate binary names. Can we resolve the configured lazy candidates in the same toolset/override order instead of using "installed provider exists" as a hard stop?

settings.toml
```diff
@@ -578,6 +578,31 @@ right, try disabling it if you can.
...
+[experimental_lazy_shims]
+description = "[experimental] Generate shims for aqua-backed tools before they are installed and auto-install on first invocation."
+docs = """
+[experimental] When set together with `experimental = true`, mise will generate shims for
+aqua-backed tools that are configured but not yet installed, when the executable names
+can be determined unambiguously from the aqua registry metadata.
+
+When such a shim is executed, the underlying tool is installed automatically and then
+the requested binary is run. The same code path is used by `mise activate`'s
+`command_not_found_handler`, so users without shims on PATH are also covered.
```
risu729, line 590: This adds the setting reference, but `docs/troubleshooting.md` still says `not_found_auto_install` cannot handle never-installed tools. With this feature enabled that is no longer universally true for aqua static names, so users reading the docs will get conflicting guidance. Could this PR update that limitation to mention the experimental aqua lazy-shim path and its caveats?

risu729 (changes requested):
I used Codex to generate review comments, but some are my own findings, and I confirmed they're all valid. Could you have a look when you have a time?

jdx:
i like this feature idea but unfortunately I think it's going to cause a lot of regressions and I don't have time right now to deal with that

yumafuu:
@jdx 
Do you have any long-term plans around supporting lazy install workflow in mise?

I completely understand that the regression surface is too large right now — I’m mainly curious about the long-term architecture / design direction.

jdx:
no, i'm not sure if it's even possible

yumafuu:
Understood.

@jdx @risu729  Thanks for taking the time to review and discuss this.

PR #9759, opened 2026-05-09
fix(aqua): resolve bin paths for prefixed v tags

risu729:
## Summary
- Keep Aqua `list_bin_paths()` local-only while allowing file path templates to try both canonical and prefixed-`v` tag-shaped versions.
- Use the already-resolved install version when creating Aqua file links during install.
- Add unit coverage for version-prefix packages whose actual tags include an extra `v` after the prefix, such as `tool-v1.2.3` for canonical `1.2.3`.

## Why
PR #5562 aligned `list_bin_paths()` with install-time version resolution by sharing the full resolver, but it was reverted in #5574 after the hyperfine check showed a serious slowdown.

The slowdown came from putting remote resolution on the PATH/bin-path path: the shared resolver could fetch GitHub tags/releases, probe URLs/HEADs, and build fallback errors even when `list_bin_paths()` only needed local installed file paths.

This keeps the fix local-only. `list_bin_paths()` still reads registry metadata and installed files through the existing cache, but does not call `get_version_tags()`, `get_url()`, `github_release_*`, or `HTTP.head()`. Install already has the resolved tag/version, so it passes that directly into file-link creation instead of recomputing from the canonical version.

## Tests
- `/home/risu/.cargo/bin/cargo test -p mise test_version_candidates`
- `/home/risu/.cargo/bin/cargo test -p mise test_srcs_`
- `/home/risu/.cargo/bin/cargo test -p mise backend::aqua::tests`
- `git diff --check`


risu729:
CI status after 0958926ba:

The failing checks look unrelated to this Aqua bin-path change. The e2e failures across Linux shards and Windows e2e are all caused by GitHub API 403/rate-limit errors while installing external GitHub/Aqua-backed tools, for example `aqua:mozilla/sccache`, `aqua:jqlang/jq`, `github:docker/compose`, `github:ip7z/7zip`, and `aqua:tree-sitter/tree-sitter`.

The benchmark job also failed after the benchmark table, while trying to install `aqua:sharkdp/hyperfine@latest`, with a GitHub API 403. The benchmark measurements themselves did not show a large regression: cached `bin-paths` was 60ms baseline vs 62ms on this branch, and the other cached commands were within about 0-3%.

Build, lint, unit, autofix, registry, Socket, and release checks passed.

PR #9760, opened 2026-05-09
fix(exec): demote parent mise paths for nested exec

ikma-ricon:
## Summary

Nested `mise exec`/`mise env` calls inside an existing `mise run` task can inherit the outer task's direct install path before shims. `PathEnv::from_iter()` treats everything before shims as user/system `pre` paths, so the outer task's tool version can keep winning over the nested config's resolved tool version.

This reclassifies inherited mise-managed install paths that `PathEnv` places in `pre`:

- build the inherited PATH through `PathEnv` first, preserving its existing shims/no-shims/`activate_aggressive` semantics
- remove mise-managed install paths from `pre`
- add the current/nested user/tool paths first
- re-add the inherited mise install paths after them, but still before shims, so they remain direct fallbacks

That makes the nested config's selected tool win without dropping the parent task's tool path from the child environment.

Safety boundary: this only reorders paths under mise install roots that `PathEnv` would otherwise classify as pre-shims entries. Non-mise pre-PATH entries, no-shims PATHs, and `activate_aggressive` keep their existing inherited ordering.

Discussion: https://github.com/jdx/mise/discussions/9754

## Validation

- `cargo fmt --check`
- `cargo build --bin mise`
- `cargo test path_env`
- `mise run lint`
- `./e2e/run_test cli/test_exec_nested_mise_run_path`
- `./e2e/run_test cli/test_exec_venv_path_order`
- `./e2e/run_test cli/test_exec_wrapper_recursion_with_shims`
- `./e2e/run_test env/test_path_reorder_after_activate`
- Original discussion repro with `target/debug/mise` first in `PATH`:
  - `usage@3.0.0` task probing `usage@3.2.1` now resolves nested `mise exec -- command -v usage` to `usage@3.2.1`
  - `shiv:sessions@0.4.0` task probing `shiv:sessions@0.4.1` now resolves nested `mise exec -- command -v sessions` to `shiv:sessions@0.4.1`

### DIFF
--- a/e2e/cli/test_exec_nested_mise_run_path
+++ b/e2e/cli/test_exec_nested_mise_run_path
@@ -1,0 +1,120 @@
+#!/usr/bin/env bash
+
+# Regression test: nested `mise exec` inside a `mise run` task should put the
+# nested target tool before the outer task's mise-managed tool path, while still
+# preserving the outer path as a fallback.
+
+mise install dummy@1.0.0 dummy@2.0.0
+
+old_dir="$HOME/old"
+new_dir="$HOME/new"
+mkdir -p "$old_dir/.mise/tasks" "$new_dir"
+
+cat >"$old_dir/mise.toml" <<'EOF'
+[tools]
+dummy = "1.0.0"
+EOF
+
+cat >"$new_dir/mise.toml" <<'EOF'
+[tools]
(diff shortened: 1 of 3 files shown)


jdx:
this is not a good idea, sorry

PR #9762, opened 2026-05-10
registry: use aqua backend for npm

risu729:
## Summary
- Reopens https://github.com/jdx/mise/pull/8185 with only the `registry/npm.toml` change.
- Prefer `aqua:npm/cli` while keeping `npm:npm` as the fallback backend.
- Add the `node` dependency required to run the npm JavaScript entrypoint.
- Keep the registry test as `npm --version`; it should not check `node --version`.
- Update the backend-arg unit test now that bare `npm` resolves through the registry to `aqua:npm/cli`.
- Keep explicit coverage for `npm:npm` as the npm backend form.
- Clarify why the npm backend still waits for node when bootstrapping `npm:npm`.

## Notes
- The aqua registry entry has not merged yet: https://github.com/aquaproj/aqua-registry/pull/48803
- This relies on recent mise aqua file-link / `hard: true` support from https://github.com/jdx/mise/pull/9610

## Testing
- `cargo test test_bare_`
- `cargo test test_get_dependencies_for_npm_itself`
- `cargo fmt --check`
- `git diff --check`


risu729:
CI note: the remaining failures are from the stacked #9755 registry change, not from this follow-up's test/comment changes.

- `test-tool-0` / `registry-ci` fail while installing `npm` because bare `npm` now resolves to `aqua:npm/cli`, but the checked-in aqua registry used by CI does not yet contain `npm/cli`: `registry not available: no aqua-registry found for npm/cli`.
- `e2e-3` fails in `registry/test_overrides` for the same reason while installing `npm@11.7.0`.
- `e2e-4` fails in `registry/test_overrides_shims` for the same reason while installing `npm@10.9.0`.
- `test-ci` is the aggregate failure from those e2e jobs.

The backend-arg unit expectation that failed on #9755 now passes here, and the npm backend dependency test still passes locally. The e2e override tests should remain bare `npm` because they verify the registry-managed npm still overrides node's bundled npm after #9755.

PR #9770, opened 2026-05-10
fix(backend): retry ubi installs without v prefix

risu729:
## Summary
- Preserve UBI install failures as retryable errors for `try_with_v_prefix`, so `vX` can fall back to raw `X` when the release tag is missing or has no usable asset.
- Keep direct URL and lockfile URL UBI installs terminal instead of applying tag fallback to them.
- Add an e2e regression for `ubi:metalbear-co/mirrord@3.209.1`, which is published under the raw tag `3.209.1`.

## Investigation
- `aqua` CLI installs `metalbear-co/mirrord@3.209.1`, `3.209.2`, and `3.171.0` successfully.
- mise's aqua backend installs the same mirrord versions on the reported 2026.5.0 build and current main, including `latest` resolved with `--before 2026-05-05` to match the report window.
- The exact `releases/tags/v3.209.1` failure comes from the deprecated UBI path. UBI returns an `anyhow::Error`, and mise was flattening it into an `eyre` string before the shared prefix fallback could classify it as retryable.
- This also covers releases where `vX` exists but has no matching platform asset, such as mirrord `v3.171.0` on Linux, while raw `3.171.0` is installable.

## Tests
- `mise run format`
- `git diff --check`
- `mise run test:e2e e2e/backend/test_ubi`
- Manual: `target/debug/mise x ubi:metalbear-co/mirrord@3.171.0 -- mirrord --version`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5679,6 +5679,7 @@ version = "2026.5.5"
 dependencies = [
  "age",
  "aho-corasick",
+ "anyhow",
  "aqua-registry",
  "async-backtrace",
  "async-trait",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,6 +62,7 @@ opt-level = 3
 [dependencies]
 age = { version = "0.11", features = ["ssh"] }
 aho-corasick = "1"
+anyhow = "1"
 async-backtrace = "0.2"
 async-trait = "0.1"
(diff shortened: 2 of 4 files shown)


risu729:
mirrord is not using UBI backend. We must not fix deprecated, unrelated backend.

PR #9779, opened 2026-05-11
fix(upgrade): remove completed progress jobs to prevent duplicate output

jdx:
## Summary

Fixes #9774 — `mise upgrade` (and its alias `mise up`) was printing the installed-tools block twice when it also needed to uninstall an old version.

## Root cause

`MultiProgressReport` is a process-global singleton on top of clx's global `JOBS` registry.

1. [`install_all_versions`](https://github.com/jdx/mise/blob/main/src/toolset/toolset_install.rs) calls `mpr.init_footer(...)` which pushes a header job + per-tool jobs into clx's `JOBS`. They run to completion and `mpr.footer_finish()` calls `progress::stop()`, which renders the final state and resets `LINES = 0` but does **not** remove jobs from `JOBS` (the default `ProgressJobDoneBehavior` is `Keep`).
2. [`upgrade.rs`](https://github.com/jdx/mise/blob/main/src/cli/upgrade.rs) then calls `mpr.add("uninstall ...")`. clx's background thread restarts, sees four jobs in `JOBS` (header + uv + python + uninstall), and renders them all. Because `LINES = 0`, the new render is appended *below* the previously-printed final state — producing the duplicate block reported.

## Fix

Track every job created via `add_with_options` in `MultiProgressReport`, and in `footer_finish` call `ProgressJob::remove()` on each (plus the header) after `progress::stop()`. Subsequent `mpr.add(...)` calls then start from a clean `JOBS` list and render only the new job.

## Test plan
- [x] `cargo build` clean
- [x] `cargo clippy --bin mise -- -D warnings` clean
- [ ] Manual: run `mise upgrade` against a tool that needs both an install and an uninstall — expect a single installed-tools block followed by the uninstall line
- [ ] Manual: confirm normal `mise install` and `mise upgrade` (no uninstalls) still render correctly

The bug only manifests when stderr is a real TTY (`use_progress_ui = true`), so the e2e harness can't easily assert on it.

---

> [!NOTE]
> **Low Risk**
> Low risk: changes are limited to progress UI lifecycle, clearing clx’s global job registry after completion/stop to prevent duplicated output; main impact is on terminal rendering behavior.
> 
> **Overview**
> Prevents duplicated progress output across sequential operations by explicitly clearing clx’s global progress job list when a progress session ends.
> 
> `MultiProgressReport` now centralizes cleanup in `reset_jobs()`, called from both `footer_finish()` and `stop()`, which clears header state, resets counters, and calls `progress::clear_jobs()` so subsequent `mpr.add(...)` starts from a clean slate. Also bumps the `clx` dependency in `Cargo.lock` to `2.1.0`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 80ee4ef866ee54dcb15cded16f4d175652441e09. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1541,9 +1541,9 @@ dependencies = [
 
 [[package]]
 name = "clx"
-version = "2.0.3"
+version = "2.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "00505ebcbbb9f5ce9e205e324db6c44518d97e9d6278f6dab8e0bff1937d2edf"
+checksum = "20daab9df9e49b0478215055d38362d5973043b7a4af836a58968760755e9a9f"
 dependencies = [
  "console",
  "nix 0.31.2",
--- a/src/ui/multi_progress_report.rs
+++ b/src/ui/multi_progress_report.rs
@@ -1,6 +1,6 @@
 use std::sync::{Arc, Mutex};
 
-use clx::progress::{self, ProgressJobBuilder, ProgressOutput};
+use clx::progress::{self, ProgressJob, ProgressJobBuilder, ProgressOutput};
 
 use crate::cli::version::VERSION_PLAIN;
 use crate::config::Settings;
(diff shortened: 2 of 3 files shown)


jdx:
Upstream alternative open at jdx/clx#94 — adds a small `clear_jobs()` primitive so this fix doesn't need a parallel tracking Vec on the mise side. Plan: once that lands and clx is released, I'll bump clx here and simplify this PR to call `progress::clear_jobs()` from the new `reset_jobs()` helper. Leaving this PR as-is in the meantime as the immediate fix.

jdx:
Updated to use `clx::progress::clear_jobs()` now that [clx#94](https://github.com/jdx/clx/pull/94) has shipped in [clx v2.1.0](https://crates.io/crates/clx/2.1.0). Dropped the mise-side `tracked_jobs` Vec and the `pub(crate)` widening of `ProgressReport.job` — both were only needed because clx didn't have a way to drain its global JOBS list externally.

As a bonus, the clx fix also resets the `STOPPING` flag, so a fresh session after `stop()` actually animates again (previously only the final terminal-state flush rendered, via the synchronous `refresh_once()` path that bypasses STOPPING). Doesn't change behavior here, but the "start a new session" pattern is now sound for anything that wants spinners or incremental updates after a stop.

PR #9791, opened 2026-05-11
fix: github oauth device flow paths

jasisk:
Mentioned in chat.

Github device flow oauth paths are a little off.

[Documentation](https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps#device-flow).


pdecat:
Can confirm this fixes the issue for me too.

Without the change, I get:

```
# mise token github --oauth
mise ERROR HTTP status client error (422 Unprocessable Entity) for url (https://github.com/login/oauth/device/code)
```

jasisk:
Sorry, took me a minute to see the error in the logs. Github on the web kept putting me at the bottom where the permissions errors are—only learned about the real reason for the e2e failure when checking the action run log from my phone.

PR #9795, opened 2026-05-12
fix(config): reject ambiguous env directive ordering

yordis:
- `[env]` ordering is intentional, so configs should not silently change directive evaluation semantics to make one TOML edge case work.
- TOML dotted-key grouping can make mixed directive/env ordering ambiguous, and users need a clear failure instead of surprising secret or virtualenv behavior.


jdx:
this is exactly the behavior we don't want, [env] is supposed to go in order by design

PR #9807, opened 2026-05-12
registry: alias dotnet-core to dotnet

risu729:
## Summary

`dotnet-core` is a legacy shorthand for the .NET SDK. This PR aliases it to the existing `dotnet` registry tool instead of adding another plugin backend to `registry/dotnet-core.toml`.

Changes:

- add `dotnet-core` as an alias of `dotnet`
- remove the separate `registry/dotnet-core.toml` entry
- canonicalize `dotnet-core` to `dotnet` in backend parsing, matching the existing `nodejs -> node` and `golang -> go` compatibility aliases

## Why this is an alias, not a runtime option

The old `mise-plugins/mise-dotnet-core` asdf plugin installs SDKs, not runtime-only packages:

- `bin/list-all` reads Microsoft's release metadata and returns each channel's `latest-sdk`
- `bin/install` downloads Microsoft's `dotnet-install.sh` and runs it with `--version <version>` and no `--runtime`

So `core:dotnet` already covers the behavior that `dotnet-core` users expect. I did not use `core:dotnet[runtime=dotnet]` because that option means the .NET runtime-only package (`Microsoft.NETCore.App`), while `core:dotnet` latest/prefix resolution currently resolves SDK versions. Mixing SDK version resolution with `--runtime dotnet` would not preserve the old plugin behavior and can ask the installer for runtime versions that do not exist.

## Why canonicalization is included

A separate `registry/dotnet-core.toml` pointing directly at `core:dotnet` would make two registry shorthands resolve to the same core plugin. Core plugins are singleton backend implementations keyed by their canonical tool name, so leaving `dotnet-core` as its own backend identity risks split behavior where a request uses `dotnet-core` paths/options while the core plugin reports metadata under `dotnet`.

Canonicalizing `dotnet-core` to `dotnet` avoids that edge case. It keeps the compatibility spelling available, but internally uses the same tool identity, install path, registry defaults, and metadata path as `dotnet`.

## Compatibility impact

This is intentionally a consolidation, not a second `core:dotnet` registry identity:

- `mise install dotnet-core` and `mise install dotnet` now both use the canonical `$MISE_INSTALLS_DIR/dotnet` install directory.
- Existing installs that only exist under the legacy `$MISE_INSTALLS_DIR/dotnet-core` directory will not be discovered through the `dotnet-core` shorthand after this change; users will need to reinstall with `mise install dotnet-core`/`mise install dotnet` or migrate that install manually.
- The old default backend for `dotnet-core` was `asdf:mise-plugins/mise-dotnet-core`; the new default is the built-in `core:dotnet` SDK backend.
- This keeps one .NET SDK identity in mise rather than supporting two install roots for the same SDK tool.

## Popularity

- GitHub `dotnet/runtime`: 17,883 stars, 5,442 forks, latest release `v11.0.0-preview.4.26230.115` published 2026-05-12, pushed 2026-05-14.
- GitHub `dotnet/sdk`: 3,125 stars, 1,242 forks, latest release `v8.0.421` published 2026-05-12, pushed 2026-05-14.
- This is an existing registry shorthand; this PR does not add a new tool, it aliases the legacy name to the existing `dotnet` SDK tool.

## Local verification

```sh
mise run format
git diff --check
cargo test test_unalias_backend
cargo test cli::args::backend_arg::tests::test_backend_arg -- --exact
cargo test cli::args::backend_arg::tests::test_backend_arg_pathname -- --exact
target/debug/mise registry dotnet-core
target/debug/mise registry dotnet
target/debug/mise registry --json dotnet-core
target/debug/mise registry --hide-aliased | rg '^dotnet\\s'
target/debug/mise registry | rg '^dotnet(-core)?\\s'
MISE_USE_VERSIONS_HOST=false target/debug/mise latest dotnet-core # 10.0.300
MISE_USE_VERSIONS_HOST=false target/debug/mise latest dotnet      # 10.0.300
mise run test:e2e e2e/cli/test_registry
```

## Compatibility checks

- `mise registry dotnet-core` still resolves and reports the same backends as `dotnet`.
- `mise registry --hide-aliased` hides `dotnet-core`, so the canonical registry listing remains `dotnet`.
- Existing `dotnet-core` CLI/config input is normalized to `dotnet`, avoiding a second core-plugin install identity.
- The old asdf plugin installed SDK versions, so `core:dotnet` is the correct replacement; `core:dotnet[runtime=dotnet]` would be a runtime-only install and is not equivalent.

*This PR was prepared with an AI coding assistant.*


risu729:
Greptile raised a valid compatibility concern: aliasing `dotnet-core` to `dotnet` changes the canonical install root from `/dotnet-core` to `/dotnet`, so existing legacy installs under the old directory will not be reused automatically.

I updated the PR body to document that impact explicitly. I left the implementation as a canonical alias because preserving a separate `dotnet-core` install identity would recreate the two-core-plugin-identity issue this change is meant to avoid.

PR #9819, opened 2026-05-12
perf(task): cache per-file content hashes for source_freshness_hash_contents

jdx:
## Summary

- When `task.source_freshness_hash_contents = true`, every freshness check currently re-reads and re-blake3-hashes every source file ([discussion #9802](https://github.com/jdx/mise/discussions/9802)).
- This caches each file's content hash keyed by `(size, mtime_secs, mtime_nanos)` — git's stat-info trick — in a per-task file under \`STATE/task-sources/<key>-content-cache\`. Unchanged files are skipped on subsequent runs; the file is only re-hashed when its size or mtime moves.
- The cache is rebuilt from scratch each run, so entries for files no longer in \`sources\` are pruned and the cache file stays bounded by the current source set.

## Design notes

- **Serialization**: \`rmp_serde + zlib\`, matching [src/cache.rs](src/cache.rs) and [src/env_diff.rs](src/env_diff.rs). Considered \`rkyv\` for zero-copy reads, but the bottleneck this PR addresses is blake3 hashing of file *contents* (ms–s); parsing a \`BTreeMap\` of ≤1000s of entries with rmp_serde is microseconds and not on the critical path. Adding rkyv would mean a new heavy dep and a brittle byte format inconsistent with the rest of the codebase.
- **Per-task scoping**: cache key reuses \`sources_hash_path\`'s identity hash (task + config_source + root), so any change to the task definition invalidates the cache automatically.
- **Atomic writes**: \`*.part-XXXX\` + rename, same pattern as [src/cache.rs:262](src/cache.rs:262). Two concurrent runs may lose each other's writes but cannot corrupt the cache.
- **Failure handling**: corrupt/unreadable cache → silently fall back to empty (\`unwrap_or_default\`); failed write → trace-level log. Correctness is preserved either way; only speed is at stake.
- **Cache hit predicate**: \`size + mtime_secs + mtime_nanos\`. False positives only if a file is rewritten with identical size *and* mtime preserved to the nanosecond — vanishingly rare and the same trick git uses safely.

## Test plan

- [x] New unit tests in [src/task/task_source_checker.rs](src/task/task_source_checker.rs):
  - \`content_hash_cache_reuses_unchanged_files\` — second call reuses cached hashes; mutating a file invalidates.
  - \`content_hash_cache_prunes_dropped_files\` — files removed from \`sources\` drop out of the cache.
  - \`content_hash_cache_round_trips_through_disk\` — save/load round-trips; corrupt file falls back to empty.
- [x] Existing e2e tests still pass: \`test_task_source_freshness\`, \`test_task_source_freshness_with_cwd\`, \`test_task_run_sources\`, \`test_task_run_sources_negation\`, \`test_task_dep_invalidates_sources\`.

---

> [!NOTE]
> **Medium Risk**
> Touches task invalidation/freshness logic by persisting and reusing per-file content hashes; bugs could cause tasks to incorrectly be treated as fresh or stale, though it falls back safely on cache errors.
> 
> **Overview**
> When `task.source_freshness_hash_contents` is enabled, freshness checks now **persist a per-task, per-file blake3 hash cache** (keyed by file size + mtime secs/nanos) so unchanged sources aren’t re-read/re-hashed on every run.
> 
> This introduces a stable `task_state_key` to scope state under `STATE/task-sources/`, adds msgpack+zlib load/save with atomic `*.part-*` writes (and silent fallback on corrupt cache / trace-only write failures), and expands unit tests to cover cache reuse, pruning, and disk round-tripping.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f1f039b3990f2148db409dcf5213c9ff6bf2ed1d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jdx:
Thanks @gemini-code-assist, but I'm going to pass on both of these.

The buffered \`read_to_end\` → \`from_slice\` and \`to_vec_named\` → \`write_all\` pattern is the established convention everywhere else in this codebase that does \`rmp_serde + zlib\` ([src/cache.rs:248-251](https://github.com/jdx/mise/blob/main/src/cache.rs#L248-L251), [src/env_diff.rs:136](https://github.com/jdx/mise/blob/main/src/env_diff.rs#L136), [src/hook_env.rs:363](https://github.com/jdx/mise/blob/main/src/hook_env.rs#L363)). Diverging just here adds inconsistency for a theoretical memory win that doesn't materialize at this scale — a cache entry is ~100 bytes, so even a task with 10k source files is ~1MB uncompressed. Streaming msgpack also produces less precise error messages than \`from_slice\` on a flat byte buffer.

If we ever want to switch the convention across the codebase, that should be a separate PR touching all four call sites at once.

PR #9839, opened 2026-05-13
fix(rust): reinstall incomplete rustup toolchains

risu729:
## Summary

- add an install-only backend completeness predicate that defaults to the existing cheap installed check
- make install decisions use that predicate for missing detection, dry-run exit checks, pre-install skips, and post-lock double-checks
- teach the Rust backend to treat the mise `installs/rust/<version>` marker as incomplete when the matching rustup/Cargo state is absent
- add an e2e regression that covers missing rustup toolchain state, `profile = "default"` clippy/rustfmt, explicit components, explicit targets, custom `MISE_RUSTUP_HOME`/`MISE_CARGO_HOME`, and the hook-env fast path

## Why this shape

This supersedes the action-side cache proposal in jdx/mise-action#467. In that review, @jdx pushed back on making mise-action own a Rust cache model: https://github.com/jdx/mise-action/pull/467#discussion_r3218912506

This PR avoids that model. mise-action can keep caching mise state the same way it already does; mise itself now notices, only during install decisions, that the restored Rust install marker is not enough if rustup's toolchain/components are missing. The interactive paths still use the cheap check: hook-env, PATH generation, status display, list_current_installed_versions, and normal command lookup do not perform Rust filesystem validation.

The Rust completeness check is filesystem-only and scoped to install. It first requires the existing mise install marker, then verifies the configured `MISE_CARGO_HOME`/`MISE_RUSTUP_HOME` contain the proxy bins, requested toolchain directory, `profile = "default"` rustfmt/clippy, explicitly configured components, and explicitly configured targets. If that state is missing, `mise install` reruns `rustup toolchain install` instead of reporting `all tools are installed`.

Related action failure: https://github.com/jdx/mise-action/issues/215
Supersedes: https://github.com/jdx/mise-action/pull/467

## Test plan

- [x] `cargo fmt --check`
- [x] `git diff --check`
- [x] `$(rustup which cargo) build -p mise --all-features`
- [x] `mise run test:e2e e2e/core/test_rust_install_completeness`

*This PR body was generated by an AI coding assistant.*

### DIFF
--- a/src/toolset/tool_request_set.rs
+++ b/src/toolset/tool_request_set.rs
@@ -55,6 +55,16 @@ impl ToolRequestSet {
         tools
     }
 
+    pub async fn missing_tools_for_install(&self, config: &Arc<Config>) -> Vec<&ToolRequest> {
+        let mut tools = vec![];
+        for tr in self.tools.values().flatten() {
+            if tr.is_os_supported() && !tr.is_installed_for_install(config).await {
+                tools.push(tr);
+            }
+        }
+        tools
+    }
+
     pub fn list_tools(&self) -> Vec<&Arc<BackendArg>> {
         self.tools.keys().collect()
     }
--- a/e2e/core/test_rust_install_completeness
+++ b/e2e/core/test_rust_install_completeness
@@ -1,0 +1,179 @@
+#!/usr/bin/env bash
+
+export MISE_RUSTUP_HOME="$MISE_DATA_DIR/rustup-custom"
+export MISE_CARGO_HOME="$MISE_DATA_DIR/cargo-custom"
+export MISE_OFFLINE=1
+
+VERSION="1.81.0"
+HOST="$(rustc -vV | sed -n 's/^host: //p' | head -n1)"
+TOOLCHAIN="$VERSION-$HOST"
+CALLS="$MISE_DATA_DIR/fake-rustup-calls"
+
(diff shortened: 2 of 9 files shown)


src/toolset/tool_request_set.rs
```diff
@@ -55,6 +55,16 @@ impl ToolRequestSet {
         tools
     }
 
+    pub async fn missing_tools_for_install(&self, config: &Arc<Config>) -> Vec<&ToolRequest> {
```
jdx, line 58: it's too hacky and this isn't a bug I really care that much to resolve

risu729: I'm sorry, I thought I told Codex to create a draft PR. I agree with you, it doesn't affect performance that much, but it's too much for just Rust.
What about https://github.com/jdx/mise-action/pull/474? It's still hacky as we can't simply exclude a child dir... It might be better to just document it doesn't work well with Rust so cache by yourself or force reinstall Rust.

jdx: not sure, I'm half-tempted to just deprecate rust since I don't use it and don't really think others should. Though I know some do like it.

risu729: That's fair. I use mise to manage Rust because I prefer managing everything with mise, but just because of it.
I think a config like nightly for clippy and stable for other components is not supported in `.rust-toolchain.toml` and could be a case mise can be used for, but I don't think it's a good use case. I'm not even sure if it actually works.

PR #9847, opened 2026-05-14
chore(ci): downgrade noisy hyperfine regressions

risu729:
## Summary
- capture hyperfine output in the benchmark workflow so statistical outlier warnings can be detected
- treat >10% slower measurements with hyperfine outlier warnings as inconclusive warnings instead of failing the check
- keep clean >10% slower measurements failing, and keep pipefail so hyperfine command failures still fail

## Testing
- `sed -n '71,125p' .github/workflows/hyperfine.yml | sed 's/^          //' | bash -n`\n- `actionlint .github/workflows/hyperfine.yml`\n- `git diff --check`\n\n*This PR was generated by an AI coding assistant.*

### DIFF
--- a/.github/workflows/hyperfine.yml
+++ b/.github/workflows/hyperfine.yml
@@ -70,7 +70,9 @@ jobs:
       - uses: ./.github/actions/mise-tools
       - run: |
           set -x
+          set -o pipefail
           failed=false
+          outlier_warning="Statistical outliers were detected"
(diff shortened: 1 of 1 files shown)


risu729:
Applied the Greptile feedback in 54f58ed02: the workflow now uses explicit `set -exo pipefail`, and the `::warning` annotation is emitted on stdout.

PR #9861, opened 2026-05-14
fix(aqua): respect supported_envs for latest

risu729:
## Summary
- filter aqua `latest` candidates by both asset availability and current-platform `supported_envs`
- apply the same platform-aware candidate check to the GitHub `/releases/latest` fast path and the full version-list fallback path
- keep `versioned_package_from_tag()` unfiltered so explicit pinned versions, install-time validation, and lockfile tag lookup still use the existing path

## Lockfile / shims impact
- Shims and normal install/exec resolution are current-platform operations, so `latest` now resolves to the newest aqua release that is valid for the current platform before install-time validation runs.
- Lockfile URL/checksum resolution for a concrete version still works the same way: `resolve_lock_info()` applies target-platform overrides and returns empty platform metadata when that concrete version is unsupported for a requested target.
- Important limitation: `mise lock` resolves `latest` into a single concrete `ToolVersion.version` before it fans out over target platforms. If aqua's latest supported version differs by platform, the current lock pipeline cannot automatically write different concrete versions for different platforms from one `latest` request. It will lock the version resolved on the host platform and will not pick an older supported version for another target platform. That limitation is pre-existing in the lockfile flow and is not worked around in this PR.

## Tests
- `cargo fmt --all --check`
- `git diff --check`
- `/home/risu/.cargo/bin/cargo test -j1 backend::aqua::lock_candidate_tests`

### DIFF
--- a/src/backend/aqua.rs
+++ b/src/backend/aqua.rs
@@ -262,14 +262,14 @@ impl Backend for AquaBackend {
         let target_libc = Self::target_variant_libc(&target);
         let mut versions = Vec::new();
         for (tag, created_at, prerelease) in tags_with_timestamps.into_iter().rev() {
-            let (version, versioned_pkg) = match versioned_package_from_tag(
+            let version = match latest_candidate_from_tag(
                 &pkg,
                 &tag,
                 target_os,
                 target_arch,
                 target_libc.as_deref(),
             ) {
-                Ok(Some(versioned)) => versioned,
+                Ok(Some((version, _))) => version,
                 Ok(None) => continue,
                 Err(e) => {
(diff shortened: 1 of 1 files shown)


risu729:
CI note: the benchmark check failed for an unrelated infrastructure reason. The job failed while installing aqua:sharkdp/hyperfine@latest because the GitHub API rate limit was exceeded and the release API returned 403, before benchmark comparison ran. All functional test/lint checks passed; only this benchmark infrastructure failure remains.

risu729:
It doesn't work with lockfiles so close.

PR #9913, opened 2026-05-16
feat(npm): disable npm lifecycle scripts by default

risu729:
This is a breaking change. We might not want to change the default. However, I added `--trust` flag when I added Bun support before, which I now think was a mistake. At least we should remove that.

## Summary
- Disable npm lifecycle scripts by default for npm-backed global installs by passing `--ignore-scripts=true`.
- Keep `npm_args` as the user escape hatch; user args are appended after the default, so `npm_args = "--ignore-scripts=false"` opts back into npm's script behavior.
- Stop adding Bun `--trust` automatically; explicit `bun_args = "--trust"` is still forwarded.
- Refresh npm backend docs around lifecycle behavior, ordered as aube -> pnpm -> bun -> npm, with current aube and exact pnpm CLI support notes.

## Behavior changes
- Bun installs remove mise's implicit `--trust`.
- npm installs set `--ignore-scripts=true` by default.

## npm prepare note
npm docs have historically called out `prepare` separately for git dependency installs. This PR uses npm's documented `--ignore-scripts=true` install switch for npm-backed global registry installs, rather than adding prepare-specific handling.

## Testing
- `cargo fmt --check`
- `cargo test backend::npm`
- `bash -n e2e/backend/test_npm_package_manager`
- `git diff --check`
- `CARGO_TARGET_DIR="$(cargo metadata --format-version 1 --no-deps | jq -r .target_directory)" mise run test:e2e e2e/backend/test_npm_package_manager`


risu729:
The remaining CI failure appears unrelated to this PR: the `lint` job completed render without producing a diff, then `cargo deny check` crashed with exit 139. I verified the same deny step locally with `mise x aqua:EmbarkStudios/cargo-deny -- cargo deny check`, and it passed (`advisories ok, bans ok, licenses ok, sources ok`). I attempted to rerun the failed GitHub job, but this token does not have repository admin rights to rerun checks.

PR #9918, opened 2026-05-16
fix(aqua): resolve prefixed vars with manifest opts

risu729:
Stacked on #9958. Target branch remains `main`.

## Summary
- preserve Aqua `symlink_bins` from the install manifest while current config overrides ordinary options
- resolve flat `vars.<name>` option keys in `AquaOptions::var`, matching bracket and TOML dotted-key syntax
- add regression coverage for both manifest-only `symlink_bins` and prefixed Aqua vars

## Verification
- `git diff --check fix/install-manifest-option-source...fix/aqua-symlink-bins-manifest-only`


risu729:

PR #9919, opened 2026-05-16
fix(npm): disable npm scripts in paranoid mode

risu729:
## Stack
Depends on #9913 (`fix(npm): stop implicitly trusting bun installs`). GitHub cannot set this PR's base to the fork-only PR branch, so this branch is based on #9913 and still targets `main`.

## Summary
- When the npm backend resolves to the `npm` package manager under paranoid mode, append `--ignore-scripts=true` to the npm install command.
- Append the paranoid flag after raw `npm_args`, so `MISE_PARANOID=1` wins over earlier `--ignore-scripts=false` args or `NPM_CONFIG_IGNORE_SCRIPTS=false` environment defaults.
- Let `npm.package_manager = "auto"` fall back to npm under paranoid with scripts disabled when aube is unavailable.
- Document that npm-backed paranoid installs may break packages that require lifecycle build hooks.

## Testing
- `cargo fmt --check`
- `bash -n e2e/backend/test_npm_package_manager`
- `git diff --check`
- Full `cargo test backend::npm` and `mise run test:e2e e2e/backend/test_npm_package_manager` are pending because unrelated cargo jobs are currently occupying the shared build queue.


risu729:
Superseded by #9913, which now disables npm lifecycle scripts by default and includes the npm backend docs/tests from this branch.

PR #9946, opened 2026-05-17
fix(github): asset picker handles ambiguous patterns and runtime archives

jdx:
## Summary

Two small, related asset-picker fixes that surfaced in recent discussions:

### 1. `asset_pattern` shortest-match tiebreaker

Fixes [discussion #9358 (br-arruda's comment)](https://github.com/jdx/mise/discussions/9358).

When a user-supplied `asset_pattern` matches multiple release assets, the previous code returned whichever GitHub's API listed first. Concretely, `cloudflared` with `asset_pattern = "*linux*64"` would sometimes pick `cloudflared-fips-linux-amd64` instead of `cloudflared-linux-amd64`, producing inconsistent lockfile entries across platforms.

Mirrors the auto-detection tiebreaker from #9361 (`bebab7f39`): shortest name wins, then lexicographic. The three pattern-matching call sites (GitHub, GitLab, Forgejo) now share a `pick_by_pattern` helper.

### 2. Platform-agnostic runtime archives (`.phar`, `.jar`, etc.)

Fixes [discussion #9936](https://github.com/jdx/mise/discussions/9936).

`composer` ships a single `composer.phar` and that's it — no per-platform binaries. The github backend rejected this asset (score 0, filtered by `score > 0`), so the deprecation notice telling users to migrate from `ubi:composer/composer` to `github:composer/composer` led to a broken install:

```
mise ERROR Failed to install github:composer/composer@latest:
  No matching asset found for platform linux-x64
  Available assets: composer.phar, composer.phar.asc
```

`.phar`/`.jar`/`.pyz`/`.whl`/`.gem` are language-runtime archives — they run anywhere the runtime runs. Award them the same score as a regular archive (`+10`) so they're picked when present, but still lose to platform-specific binaries when those exist.

### Why not the broader \"single-asset fallback\" from #8756

[PR #8756](https://github.com/jdx/mise/pull/8756) proposed accepting any single non-metadata asset as a fallback. It was closed because a Windows-only `foo.exe` with no platform tokens would then be auto-picked on Linux — wrong. This PR sidesteps that footgun by gating only on extensions that *are* themselves platform-agnostic signals. A regression test (`test_exe_on_non_windows_not_picked`) guards the distinction.

## Test plan

- [x] New unit tests for the asset_pattern tiebreaker (3 tests in `github::tests`)
- [x] New unit tests for platform-agnostic archive picking (3 positive + 1 regression in `asset_matcher::tests`)
- [x] All 254 backend unit tests still pass
- [x] `mise run lint` clean

---

> [!NOTE]
> **Medium Risk**
> Changes asset selection logic used for tool downloads, which can alter which release artifacts are installed across GitHub/GitLab/Forgejo targets. Risk is mitigated by added unit tests and the changes are narrowly scoped to picker heuristics.
> 
> **Overview**
> Fixes two asset-picking edge cases.
> 
> **Pattern-based selection is now deterministic**: when a user `asset_pattern` matches multiple assets, the backend now picks the *shortest name* (then lexicographic) via a shared `pick_by_pattern` helper, instead of relying on provider API ordering.
> 
> **Runtime archives are treated as valid cross-platform assets**: the auto-detection scorer now gives `.phar`, `.jar`, and `.pyz` the same format bonus as archives so single-asset releases can be installed; `.whl`/`.gem` are explicitly excluded, and new regression tests ensure platform-specific binaries still win and Windows-only `.exe` isn’t picked on non-Windows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1476ed5fa1487e148869734ac327e9b4dd44d8fd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/backend/asset_matcher.rs
+++ b/src/backend/asset_matcher.rs
@@ -338,7 +338,27 @@ impl AssetPicker {
             }
         }
 
-        if format.is_archive() { 10 } else { 0 }
+        if format.is_archive() {
+            return 10;
+        }
+
+        // Platform-agnostic runtime archives (composer.phar, foo.jar, bar.pyz)
+        // run on the language runtime, not the OS — give them the same score as
+        // a regular archive so single-asset releases like composer's
+        // `composer.phar` are picked instead of failing platform matching.
+        // See: https://github.com/jdx/mise/discussions/9936
+        //
+        // `.whl` and `.gem` are intentionally NOT in this list: both have
+        // platform-tagged variants whose tokens OS_PATTERNS doesn't reliably
+        // catch (`manylinux2014_x86_64`, `mingw32`), so granting the bonus
+        // could let a wrong-platform variant be picked. Those cases should
+        // use an explicit `asset_pattern`.
+        let lower = asset.to_lowercase();
+        if lower.ends_with(".phar") || lower.ends_with(".jar") || lower.ends_with(".pyz") {
+            return 10;
+        }
+
+        0
     }
 
     fn score_build_penalties(&self, asset: &str) -> i32 {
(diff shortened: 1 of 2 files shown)


jdx:
Addressed both review comments and rebased:

- **Greptile (.whl)**: Removed `.whl` from the platform-agnostic list. The concern is real — wheels can encode platform info in tags like `manylinux2014_x86_64`, and `OS_PATTERNS` doesn't catch the embedded `linux` in `manylinux` (no word boundary in front of it). Comment in the code explains why it's intentionally excluded. New regression test (`test_whl_does_not_get_platform_agnostic_bonus`) guards the distinction: a `.whl` with no other tokens scores 0 (filtered out), while a `.jar` with no other tokens scores 10 (picked).

- **Gemini (regex compile)**: Compile the glob → regex once in `pick_by_pattern` outside the filter closure, instead of recompiling per asset. Falls back to `contains` on compile failure, same as `matches_pattern`.

- **Rebased on latest main** to pick up #9945 (verify provenance during lock), which should resolve the `mise run render` lockfile-diff lint failure.

jdx:
Round 2 of review fixes:

- **Greptile (.gem)**: Same risk as `.whl` — Ruby gems use `mingw32` for Windows (e.g., `nokogiri-1.15.5-x86_64-mingw32.gem`), which isn't in `OS_PATTERNS`. Removed `.gem` from the platform-agnostic list. The list is now just `.phar` / `.jar` / `.pyz`, all of which have no platform-tagged variant convention. Test renamed to `test_platform_tagged_extensions_excluded_from_bonus` and now covers both `.whl` and `.gem`.

- **Cursor Bugbot (dead code)**: `matches_pattern` lost all production callers when the three call sites moved to `pick_by_pattern`. Removed `matches_pattern` (its glob→regex logic now lives only inside `pick_by_pattern`) and rewrote its unit test to exercise `pick_by_pattern` instead.

PR #9915, opened 2026-05-16
test(http): cover current config over install manifest opts

risu729:
Stacked on #9958. Target branch remains `main`.

## Summary
- test-only follow-up on #9958
- add an HTTP `ls-remote` regression for discussion #8880
- verify current config options override stale install-manifest options
- harden the regression by probing server readiness and creating the installs directory

## Verification
- `git diff --check fix/install-manifest-option-source...fix/http-current-config-manifest-options`

PR #9958, opened 2026-05-18
fix(config): track install manifest option source

risu729:
## Summary
- track install manifest tool options as their own source layer
- keep install manifest options lower precedence than config and inline backend args
- avoid serializing install manifest options as inline backend args
- add shared source-attribution coverage for config-only and manifest-only keys

Extracted from the shared install-manifest option source work in #9915, #9917, and #9918.

## Tests
- `git diff --check upstream/main...fix/install-manifest-option-source`


risu729:
CI note: `lint`/`test-ci` are failing because `cargo deny check` now detects `RUSTSEC-2026-0145` for `astral-tokio-tar 0.6.1` via `rattler_package_streaming`. This appears unrelated to this PR: the latest `main` `test` workflow is failing with the same advisory, and recent test workflow runs on other branches show the same failure. All other checks on this PR passed after `0f903d79a`.

PR #9970, opened 2026-05-18
feat(core): deprecate default package files

jdx:
## Summary
- deprecate node/python/ruby/go default package file settings with 2026.11.0 warning and 2027.11.0 removal metadata
- warn when non-empty default package files are consumed during installs
- document replacements with package-manager backends for CLIs and tool-level postinstall hooks for per-runtime installs

## Context
- Recent discussion: https://github.com/jdx/mise/discussions/9951

## Validation
- mise run render:schema
- cargo fmt --check
- cargo check
- mise run docs:build

---

> [!NOTE]
> **Medium Risk**
> Changes core runtime install flows to emit new deprecation warnings when legacy default package files are present, which could affect user output/automation. Otherwise it is mostly documentation and schema/setting metadata updates.
> 
> **Overview**
> Adds a *planned deprecation* for Go/Node/Python/Ruby default package files, including new docs guidance to use package-manager backends (e.g. `npm:`, `pipx:`, `gem:`, `go:`) or per-tool `postinstall` hooks instead.
> 
> Marks the related settings/schema fields as deprecated with `2026.11.0` warn and `2027.11.0` removal metadata, and updates the core plugins to detect non-empty default package files and emit a single deprecation warning when they are consumed during installs (with shared parsing/helpers in `Settings`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8ca479b2eddcef37109998d1ca1e38be3adebbf3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/lang/go.md
+++ b/docs/lang/go.md
@@ -34,8 +34,29 @@ See [idiomatic version files](/configuration.html#idiomatic-version-files)
 
 ## Default packages
 
+::: warning Planned deprecation
+Default package files are deprecated. They are still supported for now, but mise will start warning
+in `2026.11.0` and support will be removed in `2027.11.0`.
+
+For Go CLIs, install the tool directly with the `go:` backend:
+
+```toml
+[tools]
+"go:github.com/jesseduffield/lazygit" = "latest"
+```
+
+For packages that really should be installed into every Go version, use a tool-level `postinstall`
+hook:
+
+```toml
+[tools]
+go = { version = "1.25", postinstall = "go install github.com/daixiang0/gci@latest" }
+```
+
+:::
+
 mise can automatically install a default set of packages right after installing a new go version.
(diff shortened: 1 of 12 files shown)


dhanak:
I tried using the postinstall hook to install some python packages, but since the python and pip was just installed, it doesn't find the correct version of pip on the PATH. Hence, it invokes pip from another python installation, or from the system. I also tried writing `mise x python -- pip`, but that didn't help either, I'm guessing because the new install wasn't registered with mise yet.

What is the proper solution to make mise find the pip binary that it just installed?

dhanak:
Answering my previous comment: this seems to be fixed in mise 2026.5.15

PR #9977, opened 2026-05-19
fix(doctor): honor http timeout for version checks

risu729:
## Summary
- use the configured `http_timeout` for the `mise doctor`/version endpoint check instead of a hardcoded 3s timeout
- remove the dedicated version-check client so timeout hints point users at `http_timeout` / `MISE_HTTP_TIMEOUT`

## History / Rationale
- `edb8a32ea` (`feat(timeout): show duration, URL, and config hint on timeouts; increase fetch timeout default to 10s`) introduced `ClientKind::VersionCheck => ("version_check_timeout", "")` alongside the hardcoded 3s `HTTP_VERSION_CHECK` client.
- That empty env var case was intentional: there was no real `version_check_timeout` setting/env var, so the timeout formatter avoided printing a misleading "change with ..." hint for version checks.
- This PR removes that special version-check client path. The version check now uses the regular `HTTP` client, so the former empty-env-var branch is no longer reachable and timeout messages correctly point to the real `http_timeout` / `MISE_HTTP_TIMEOUT` controls.

## Discussion
- Addresses https://github.com/jdx/mise/discussions/9934#discussioncomment-16959794

## Tests
- `git diff --check`
- `MISE_TRUSTED_CONFIG_PATHS=$PWD mise run build`
- Removed the e2e test because this PR is small and there is no separate way to configure only the version-check timeout; forcing a slow timeout path in test coverage could be troublesome for low-speed internet users.


Patterner:
this is by far one of the best reaction a bug report i've seen

risu729:
CI is failing in `lint` because `cargo deny check` now reports `RUSTSEC-2026-0145` for `astral-tokio-tar 0.6.1` through `rattler_package_streaming`.

This appears unrelated to this PR: the latest `main` test workflow (`26093363275`) fails on the same advisory in the same `cargo deny check` step. The code/test changes here only remove the version-check e2e coverage and route the existing version check through the shared HTTP client, so I am leaving the dependency advisory out of scope for this PR.

PR #9984, opened 2026-05-19
fix(conda): include channel in lock identity

risu729:
## Summary
- include Conda channel values in lockfile option identity
- reuse typed channel parsing for lock options
- add regression coverage for custom channel lock options

## Why this belongs in lock identity
The Conda channel is part of the package universe being solved. The same tool name and version can resolve to different package builds, dependency URLs, and checksums on `conda-forge`, `bioconda`, or a private channel. If the channel is not part of the lock identity, `mise lock` can reuse solve metadata from one channel while the config asks for another, which breaks reproducibility and can install the wrong artifact set.

## Verification
- `cargo fmt --check`
- `cargo test conda_lockfile_options_include_channel`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved conda channel handling: Channel configurations specified in tool requests are now properly resolved and included in lockfile options during generation.

### DIFF
--- a/src/backend/conda.rs
+++ b/src/backend/conda.rs
@@ -75,6 +75,10 @@ impl<'a> CondaOptions<'a> {
         Channel::from_str(&name, &config)
             .map_err(|e| eyre::eyre!("invalid conda channel '{}': {}", name, e))
     }
+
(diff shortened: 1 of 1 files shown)


risu729:

PR #9985, opened 2026-05-19
fix(github): include target artifact options in lock identity

risu729:
## Summary
- include GitHub/GitLab/Forgejo target artifact selector inputs in lock identity
- resolve lock options per target platform instead of host-only strings
- add regression coverage for `api_url`, `version_prefix`, platform `asset_pattern`, direct `url`, and `no_app`

## Classification
Stale-lock invalidation fix.

These selector inputs are generally reflected in the resolved lockfile output (`platforms.<platform>.url`, `url_api`, checksum, and provenance metadata). They are not needed to perform the download once a matching lock entry has been selected.

They are still needed in the current lock identity because mise selects lock entries by exact `(version, options)` before trusting the locked URL. If the selector changes but the options key does not, a lock entry produced for an older selector can be reused and install the stale locked artifact for the same version.

## Verification
- `cargo fmt --check`
- `cargo test test_lockfile_options_use_target_artifact_inputs`

### DIFF
--- a/src/backend/github.rs
+++ b/src/backend/github.rs
@@ -104,12 +104,22 @@ impl<'a> GitBackendOptions<'a> {
             })
     }
 
-    fn lockfile_options(&self) -> BTreeMap<String, String> {
+    fn lockfile_options(&self, target: &PlatformTarget) -> BTreeMap<String, String> {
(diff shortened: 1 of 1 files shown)


risu729:

PR #9986, opened 2026-05-19
fix(spm): classify artifact bundle lock options

risu729:
## Summary
- include SPM provider/API/artifact bundle inputs in lock identity
- keep `filter_bins` out of lock identity while preserving it as an install/layout option
- wire SPM install-time option keys into the shared manifest precedence filter

Stacked on #9959. Target branch remains `main`.

## Verification
- `cargo fmt --check`
- `cargo test options_include`


risu729:

PR #9987, opened 2026-05-19
fix(dotnet): support runtime tool option

risu729:
## Summary
- add direct .NET backend support for the `runtime` tool option
- pass `--runtime` to `dotnet tool install`
- include runtime in lockfile option identity

## Why this belongs in lock identity
The .NET `runtime` option changes the runtime-specific asset restored by `dotnet tool install` for the same package and version. A lock entry for `linux-x64` is not equivalent to one for `linux-musl-x64` or another runtime identifier, because the installed tool may bind to a different target runtime. Including `runtime` prevents `mise lock` and `mise install --locked` from reusing metadata for a different runtime-specific install.

## Verification
- `cargo fmt --check`
- `cargo test dotnet_`

### DIFF
--- a/src/backend/dotnet.rs
+++ b/src/backend/dotnet.rs
@@ -1,13 +1,15 @@
+use std::collections::BTreeMap;
 use std::sync::Arc;
 
 use crate::backend::VersionInfo;
 use crate::backend::backend_type::BackendType;
 use crate::backend::options::BackendOptions;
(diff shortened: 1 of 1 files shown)


risu729:

risu729:
Sorry for the noise. This is for dotnet core plugin, not for the backned.

PR #9988, opened 2026-05-19
fix(rust): include toolchain install options in lock identity

risu729:
## Summary
- include rustup `profile`, `components`, and `targets` in lock identity
- read those values from both tool options and `rust-toolchain.toml`
- sort component and target lists before recording them for stable identity

## Classification
Actual lock/install identity fix.

Rust installs are not just a channel/version. `profile`, `components`, and `targets` change the toolchain contents that rustup installs, and there is no locked artifact URL in `mise.lock` that fully represents those extra installed components or cross targets. A lock entry for `stable` with only the default profile is not equivalent to one that also installs `clippy`, `rustfmt`, or additional targets.

## Verification
- `cargo fmt --check`
- `cargo test rust_lockfile_options`

## Summary by CodeRabbit

* **New Features**
  * Added lockfile option mappings support for Rust toolchain installations, enabling configuration of profile, components, and targets from toolchain settings or backend options.

* **Tests**
  * Extended test coverage for lockfile options handling in both standard and idiomatic configuration modes.

### DIFF
--- a/src/plugins/core/rust.rs
+++ b/src/plugins/core/rust.rs
@@ -73,11 +73,15 @@ impl<'a> RustOptions<'a> {
         if let Some(components) = components
             && !components.is_empty()
         {
+            let mut components = components;
+            components.sort();
             opts.insert("components".into(), components.join(","));
         }
         if let Some(targets) = targets
             && !targets.is_empty()
         {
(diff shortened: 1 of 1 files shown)


risu729:

PR #9990, opened 2026-05-19
fix(python): include sysconfig patch option in lock identity

risu729:
## Summary
- include non-default `patch_sysconfig = false` in Python lock identity
- keep `virtualenv` out of lock identity as an activation-only option

Stacked on #9958. Target branch remains `main`.

## Verification
- `cargo fmt --check`
- `cargo test python_lockfile_options_include_patch_sysconfig_but_not_virtualenv`


risu729:
CI note: `lint`/`test-ci` are failing because `cargo deny check` now detects `RUSTSEC-2026-0145` for `astral-tokio-tar 0.6.1` via `rattler_package_streaming`; the current `main` branch is failing the same advisory, so that part appears unrelated.

The additional `e2e-5` failure is `lockfile/test_lockfile_python_skip_provenance_verify`, which fails while installing `python@3.13.5` because GitHub artifact attestation verification gets `401 Unauthorized` / `Bad credentials`. The same e2e-5 failure was already present on this branch in the previous run before the latest review-fix commit, and `main`'s e2e-5 passed in run 26093363275, so this looks like a CI credentials/environment issue rather than a regression from the lockfile option test change.

PR #10019, opened 2026-05-21
fix(backend): strip system shims dir from dependency_env PATH

andrewjamesbrown:
## Summary

#8475 fixed a fork-bomb where backend version-resolution subprocesses re-entered mise through a shim, by stripping `dirs::SHIMS` (the **user** shims dir) from `dependency_env`'s PATH.

Devcontainer / Docker setups that use `mise install --system` end up with a **second** shims dir at `$MISE_SYSTEM_DATA_DIR/shims` (e.g. `/usr/local/share/mise/shims`). That directory is conventionally on PATH alongside the user shims dir — but `dependency_env` only filters the user one, so the system dir's mise-symlink still wins PATH lookup. The same recursion that #8475 closed for the user dir re-opens through the system dir.

## Concrete repro

A workspace `mise.toml` pinning `"npm:yarn" = "1"` together with an uninstalled `node` (in a wsdev/devcontainer image with `mise install --system node@lts npm:@withgraphite/graphite-cli` baked in) reliably forks indefinitely.

Process tree captured live before the OOMKill:

```
mise install (PID 829)
└─ npm view yarn versions time --json (841)
   └─ npm view yarn versions time --json (851)
      └─ npm view yarn versions time --json (861)
         └─ ... [hundreds of descendants until cgroup OOM]
```

Each `npm view` is the child of the previous — classic shim-re-entry signature. Chain: npm backend's `_list_remote_versions` calls `npm view`, PATH lookup hits `/usr/local/share/mise/shims/npm` (the system shim), mise re-enters as `npm`, sees the same uninstalled `npm:yarn`, auto-installs via the npm backend, calls `npm view` again — loop.

The exact same path described in #8475's root-cause section, just landing in the *other* shims dir.

## Fix

Apply the same filter to the system shims dir in `dependency_env`, mirroring the dual-dir guard that `shims::which_shim` already adopts after #8816:

```rust
let sys_shims = env::MISE_SYSTEM_DATA_DIR.join("shims");
let filtered: Vec<_> = paths
    .into_iter()
    .filter(|p| {
        let resolved = file::replace_path(p);
        !file::paths_eq(&resolved, &dirs::SHIMS)
            && !file::paths_eq(&resolved, &sys_shims)
    })
    .collect();
```

`paths_eq` keeps the case-insensitive semantics #8475 documented for macOS/Windows.

## Regression test

`e2e/backend/test_npm_shim_recursion_system_dir` — same shape as `test_go_shim_recursion`, but targets `$MISE_SYSTEM_DATA_DIR/shims` instead of the user shims dir. Plants a fake `npm` shim there, configures `npm:left-pad` + `node = "99.0.0"` (uninstalled), clears the version cache, and asserts `mise ls --json -c` completes within 15s. Before this fix it would fork-bomb out via the system shims dir; after it falls through cleanly.

## Test plan

- [x] `cargo check --bin mise` — clean
- [x] `cargo clippy --bin mise -- -D warnings` — clean
- [x] `cargo test --bin mise backend::` — 269/269 pass
- [ ] CI: full e2e suite including the new `test_npm_shim_recursion_system_dir`

## Related

- #8475 — the original `dependency_env` fork-bomb fix (covered user shims dir only)
- #8816 — fixed the dual-dir case for `which_shim`; this PR is the analogous patch for `dependency_env`
- #8189, #8276, #8560, #8802 — prior shim-recursion fixes in the same class

### DIFF
--- a/e2e/backend/test_npm_shim_recursion_system_dir
+++ b/e2e/backend/test_npm_shim_recursion_system_dir
@@ -31,11 +31,14 @@ export PATH="$MISE_SYSTEM_DATA_DIR/shims:$PATH"
 
 # Configure an npm:* backend tool + node uninstalled, so the npm backend's
 # _list_remote_versions runs `npm view ... --json` and the only `npm` on
-# PATH is the system shim we just planted.
+# PATH is the system shim we just planted. Use a real-but-uninstalled node
+# version (matches the pattern in test_go_shim_recursion using
+# `go = "1.23.3"`) — picking a non-existent version like `99.0.0` would risk
+# `mise ls` failing on version resolution and masking a fork-bomb regression.
 cat >>mise.toml <<'EOF'
 [tools]
 "npm:left-pad" = "latest"
-node = "99.0.0"
+node = "22.0.0"
 EOF
 mise trust --yes
 
@@ -50,8 +53,11 @@ mise cache clear
 # With the fix: both shim dirs are stripped, no `npm` is reachable in the
 # dependency PATH, `npm view` fails cleanly, mise reports the resolution
 # failure and exits.
-output="$(run_with_timeout 15 mise ls --json -c)" || {
-  echo "ERROR: mise ls timed out or failed (likely infinite recursion fork bomb via system shims dir)"
-  exit 1
-}
+#
+# `|| true` keeps the test focused on "no fork bomb" rather than "ls returns
+# zero" — `mise ls` may exit non-zero when npm:left-pad's version resolution
+# fails (the expected failure when shims are stripped), and we don't want
+# that to be confused with a recursion regression. The `assert_contains`
+# on the actual output still verifies mise made it past resolution.
+output="$(run_with_timeout 15 mise ls --json -c || true)"
 assert_contains "echo '$output'" '"node"'
--- a/e2e/backend/test_npm_shim_recursion_system_dir
+++ b/e2e/backend/test_npm_shim_recursion_system_dir
@@ -1,0 +1,57 @@
+#!/usr/bin/env bash
+# Regression test for the dual-dir gap left by #8475.
+#
+# #8475 fixed `dependency_env` to strip the user shims dir (`dirs::SHIMS`) so
+# backend version-resolution subprocesses don't recurse through a mise shim.
+# But it only stripped the user dir — when tools are installed via
+# `mise install --system` (devcontainer / Docker setups), there's a *second*
+# shims dir at `$MISE_SYSTEM_DATA_DIR/shims` that's also on PATH. The npm
+# backend's `npm view yarn versions time --json` call found that system shim,
+# re-entered mise, and the same fork bomb fired — reported as "ws agent run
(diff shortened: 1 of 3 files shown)


andrewjamesbrown:
Maybe fixes #8499 @andrewthauer ?

andrewjamesbrown:
Pushed [`6c58163`](https://github.com/jdx/mise/pull/10019/commits/6c5816379) addressing the bot reviews:

**gemini (HIGH) — centralize shim-dir filter, cover symlinked roots.** Pulled the dual-dir predicate into `file::is_mise_shims_dir(&Path) -> bool` and routed all four shim-stripping call sites through it (`dependency_env`, `path_env_without_shims`, `strip_shims_from_path`, `which_no_shims`). The existing helpers had been silently incomplete since #8475 — they only ever filtered the user dir, so callers like `gh auth token` / `git credential fill` (the consumers of `path_env_without_shims`) were still exposed. They are now covered too. Added a `fs::canonicalize` fallback per the macOS symlinked-roots concern.

**greptile (P2) — fake version may mask the regression.** Switched the test to `node = "22.0.0"` (real-but-uninstalled), matching the `go = "1.23.3"` pattern in `test_go_shim_recursion`.

**gemini (MEDIUM) — test robustness.** `output="$(... || true)"` and double-quoted `"$output"` in the `assert_contains`, so a non-zero exit from `mise ls` (the expected post-fix behaviour when `npm view` can't resolve) can't bypass the assertion.

Verified locally on the patched tree:
- `cargo check --bin mise` ✅
- `cargo clippy --bin mise -- -D warnings` ✅
- `cargo test --bin mise file::` — 108/108 ✅
- `cargo test --bin mise backend::` — 269/269 ✅

---

On **#8499**: definitely overlapping symptom (orphan npm processes), different root cause. The three issues identified in that discussion are:

1. `prefer_offline` doesn't cover the `"latest"` branch in `tool_version::resolve_version`
2. `tokio::time::timeout` can't interrupt `duct::cmd!(...).read()` because it blocks the worker thread with no `.await` points
3. `std::process::Child::drop()` doesn't signal the child, so killed mise leaves npm orphans behind

This PR fixes a fourth, adjacent bug in the same neighbourhood: when the npm backend's `_list_remote_versions` does fire `npm view`, the *system* mise-shim in PATH was re-entering mise, and each re-entry kicked off another `npm view` — turning what #8499 describes as a single hang-and-orphan into an exponential fork-bomb. With this PR, you get at most one hanging `npm view` per resolution attempt; the three #8499 issues still need their own fixes to make that one hang exit cleanly and not leave an orphan.

So: narrows the failure mode of #8499 (linear ≤1 orphan instead of dozens-to-hundreds), but doesn't close it. Probably worth resolving both together; happy to follow up on the `prefer_offline` / `kill_on_drop` fixes once this lands if no-one else is on it.

jdx:
see https://github.com/jdx/mise/pull/10068 i think we will need to memoize this

andrewjamesbrown:
> see #10068 i think we will need to memoize this

@jdx done!

PR #10020, opened 2026-05-21
fix(backend): adjust libc detection for musl distros

thespags:
Hi,

I was running into issues where if `gcompat` was installed on Alpine, mise would select the wrong environment. The workaround is to explicitly set with `MISE_LIBC`. and refactored out an existing piece that followed the same logic.

Thank you

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="$PROJECT_DIR$" vcs="Git" />
+  </component>
   <component name="CommitMessageInspectionProfile">
     <profile version="1.0">
       <inspection_tool class="BodyLimit" enabled="true" level="WARNING" enabled_by_default="true" />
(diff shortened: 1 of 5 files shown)


crates/vfox/Cargo.toml
```diff
@@ -55,6 +55,7 @@ xx = { version = "2", default-features = false, features = ["archive", "hash"] }
 env_logger = { version = "0.11", optional = true }
 clap = { version = "4", features = ["derive"], optional = true }
 tempfile = "3"
+mise-libc-detect = { path = "../mise-libc-detect" }
```
jdx: a whole new crate for this seems excessive

thespags: Hi @jdx thanks for the view.
Where would you prefer the code live? 
1. Duplicate the code between vfox and mise
2. Move the code into vfox
3. An independent crate

I don't have a strong opinion and want to respect your preference. Thank you.

thespags: Or something else?

jdx: I would probably just configure vfox from the mise crate

thespags: Updated @jdx

PR #10045, opened 2026-05-22
fix(bun): Use ARM64 Windows release asset when possible

brianmichel:
As of Bun 1.3.10 a native arm64 version of Bun is available on their releases. This native binary should be preferred by default when running mise on an arm64 version of Windows to not have to run Bun under emulation. 

This should fix https://github.com/jdx/mise/discussions/10044

### DIFF
--- a/src/plugins/core/bun.rs
+++ b/src/plugins/core/bun.rs
@@ -393,7 +393,17 @@ impl BunPlugin {
             .or_else(|| version.strip_prefix('v'))
             .unwrap_or(version);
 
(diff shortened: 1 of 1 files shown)


brianmichel:
There were some issues getting the full test suite running on Windows arm64 with the Lua Language Server not having an arm64 variant, and some issues with aube not being able to install packages for some reason. I'd be happy to contribute back the language server package fix (it just redirects Windows arm64 to use the x64 release since that's what's available).

brianmichel:
Sorry for the CI issues, will get these fixed up.

PR #10054, opened 2026-05-23
fix: watch cwds of dependencies

43081j:
When I changed watch mode to also watch dependencies' sources, I didn't
take into account the fact that the `sources` will be used _as-is_ (i.e.
not resolved to anything).

This meant the following would happen:

- `build:a` has `["src/*.ts"]`
- `build:b` has `["lib/*.js"]` and depends on `build:a`
- We pass `["src/*.ts", "lib/*.js"]` to watchexec _in the directory of
  `build:b`_

This made `lib/*.js` basically no-op, or worse, watch the wrong files.

This change resolves the paths to their owning directory so we end up
with `["wherever-builda-lives/src/*.ts",
"wherever-buildb-lives/lib/*.js"]`.

**Notable Changes:**

- Instead of passing relative globs to watchexec, we now pass resolved
  ones (relative to the root)
- We now pass `--project-origin` to watchexec which comes with some perf
  gains but also means globs are now relative to it
- We pass `--watch {cwd}` for the cwd of each dependency
- This new `resolve_source` function is basically turning a source glob
  into a relative-to-the-root glob while retaining negations
- The new `common_ancestor` function  tries to find the common ancestor of two directories

---

cc @jdx i don't usually delve into rust so i'd love some help here if you can.

especially if there's a better way to find the common ancestor, or if you think we should deal with that differently.

basically, watchexec doesn't seem to support absolute paths, so we have to set the `project-origin` to something above all globs. which is why this new function tries to find a common root (if not the configured one)

## Summary by CodeRabbit

* **Bug Fixes**
  * Watch now reliably detects changes for source patterns outside a task’s working directory; out-of-tree patterns trigger rebuilds and watched directories cover relevant files.

* **Improvements**
  * Watch generates accurate file filters and recursive watch directories so file-change matching is correct across varied path patterns and anchors.

* **Tests**
  * Added end-to-end regression test plus unit tests covering path relativization, filter generation, and watch-directory derivation.

### DIFF
--- a/e2e/tasks/test_task_watch_cwd_escape
+++ b/e2e/tasks/test_task_watch_cwd_escape
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+
+# Regression: a task whose cwd is a subdirectory (`dir`) but whose `sources`
+# escape that cwd via `..` must still be watched. The anchor/project-origin
+# has to widen up to the common ancestor so watchexec's filter actually
+# matches the out-of-cwd source file when it changes.
+
+mise use -g watchexec
+
+mkdir -p packages/foo shared/src
+
+cat <<EOF >mise.toml
+[tasks.build]
+dir = "packages/foo"
+sources = ["../../shared/src/*.ts"]
+run = "echo built"
+EOF
+
+touch shared/src/index.ts
+
+LOG_FILE=$(mktemp)
+mise watch build >"$LOG_FILE" 2>&1 &
+WATCH_PID=$!
+
+trap 'kill $WATCH_PID' EXIT
+
+# Wait for the initial run before asserting the re-run.
+for _ in $(seq 1 40); do
+  grep -q "built" "$LOG_FILE" && break
+  sleep 0.5
+done
+if ! grep -q "built" "$LOG_FILE"; then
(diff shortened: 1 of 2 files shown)


jdx:
This mostly looks good, but I think there is one missed case before merge.

When a source escapes the task cwd, the filter anchor is widened correctly, but the watched paths are still only the task cwds. For example:

```toml
[tasks.build]
dir = "packages/foo"
sources = ["../../shared/src/*.ts"]
```

This becomes roughly:

```text
--project-origin /repo
--watch /repo/packages/foo
-f shared/src/*.ts
```

The filter now points at `/repo/shared/src/*.ts`, but watchexec is not watching `/repo/shared`, so changes there may not trigger.

Could we either watch the computed anchor/common ancestor, or add watch dirs for source paths that fall outside their task cwd? A regression test for `dir` plus `../..` sources would be great too.

43081j:
i've added logic to do the latter - basically slice the source dirs upto the first glob character and watch that.

so `../../src/foo/*.ts` becomes `/wherever/src/foo` as a watch dir

and i added an e2e test similar to the existing ones but with a `../` source

43081j:
@jdx this should be good to go now 🎉

PR #10082, opened 2026-05-26
feat(node): add node.npm_shim setting to opt out of npm wrapper

jjb:
## Problem

When mise installs Node, `install_npm_shim` unconditionally writes a bash wrapper at `installs/node/<v>/bin/npm` that detects `npm install -g` and triggers `mise reshim` afterwards. This wrapper is created on every install and cannot currently be disabled.

This makes it impossible to let **corepack** manage npm under mise. With `corepack enable`, mise correctly installs shims for `yarn` / `pnpm` / `pnpx` / `yarnpkg`, but corepack's safety logic refuses to clobber the existing non-symlink `bin/npm` file — so `packageManager: "npm@X.Y.Z"` in `package.json` is silently ignored for npm specifically, while it works for the other package managers.

It also gets in the way of users who'd rather manage npm directly via `npm install -g npm@<pin>` — that does replace the wrapper with a symlink, but a subsequent `mise install --force node@<v>` overwrites it again.

The existing `node.corepack` setting only gates whether `corepack enable` runs as a post-install step; it doesn't affect the wrapper.

## Change

Add a new boolean setting:

- config: `node.npm_shim`
- env: `MISE_NODE_NPM_SHIM`
- default: `true` (preserves existing behavior)

When set to `false`, `install_npm_shim` is skipped. Combined with `node.corepack = true`, this lets corepack manage `bin/npm` cleanly.

The change is surgical: one new setting entry in `settings.toml`, one new conditional in `src/plugins/core/node.rs`, and the auto-regenerated `schema/mise.json`.

## Questions for the maintainer

- **Naming**: I went with `node.npm_shim` (positive, default `true`) to match the style of `node.corepack`. Happy to rename — e.g. `auto_install_npm_shim`, or a negative `node.no_npm_shim` — if you'd prefer.
- **Default**: kept at `true` for backward compatibility. If you'd rather flip it to `false` (corepack-friendly out of the box for users who have `node.corepack` on), I can do that.

## Verification

Verified locally that with `MISE_NODE_NPM_SHIM=0` + `MISE_NODE_COREPACK=1`, `corepack enable` successfully installs an npm shim alongside yarn/pnpm and the `packageManager` pin in `package.json` is respected.

---

*This PR was prepared with the help of an AI coding assistant.*


schema/mise.json
```diff
@@ -1113,6 +1113,11 @@
               "description": "Installs the default corepack shims after installing any node version.",
               "type": "boolean"
             },
+            "npm_shim": {
+              "default": true,
+              "description": "Install a bash wrapper at bin/npm that triggers `mise reshim` after `npm install -g`. Disable to let corepack or a global `npm install -g npm@…` manage `bin/npm` directly.",
```
jjb:
```suggestion
              "description": "Install a bash wrapper at bin/npm that triggers `mise reshim` after `npm install -g`. Disable to let corepack or a global `npm install -g npm@...` manage `bin/npm` directly.",
```

settings.toml
```diff
@@ -1434,6 +1434,12 @@ description = "Installs the default corepack shims after installing any node ver
 env = "MISE_NODE_COREPACK"
 type = "Bool"
 
+[node.npm_shim]
+default = true
+description = "Install a bash wrapper at bin/npm that triggers `mise reshim` after `npm install -g`. Disable to let corepack or a global `npm install -g npm@…` manage `bin/npm` directly."
```
jjb:
```suggestion
description = "Install a bash wrapper at bin/npm that triggers `mise reshim` after `npm install -g`. Disable to let corepack or a global `npm install -g npm@...` manage `bin/npm` directly."
```

dcstone09 (approved):

jdx:
needs a test

jjb:
@jdx okay, ready! all feedback welcome

PR #10091, opened 2026-05-27
docs: document missing `mise.lock` as potential cause of slowness

andreynering:
I started having a slow shell, and what fixed it for me was locking my dependencies. I noticed this is missing in the docs, so adding it.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -43,6 +43,7 @@ Replace `bash` with your shell. Common causes of slow prompts:
 - Expensive `_.source` scripts in `mise.toml` — these re-run on every prompt
 - Large numbers of tools or plugins
 - Network-dependent operations in env directives
+- Missing `mise.lock` - run `mise lock` or `mise lock -g`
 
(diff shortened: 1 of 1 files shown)


jdx:
mise.lock definitely isn't required though, could you share your mise.toml with me so we could figure out why that was slow?

andreynering:
@jdx

Project `mise.toml`:

```toml
[tools]
air = "latest"
gcloud = "latest"
go = "1.26"
gofumpt = "0.10"
golangci-lint = "2.11"
goreleaser = "latest"
node = "22"
stripe = "latest"
task = "latest"
```

Global `mise.toml`:

https://github.com/andreynering/mise/blob/8710a333e8b9b4544992f6c917609b03af72b003/config.toml

My guess is that Mise was looking for new versions on some of the these `latest` tools maybe, but I didn't expect it. `mise lock` made it instant again.

jdx:
closing since this is actually a bug if mise is fetching versions in shell hooks

PR #10100, opened 2026-05-27
registry: use aqua for aube

scop:


jdx:
using github is intentional so we don't use the musl version

PR #10118, opened 2026-05-28
registry: add imagemagick (aqua:ImageMagick/ImageMagick)

thernstig:
## Summary by CodeRabbit

* **Chores**
  * ImageMagick registry adds an additional installation backend supporting Linux and Windows alongside existing options.
  * CI and test containers now run with FUSE device access and elevated mount privileges; test images include FUSE support and a compatibility symlink for fusermount.
  * CI change-detection broadened to recognize additional registry-related paths.

* **Tests**
  * Automated verification added that runs ImageMagick and checks its reported version.

### DIFF
--- a/.github/workflows/registry.yml
+++ b/.github/workflows/registry.yml
@@ -43,7 +43,7 @@ jobs:
           diff_base_sha="$(git merge-base "$PR_BASE_SHA" "$PR_HEAD_SHA")"
           echo "diff-base-sha=$diff_base_sha" >> "$GITHUB_OUTPUT"
           # Check if relevant files changed
-          if git diff --name-only "$diff_base_sha" "$PR_HEAD_SHA" | grep -qE '^(registry/.*\.toml|\.github/workflows/registry\.yml|src/cli/test_tool\.rs)$'; then
+          if git diff --name-only "$diff_base_sha" "$PR_HEAD_SHA" | grep -qE '^(registry/.*\.toml|\.github/workflows/registry\.yml|src/cli/test_tool\.rs|e2e/run_test|packaging/e2e/Dockerfile)$'; then
             echo "registry-changed=true" >> "$GITHUB_OUTPUT"
           else
             echo "registry-changed=false" >> "$GITHUB_OUTPUT"
@@ -105,7 +105,7 @@ jobs:
           DIFF_BASE_SHA: ${{ needs.check-changes.outputs.diff-base-sha }}
           PR_HEAD_SHA: ${{ github.event.pull_request.head.sha }}
         run: |
-          if git diff --name-only "$DIFF_BASE_SHA" "$PR_HEAD_SHA" | grep -q ".github/workflows/registry.yml"; then
+          if git diff --name-only "$DIFF_BASE_SHA" "$PR_HEAD_SHA" | grep -qE '^(\.github/workflows/registry\.yml|e2e/run_test|packaging/e2e/Dockerfile)$'; then
             echo "modified=true" >> "$GITHUB_OUTPUT"
           fi
       - uses: ./.github/actions/registry-diff
--- a/.github/workflows/registry.yml
+++ b/.github/workflows/registry.yml
@@ -238,6 +238,8 @@ jobs:
 
           set +e
           docker run --rm \
+            --device /dev/fuse \
+            --cap-add SYS_ADMIN \
             --read-only \
             -w /tmp \
             -v "$container_tmp_dir:/tmp" \
@@ -301,6 +303,8 @@ jobs:
 
           set +e
           docker run --rm \
+            --device /dev/fuse \
+            --cap-add SYS_ADMIN \
             --read-only \
             -w /tmp \
             -v "$container_tmp_dir:/tmp" \
--- a/.github/workflows/registry.yml
+++ b/.github/workflows/registry.yml
@@ -43,7 +43,7 @@ jobs:
           diff_base_sha="$(git merge-base "$PR_BASE_SHA" "$PR_HEAD_SHA")"
           echo "diff-base-sha=$diff_base_sha" >> "$GITHUB_OUTPUT"
           # Check if relevant files changed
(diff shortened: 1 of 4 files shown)


thernstig:
I am not able to run `mise run format` locally,  it errors out on things, so I would guess some things are setup manually that I am unaware about. I am running in a plain Ubuntu machine.

**edit:** wrote an issue here about this https://github.com/jdx/mise/discussions/10121

Regarding the other failure imagemagick has not been released yet: https://github.com/aquaproj/aqua-registry/releases

thernstig:
It was released here: https://github.com/aquaproj/aqua-registry/releases/tag/v4.518.0. Not sure why it says it is not in the registry.

thernstig:
@risu729 help me here.

risu729:
@thernstig Check https://github.com/jdx/mise/blob/main/vendor/aqua-registry/metadata.json

thernstig:
@risu729 updating it manually did not make this repo happy. I thought that was what you indicated I should do.

risu729:
@thernstig You need to wait until the next mise release.

jdx:
https://github.com/aquaproj/aqua-registry/blob/cea4477e3bf8e8e6296e5c117c94b256f69f5dc8/pkgs/ImageMagick/ImageMagick/registry.yaml#L10

thernstig:
I am checking this, as I thought they meant to support everything _but_ Windows on Arm.

Or will mise not even allow this in even in that case, if everything is supported but WIndows on Arm 64?

risu729:
That's the opposite. Check aqua docs

thernstig:
@risu729 I saw that, but I thought their intention was to support everything _but_ windows/amd64, but it seems only windows/amd64  is supported.

I have asked here if the rest of the support will be added, so we can keep this on hold, see https://github.com/aquaproj/aqua-registry/pull/50281. They now added support in https://github.com/aquaproj/aqua-registry/pull/54605/changes.

My question is: Is this support now enough to make it into mise? Or what is required for it to be allowed to be added here? Is the support now enough?

risu729:
mise registry supports os/arch pairs. I think you can use it

thernstig:
> mise registry supports os/arch pairs. I think you can use it

I updated it to only allow for `linux` and `windows`. But I think maybe that schema/mise-registry-tool.json is wrong. I think maybe `os` supports os/arch pairs? Which would have been better. But if I try that now I get a schema error.

thernstig:
@jdx I am unsure what to do here, as AppImage requires fuse, see:

- https://imagemagick.org/download/
- https://github.com/AppImage/AppImageKit/wiki/FUSE

It should be possible to get around this to run the AppImage with [`--appimage-extract`](https://github.com/AppImage/AppImageKit/wiki/FUSE#fallback).

Is this possible here?

jdx:
I don't know either

thernstig:
@jdx this cannot be solved without your help. We need to update the Docker image to include fuse, which I tried to do, as done here https://github.com/AppImage/AppImageKit/wiki/FUSE

But I cannot build and push the new image myself.

jdx:
If the tool would need FUSE as a prereq I don't think it's an improvement over conda to begin with

thernstig:
@jdx I think I have a better solution for this, can you contemplate https://github.com/jdx/mise/discussions/10267.

jdx:
why are we getting rid of conda here? I don't understand the value

thernstig:
@jdx I am unsure what you mean with "getting rid of"? Is it not supported still via https://github.com/jdx/mise/pull/10118/changes#diff-f322a53bf146c72239948186ecada0e05f5189a503c589da27ba156c22e41856R1-R8?

If you wonder why we wanted to add aqua, I thought it was preferred? I read all docs and that is the general gist I got out of it, especially reading https://mise.jdx.dev/registry.html where e.g. conda is tier 2. It depends on if we can support AppImage in this case or not.

Seems you do not want this?

I can mention I personally had trouble with conda, even though I did not write a bug report. It worked eventually, but it took forever to install and timed out. Worked on retries.

My only goal is/was to help future people of imagemagick installs 😃

iki:
> If you wonder why we wanted to add aqua, I thought it was preferred?

I'll try to add the context here:
- The primary reason is that `mise use imagemagick` doesn't work on Windows, as neither of currently configured conda/asdf backends provides ImageMagick on Windows
- I opened #8476 to add github backend, but it requires specifying platforms and @jdx [suggested](https://github.com/jdx/mise/pull/8476#issuecomment-4006997806) to add it to aqua directly. That was completed in https://github.com/aquaproj/aqua-registry/pull/50281 and after @risu729 added aqua 7z support in #10224 (v2026.6.1) it works OK on Windows: `mise use aqua:ImageMagick/ImageMagick`

Imo the options are:
- Either use aqua ImageMagick on Windows only and use conda/asdf on Linux (not tested)
- Or use @thernstig #10267 AppImage type 2 support and use aqua ImageMagick on Linux too (supposedly faster to install?)

WDYT?

jdx:
I would just use it for windows. If there isn't a reason for moving away from Conda I don't understand why we would

thernstig:
@jdx it is much faster, and less error-prone. I know this for a fact since I installed imagemagick via conda 3-4 weeks ago and it was a) very slow compared to other aqua tools b) failing at first at timeouts, twice, but later workd. Did not investigate why.

Seems you do not wish to support AppImage type2?

jdx:
I think there is a lot of risk with adding AppImage support that I don't want to take on without a good reason

iki:
@thernstig maybe just add aqua:ImageMagick as Windows-only for now, and open issue on ImageMagick repo to build linux archive as 7z as well (in addition to AppImage), so we can add it to mise registry for faster install as well?

thernstig:
> @thernstig maybe just add aqua:ImageMagick as Windows-only for now, and open issue on ImageMagick repo to build linux archive as 7z as well (in addition to AppImage), so we can add it to mise registry for faster install as well?

Will do this. Do you mind opening that issue since you wrote and fixed the original issue on Aqua?

thernstig:
Don't think the latest errors are caused by my change.

iki:
@suchipi reports long build times on Linux too: https://github.com/jdx/mise/pull/8476#issuecomment-4636331291

> I install imagemagick with mise on Linux as part of my build but it takes a long time to install because it has to compile from source

Let's hope ImageMagick will add the static Linux build archive: https://github.com/ImageMagick/ImageMagick/issues/8792 (comment there if you need it)

PR #10120, opened 2026-05-28
fix(pipx): combine --pip-args value to avoid argparse error

iloveitaly:
## Summary

Reopening #9982

pipx's argparse rejected `--pip-args --uploaded-prior-to=...` because the value starts with `--`. Pack into a single `--pip-args=VALUE` element. Regression from #9190.

Closes #9976

## Context

I'm seeing this across a bunch of failures on railpack. [Here's an example](https://github.com/railwayapp/railpack/actions/runs/26365713405/job/77609431920?pr=564).

This occurs [because `install_before` is set](https://github.com/railwayapp/railpack/blob/ffdb0331677897577eafe61edc600aafdec387be/core/generate/mise_step_builder.go#L293-L294) and the `pipx` backend is used. [Here's the project](https://github.com/railwayapp/railpack/tree/main/examples/python-poetry) to reproduce the failure (although the mise config which triggers it is _not_ shown there, it's injected into the project when it's built by railpack).

### DIFF
--- a/src/backend/pipx.rs
+++ b/src/backend/pipx.rs
@@ -387,10 +387,9 @@ impl PIPXBackend {
 
     fn pip_uploaded_prior_to_args(before_date: Option<Timestamp>) -> Vec<OsString> {
         match before_date {
-            Some(before_date) => vec![
-                "--pip-args".into(),
-                format!("--uploaded-prior-to={before_date}").into(),
-            ],
+            Some(before_date) => {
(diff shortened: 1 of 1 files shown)


iloveitaly:
@jdx I _think_ this is ready to go! Happy to make any other fixes or improvements required.

risu729 (approved):

PR #10127, opened 2026-05-28
feat(github): use versions host for release metadata

jdx:
## Why this matters

This tackles one of mise's oldest and most painful operational problems: public GitHub API rate limits during installs. A huge amount of the registry ultimately resolves through GitHub releases, especially via `aqua:` and `github:` tools. In CI and Docker builds there is often no GitHub token, so many independent installs all compete for GitHub's low anonymous API limit just to discover release assets or check provenance.

With this PR, normal public GitHub-backed installs can use mise's shared versions host for release metadata and GitHub artifact attestations before falling back to `api.github.com`. That means common installs like `gh`, `gum`, `eza`, `starship`, `lazygit`, `lazydocker`, and many aqua-backed tools no longer need to hit GitHub directly just to resolve release assets. This should remove a long-running source of flaky Docker builds, CI failures, and confusing "403 API rate limit exceeded" reports for users who are not doing anything unusual.

## Summary

- try `mise-versions` before `api.github.com` for public GitHub release metadata
- fetch cached GitHub attestation payloads from `mise-versions` before using GitHub's attestation API
- keep verification local: cached attestation bundles are still verified against the downloaded artifact
- validate mirrored release asset URLs before using them, requiring same-repo GitHub release download URLs and same-repo GitHub release asset API URLs
- fall back to GitHub's API when the versions host misses, fails, is disabled, or when a non-public GitHub API host is configured
- keep versions-host metadata fallback quiet, so private GitHub repos that the shared cache cannot access still work out of the box with normal GitHub authentication and no spurious warning
- add an e2e test using real `mise-versions` that fails if `api.github.com` is contacted
- update docs for `use_versions_host`, GitHub token guidance, and rate-limit troubleshooting

## Security / correctness notes

- `mise-versions` is a metadata cache, not a trust root. Release asset URLs are validated before use, and artifact attestations are verified locally by mise.
- The versions host is only used for public `api.github.com` flows. GitHub Enterprise/custom API hosts continue to use their configured API path.
- Private github.com repositories naturally miss the shared cache and fall back to GitHub's API without warning.
- Empty release asset lists are not cached as usable release metadata, so transient GitHub/API weirdness does not poison installs.
- If the shared cache does not have data yet, existing authenticated/unauthenticated GitHub fallback behavior remains in place.

## Testing

- `cargo fmt`
- `cargo check -q`
- `cargo test -q versions_host::tests`
- `cargo test -q -p mise-sigstore`
- `bash -n e2e/backend/test_github_versions_host_no_api`

Note: the new e2e depends on the production `mise-versions` mirror endpoints from `jdx/mise-versions#201`. Production currently needs the follow-up Worker fix in `jdx/mise-versions#202` to stop returning 502 for mirrored GitHub release metadata; after that deploy, this e2e should be run against the live service.

---

> [!NOTE]
> **Medium Risk**
> Changes install and provenance paths for many public GitHub tools; mitigations include strict URL validation, local attestation verification, and unchanged GitHub API fallback when the cache misses or is disabled.
> 
> **Overview**
> Public GitHub installs can now use **mise-versions** for release metadata and artifact attestations before calling `api.github.com`, cutting anonymous API use in CI and Docker when `use_versions_host` is on.
> 
> **Release resolution:** `get_release_` tries `versions_host::github_release` for public `api.github.com` URLs, with URL/tag validation on mirrored assets. The GitHub backend adds `resolve_exact_version` so pinned semver-like versions can be proven via a single release lookup instead of listing all releases. `ToolVersion` resolution calls that hook before full remote version lists.
> 
> **Attestations:** `verify_attestation` and `detect_attestations` fetch cached bundles from mise-versions when enabled; inline bundles are verified locally via new `mise-sigstore` entry points (`verify_github_attestation_with_attestations`, optional precomputed digest). Missing or bundle-URL-only cache entries fall back to GitHub’s API without treating cache miss as failure for private repos.
> 
> **Docs/tests:** `use_versions_host` descriptions and troubleshooting/GitHub token docs reflect the broader cache role. E2e blocks `api.github.com` while installing from real mise-versions metadata; `mise ls` cache test expects versions-host GitHub release URLs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 34bae18a2ac2cb6e9382aa77f5d54e4bc954a2ff. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/mise-sigstore/src/lib.rs
+++ b/crates/mise-sigstore/src/lib.rs
@@ -219,6 +219,12 @@ pub struct Attestation {
     bundle_url: Option<String>,
 }
 
+impl Attestation {
+    pub fn has_inline_bundle(&self) -> bool {
+        self.bundle.is_some()
+    }
+}
+
 impl AttestationClient {
     pub fn builder() -> AttestationClientBuilder {
         AttestationClientBuilder::default()
@@ -332,7 +338,16 @@ pub async fn verify_github_attestation(
     token: Option<&str>,
     signer_workflow: Option<&str>,
 ) -> Result<bool> {
-    verify_github_attestation_inner(artifact_path, owner, repo, token, signer_workflow, None).await
+    verify_github_attestation_inner(
+        artifact_path,
+        owner,
+        repo,
+        token,
+        signer_workflow,
+        None,
+        None,
+    )
+    .await
 }
 
 pub async fn verify_github_attestation_with_base_url(
@@ -350,8 +365,44 @@ pub async fn verify_github_attestation_with_base_url(
         token,
         signer_workflow,
         Some(base_url),
+        None,
     )
     .await
+}
+
+pub async fn verify_github_attestation_with_base_url_and_digest(
+    artifact_path: &Path,
+    owner: &str,
+    repo: &str,
+    token: Option<&str>,
+    signer_workflow: Option<&str>,
+    base_url: &str,
+    digest: &str,
+) -> Result<bool> {
+    verify_github_attestation_inner(
+        artifact_path,
+        owner,
+        repo,
+        token,
+        signer_workflow,
+        Some(base_url),
+        Some(digest),
+    )
+    .await
+}
+
+pub async fn verify_github_attestation_with_attestations(
+    artifact_path: &Path,
+    attestations: &[Attestation],
+    signer_workflow: Option<&str>,
+) -> Result<bool> {
+    if attestations.is_empty() {
+        return Err(AttestationError::NoAttestations);
+    }
+
+    let artifact = tokio::fs::read(artifact_path).await?;
+    let mut trust_roots = TrustRoots::default();
+    verify_attestation_bundles(attestations, &artifact, signer_workflow, &mut trust_roots).await
 }
 
 async fn verify_github_attestation_inner(
@@ -361,6 +412,7 @@ async fn verify_github_attestation_inner(
     token: Option<&str>,
     signer_workflow: Option<&str>,
     base_url: Option<&str>,
+    digest: Option<&str>,
 ) -> Result<bool> {
     let mut builder = AttestationClient::builder();
     if let Some(token) = token {
@@ -370,7 +422,10 @@ async fn verify_github_attestation_inner(
         builder = builder.base_url(base_url);
     }
     let client = builder.build()?;
-    let digest = calculate_file_digest_async(artifact_path).await?;
+    let digest = match digest {
+        Some(digest) => digest.to_string(),
+        None => calculate_file_digest(artifact_path).await?,
+    };
     let attestations = client
         .fetch_attestations(FetchParams {
             owner: owner.to_string(),
@@ -1261,7 +1316,7 @@ fn verify_raw_signature(
         .map_err(|e| AttestationError::Verification(format!("signature verification failed: {e}")))
 }
 
-async fn calculate_file_digest_async(path: &Path) -> Result<String> {
+pub async fn calculate_file_digest(path: &Path) -> Result<String> {
     let mut file = tokio::fs::File::open(path).await?;
     let mut hasher = Sha256::new();
     let mut buffer = [0; 8192];
--- a/docs/dev-tools/github-tokens.md
+++ b/docs/dev-tools/github-tokens.md
@@ -1,6 +1,8 @@
 # GitHub Tokens
 
-Many tools in mise are hosted on GitHub, and mise uses the GitHub API to list versions and download releases. Unauthenticated requests are subject to low [rate limits](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api), which can cause `403 Forbidden` errors — especially in CI. This page explains how to configure GitHub authentication in mise.
+Many tools in mise are hosted on GitHub. For public releases, mise uses [mise-versions](https://mise-versions.jdx.dev) by default as a shared cache for version lists, release metadata, and GitHub artifact attestations. This avoids most unauthenticated GitHub API calls during normal installs, including CI and Docker builds.
+
+GitHub tokens are still useful when mise has to fall back to GitHub's API, when `MISE_USE_VERSIONS_HOST=0` is set, or when installing tools from private repositories, GitHub Enterprise, or custom GitHub API hosts. Unauthenticated requests are subject to low [rate limits](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api), which can cause `403 Forbidden` errors. This page explains how to configure GitHub authentication in mise.
 
 ## Token Priority
 
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -108,6 +108,12 @@ get around GitHub rate limits when querying for new versions. Check that repo fo
 see if it has an updated version. This service can be disabled by
 setting `MISE_USE_VERSIONS_HOST=0`.
 
+mise also uses the versions host as a shared cache for public GitHub release metadata and
+GitHub artifact attestations. This means normal installs of public `github:` and many
+`aqua:` tools can avoid unauthenticated GitHub API calls even in Docker builds or CI jobs
+that do not have a token configured. If the versions host does not have the requested
+metadata yet, mise falls back to GitHub's API.
+
 mise-versions itself also struggles with rate limits but you can help it to fetch more frequently by authenticating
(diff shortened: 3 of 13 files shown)


jdx:
Addressed relevant feedback in the latest push:

- `github::get_release_` now treats mise-versions lookup errors as cache misses and falls back to GitHub.
- `calculate_file_digest` now streams the file instead of reading the full artifact into memory.
- Kept empty attestation arrays as authoritative negative cache responses. That is intentional for this feature: otherwise tools without attestations would still hit `api.github.com` during normal installs, which defeats the rate-limit fix this PR is targeting.

Validation rerun:
- `cargo fmt`
- `cargo test -q versions_host::tests`
- `cargo check -q`
- `bash -n e2e/backend/test_github_versions_host_no_api`

jdx:
Addressed the latest feedback by making the versions-host release asset validator reject empty asset lists before accepting mirrored metadata. Added a unit test for the empty-assets case.

Verification:
- `cargo test -q versions_host::tests`
- `cargo check -q`
- `bash -n e2e/backend/test_github_versions_host_no_api`

jdx:
Addressed two additional relevant feedback items in the latest push:

- `use_versions_host_for_attestations` now respects `use_versions_host`/offline settings before hashing the artifact, avoiding wasted file reads when the feature is disabled.
- Mirrored positive attestation responses are now required to contain inline hydrated bundles before the fast path verifies them; malformed positive mirror responses fall back to the GitHub API path instead of failing inside bundle verification. Empty attestation responses remain authoritative negative cache responses by design.

Verification:
- `cargo test -q github::sigstore::tests`
- `cargo test -q -p mise-sigstore`
- `cargo check -q`
- `bash -n e2e/backend/test_github_versions_host_no_api`

jdx:
Addressed the new release metadata feedback in the latest push: mirrored non-`latest` releases now have to report the same `tag_name` that was requested. If the mirror returns a stale/wrong tag, mise warns and falls back to GitHub API instead of using the mirrored release.

Verification:
- `cargo fmt`
- `cargo test -q versions_host::tests`
- `cargo check -q`
- `bash -n e2e/backend/test_github_versions_host_no_api`

jdx:
Addressed the latest relevant Cursor feedback in the newest push:

- Moved the streamed artifact digest helper into `mise-sigstore` and reused it from the mise wrapper instead of duplicating the implementation.
- Added a digest-aware GitHub attestation verification path so if the versions-host lookup misses or falls back, the GitHub API fallback reuses the already-computed digest instead of hashing the artifact a second time.

Verification:
- `cargo test -q github::sigstore::tests`
- `cargo test -q -p mise-sigstore`
- `cargo check -q`
- `bash -n e2e/backend/test_github_versions_host_no_api`

jdx:
I reviewed the current PR feedback again. I do not see new mise-side code changes to make from the latest comments:

- The release lookup fallback, streamed digest, inline-bundle guard, empty release asset rejection, tag-name validation, and digest reuse items have already been addressed in prior pushes.
- The remaining empty attestation-list feedback is intentionally not changed: empty arrays from mise-versions are treated as authoritative negative cache responses so normal installs for tools without attestations do not fall back to `api.github.com`, which is central to this rate-limit fix. Positive malformed responses still fall back.

jdx:
Good point. I pushed a small follow-up so versions-host GitHub metadata rate limits are debug-only, not warnings. Private github.com repositories already miss the shared cache with 404 and fall back silently; this closes the other optional-cache case so private installs keep using normal GitHub authentication without spurious mise-versions warnings.

Verification:
- `cargo fmt`
- `cargo test -q versions_host::tests`
- `cargo check -q`

jdx:
Addressed the lint failure and tightened the attestation mirror parsing:

- Ran `mise run render` and committed the generated `schema/mise.json` update, which should fix the failing render/no-diff lint job.
- Removed the serde default from the versions-host attestation response so malformed `{}` responses are treated as mirror misses and fall back to GitHub API. Explicit `{"attestations":[]}` responses remain authoritative negative cache entries by design.

Verification:
- `mise run render`
- `cargo fmt`
- `cargo test -q versions_host::tests`
- `cargo check -q`

jdx:
mise-versions is deployed and working now. I verified the live endpoints and fixed the mise e2e gaps that showed up once production was healthy:

- Live tagged release mirror now returns 200 with `Cache-Control: public, max-age=600, s-maxage=31536000, immutable`.
- Live latest release mirror returns 200 with the shorter CDN TTL.
- Pinned `github:` versions now avoid the full GitHub `/releases` list and let install-time release resolution use the mirrored specific release metadata.
- Explicit empty attestation arrays from mise-versions now stop the attestation API fallback as intended.
- The attestation digest URL keeps the `sha256:` separator unescaped so the deployed Worker route accepts it.

Verification:
- `cargo fmt`
- `cargo test -q github::sigstore::tests`
- `cargo test -q versions_host::tests`
- `cargo check -q`
- `mise run test:e2e e2e/backend/test_github_versions_host_no_api`

jdx:
Greptile was 3/5 because it found a real validation gap: `release.tag_name` was checked against the requested tag, but each mirrored `browser_download_url` still accepted any tag path segment under the same owner/repo. A manipulated mirror response could therefore claim `tag_name = v1.0.0` while pointing the actual browser download at `/releases/download/v0.9.0/...`.

Fixed in the latest push:

- `valid_github_browser_download_url` now requires the download URL tag segment to match `release.tag_name`.
- The comparison handles percent-encoded tag segments, e.g. `release%2F2026` matching `release/2026`.
- Added tests for matching tag, mismatched tag, and encoded tag cases.

Verification:
- `cargo fmt`
- `cargo test -q versions_host::tests`
- `cargo check -q`
- `mise run test:e2e e2e/backend/test_github_versions_host_no_api`

jdx:
Updated the PR for the latest feedback and failing check:

- Replaced the unconditional dotted GitHub version shortcut with a backend exact-release fast path. It now only skips the full `/releases` listing after the specific release endpoint confirms the tag exists, so prefix resolution still falls through correctly.
- Updated `e2e/cli/test_ls_cache` to expect the versions-host release endpoint and to fail if `mise ls` re-fetches that endpoint instead of using cached metadata.

Local verification passed:

- `cargo fmt`
- `cargo check -q`
- `cargo test -q versions_host::tests`
- `cargo test -q github::sigstore::tests`
- `mise run test:e2e e2e/cli/test_ls_cache`
- `mise run test:e2e e2e/backend/test_github_versions_host_no_api`

jereviikari:
I’m seeing a regression that looks related to this versions-host release metadata change.

### Summary

After upgrading to `mise 2026.5.16 linux-x64 (2026-05-28)`, `mise upgrade --bump` started printing repeated warnings from `mise-versions.jdx.dev`.

Setting `use_versions_host = false` avoids the warnings.

### Environment

```text
mise version: 2026.5.16 linux-x64 (2026-05-28)
OS: Linux x64
Shell: bash
```

Relevant setting:

```toml
[settings]
use_versions_host = true # default
```

### Reproduction

Run:

```sh
mise upgrade --bump
```

### Actual behavior

`mise` repeatedly warns about failed requests to `mise-versions.jdx.dev`, for example:

```text
mise WARN  HTTP GET https://mise-versions.jdx.dev/api/github/repos/npm/cli/releases/latest attempt 1 failed (transient): HTTP status server error (502 Bad Gateway)
mise WARN  HTTP GET https://mise-versions.jdx.dev/api/github/repos/npm/cli/releases/latest attempt 2 failed (transient): HTTP status server error (502 Bad Gateway)
mise WARN  HTTP GET https://mise-versions.jdx.dev/api/github/repos/npm/cli/releases/latest attempt 3 failed (transient): HTTP status server error (502 Bad Gateway)
```

Direct check:

```sh
curl -I https://mise-versions.jdx.dev/api/github/repos/npm/cli/releases/latest
```

returns:

```text
HTTP/2 502
```

The canonical upstream endpoint works:

```sh
curl -I https://api.github.com/repos/npm/cli/releases/latest
```

returns:

```text
HTTP/2 200
```

### Expected behavior

If `mise-versions.jdx.dev` is unavailable, `mise upgrade --bump` should fall back cleanly or emit a single concise warning instead of repeated retry warnings.

### Workaround

```toml
[settings]
use_versions_host = false
```

chadlwilson:
Also, see https://github.com/jdx/mise/discussions/10284 - repeated `403` warnings for some projects as well.

PR #10130, opened 2026-05-29
fix(upgrade): handle lone v prefix in --bump latest queries

zeitlinger:
## Summary
- fix `mise upgrade --bump` / `mise outdated --bump` when a tool is pinned with a lone leading `v` prefix
- treat `v`/`V` as version syntax rather than a backend or vendor prefix when building the latest-version query
- add a regression e2e test covering `shfmt = "v3.12.0"`

## Root cause
`--bump` splits the configured version into a prefix and numeric portion before asking the backend for the latest matching version.
For a pin like `v3.13.1`, that produced the fuzzy query `v3`.

The aqua backend normalizes tags like `v3.13.1` to bare versions like `3.13.1`, so querying for `v3` returned no match and emitted `no latest version found` even though the tool was up to date.

## Validation
- `cargo test test_prefixed_latest_query`
- manual repro with a built binary before/after the fix using:
  - `[tools] shfmt = "v3.13.1"`
  - `mise up -lvn shfmt`

## Notes
I also added an e2e regression test, but I could not run the full `mise run test:e2e e2e/core/test_v_prefix_bump_query` task locally because the environment failed while installing `npm:prettier@3` through `aube` with a permission error under `~/.local/share/aube/...`.

### DIFF
--- a/e2e/core/test_v_prefix_bump_query
+++ b/e2e/core/test_v_prefix_bump_query
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+# Regression test for aqua-backed tools pinned with a leading v prefix.
+# `mise upgrade --bump` used to derive the fuzzy latest query `v3` from
+# `v3.12.0`, but aqua normalizes shfmt releases to bare versions like 3.13.1.
+# That mismatch triggered: "no latest version found".
+cat <<'TOML' > mise.toml
+[tools]
+shfmt = "v3.12.0"
+TOML
+
+assert_not_contains "mise upgrade shfmt --bump --dry-run 2>&1" "no latest version found"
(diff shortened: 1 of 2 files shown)


zeitlinger:
Addressing the Greptile finding and lint failure:

- preserve the full bare latest version when the configured prefix is a lone `v`/`V` and the backend returns `3.13.1`
- extend the regression coverage to assert the dry-run bump target is `v3.13.1`
- fix the new e2e test formatting so `shfmt` passes

Validated with targeted unit tests plus a manual repro using the built binary.

jdx:
to be clear, putting "v" prefixes in mise.toml is not a supported use-case, I went ahead and merged this because it's a minor fix but generally speaking you should not be doing that

zeitlinger:
> to be clear, putting "v" prefixes in mise.toml is not a supported use-case, I went ahead and merged this because it's a minor fix but generally speaking you should not be doing that

got it - maybe is should print a warning to tell users to migrate? or at least in doctor.

PR #10138, opened 2026-05-29
fix(pipx): upgrade shared pip environment when using version constraints

iloveitaly:
- Trigger an upgrade of the shared pip environment to meet minimum requirements (pip>=26.0) when `before_date` constraints are applied.
- Add an e2e regression test to verify installation flows for packages with `minimum_release_age`.

### DIFF
--- a/docs/dev-tools/backends/pipx.md
+++ b/docs/dev-tools/backends/pipx.md
@@ -30,6 +30,7 @@ If you have `uv` installed, mise will use `uv tool install` under the hood and y
 When [`minimum_release_age`](/configuration/settings.html#minimum_release_age) is set, mise forwards the cutoff
 to transitive Python dependency resolution during install. The uv install path uses uv's
 `--exclude-newer` flag, and the `pipx` fallback passes pip's `--uploaded-prior-to` flag.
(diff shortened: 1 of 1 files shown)


risu729:
~~Why do we need this? I thought pipx creates a new pip environment every time. See https://github.com/jdx/mise/pull/9190~~

Thanks for the PR! I didn't notice this case.

risu729:
Yes, my understanding in #9190 was incomplete.

mise does give pipx a fresh `PIPX_HOME` for a fresh tool install, so the shared env path is fresh too. The part I missed is that this does not guarantee the first pip process in that shared env already supports `--uploaded-prior-to`.

In pipx's pip backend, `create_venv()` calls `shared_libs.create(verbose=..., pip_args=pip_args)`. `shared_libs.create()` creates the shared venv and then upgrades shared pip using the same `pip_args` from the current `pipx install`. So when mise passes `--pip-args=--uploaded-prior-to=...`, that flag can be used while bootstrapping/upgrading shared pip itself, before pip has been upgraded to `>=26.0`. If the ensurepip/bootstrap pip is older, it errors with `no such option: --uploaded-prior-to` before it can install a newer pip.

So the separate `pipx upgrade-shared --pip-args pip>=26.0` step is useful: it initializes/upgrades the shared pip without passing `--uploaded-prior-to`, then the actual package install can safely use that flag.

PR #10140, opened 2026-05-29
registry: add moon (aqua:moonrepo/moon)

agdphd:
moon is a monorepo build system which my organization uses via Mise. Installing it via the Aqua backend works perfectly, so this file just aliases that for convenience.

### DIFF
--- a/registry/moon.toml
+++ b/registry/moon.toml
@@ -4,4 +4,4 @@ idiomatic_files = [".moon/*"]
 
 [test]
 cmd = "moon --version"
-expected = "moon"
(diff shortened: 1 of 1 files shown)


registry/moon.toml
```diff
@@ -0,0 +1,7 @@
+description = "A build system and monorepo management tool for the web ecosystem, written in Rust"
+backends = ["aqua:moonrepo/moon"]
+idiomatic_files = [".moon/*"]
```
jdx, line 3: this doesn't make any sense

PR #10153, opened 2026-05-30
fix(completion): keep global -C/--cd usable in task argument completion (#10069)

JamBalaya56562:
Fixes the completion failure reported in #10069.

## Problem

When the global `-C/--cd` flag is placed before `run` and the task has a positional argument with `choices`, tab-completion errors instead of offering the choices:

```toml
[tasks."sample:run"]
usage = 'arg "<profile>" { choices "alpha" "beta" "gamma" }'
run   = 'printf "%s" "$usage_profile"'
```

```console
$ mise -C /tmp/repro run sample:run -- <TAB>
× Invalid choice for arg profile: -C, expected one of alpha, beta, gamma   # actual
alpha beta gamma                                                           # expected
```

This makes `mise -C <dir> ...` (a common way to drive a project's tasks from anywhere) unusable with task-argument completion.

## Root cause

Runtime completion is done by the external `usage` binary's `complete-word`, which calls `usage::parse::parse_partial(spec, words)`. The spec comes from `mise usage` (`src/cli/usage.rs`); its `run` subcommand mounts `mise tasks --usage` to supply task subcommands (e.g. `sample:run`) with their `choices` args.

Two facts combine:

1. mise declares `-C/--cd` twice with different `global` flags:
   - `src/cli/mod.rs` — root `Cli.cd`, `global = true` (changes mise's cwd).
   - `src/cli/run.rs` — `Run.cd`, **non-global**, functional: feeds `TaskExecutorConfig.cd` (the task's execution dir). Not redundant.
2. The usage parser, when descending into a subcommand, keeps only global flags: `available_flags.retain(|_, f| f.global)`, and Phase 1 collects but does not remove consumed global-flag tokens from the input.

So for `mise -C /tmp/repro run sample:run --`: `-C /tmp/repro` is collected for the mount (`mise --cd /tmp/repro tasks --usage`, correct) but left in the input; descending `run` overwrites the `-C` key with run's **non-global** `-C`; descending `sample:run` then drops `-C` entirely; in the main parse the leftover `-C` is no longer a known flag and is consumed as the `profile` positional → `Invalid choice`.

If `-C` had stayed global it would simply be re-consumed harmlessly and the choices would be offered.

## Fix

Completion-spec-only mitigation in `src/cli/usage.rs`: when emitting the spec, promote any flag on the mounted `run` / `tasks run` subcommands that collides with a root-level **global** flag back to `global = true`. This restores the invariant the parser relies on, so `-C` survives the task descent.

- Fixes the bug with the **currently released** `usage` parser (it changes spec data, not parser code — every `usage` version honors `retain(|_, f| f.global)`).
- Completion-spec only: no change to clap runtime parsing, `mise run` behavior, `Run.cd`, or `--help`.
- Written as a generic set-intersection, so any future global redeclaration is covered automatically. Today only `-C/--cd` collides.

## Verification

Reproduced against the installed `usage` 3.3.0:

| spec | `mise -C . run sample:run -- <TAB>` |
| --- | --- |
| before (run-level `-C` non-global) | ❌ `Invalid choice for arg profile: -C` |
| after (`global=#true`) | ✅ `alpha / beta / gamma` |

Baseline (no `-C`), `mise run`, and `mise tasks run` paths all pass. Added e2e regression `e2e/tasks/test_task_completion_global_cd`.

## Note / upstream follow-up

The deeper root cause is in the `usage` parser (consumed global-flag tokens left in the input, and a non-global redeclaration dropping an inherited global). That will be addressed upstream in jdx/usage; once fixed, this `global = true` promotion becomes a harmless no-op.

### DIFF
--- a/e2e/tasks/test_task_completion_global_cd
+++ b/e2e/tasks/test_task_completion_global_cd
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+# Regression test for https://github.com/jdx/mise/discussions/10069
+# Completion must work when the global -C/--cd flag precedes `run` and the
+# task has a positional arg with choices.
+
+cat <<'EOF' >mise.toml
+[tools]
+"usage" = { version = "latest", os = ["linux", "macos"] }
+
+[tasks."sample:run"]
+usage = 'arg "<profile>" { choices "alpha" "beta" "gamma" }'
+run = 'printf "%s" "$usage_profile"'
+EOF
+
+mise usage >./mise.usage.kdl
+
+# baseline: completion of the profile arg works without the global flag
+assert "mise exec -- usage complete-word --shell zsh -f ./mise.usage.kdl -- mise run sample:run -- ''" "alpha
+beta
+gamma"
+
+# regression: with the global -C <dir> flag before `run`, completion must still
+# offer the choices (previously errored: Invalid choice for arg profile: -C)
+assert "mise exec -- usage complete-word --shell zsh -f ./mise.usage.kdl -- mise -C . run sample:run -- ''" "alpha
+beta
+gamma"
+
+# also cover the `tasks run` path, which shares the same flags/mount
+assert "mise exec -- usage complete-word --shell zsh -f ./mise.usage.kdl -- mise -C . tasks run sample:run -- ''" "alpha
+beta
+gamma"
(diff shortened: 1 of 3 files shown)


JamBalaya56562:
Thanks @gemini-code-assist — addressed in c15bd7a. Replaced the `.unwrap()` subcommand lookups (including the pre-existing ones) with graceful `if let` handling, and reused the existing `run`/`tasks run` mutable bindings to apply the flag promotion so there are no extra lookups. Spec generation now skips gracefully if a subcommand is ever absent rather than panicking.

PR #10159, opened 2026-05-31
test(aqua): cover manifest symlink bins precedence

risu729:
Stacked on #10187. Target branch remains the default branch.

## Summary
- add regression coverage that Aqua install-manifest `symlink_bins` survives when current config overrides ordinary Aqua options
- keep this PR focused on the manifest/config precedence coverage; the Aqua vars canonicalization fix lives in #10187

## Details
The install manifest can carry layout-affecting Aqua options from the installed tool. This test covers the case where current config overrides a normal option such as `channel`, while preserving manifest-only `symlink_bins = true` so the installed runtime layout remains exposed correctly.

## Verification
- `cargo test test_resolve_tool_opts_preserves_aqua_symlink_bins_from_install_manifest`
- `git diff --check`

### DIFF
--- a/src/backend/aqua.rs
+++ b/src/backend/aqua.rs
@@ -65,47 +65,77 @@ impl<'a> AquaOptions<'a> {
     }
 
     fn var(&self, name: &str) -> Result<Option<String>> {
-        let opts = self.values.raw();
-        if let Some(toml::Value::Table(vars)) = opts.opts.get("vars")
-            && let Some(value) = vars.get(name)
-        {
-            return toml_string_var(&format!("vars.{name}"), value).map(Some);
-        }
-        opts.opts
+        self.canonical_var_options()
             .get(name)
-            .map(|value| toml_string_var(name, value).map(Some))
+            .map(|value| toml_string_var(&format!("vars.{name}"), value).map(Some))
             .unwrap_or(Ok(None))
     }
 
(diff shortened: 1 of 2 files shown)


risu729:
CI follow-up after 9231d927:

- All non-task-completion checks are green or skipped as expected, including lint, build-ubuntu, build-windows, windows-unit, windows-e2e, benchmark, registry, release, and Socket.
- The remaining failures are e2e-2 and e2e-3. Both fail in task completion tests, not Aqua code:
  - `tasks/test_task_completion`: `usage complete-word` returned `mise.toml	mise.toml` and `mise.usage.kdl	mise.usage.kdl` where the test expected bare candidates.
  - `tasks/test_task_completion_global_cd`: `usage complete-word` returned `alpha	alpha`, `beta	beta`, `gamma	gamma` where the test expected bare candidates.
- The PR diff only touches Aqua option/lockfile handling and a config regression test. I also checked recent `main` test runs; the latest `main` test run is still pending and the immediately preceding ones were cancelled, so there is no completed base failure to cite yet. Based on the logs, these failures appear unrelated to this PR.

risu729:
Closing this PR because its remaining branch delta is test-only (`src/config/mod.rs`). The original behavior change, including canonical Aqua vars in lock/cache identity and rejecting ambiguous duplicate var spellings, is now handled by #10187.

PR #10168, opened 2026-05-31
fix(tasks): show non-string task run entries in json output

roele:
Fixes a bug mentioned in #10013


roele:
Closing as duplicate of #10163

PR #10173, opened 2026-05-31
fix(aqua): support deprecated signer-workflow field

risu729:
## Summary
- parse deprecated aqua YAML field `github_artifact_attestations.signer-workflow`
- keep `signer_workflow` as the preferred canonical field
- route call sites through a `signer_workflow()` accessor so deprecated-field-only registry entries are honored

## Context
Aqua kept the old hyphenated YAML field for compatibility while moving to canonical `signer_workflow`. Its accessor prefers `signer_workflow` and falls back to deprecated `signer-workflow`.

This PR mirrors that behavior in mise's aqua-registry model. It intentionally does not include `predicate_type` support; that remains scoped to #10169.

## Tests
- `cargo check -p mise --all-features`
- `cargo test -p aqua-registry test_github_artifact_attestations -- --nocapture`


risu729:
Closing as it's no longer used in the standard registry.

PR #10175, opened 2026-05-31
fix(dotnet): include runtime in lock identity

risu729:
## Summary
- add core dotnet lockfile option identity for the `runtime` tool option
- keep SDK installs using the default empty lock option set
- add unit coverage for the dotnet option mapper and backend hook, including SDK/no-runtime behavior

## Context
This implements the missing PR D follow-up from https://github.com/users/risu729/projects/3/views/1?pane=issue&itemId=185472417. The direct `dotnet:` backend was not the target; this change applies to the core .NET plugin in `src/plugins/core/dotnet.rs`.

## Why `runtime` belongs in the lock identity
The core .NET plugin installs an SDK by default, but `runtime = "..."` changes the artifact class passed to `dotnet-install` and installs a runtime-only shared framework instead. Those are different install results even when the requested .NET version string is the same.

Lockfile entries are looked up by resolved version plus backend lockfile options. Without including `runtime`, an SDK request and a runtime request for the same version can share a lock entry, leaving one install to reuse stale identity from the other. Keeping SDK installs as the empty option set preserves existing lock entries, while adding `runtime` only for runtime installs makes SDK and runtime lock entries distinct.

## Tests
- `cargo fmt --check`
- `cargo test dotnet_lockfile_options_include_runtime`
- `cargo test resolve_lockfile_options_include_runtime`

## Summary by CodeRabbit

* **New Features**
  * Improved .NET tool support for lockfile-specific options and runtime configuration.

* **Tests**
  * Added test coverage for .NET lockfile handling scenarios.

### DIFF
--- a/src/plugins/core/dotnet.rs
+++ b/src/plugins/core/dotnet.rs
@@ -8,15 +8,15 @@ use serde::Deserialize;
 use versions::Versioning;
 
 use crate::backend::Backend;
-use crate::backend::VersionInfo;
 use crate::backend::options::BackendOptions;
+use crate::backend::{VersionInfo, platform_target::PlatformTarget};
 use crate::cli::args::BackendArg;
 use crate::cmd::CmdLineRunner;
 use crate::config::{Config, Settings};
 use crate::http::{HTTP, HTTP_FETCH};
 use crate::install_context::InstallContext;
 use crate::parallel;
-use crate::toolset::{ToolVersion, ToolVersionOptions, Toolset};
+use crate::toolset::{ToolRequest, ToolVersion, ToolVersionOptions, Toolset};
 use crate::ui::progress_report::SingleReport;
 use crate::{dirs, env, file, plugins};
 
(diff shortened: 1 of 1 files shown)


risu729:
CI note: `e2e-2` and `e2e-3` are failing in task completion tests because `usage complete-word --shell zsh` is returning `candidate<TAB>description` pairs where the tests expect plain candidates:

- `tasks/test_task_completion`: `mise.toml<TAB>mise.toml`, `mise.usage.kdl<TAB>mise.usage.kdl`
- `tasks/test_task_completion_global_cd`: `alpha<TAB>alpha`, `beta<TAB>beta`, `gamma<TAB>gamma`

This appears unrelated to this PR, which only changes `src/plugins/core/dotnet.rs` lockfile option identity. The same `e2e-2`/`e2e-3` failures are present on #10161, and #10176 appears to be the targeted completion fix. The aggregate `test-ci` job fails only because the e2e group is not green; Windows e2e passed.

PR #10187, opened 2026-05-31
fix(aqua): canonicalize and reject duplicate vars

risu729:
## Summary

- canonicalize Aqua var options from plain keys, nested `vars`, and literal `vars.<name>` backend-option keys into one lock/cache identity
- reject duplicate final Aqua vars with a source-neutral `conflicting aqua var` error instead of picking one spelling
- add unit/e2e coverage for literal dotted-key parsing, duplicate detection, and normal higher-precedence config overrides

## Details

TOML already treats unquoted dotted keys and tables as the same structure: `vars.channel = "stable"` and `[vars] channel = "stable"` both parse as a nested `vars` table. This PR does not distinguish those TOML forms.

The extra spelling mise has to normalize is the literal backend-option key `"vars.channel"`, which can come from backend arg/manual option parsing or persisted raw option maps rather than TOML dotted-key parsing. Aqua now canonicalizes that literal key to the same `vars.channel` lock/cache identity as nested TOML `vars.channel` and the documented plain `channel` var option.

If two entries in the final flattened options map collapse to the same Aqua var, Aqua stops with `conflicting aqua var`. If a higher-precedence config layer overrides a lower-precedence layer before Aqua sees the options, only the winning value reaches the backend, so normal config precedence still applies silently.

## Note

`mise lock` still has existing best-effort behavior that can ignore invalid config/tool request conversion or skip failed per-platform lock resolution while writing successful entries. This PR does not change that broader policy; making invalid in-scope config fatal for `mise lock` should be handled separately.

## Test plan

- [ ] CI (lint, unit, e2e)
- [ ] `mise run test:e2e e2e/backend/test_aqua_vars`

### DIFF
--- a/e2e/backend/test_aqua_vars
+++ b/e2e/backend/test_aqua_vars
@@ -63,6 +63,27 @@ EOF_MISE
 MISE_TIMINGS=1 mise lock --platform "$MISE_PLATFORM"
 assert_succeed "cmp mise.lock nested-vars.lock"
 
+rm -f mise.toml mise.lock
+mkdir -p .config/mise/conf.d
+
+cat <<'EOF_MISE' >.config/mise/conf.d/01-base.toml
+[tools."aqua:example/vars-tool"]
+version = "1.0.0"
+vars.fixture_version = "1.0.0"
+EOF_MISE
+
+cat <<'EOF_MISE' >.config/mise/conf.d/02-override.toml
+[tools."aqua:example/vars-tool"]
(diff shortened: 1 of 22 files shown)


risu729:
@coderabbitai review

risu729:
@coderabbitai review

risu729:
@coderabbitai review

risu729:
@coderabbitai review

PR #10190, opened 2026-06-01
fix(task): honor MISE_CYGDRIVE_PREFIX for Git Bash, not only Cygwin

JamBalaya56562:
## Problem

PR #10147 added Cygwin support to the Windows→Unix PATH conversion, with `MISE_CYGDRIVE_PREFIX` as an escape hatch for a non-default `cygdrive` automount root configured in `/etc/fstab`.

As pointed out in [review feedback](https://github.com/jdx/mise/pull/10147#issuecomment-4589301205) (thanks @lewis-yeung), the `cygdrive` automount mechanism is **also** present in MSYS2 / Git Bash (a Cygwin derivative), where the mount root is likewise configurable in `/etc/fstab` (default `/`, i.e. `/c/...`).

`msys_drive_prefix_for` returned an empty prefix immediately for any non-Cygwin bash and **never read `MISE_CYGDRIVE_PREFIX`**, so a Git Bash user with a custom automount root got a corrupted `/c/...` PATH with no way to fix it — the same class of bug #10147 set out to address, still present for Git Bash.

## Fix

- `is_cygwin_shell` now only selects the **default** prefix when no override is set: empty for Git Bash / MSYS2 (`/c/...`), `/cygdrive` for Cygwin (`/cygdrive/c/...`).
- `MISE_CYGDRIVE_PREFIX` is now honored for **both** shells. The relative-value warning reports the shell-appropriate default.

No behavior change for existing users: Git Bash without an override still emits `/c/...`, Cygwin without an override still emits `/cygdrive/c/...`. The only new behavior is that a Git Bash user *can* now set a custom prefix.

## Tests

- Added `test_maybe_convert_env_for_msys_shell_honors_cygdrive_prefix_for_git_bash` (Git Bash + `MISE_CYGDRIVE_PREFIX=/mnt` → `/mnt/c/...`).
- Existing Git Bash (`/c/...`) and Cygwin (`/cygdrive/...`) tests still guard the defaults.
- Verified the full binary builds and all `path::` unit tests pass; the `#[cfg(windows)]` prefix logic was exercised standalone across all input combinations.

Follow-up to #10147.

---

## Summary by CodeRabbit

* **Documentation**
  * Expanded Windows troubleshooting guide with clearer instructions for configuring cygdrive mount roots and using MISE_CYGDRIVE_PREFIX, including absolute-path requirements and behavior when set to "/".

* **Bug Fixes**
  * Improved Windows PATH conversion for POSIX shells (Cygwin, MSYS2/Git Bash) to honor custom cygdrive prefixes and produce correct converted paths.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -204,8 +204,12 @@ the intended one is used:
 $env:MISE_BASH_PATH = "C:\cygwin64\bin\bash.exe"
 ```
 
-If you changed the `cygdrive` prefix in `/etc/fstab` (the default is `/cygdrive`),
-set `MISE_CYGDRIVE_PREFIX` to match — mise does not read `/etc/fstab`:
+#### Custom `cygdrive` mount root (Cygwin **and** Git Bash / MSYS2)
+
+The `cygdrive` automount mechanism is shared by Cygwin and MSYS2 / Git Bash — both let
+you change the mount root in `/etc/fstab` (Cygwin's default is `/cygdrive`, Git Bash /
+MSYS2's is `/`, i.e. `/c/...`). mise does not read `/etc/fstab`, so if you changed it,
+set `MISE_CYGDRIVE_PREFIX` to match — this works for **either** shell:
 
 ```powershell
 # e.g. for an fstab that mounts drives under /mnt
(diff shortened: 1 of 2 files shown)


JamBalaya56562:
> The comment in `test_maybe_convert_env_for_msys_shell_rejects_relative_cygdrive_prefix` still says "fall back to the default `/cygdrive` instead"...

Good catch — fixed. The comment now states that the fallback is the *shell's* default: `/cygdrive` for the Cygwin binary the test exercises, and an empty prefix (the `/c/...` form) for Git Bash. The assertion is unchanged.

PR #10191, opened 2026-06-01
fix(task): honor task_config.includes order so a local task can override a git::include

vmaleze:
PR #9147 introduced an early last-wins dedup in merge_file_and_config_tasks and, combined with load_tasks_in_dir loading all non-git includes before all git:: includes, made a git:: include always clobber a same-named local task regardless of its position in task_config.includes.

Load includes in their declared order via a single ordered pass (mirroring load_file_tasks) and keep the file-task dedup first-wins, so the earlier include in the list takes precedence uniformly for directory, toml-file, and git:: includes. The TOML [tasks.*] overlay behavior from #9147 is unchanged.

Document the include precedence and add an e2e test covering local-first, git-first, and two same-named local includes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed task include precedence: when multiple includes define the same task name, the last include listed wins.

* **Documentation**
  * Updated docs to clarify include evaluation order and show how to override git-hosted tasks by listing local includes after them.

* **Tests**
  * Added an end-to-end test verifying include-order precedence across local and git-hosted includes.

### DIFF
--- a/docs/tasks/task-configuration.md
+++ b/docs/tasks/task-configuration.md
@@ -690,6 +690,17 @@ For local and monorepo task discovery, mise uses the nearest config file that de
 That means a child config's `includes` replaces both the defaults and any `includes` defined by parent configs for that directory.
 Global config files are loaded independently, so each global config file uses its own `task_config.includes` or the default directories if `includes` is unset.
 
+Entries are evaluated in order, and when more than one include defines a task with the same name the **last** entry in the list wins.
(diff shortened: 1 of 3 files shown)


risu729:
I feel last win is more intuitive as other config like tools or env are last win. I'm just curious but are there any reasons for the first win?

vmaleze:
> I feel last win is more intuitive as other config like tools or env are last win. I'm just curious but are there any reasons for the first win?

There is no real reason for first to win. I just wanted to fix the issue that was introduced were remote tasks would always take precedence over local ones. I will update the PR to have last define wins :)

risu729:
I see, thanks!

PR #10200, opened 2026-06-02
fix(aqua): support override env selectors

risu729:
## Summary
- parse `overrides[].envs` in Aqua registry package overrides
- apply `envs` as an additional override selector alongside `goos`, `goarch`, and supported `variants`
- bump the compiled Aqua registry cache version because the archived override layout changed

## What is `overrides[].envs`?
Aqua registry packages can define `overrides` to replace parts of a package definition for specific runtime targets. The `envs` field is one of Aqua's override selectors: each entry can match a GOOS value such as `windows`, a GOARCH value such as `arm64`, a combined `GOOS/GOARCH` value such as `windows/arm64`, or `all`.

Aqua treats `envs` as an optional selector alongside `goos`, `goarch`, and `variants`: if it is present, at least one listed environment must match the target runtime; if other selectors are also present, they must match too.

Relevant Aqua references:
- https://github.com/aquaproj/aqua/blob/main/pkg/config/registry/package_info.go
- https://github.com/aquaproj/aqua/blob/main/pkg/config/registry/override.go
- https://github.com/aquaproj/aqua/blob/main/pkg/config/registry/supported_envs.go

## Packages using this field
Two live aqua-registry entries show why this matters:

- `apache/tomcat` uses `overrides[].envs: [windows]` to switch the installed file from `catalina.sh` to `catalina.bat` on Windows.
  https://github.com/aquaproj/aqua-registry/blob/main/pkgs/apache/tomcat/registry.yaml
- `eza-community/eza` uses `overrides[].envs: [darwin, windows/arm64]` to route those targets through its `type: cargo` package configuration instead of the default release-asset path.
  https://github.com/aquaproj/aqua-registry/blob/main/pkgs/eza-community/eza/registry.yaml

## Why mise needs support
Before this change, mise parsed and applied Aqua overrides for `goos`, `goarch`, and supported `variants`, but ignored `overrides[].envs`. Env-only overrides therefore never applied. That could make mise choose the wrong file or package configuration for valid Aqua registry packages, especially on Windows and Darwin targets.

Supporting `envs` makes mise's native Aqua backend match Aqua's selector semantics for these registry entries and prevents platform-specific override data from being silently skipped.

## Tests
- `rustfmt --edition 2024 --check crates/aqua-registry/src/types.rs crates/aqua-registry/src/cache.rs`
- `git diff --check`
- `cargo test -p aqua-registry override_envs --no-default-features`

## Summary by CodeRabbit

* **New Features**
  * Added environment selectors for overrides, allowing rules to target specific OS, architecture, or combined "os/arch" identifiers (including an "all" option) for more precise selection of files, URLs, and variants.

* **Chores**
  * Bumped the compiled registry cache layout version; previously compiled caches will be ignored and rebuilt automatically on next use.

### DIFF
--- a/crates/aqua-registry/src/cache.rs
+++ b/crates/aqua-registry/src/cache.rs
@@ -7,7 +7,7 @@ use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::time::{Duration, SystemTime};
 
-const COMPILED_REGISTRY_CACHE_VERSION: &str = "v2";
+const COMPILED_REGISTRY_CACHE_VERSION: &str = "v3";
 
 #[derive(Debug, Clone)]
 pub struct RegistryCache {
--- a/crates/aqua-registry/src/types.rs
+++ b/crates/aqua-registry/src/types.rs
@@ -1965,6 +1965,7 @@ packages:
 packages:
   - url: https://example.com/tool-default
     format: raw
+    complete_windows_ext: false
     overrides:
       - envs:
           - darwin
--- a/crates/aqua-registry/src/types.rs
+++ b/crates/aqua-registry/src/types.rs
@@ -826,6 +826,7 @@ impl AquaOverride {
 
 fn envs_match(envs: &[String], os: &str, arch: &str) -> bool {
     let os_arch = format!("{os}/{arch}");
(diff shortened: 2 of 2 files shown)


risu729:
@coderabbitai review

risu729:
@coderabbitai review

risu729:
@coderabbitai review

PR #10203, opened 2026-06-02
fix(upgrade): show current latest in summary

risu729:
## Summary
- Show a meaningful current value in `mise upgrade` summaries for `latest` requests instead of `(none)` when an installed current value is observable.
- Prefer the backend-reported concrete installed version for `latest`; fall back to the literal `latest` only for the legacy stale real `installs/<tool>/latest/` directory case where the old concrete version is not knowable.
- Keep existing prefix/concrete-version summary behavior intact and add focused e2e coverage for concrete `latest`, stale real `latest/`, lockfile-backed cases, already-up-to-date `latest`, and prefix upgrades.

## Reported Case
Discussion: https://github.com/jdx/mise/discussions/10061

The exact follow-up case is https://github.com/jdx/mise/discussions/10061#discussioncomment-17131240. On `mise 2026.5.16`, `thernstig` upgraded five tools and every summary row showed `(none)`:

```text
npm:vercel (none) → 54.6.1
ruff (none) → 0.15.15
rust (none) → 1.96.0
ty (none) → 0.0.40
uv (none) → 0.11.17
```

The original discussion showed the same symptom for `fzf (none) → 0.73.0`, and the attached `mise doctor` output showed active tool paths under `~/.local/share/mise/installs/<tool>/latest/...` such as `fzf/latest`, `ruff/latest`, `uv/latest`, and `npm-vercel/latest/bin`.

The reproduced local case is a stale real `installs/<tool>/latest/` directory, not a runtime symlink. That state can exist from older fuzzy-version/runtime-symlink behavior. `mise upgrade` can now install the concrete target and replace that real `latest/` directory with a symlink, but the upgrade summary is computed from the pre-upgrade state. Before this PR, `OutdatedInfo::current_version` returned `None` immediately for `ToolRequest::Version { version: "latest" }`, so the summary printed `(none)` even though `installs/<tool>/latest/` existed.

## Previous Behavior
For a `latest` request where the resolved target was not already installed, the summary logic did not inspect installed versions. It returned no current value for `latest` and printed `(none)` in the stale real `latest/` case:

```text
Upgraded 1 tool:
  dummy (none) → 2.0.0
```

`#9973` fixed the adjacent prefix/current-version problem, but intentionally kept `latest` conservative. So this is not a regression introduced by `#9973`; it is a remaining display gap for legacy `latest/` install state. The stale real `latest/` directory itself is legacy broken state from the runtime-symlink/fuzzy-version history addressed by `#9276`.

## Behavior After This PR
For `latest` requests, the summary now reports the best installed current value that the backend can prove:

```text
# concrete previous install exists
Upgraded 1 tool:
  dummy 1.0.0 → 2.0.0

# only a legacy real latest/ directory exists, so the old concrete version is unknown
Upgraded 1 tool:
  dummy latest → 2.0.0

# prefix requests keep reporting the installed concrete version
Upgraded 1 tool:
  tiny 2.0.0 → 2.1.0
```

If both a stale real `latest/` directory and an installed concrete version exist, the concrete version wins. If only the stale real `latest/` directory exists, the summary says `latest`, not a guessed concrete version. If the requested `latest` version is already installed, `mise upgrade` remains a no-op and does not print an upgrade summary.

This keeps the behavior aligned with existing history: `#9276` stopped treating real `latest/` directories as valid concrete resolution targets, and this PR does not restore that. It only makes the already-observable pre-upgrade state visible in the summary.

## Verification
- `MISE_TRUSTED_CONFIG_PATHS=$PWD mise run build`
- PR binary: `CARGO_TARGET_DIR=/home/risu/.cache/cargo-target/mise-409845c58888dcc6 e2e/run_test e2e/cli/test_upgrade_latest_summary_current` passed.
- Without this PR: `CARGO_TARGET_DIR=/nonexistent-mise-pr10203-base e2e/run_test e2e/cli/test_upgrade_latest_summary_current` used `/usr/bin/mise 2026.6.0` and failed at the stale real `latest/` case with `dummy (none) → 2.0.0`.
- CI previously passed the new focused test in `e2e-3`; the remaining `e2e-4` failure is in existing `cli/test_upgrade_parallel_failure` and is tracked in the PR comments as unrelated.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive end-to-end test for `mise upgrade` functionality, validating upgrade behavior across multiple scenarios including `latest` version specifications, lockfile configurations, and proper status reporting.

* **Bug Fixes**
  * Enhanced version detection for tools using `latest` specifications to accurately determine currently installed versions and provide correct upgrade status information.


risu729:
CI note: `e2e-4` failed in `cli/test_upgrade_parallel_failure` because `mise ls --installed tiny` did not find the intentionally successful `tiny` install after the expected `dummy@other-dummy` failure. I inspected the job log and reran the focused test locally both through the normal e2e harness and with `MISE_E2E_DOCKER=1`; both passed on this branch. The new regression test `cli/test_upgrade_latest_summary_current` passed locally via the official `mise run test:e2e` task, and all other CI jobs on this head are passing.

I tried to rerun the failed job, but GitHub rejected it because rerunning Actions jobs requires admin rights on this repository. This appears to be a flaky shard failure rather than caused by this PR.

risu729:
CI update for commit `3003cb170adb21edcef50acfcbd4c7f2e4637bcf`:

- The new regression test passed in CI: `e2e-3` logged `PASS: cli/test_upgrade_latest_summary_current`.
- The only failing test-workflow check is `e2e-4`, which failed in existing `cli/test_upgrade_parallel_failure` with `tiny is not installed` after the intentionally failing parallel `mise up` path. This is the same unrelated shard failure seen on the previous run for this PR.
- I tried rerunning the failed checks with `gh run rerun 26980191841 --failed` and the failed job with `gh run rerun --job 79617583280`; GitHub rejected both rerun attempts.

PR #10225, opened 2026-06-04
fix(task): render task config includes

risu729:
## Summary
- render `task_config.includes` entries with the config-file Tera context before resolving local task include paths or `git::` task sources
- expand `~/` for local task include paths and globs before deciding whether they are absolute or config-root-relative
- add an e2e regression for `{{ config_root }}`, `{{ env.HOME }}`, and `~/` includes

Addresses discussion https://github.com/jdx/mise/discussions/10209.

## Tilde support
I checked the nearby path-like config behavior before adding this. `~/` is already supported for task execution paths such as task `dir`, task `file`, and task sandbox `allow_read` / `allow_write`, plus env path directives that resolve config paths such as `env._.file`, `env._.source`, and `env._.python.venv`.

`task_config.includes` is also a config-provided local path/glob list, so this PR follows the existing local-path convention and expands only leading `~/` for local includes. Remote `git::` includes are still treated as remote source strings after template rendering.

## Template context
`task_config.includes` is rendered while task files are being discovered, before any specific task exists. The available context is therefore the config-file context, not the task runtime context.

Available values include:
- `env`: the pristine process environment plus plain key/value entries from the same config file's `[env]`
- `config_root`: the root directory for the config file that defines `task_config.includes`
- `cwd`: the current working directory captured in the base Tera context
- mise/base fields from the shared template context, such as `mise_bin`, `mise_pid`, `mise_env`, and the XDG home paths
- standard mise Tera functions and filters

Task-only values such as `usage`, a concrete task object, task-level vars/env, and resolved tool information are not available here because includes decide which tasks are loaded.

## Tests
- `cargo fmt --check`
- `git diff --check`
- `bash -n e2e/tasks/test_task_config_includes_templates`
- `mise --cd /tmp x shellcheck -- shellcheck -x /home/risu/.worktrunk/risu729/mise/mise.codex-20260604-092743-8567ab/e2e/tasks/test_task_config_includes_templates`

Not completed locally:
- `cargo check --all-targets` was queued behind existing machine-level Cargo jobs for over 20 minutes, so I stopped only my queued check rather than bypassing the Cargo wrapper.
- `mise run test:e2e e2e/tasks/test_task_config_includes_templates` depends on `build`, so it would have hit the same Cargo queue.

## Summary by CodeRabbit

* **New Features**
  * Task includes now support template expansion (e.g., `{{ config_root }}`, `~`, `{{ env.HOME }}`).

* **Bug Fixes**
  * Improved error handling and propagation when resolving task includes.

* **Tests**
  * Added end-to-end test validating task include template expansion and file globbing.

### DIFF
--- a/.github/workflows/pr-closer.yml
+++ b/.github/workflows/pr-closer.yml
@@ -6,34 +6,12 @@ on:
   workflow_dispatch:
 
 jobs:
-  close-stale-prs:
+  pr-closer:
     runs-on: ubuntu-latest
     permissions:
+      issues: write
       pull-requests: write
+      checks: read
+      statuses: read
     steps:
-      - name: Close stale PRs
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          REPO: ${{ github.repository }}
-        run: |
-          CUTOFF=$(date -u -d '7 days ago' +%Y-%m-%d)
-          gh pr list -R "$REPO" --state open --search "updated:<$CUTOFF -author:jdx -label:keep-open sort:updated-asc" --json number,mergeStateStatus,statusCheckRollup --limit 500 | \
-          jq -r '
-            def failed_check:
-              (.statusCheckRollup | length > 0) and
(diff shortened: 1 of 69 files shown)


risu729:
CI settled cleanly after the follow-up commit `e104043d7`. All checks completed successfully on the latest run, including `windows-e2e`, `lint`, `benchmark`, and aggregate `test-ci`.

PR #10167, opened 2026-05-31
fix(aqua): mirror aqua windows extension semantics

risu729:
## Summary
- model aqua `complete_windows_ext` as tri-state instead of a defaulted boolean
- support registry `windows_ext` and `append_ext` when rendering aqua assets, URLs, and Windows install links
- treat `.exe` and `.jar` assets as already complete, preserve existing extensions when `format` is omitted, and keep version dots from suppressing Windows completion
- centralize the version-aware extension rules in the `aqua-registry` crate so the backend delegates aqua semantics instead of duplicating them

## Aqua behavior checked
Current aqua source and upstream history checked:
- Source: aqua's [`windowsExt`, `completeWindowsExtToAsset`, and `completeWindowsExtToFileSrc`](https://github.com/aquaproj/aqua/blob/0b7030945d0442de65b2aee7967c84405998ee74/pkg/config/windows_ext.go#L55-L110), [`osfile.Ext`](https://github.com/aquaproj/aqua/blob/0b7030945d0442de65b2aee7967c84405998ee74/pkg/osfile/ext.go#L8-L9), and [`appendExt`](https://github.com/aquaproj/aqua/blob/0b7030945d0442de65b2aee7967c84405998ee74/pkg/config/package.go#L392-L402).
- `append_ext` defaults to true. Aqua appends the configured `format` to rendered asset/URL strings when the string does not already look like an archive/compressed asset and `format` is neither empty nor `raw`; `append_ext: false` disables that behavior.
- `complete_windows_ext` is tri-state. Explicit `true` completes Windows asset/URL/file source names, explicit `false` leaves them alone, and an omitted value completes by default except for `github_content` and `github_archive` packages.
- `windows_ext` controls the suffix used by Windows completion. Aqua defaults to `.exe`, but defaults to `.sh` for `github_content` and `github_archive`; registry entries can override it, such as `.bat`.
- `.exe` is hardcoded upstream as the Windows executable suffix. It started with aqua's Windows support work in [aquaproj/aqua#858](https://github.com/aquaproj/aqua/pull/858) for assets and [aquaproj/aqua#874](https://github.com/aquaproj/aqua/pull/874) for URLs/files, both under the broader Windows support request [aquaproj/aqua#850](https://github.com/aquaproj/aqua/issues/850).
[...]

PR #10227, opened 2026-06-04
fix(aqua): avoid double raw windows extensions

risu729:
## Summary
- keep raw Windows aqua assets/URLs with an existing file extension unchanged before applying `windows_ext`
- continue completing extensionless raw Windows assets/URLs as before
- add regression coverage for raw `.bat` asset and direct URL rendering

## Review validation
Validated the remaining CodeRabbit comment from #10167 with a focused reproduction. Before the fix, this allowed registry shape rendered a double extension on Windows:

```yaml
asset: tool-{{.Version}}-{{.OS}}-{{.Arch}}.bat
format: raw
windows_ext: .bat
```

For `1.0.0` on `windows/amd64`, `pkg.asset(...)` produced `tool-1.0.0-windows-amd64.bat.bat` instead of `tool-1.0.0-windows-amd64.bat`. The same issue is observable for direct URLs such as `https://example.com/tool/{{.Version}}/tool.bat`, which would make mise request `tool.bat.bat`.

I also checked the current aqua registry: `sass/dart-sass` uses `windows_ext: .bat`, but it is archive-based rather than this `format: raw` shape. So this is not a current dart-sass regression; it is a valid edge case for raw assets/URLs that the registry schema permits.

## Upstream aqua comparison
Current `aquaproj/aqua` at `1e4b1cf0` does not avoid this edge case. Its `completeWindowsExtToAsset` returns `completeWindowsExt(asset)` unconditionally for `format: raw` after only skipping assets ending in `.exe` or `.jar`.

I verified that with a temporary local reproduction against aqua's `RenderAsset` path: the same `format: raw` / `windows_ext: .bat` asset rendered as `tool-1.0.0-windows-amd64.bat.bat`.

## Tests
- `cargo test -p aqua-registry raw_preserves_custom_windows_ext -- --nocapture`
- `cargo test -p aqua-registry`
- `cargo fmt -p aqua-registry --check`

## Current aqua-registry packages
None. I scanned current `aquaproj/aqua-registry` at `830124e62b` for packages matching the exact risky shape: `format: raw`, Windows extension completion enabled, a custom `windows_ext`, and an asset/URL that already carries that extension. No packages match.

The only current `windows_ext: .bat` package found is `sass/dart-sass`, and it uses archive formats (`zip`/`tar.gz`) rather than `format: raw`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced Windows extension handling for raw-format assets to more conservatively preserve existing file extensions and prevent unintended modifications.

* **Tests**
  * Added test coverage for raw assets with custom extensions to verify they remain unmodified during processing.

### DIFF
--- a/crates/aqua-registry/src/types.rs
+++ b/crates/aqua-registry/src/types.rs
@@ -611,7 +611,7 @@ impl AquaPackage {
     }
 
     /// Apply Windows executable extension to an asset or URL string if appropriate.
-    /// Mirrors upstream aqua's `completeWindowsExtToAsset` decision tree.
+    /// Keeps already-extended raw assets unchanged to avoid doubling custom Windows extensions.
     fn complete_windows_ext_to_asset(&self, s: &str, v: &str, os: &str) -> Result<String> {
         if os != "windows" || s.ends_with(".exe") || s.ends_with(".jar") {
             return Ok(s.to_string());
--- a/crates/aqua-registry/src/types.rs
+++ b/crates/aqua-registry/src/types.rs
@@ -617,7 +617,10 @@ impl AquaPackage {
             return Ok(s.to_string());
         }
         if self.format == "raw" {
-            return Ok(self.complete_windows_ext(s));
+            if self.os_file_ext_is_empty(s, v) {
+                return Ok(self.complete_windows_ext(s));
+            }
+            return Ok(s.to_string());
         }
         if !self.format.is_empty() {
             return Ok(s.to_string());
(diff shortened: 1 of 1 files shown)


risu729:
@coderabbitai review

risu729:
@coderabbitai review

risu729:
Closing this because current upstream aqua does not avoid this edge case either. I checked `aquaproj/aqua` at `1e4b1cf0`: `completeWindowsExtToAsset` still appends the Windows extension unconditionally for `format: raw` after only skipping `.exe` and `.jar`, and a local `RenderAsset` reproduction rendered `tool-1.0.0-windows-amd64.bat.bat`.

I also scanned current `aquaproj/aqua-registry` at `830124e62b` and found no packages matching the exact risky shape (`format: raw` with custom `windows_ext` and an already-extended Windows asset/URL).

PR #9992, opened 2026-05-19
fix(ruby): lock resolved install options

risu729:
## Summary
- include resolved Ruby install options in lock identity, including defaults such as compile mode, installer choice, the active build repo, and precompiled URL
- include source-build options and patch inputs that can change compiled output or fallback output
- add backend-level regression coverage for precompiled, experimental-default, resolved-default, source-build, fallback, and ruby-install paths

## Rationale
Lock option values are part of the artifact identity. Recording resolved defaults keeps existing lockfiles stable if mise defaults change later, including Ruby's planned move toward precompiled binaries by default.

## Classification
Mixed actual lock/install identity and stale-lock invalidation fix.

Actual identity: source-build settings such as installer choice, build repositories, build options, and patch inputs can change the compiled Ruby tree for the same version. Those inputs are not represented by a locked artifact URL.

Stale-lock invalidation: precompiled URL/platform selector settings mostly choose the binary artifact URL/checksum. Once the lock entry is selected, the locked URL identifies the binary, but the selector still belongs in the exact-options key so changing precompiled source or platform settings does not reuse an older lock entry.

## Verification
- `cargo fmt --check`
- `git diff --check`
- `cargo test ruby_lockfile_options`
- GitHub Actions `test` workflow on `0d98b51f1` passed, including unit tests, lint, e2e shards, Windows, and `test-ci`
- GitHub Actions `hyperfine`, `registry`, and `release` workflows on `0d98b51f1` passed or skipped only expected matrix jobs

## Summary by CodeRabbit

* **Improvements**
  * Ruby lockfile resolution now honors explicit precompiled OS/arch overrides, consistently derives precompiled vs source-build behavior from settings, always includes install/build options and optional patch application, and injects precompiled download info when precompiled mode is used; Windows targets return an empty result.

* **Tests**
  * Added deterministic tests and scaffolding covering explicit precompiled, source-build (with opts/patches), experimental/default behaviors, non-current targets, fallback cases, and override-driven precompiled URL resolution.

PR #10239, opened 2026-06-05
fix(erlang): record lockfile fallback mode

risu729:
## Summary
- resolve Erlang precompiled lockfile URLs for Linux, macOS, and Windows without requiring a prior install
- record Erlang default fallback mode in lockfile options so kerl fallback has a distinct lock identity
- keep `erlang.compile=false` strict: missing precompiled artifacts still fail lock resolution instead of falling back
- add an Erlang-only e2e regression for precompiled URL locking and default kerl fallback locking

## Notes
- Follows the lock identity approach from #9992: install-mode options that can affect source fallback output are part of the lock entry identity.
- This PR is separate from #10226, which is now Elixir-only.

## Verification
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Enhanced Erlang lockfile support: platform-specific precompiled URLs and checksums for Linux, macOS, and Windows; improved resolution and persisted metadata to speed locked installs.

* **Tests**
  * Added Linux-only end-to-end test validating Erlang lockfile generation, platform-specific fields, musl handling, and locked-install dry-run behavior.

* **Documentation**
  * Clarified ImageOS → build-target mappings and precompiled resolution fallbacks in Erlang docs.


risu729:
@coderabbitai full review

PR #10243, opened 2026-06-05
fix: nix flake build test failure

okuuva:
Adding the `git::tests::clone_by_sha_does_not_panic` broke nix flake builds as `git` was not available during tests. Adding it to `nativeCheckInputs` fixes the issue.

## Summary by CodeRabbit

* **Chores**
  * Updated build configuration to include Git in the test environment to support validation during the build process.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -17,6 +17,9 @@ rustPlatform.buildRustPackage {
     pkg-config
     rustPlatform.bindgenHook
   ];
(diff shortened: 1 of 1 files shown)


okuuva:
Basically the same issue as with https://github.com/jdx/mise/discussions/10005, only with nix instead of Arch.

PR #10302, opened 2026-06-11
feat(backend): allow version_expr to post-process version_regex results

konono:
## Summary

  - Allow `version_expr` and `version_regex` to work together as a pipeline: regex extracts versions, then expr post-processes them (e.g. sorting)
  - Previously these were mutually exclusive in an `if/else if` chain (#7723). This PR allows combining them, but if there was a reason for keeping them separate, happy to take a different approach.
  - Fix `oc` and `openshift-install` version sorting by adding `sortVersions(versions)` to their registry entries

  ## Motivation

  The upstream HTML directory at `mirror.openshift.com` lists versions lexicographically. `find_match_in_list` returns the last element assuming ascending order, so `latest` resolves to `4.9.9` instead of `4.22.0`.

  A registry-side fix was the only option because:
  - These tools are distributed via `mirror.openshift.com`, not GitHub Releases, so switching to `aqua:` or `github:` is not possible
  - `version_expr` alone can't extract from HTML — expr-lang has no regex capture function, and `version_regex` / `version_expr` were mutually exclusive

  ## Changes

  - `src/backend/version_list.rs`: Move `version_expr` evaluation to run after extraction. When both `version_regex` and `version_expr` are specified, the regex results are passed as a `versions` variable into the expr context. When
  `version_expr` is used alone, it operates on `body` as before. No existing registry entries combine the two, so this is a no-op for all current tools.
  - `registry/oc.toml`, `registry/openshift-install.toml`: Add `version_expr = 'sortVersions(versions)'`

  ## Test plan

  - [x] `cargo test backend::version_list` — 24 tests pass (including new pipeline test)
  - [x] `mise ls-remote oc` returns versions in ascending semantic order
  - [x] `mise latest oc` resolves to `4.22.0` (not `4.9.9`)

  ### Before (current release)
```
  $ mise ls-remote oc | tail -5
  4.9.59
  4.9.6
  4.9.7
  4.9.8
  4.9.9

  $ mise latest oc
  4.9.9
```

  ### After (this PR)
```
  $ target/debug/mise ls-remote oc | tail -5
  4.21.16
  4.21.17
  4.21.18
  4.21.19
  4.22.0

  $ target/debug/mise latest oc
  4.22.0
```

## Summary by CodeRabbit

* **Improvements**
  * oc client now applies explicit sorting when selecting available versions.
  * OpenShift installer now applies explicit sorting when selecting installer versions.
  * Version-processing pipeline can apply configurable post-processing expressions to extracted version lists.
* **Tests**
  * Added unit coverage for version extraction plus post-processing workflows.

### DIFF
--- a/registry/oc.toml
+++ b/registry/oc.toml
@@ -9,6 +9,7 @@ full = "http:oc"
 bin = "oc"
 version_list_url = "https://mirror.openshift.com/pub/openshift-v4/clients/ocp/"
 version_regex = 'href="(\d+\.\d+\.\d+)/"'
+version_expr = 'sortVersions(versions)'
 
 [backends.options.platforms.linux-x64]
 url = "https://mirror.openshift.com/pub/openshift-v4/clients/ocp/{{ version }}/openshift-client-linux-{{ version }}.tar.gz"
--- a/registry/oc.toml
+++ b/registry/oc.toml
@@ -7,9 +7,9 @@ full = "http:oc"
 
 [backends.options]
 bin = "oc"
+version_expr = 'sortVersions(versions)'
 version_list_url = "https://mirror.openshift.com/pub/openshift-v4/clients/ocp/"
 version_regex = 'href="(\d+\.\d+\.\d+)/"'
-version_expr = 'sortVersions(versions)'
 
 [backends.options.platforms.linux-x64]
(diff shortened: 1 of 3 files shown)


jdx:
nice solution!

PR #10305, opened 2026-06-11
feat(task): add os/arch-specific task overrides via [tasks.name.<platform>]

esteve:
This PR adds support for specifying platform-specific tasks and deprecates:

```toml
[tasks.my_task]
run_windows = '<my command>'
```

` in favor of:

```toml
[tasks.my_task.windows]
run = '<my command>'
```

os/arch combos can also be specified as tasks:

```toml
[task.my_task]
run = '<my general command>'

[tasks.my_task."linux/arm64"]
run = '<my arm64-specific command only for Linux>'
```

Windows-specific tasks use https://mise.jdx.dev/configuration/settings.html#windows_default_inline_shell_args by default, but it can be overridden as well. Only `run` and `shell` can be overridden in platform-specific tasks to keep them simple and functionally equivalent to the current task configuration system.

## Release Notes

* **New Features**
  * Added support for platform-specific task configuration using dot-syntax subtables (e.g., `[tasks.name.linux]`). Platform overrides follow a precedence order: most-specific match (`<os>/<arch>`) wins, followed by OS-level (`<os>`), then base configuration.

* **Documentation**
  * Updated task configuration documentation to reflect new platform-specific syntax and deprecation of `run_windows` in favor of platform subtables.

* **Deprecations**
  * `run_windows` is now deprecated; use `[tasks.<name>.windows]` instead.

### DIFF
--- a/docs/tasks/task-configuration.md
+++ b/docs/tasks/task-configuration.md
@@ -37,17 +37,52 @@ run = "echo hello"
 run = ["echo hello"]
 ```
 
-### `run_windows`
+### Platform-specific overrides
 
-- **Type**: `string | (string | { task: string, args?: string[], env?: { [key]: string } } | { tasks: string[] })[]`
+You can define OS-specific or platform-specific variants of `run` and `shell` using
+dot-syntax sub-tables. If a matching platform override exists it takes precedence;
+the base `run` value is the fallback for all unmatched platforms.
 
-Windows-specific variant of `run` supporting the same structured syntax:
+**Platform keys**: `linux`, `macos`, `windows`, or `<os>/<arch>` (e.g. `linux/x64`,
+`linux/arm64`, `macos/x64`, `macos/arm64`, `windows/x64`, `windows/arm64`).
+Task platform matching uses the host OS/architecture, not `MISE_OS`/`MISE_ARCH`
+overrides used for tool binary selection.
+
+**Matching order** (most specific wins, evaluated independently for `run` and
+`shell`):
+
+1. `<os>/<arch>` — exact platform (e.g. `linux/x64`)
+2. `<os>` — all architectures on that OS (e.g. `linux`, `macos`)
+3. Base `run` — everything else
 
 ```mise-toml
 [tasks.build]
 run = "cargo build"
-run_windows = "cargo build --features windows"
+
+[tasks.build.linux]
+run = "cargo build --release"
+
+[tasks.build."linux/x64"]
+run = "cargo build --target x86_64-unknown-linux-gnu"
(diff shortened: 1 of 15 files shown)


jdx:
Not a fan sorry

esteve:
@jdx that's fair, is it because of the implementation itself or the idea of having separate platform-specific tasks? This is the only thing missing from Pixi regarding task running that's preventing us from using mise as a task runner. If there's any other way to do it without changing the code, I'd appreciate any advice. Thanks.

jdx:
I would use MISE_ENV

esteve:
> I would use MISE_ENV

I could be mistaken, but a priori doesn't seem to cover the use case I described. Anyway, I'll look more into it and hopefully I can make it work, thanks for replying.

jdx:
it does, I've done this before. You just set it to MISE_ENV=windows and put the tasks for windows in mise.windows.toml

esteve:
> it does, I've done this before. You just set it to MISE_ENV=windows and put the tasks for windows in mise.windows.toml

Thanks, it's in the direction of what I wanted to do, but not exactly. My goal is to make it more automatic and detect the platform where mise is running to run different commands. However, I think a combination of https://mise.jdx.dev/configuration/environments.html#templates-in-miserc-toml, and `arch()` and `os()` in a Tera template might work. If I can make it work, I'm hopeful I can submit an example to the documentation or update this PR. Anyway, thanks for the guidance, much appreciated.

esteve:
Managed to do something that's more or less what I wanted with the following:

`mise.toml`:

```toml
[task_config]
includes = [
  "mise_tasks.{{ os() }}_{{ arch() }}.toml"
]

[tasks.foo]
run = "echo hello!"
depends = ["bar"]
```

`mise_tasks.linux_arm64.toml`:
```toml
[bar]
run = "echo hello from platform-specific task - linux/arm64"
```

`mise_tasks.linux_x64.toml`:
```toml
[bar]
run = "echo hello from platform-specific task - linux/x64"
```

There may still be some duplication with this approach that I  managed to avoid with the changes in this PR (e.g. I'd have to duplicate the description field), but not that big of a deal in the end.

esteve:
The following is in fact much simpler than https://github.com/jdx/mise/pull/10305#issuecomment-4684822751 :

```toml
[tasks.foo_linux_arm64]
hide = true
run = "echo hello from linux arm64"

[tasks.foo_linux_x64]
hide = true
run = "echo hello from linux x64"

[tasks.foo]
depends = ["foo_{{ os() }}_{{ arch() }}"]
```

PR #10315, opened 2026-06-11
fix(config): default release age cutoff should not disable installed-version resolution

jdx:
## Summary

Fixes the performance regression reported in #10308: upgrading 2026.6.1 → 2026.6.2 took shell startups from ~2.5s to ~65s.

The built-in 24h `minimum_release_age` default (#10279) set a `before_date` on every tool resolution, which flipped `should_filter_installed_versions` on for all non-prefer-offline commands (`mise which`, `mise use`, ...). That disabled the installed-version fast paths, so every invocation fetched remote version lists for the entire toolset — ~75ms shell-startup calls became ~2.5s each.

## The fix

Cutoff resolution now reports where the cutoff came from (`BeforeDateSource::Provided/Explicit/Default`):

- The **built-in default** only gates which versions *remote* resolution may pick — its actual purpose of delaying brand-new releases at install time. It no longer suppresses installed-version fast paths, so resolution stays on-disk when an installed version satisfies the request.
- **Explicit cutoffs** (`--minimum-release-age`, a per-tool `minimum_release_age` option, or the `minimum_release_age` setting) keep their date-aware resolution semantics from #9269, since opting in was a deliberate choice there.

`ResolveOptions` gains a `before_date_from_default` flag (set via a shared `apply_before_date_for_tool` helper, also visible in trace output as `before_date=... (default)`), and the two `should_filter_installed_versions` gates in `resolve_version`/`resolve_prefix` now require an explicit cutoff.

## Tests

- **e2e `test_which_no_remote_fetch`** pins the regression using aqua-backed jq (subject to the built-in default): installed match + fuzzy request must resolve from disk with zero remote version fetches. Verified it **fails on unfixed code** (which resolved remotely to a non-installed jq) and passes with the fix.
- **e2e `test_hook_env_no_remote_fetch`** guards the prefer-offline paths: a spy copy of the dummy plugin records a marker if its version-listing scripts ever run during `mise hook-env`, with and without a release-age cutoff, across `latest`/prefix/exact requests. Verified it fails if hook-env is removed from the prefer-offline list.
- New unit test for cutoff source provenance; all 1121 unit tests, related e2e suites (`test_install_before*`, `test_install_prefer_offline_no_refresh`, `test_ls_remote`, `test_upgrade*`, `test_outdated`, `test_which`), and lint pass.

Manually verified both directions with an aqua tool: default cutoff → local resolution, no fetch; explicit `MISE_MINIMUM_RELEASE_AGE=24h` → date-aware remote resolution (unchanged).

Fixes #10308

---

> [!NOTE]
> **Medium Risk**
> Changes core version-resolution behavior for all tools on aqua-like backends; explicit minimum_release_age semantics are preserved but the default path is subtly different from 2026.6.2.
> 
> **Overview**
> Fixes a performance regression (#10308) where the built-in 24h `minimum_release_age` default treated every resolution as date-filtered, skipped installed-version fast paths, and triggered remote version-list fetches on commands like `mise which` and `mise hook-env`.
> 
> Cutoff resolution now records **provenance** via `BeforeDateSource` and `ResolveOptions.before_date_from_default`. The built-in default still applies `before_date` for remote picks at install time, but **`filters_installed_versions()`** only runs for explicit cutoffs (CLI, setting, or per-tool option)—not for the default—so fuzzy requests can resolve from disk when an install already matches.
> 
> Resolution paths use shared **`apply_before_date_for_tool`** instead of overwriting `before_date` blindly; upgrade warnings use the same helper. **E2e** tests assert `mise which` and `mise hook-env` do not refetch remote versions when a local install satisfies the request.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0d075c09be21f6f4e391eefaad05f00d7af48667. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unnecessary remote version-list fetches for `hook-env` and `which` when an installed tool already satisfies the request, improving reliability and performance.
  * Clarified how the minimum-release-age cutoff is applied so default vs. explicit settings are distinguished, yielding more predictable version resolution and warnings.

* **Tests**
  * Added end-to-end regression tests verifying no remote fetches occur in these scenarios.


jdx:
Addressed the Greptile finding: the upgrade hidden-version warning path now uses `apply_before_date_for_tool` instead of reconstructing `ResolveOptions` with the provenance-less resolver, so `before_date_from_default` stays consistent there too (0d075c09b).

jdx:
Re the hyperfine flags (`x -- echo` +14%, `ls` +12%, both marked inconclusive): ran a controlled A/B on one machine — isolated env, 50 runs with 5 warmups, binaries built from merge-base main (4f7e26f15) vs this branch:

| Command | main | PR |
|---|---|---|
| `mise ls` | 24.1 ± 1.3 ms | 22.0 ± 0.6 ms |
| `mise x -- echo hi` | 40.2 ± 1.5 ms | 38.0 ± 1.2 ms |
| `mise hook-env` | 21.7 ± 1.0 ms | 21.3 ± 0.7 ms |

No regression (the deltas favor the PR slightly, likely noise). The CI baseline is the released 2026.6.2 binary rather than merge-base, so its delta also includes everything on main since the release.

Issue #137, opened 2023-02-13
create readme demo


jdxcode:
If I can manage to get #72 and #133 complete, it would be nice to get a demo into the readme. Perhaps using https://www.screen.studio/ That outputs video I think, but maybe I can turn it into a gif if the results look ok.

I think it would be good to keep it right at 30 seconds. Perhaps showing a few things:

- installing node@20 from the home directory
- settings node@20 as the global version
- entering a project directory
- setting project to use node@18 and prompting for install
- cloning an existing project with a large `.tool-versions` setup
- running `rtx install --all`

PR #10325, opened 2026-06-12
feat(github): add `matching` and `matching_regex` asset options

devnulled:
# feat(github): add `matching` and `matching_regex` asset options

Implements the proposal in [#10318](https://github.com/jdx/mise/discussions/10318). This is a step toward `github`-backend parity with ubi so ubi can eventually be retired ([#7464](https://github.com/jdx/mise/discussions/7464)).

## Summary

- Adds `matching` (substring) and `matching_regex` (regex) asset filters to the github / gitlab / forgejo backends, ported from ubi.
- Applied as a pre-filter before platform autodetection, so one config stays portable across OS/arch, unlike [`asset_pattern`](https://mise.en.dev/dev-tools/backends/github.html#asset-pattern), which replaces autodetection and forces you to template every platform's asset name by hand.
- The filter also scopes verification (checksums + SLSA provenance) so a multi-binary release can't verify one binary against another's digest.
- `matching`/`matching_regex` round-trip to the lockfile and are install-time keys, so a stale cached filter never silently overrides `mise.toml`.

```toml
[tools]
# oxc ships oxlint + oxfmt as separate per-platform assets; `matching` selects
# oxlint on every OS/arch without a platform-specific asset_pattern.
"github:oxc-project/oxc" = { version = "apps_v1.69.0", matching = "oxlint", rename_exe = "oxlint" }
```

## Motivation

Some repos ship multiple binaries as separate per-platform assets in one release, none named after the repo, so neither the shortest-name tiebreak ([#9358](https://github.com/jdx/mise/discussions/9358)) nor the repo-name preference ([#10008](https://github.com/jdx/mise/pull/10008)) can portably pick the one you want:

- `oxc-project/oxc` ships `oxlint` + `oxfmt`
- `bazelbuild/buildtools` ships `buildifier` + `buildozer` + `unused_deps`
- `grpc-ecosystem/grpc-gateway` ships `protoc-gen-grpc-gateway` + `protoc-gen-openapiv2`

`asset_pattern` can target one, but it discards autodetection so the config stops being portable. `matching`/`matching_regex` narrow the candidates and let autodetection still pick the right OS/arch, so the same config works everywhere.

## Verification Consistency (Checksums + SLSA Provenance)

The filter also scopes verification, so a multi-binary release can't verify one binary against another's digest:

- Checksums and GitHub attestations already follow the selected asset's name/digest.
- SLSA provenance previously did not. `pick_best_provenance` scored provenance files by platform only, over *all* assets, so a release with per-binary `*.intoto.jsonl` files could attach the wrong one (the tiebreak is a stable sort with no name fallback, so the wrong pick was positional). This PR threads `matching`/`matching_regex` into the picker and all three `pick_best_provenance` call sites, with one exception: if the filter excludes every provenance file, it falls back to the full set. That keeps a single shared provenance file (goreleaser's `multiple.intoto.jsonl`), which attests every artifact but carries no binary name, in play instead of silently skipping verification.

The two "empty result" cases are handled differently so a bad pattern can't turn into a silently skipped verification:

- **Invalid `matching_regex`**: hard error on the binary path (the error names the pattern); the provenance picker returns `None` instead of falling back.
- **Valid but over-narrow filter** (excludes every provenance file): falls back to the full set, so a shared provenance file is used and verification still runs.

The regex is compiled once and cached on the picker, and both checks read that same cached result, so whether the pattern is valid can't differ between the binary and provenance paths. The install path that skips binary selection (when the release URL is already in the lockfile) validates the pattern up front via `asset_matcher::validate_matching_regex` — otherwise a bad pattern could reach the picker, get `None`, and be read as "no provenance found," silently skipping SLSA.

*(SLSA provenance gap caught in review on [#10318](https://github.com/jdx/mise/discussions/10318); thanks @jdx.)*

## Fail-closed on an invalid `matching_regex`

A syntactically invalid `matching_regex` is a hard error wherever the filter actually applies, so it can't silently degrade into the wrong asset or an empty result:

- **Install** fails up front, naming the pattern (as above).
- **`mise lock`** (cross-platform generation) is best-effort — a platform with no matching asset is skipped, not fatal. But `resolve_lock_info` caught the invalid-regex error the same way and returned an empty `PlatformInfo`, so the tool was written to the lockfile with its options but no url/checksum and miscounted as a successful entry. It now validates up front and fails closed, so the platform is skipped (nothing written) instead.
- **Ignored when superseded:** both validations are gated on `asset_pattern` being unset for the target. When `asset_pattern` is set it replaces the filter entirely, so an invalid `matching_regex` is silently ignored, not validated — mise doesn't error on an option it won't act on.

## Backend Scope

`github`, `gitlab`, and `forgejo` share the implementation (common `asset_matcher` + three backend-specific `resolve_*_asset_url_for_target` functions reading through shared `GitBackendOptions` accessors). Provenance/attestation is GitHub-only: gitlab/forgejo bail before `pick_best_provenance`, so there's no provenance path to change there. `test_matching_plumbing_parity_across_git_backends` locks the three paths together so they can't drift.

Tool options are a flattened free-form map, so the rest of the CLI (parse, resolve, `outdated`/`upgrade`, doctor, schema) handles them generically. The only per-backend sites that hardcode option names for asset selection (`install_time_option_keys`, `lockfile_options`, the `asset_pattern` precedence guards) are exactly the three this PR touches.

## Install-Path Keying

`matching`/`matching_regex` are lockfile + install-time keys but not part of the install directory (kept as tool name + version). The supported way to install multiple binaries from one repo is `tool_alias`: each alias is a distinct tool with its own install dir. There's one drawback: reusing the bare `github:owner/repo` string with two different `matching` values resolves to the same dir. That case is called out with a warning in the backend docs.

## Testing

Two layers: fast unit tests for the asset-selection and provenance-picker logic (no network, driven by fixtures from real release JSON), plus e2e tests that install from real multi-binary releases across all three backends.

- [x] **Unit: 31 new tests** (28 in `asset_matcher`, 3 in `github.rs`, including 6 for the SLSA provenance picker). Fixtures from real releases (oxc, bazel buildtools, grpc-gateway) cover happy path, cross-platform portability, AND semantics, autodetection still applied within the narrowed set, invalid-regex hard error, over-narrow fallback, shared-prefix ordering regression, lockfile round-trip, and the malformed-regex-returns-`None` provenance guard.
- [x] **E2E: 5 new test files.** `test_github_matching` and `test_github_tool_alias_matching` run against real multi-binary releases (`oxc@apps_v1.69.0`, `bazelbuild/buildtools@7.1.2`). Each asserts the failure *reason* via an expected substring, so a test can't pass on an unrelated failure. The tool-alias test exercises the SLSA fix end-to-end and the lockfile round-trip. `test_github_matching` also asserts `asset_pattern` precedence over an *invalid* `matching_regex` (the install fails on the pattern path, not regex compilation — proving an ignored regex isn't validated), and `test_github_matching_lock` asserts `mise lock` skips a tool whose `matching_regex` is invalid instead of writing a url-less lockfile entry.
- [x] `test_gitlab_matching` and `test_forgejo_matching` exercise the gitlab/forgejo resolve paths against their real fixtures. (forgejo is committed disabled like the existing `test_forgejo`, for known Forgejo API flakiness; the unit parity test covers it meanwhile.)

> **Note for reviewers:** the only *new* external download dependency is `oxc-project/oxc`. (`bazelbuild/buildtools` is already used by the ubi `test_upgrade`.) Happy to drop the oxc assertions or add a synthetic multi-binary fixture to `jdx/mise-test-fixtures` if you'd prefer to minimize that surface.

## Docs

New `matching`/`matching_regex` sections in the github backend docs (with the `asset_pattern` precedence rule and verification-consistency note), shared-options updates to gitlab/forgejo docs, and a migration note in the ubi docs.

## Behavior & Semantics

A few decisions we made on purpose:

- **Filter before autodetection, not after.** ubi applies substring `matching` only as a tiebreaker *after* OS/arch filtering; we apply it *before*. This keeps one config portable across platforms and avoids the wrong-shared-prefix-binary problem from [ubi #137](https://github.com/houseabsolute/ubi/issues/137) / [mise #6611](https://github.com/jdx/mise/discussions/6611).
- **`matching` and `matching_regex` are AND, and `matching` is case-sensitive.** `matching_regex` supports inline `(?i)` if you want case-insensitive; if both options are set, an asset has to satisfy both.
- **`asset_pattern` wins.** When it's set, `matching`/`matching_regex` are ignored for both the binary and its provenance — silently, so even an invalid `matching_regex` is not validated or reported. That keeps a contradictory config from pointing provenance at a different binary than `asset_pattern` picked.

*(Kept this short — happy to add more on any of it, like the full ubi comparison or why these options are top-level instead of per-platform.)*

---

## Summary by CodeRabbit

* **New Features**
  * Added `matching` (substring) and `matching_regex` (regex) options for GitHub, GitLab, and Forgejo to narrow release asset candidates while preserving platform/OS/arch autodetection.

* **Documentation**
  * Expanded backend docs with usage examples, precedence rules (explicit `asset_pattern` wins), case/validation notes, migration guidance, and a warning about install-path/keying requiring distinct `tool_alias` to avoid overwrites.

* **Bug Fixes**
  * Invalid regexes now fail deterministically; lock generation skips unresolved entries.

* **Tests**
  * Added e2e tests for matching, regex, precedence, lockfile behavior, and multi-binary alias installs.

### DIFF
--- a/docs/dev-tools/backends/forgejo.md
+++ b/docs/dev-tools/backends/forgejo.md
@@ -139,7 +139,7 @@ mise install forgejo:user/repo
 ```
 
 ::: tip
-The autodetection logic is implemented in [`src/backend/asset_matcher.rs`](https://github.com/jdx/mise/blob/main/src/backend/asset_matcher.rs), which is shared by the Forgejo, GitHub and GitLab backends.
+The autodetection logic is implemented in [`src/backend/asset_matcher.rs`](https://github.com/jdx/mise/blob/main/src/backend/asset_matcher.rs), which is shared by the GitHub, GitLab, and Forgejo backends.
 :::
 
 ### `asset_pattern`
@@ -150,6 +150,51 @@ Specifies the pattern to match against release asset names. This is useful when 
 [tools]
 "forgejo:user/repo" = { version = "latest", asset_pattern = "tool_*_linux_x64.tar.gz" }
 ```
+
+### `matching`
+
+Narrows asset selection to names containing the given substring, **while keeping platform autodetection**. Unlike [`asset_pattern`](#asset_pattern) (which replaces autodetection entirely), `matching` only refines the candidate set — autodetection still chooses the correct OS/arch from the narrowed list, so a single config stays portable across platforms.
+
+This is the option to reach for when a repository ships **multiple binaries as separate per-platform assets** and autodetection can't tell which one you want.
+
+```toml
+[tools]
+# When a release ships several binaries per platform (e.g. `mytool-cli` and
+# `mytool-server`), matching picks one on every OS/arch without hardcoding a
+# platform-specific asset_pattern.
+"forgejo:user/repo" = { version = "latest", matching = "mytool-cli" }
+```
+
+Tool options can also be passed inline on the command line using `[key=value]` syntax:
+
+```sh
+mise use "forgejo:user/repo[matching=mytool-cli]"
+```
+
+`matching` is a case-sensitive substring test, so a value that is also a substring of another asset's name (e.g. `matching = "tool"` when both `tool-*` and `tool-extras-*` are published) won't uniquely select your binary. Use [`matching_regex`](#matching_regex) with an anchor when you need a precise match.
+
+If [`asset_pattern`](#asset_pattern) is also set, it takes precedence and `matching`/`matching_regex` are ignored — `asset_pattern` replaces autodetection entirely, so there is no candidate set left for them to narrow. They are ignored silently: when `asset_pattern` is set, a `matching_regex` is never consulted and an invalid one is not reported, since mise does not error on a superseded option.
+
+### `matching_regex`
+
+Like [`matching`](#matching), but the asset name must match the given regular expression. Use this when a substring isn't selective enough. The match is case-sensitive; use an inline `(?i)` flag for case-insensitive matching.
+
+```toml
+[tools]
+"forgejo:user/repo" = { version = "latest", matching_regex = "^mytool-cli-" }
+```
+
+If both `matching` and `matching_regex` are set, an asset must satisfy **both** (logical AND)
+to remain a candidate.
+
+::: warning
+`matching`/`matching_regex` are **not** part of the install path — it is keyed by the tool
+name (`user/repo`, or a `tool_alias`) and version. To install two binaries from the same
+release, give each its own [`tool_alias`](/dev-tools/backends/github.html#multiple-assets-from-the-same-release)
+so they get distinct install directories; reusing the same `forgejo:user/repo` string with
+different `matching` values resolves to the same directory and the second install overwrites
+the first.
+:::
 
 ### `version_prefix`
 
--- a/docs/dev-tools/backends/github.md
+++ b/docs/dev-tools/backends/github.md
@@ -44,7 +44,7 @@ mise install github:user/repo
 ```
 
 ::: tip
-The autodetection logic is implemented in [`src/backend/asset_matcher.rs`](https://github.com/jdx/mise/blob/main/src/backend/asset_matcher.rs), which is shared by both the GitHub and GitLab backends.
+The autodetection logic is implemented in [`src/backend/asset_matcher.rs`](https://github.com/jdx/mise/blob/main/src/backend/asset_matcher.rs), which is shared by the GitHub, GitLab, and Forgejo backends.
 :::
 
 ### `asset_pattern`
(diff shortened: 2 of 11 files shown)


devnulled:
Heads up: the failing `e2e-0` / `e2e-1` checks are unrelated to this PR — these are `usage` shell-completion tests, and this branch touches no completion code or `mise.usage.kdl`.

**Root cause:** `mise.toml` pins `usage = "latest"`, and `usage 3.5.0` changed zsh completion output (an extra tab when the description equals the value). `e2e/tasks/test_task_completion` and `test_task_completion_global_cd` still expect the single-tab format:

```
expected 'mise.toml\tmise.toml'
but got  'mise.toml\t\tmise.toml'
```

Since this branch is rebased onto current `main` and modifies no completion files, the failure reproduces independently of this change — it's the `usage` version, not this PR. `main`'s CI was last green before `usage 3.5.0` became the resolved `latest`, so a fresh run on `main` would hit the same thing.

The fix (update the completion-test expectations for usage 3.5, or pin `usage`) belongs on `main` rather than in this feature PR. Happy to open that separately if useful — otherwise I'll rebase here once it's resolved on `main`.

*Comment posted by an AI coding assistant.*

devnulled:
**Root cause found** for the `e2e-0` / `e2e-1` failures (still unrelated to this PR — for whoever picks it up):

`usage` **v3.5.0** intentionally changed its zsh completion output format ([jdx/usage#666](https://github.com/jdx/usage/pull/666), release notes say it was "surfaced via jdx/mise tasks"):

> The completion now builds its own display column and calls `compadd` directly … with descriptions padded for column alignment. Regenerate zsh completions to pick up the fix.

Concretely, completion lines went from `value\tdescription` to `value\t\tdescription` (an extra tab):

```
usage 3.5.0 emits:   mise.toml\t\tmise.toml
tests still expect:  mise.toml\tmise.toml
```

[#10313](https://github.com/jdx/mise/pull/10313) ("require usage 3.5") bumped the requirement and regenerated `completions/_mise` + `mise.usage.kdl`, but didn't update the **inline expected strings** in `e2e/tasks/test_task_completion` and `e2e/tasks/test_task_completion_global_cd`. Those tests install `usage = "latest"` (now 3.5.0) and assert the old single-tab format, so every assertion in both files fails.

Reproduces on a clean `main` checkout with usage 3.5.0 installed, so it's not introduced by this branch (which touches no completion code).

**Likely fix:** update the expected strings in those two test files to the new double-tab format (`value\t\tdescription`) — the same change applies to every assertion in both.

*Comment posted by an AI coding assistant.*

PR #10360, opened 2026-06-12
feat(config): load safe mise.toml files without trust

jdx:
## Summary

`mise.toml` files that cannot execute code at load time or change mise's behavior without an explicit user action no longer require trust. A config is exempt from the trust gate only when:

- it contains no Tera template syntax — checked both on the raw body (`{{`, `{%`, `{#`) **and** on every decoded TOML string (keys/values, any depth) so escaped delimiters can't slip through — and
- its top-level keys are limited to `min_version`, `[tasks]`, and `[tools]` entries whose values are plain version strings (or arrays of them)

Tools install and tasks run only on explicit commands (`mise install`, `mise run`), so loading these configs executes nothing. Template-free file tasks (`mise-tasks/` scripts and included TOML task files) are likewise loadable without trust — templates are the only thing that can execute while tasks load.

Everything else keeps the existing trust prompt/error: `[env]`, `[hooks]`, `[settings]`, `[alias]`, `[plugins]`, `[task_config]`, `env_file`/`dotenv`, tool entries with options (tables) — since options like `postinstall` run shell commands and `install_env` can alter the install environment — and any file with template syntax. Paranoid mode still requires trust for every file, and ignored configs stay unloaded even when safe.

The check is allowlist-based, so an unrecognized or future config construct fails safe (still prompts for trust).

## Security hardening

- **Escaped-template trust bypass (critical, caught in review).** `is_safe_config_body` originally only scanned the *raw* TOML body, so `tiny = "{{ exec(...) }}"` — which TOML decodes to `{{ exec(...) }}` — passed as "safe" and then executed via Tera when the tool version / task field rendered, with no trust prompt. Fixed by recursively re-checking every decoded TOML string for template syntax. Regression tests added (unit + e2e) for both the tools and tasks paths.
- **Task-include trust hole (pre-existing).** `load_tasks_in_dir` skipped the include trust check whenever *any* config existed at the root, even one loaded without trust (e.g. a plain `.tool-versions`), so a templated `mise-tasks/*.toml` could render `exec()` at `mise tasks` time. Includes now require trust unless a root config was actually trusted (new `config_file::is_path_trusted`).

## Other fixes

- **Tracked safe configs.** `get_tracked_config_files` skipped any path without a trust marker, which safe configs don't create — so their tool pins went missing from `mise ls --all-sources`, `mise upgrade`, and `mise prune`. Safe configs are now loaded there too (`MiseToml::path_is_trust_exempt`). Regression test: `e2e/config/test_trust_safe_config_tracked`.

## Tests

- Unit `test_is_safe_config_body`: tools/tasks safe cases, plus rejection of templates, tool-options, and **escaped** templates in tool versions, task fields, and keys
- `e2e/config/test_trust_safe_config`: safe tools+tasks (config and file tasks) load/run with no trust marker; literal- and escaped-template variants in versions, task fields, and file-task headers, plus tool-options and `[env]`, are all blocked with nothing executing
- `e2e/config/test_trust_safe_config_tracked`: safe config's pins survive tracked-config reload from another directory
- `test_task_untrusted_config_error` / `test_task_monorepo_trust` updated for the new semantics; all other trust e2e tests pass unchanged
- Trust-marker e2e assertions made precise (check for marker *files*, not just the directory)

## Summary by CodeRabbit

* **New Features**
  * Safe config files (only min_version, plain tool versions, and tasks without templates/options) auto-load without a trust prompt.

* **Documentation**
  * Help text, man page, CLI docs, and FAQ updated to define which configs auto-load and which require trust.

* **Refactor**
  * Trust and load logic enhanced to recognize safe configs and detect encoded/escaped template markers.

* **Tests**
  * Added unit and E2E tests for safe-config loading, tracked safe configs, escaped-template edge cases, and unsafe cases requiring trust.

---

> [!NOTE]
> **High Risk**
> Changes when untrusted config is parsed and when templates can run at load time; mistakes could allow code execution without a trust prompt.
> 
> **Overview**
> **Untrusted `mise.toml` files that only pin tools and define tasks can now load without `mise trust`**, when their content is allowlisted (`min_version`, plain `[tools]` strings/arrays, `[tasks]` with no tool-option tables) and has no Tera templates. Parsing skips `trust_check` via `is_trust_exempt` / `path_is_trust_exempt`; everything else (`[env]`, hooks, settings, templates, `postinstall`, etc.) still requires trust, and **paranoid** / **ignored** configs stay gated.
> 
> **Security tightening:** decoded TOML strings are scanned for templates so Unicode-escaped `{{ exec(...) }}` cannot bypass a raw-body check. Task includes and `mise-tasks` files only auto-load without trust when template-free; templated `#MISE` headers and included TOML still require trust. **Trusted** vs **trust-exempt** is split with `is_path_trusted` so safe configs cannot vouch for includes, and tracked-config reload no longer drops safe configs missing a trust marker.
> 
> Docs/CLI help and e2e tests document and cover safe load, escaped-template blocks, and tracked-config visibility.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1dd9a7e26b997603fb3fccb097c60e8cab47f3d1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/cli/trust.md
+++ b/docs/cli/trust.md
@@ -13,10 +13,10 @@ discovery paths, fail with an untrusted-config error when it cannot prompt,
 or assume trust in detected CI unless paranoid mode is enabled.
 
 Safe config files do not require trust: files that only contain
-`min_version`, `[tools]` entries with plain version strings, and
-`[tasks]` (no templates and no tool options) are loaded without
-prompting, since nothing in them executes code at load time — tools
-install and tasks run only on explicit commands like `mise install`
+`min_version`, `[tools]` entries with plain version strings (or arrays
+of them), and `[tasks]` (no templates and no tool options) are loaded
+without prompting, since nothing in them executes code at load time —
+tools install and tasks run only on explicit commands like `mise install`
 or `mise run`.
 
 ## Arguments
--- a/docs/cli/trust.md
+++ b/docs/cli/trust.md
@@ -13,9 +13,11 @@ discovery paths, fail with an untrusted-config error when it cannot prompt,
 or assume trust in detected CI unless paranoid mode is enabled.
 
 Safe config files do not require trust: files that only contain
-`min_version` and `[tools]` entries with plain version strings (no
-templates and no tool options) are loaded without prompting, since
-nothing in them can execute code or change mise's behavior.
+`min_version`, `[tools]` entries with plain version strings, and
+`[tasks]` (no templates and no tool options) are loaded without
+prompting, since nothing in them executes code at load time — tools
+install and tasks run only on explicit commands like `mise install`
+or `mise run`.
 
 ## Arguments
 
--- a/docs/cli/trust.md
+++ b/docs/cli/trust.md
@@ -12,6 +12,11 @@ parsing `mise.toml`. Without trust, mise may prompt, skip the config in some
 discovery paths, fail with an untrusted-config error when it cannot prompt,
 or assume trust in detected CI unless paranoid mode is enabled.
 
+Safe config files do not require trust: files that only contain
+`min_version` and `[tools]` entries with plain version strings (no
+templates and no tool options) are loaded without prompting, since
+nothing in them can execute code or change mise's behavior.
+
 ## Arguments
 
 ### `[CONFIG_FILE]`
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -241,7 +241,11 @@ mise upgrade --bump node
 
 ## My config file is being ignored / `mise trust` issues
 
-mise requires you to trust config files that were not created by you. Common issues:
+mise requires you to trust config files that were not created by you. Safe config files —
+those that only contain `min_version` and `[tools]` entries with plain version strings (no
+templates and no tool options) — are loaded without trust, since nothing in them can execute
+code or change mise's behavior. Everything else (env vars, tasks, hooks, settings, aliases,
+templates, tool options) requires trust. Common issues:
 
 - **Accidentally denied trust**: If mise prompted you to trust a file and you said no, it gets
   added to the ignore list. Check the `ignored-configs` directory in your
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -242,8 +242,8 @@ mise upgrade --bump node
 ## My config file is being ignored / `mise trust` issues
 
 mise requires you to trust config files that were not created by you. Safe config files —
-those that only contain `min_version`, `[tools]` entries with plain version strings, and
-`[tasks]` (no templates and no tool options) — are loaded without trust, since nothing in
+those that only contain `min_version`, `[tools]` entries with plain version strings (or
+arrays of them), and `[tasks]` (no templates and no tool options) — are loaded without trust, since nothing in
 them executes code at load time: tools install and tasks run only on explicit commands like
 `mise install` or `mise run`. Everything else (env vars, hooks, settings, aliases, templates,
 tool options) requires trust. Common issues:
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -242,10 +242,11 @@ mise upgrade --bump node
 ## My config file is being ignored / `mise trust` issues
 
 mise requires you to trust config files that were not created by you. Safe config files —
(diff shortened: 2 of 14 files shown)


jdx:
Thanks for the reviews — addressed in c70a4dc:

- **Escaped-template trust bypass (CodeRabbit, critical).** Confirmed real — `tiny = "{{ exec(...) }}"` decoded to a live template and ran without trust. `is_safe_config_body` now recursively re-checks every *decoded* TOML string (keys + values, any depth) for template syntax, not just the raw body. Unit + e2e regressions added for the tools and tasks paths (and an escaped key).

- **Tracked safe configs skipped (Cursor, medium).** Fixed — `get_tracked_config_files` now also loads safe trust-exempt configs (`MiseToml::path_is_trust_exempt`), so their tool pins remain visible to `mise ls --all-sources` / `upgrade` / `prune`. Regression test: `e2e/config/test_trust_safe_config_tracked`.

- **Trust exemption not shared / misleading "no tasks" error (Cursor, medium).** The tasks commit (46e02fd) makes tasks-only configs load, so `mise run <task>` from a safe config now works instead of erroring; `test_task_untrusted_config_error` was updated to assert this and to keep the trust error for an unsafe (`[env]`) config.

- **Trust-marker assertion false negative (Greptile, P2).** The e2e checks now assert no marker *files/symlinks* exist under `trusted-configs` (via `find`), instead of a bare directory-existence check.

- **`[tools]` array form in docs (CodeRabbit).** Doc text (CLI help, FAQ, regenerated man/usage) now says "plain version strings (or arrays of them)".

- **Double TOML parse (Greptile, P2).** Left as-is intentionally: the safety check parses to `toml::Value` and the main load deserializes into `MiseToml` (different targets), and the published hyperfine numbers show the load path within noise (≈1.03×). Happy to fold them into one parse if you'd prefer.

jdx:
Addressed the task-include variant in c01306d (Cursor, high — "Task includes skip decoded templates").

Confirmed it was real and reproduced it: a `mise-tasks/ci.toml` field and a `#MISE` script header both written with escaped delimiters (`{{ exec(...) }}`) decoded to live templates and executed during `mise tasks` without trust — the same class of bug as the `is_safe_config_body` one, but in `task_include_requires_trust`, which only scanned raw text.

Fix: `task_include_requires_trust` now also rejects template syntax found in the *decoded* TOML — the whole file for `.toml` includes, and the `#MISE` header values for scripts (the only part parsed and rendered at load). The `#MISE` header extraction is refactored out of `Task::from_path` into a shared `parse_mise_header_toml` so the trust check and the loader agree on exactly what gets parsed, and it reuses the recursive `toml_value_has_template` helper from the earlier fix.

Tests: unit `test_file_has_decoded_template` (both file types, plain + escaped) and e2e `e2e/tasks/test_task_include_trust_escaped` (non-paranoid; verifies escaped `.toml` and `#MISE` includes are blocked with nothing executing, and a plain file task still loads untrusted).

jdx:
Two follow-ups from the latest review (1dd9a7e):

- **`e2e/tasks/test_task_include_trust_escaped` script-header case (correct, minor).** Good catch — case 2 used a literal `{{` template, so the raw-text gate rejected it and the decoded-`#MISE`-header path wasn't actually exercised. Switched it to escaped `{{` delimiters via `printf`, so it now locks down `file_has_decoded_template` for the script path. (Case 1, the `.toml` include, already used escapes.)

- **`parse_mise_header_toml` header regex (src/task/mod.rs, flagged Major) — not a regression.** The regex is a verbatim copy of the one previously inline in `Task::from_path` (`r"^(?:#|//|::)(?:MISE| ?\[MISE\]) (...)$"`), and `test_from_path_toml_headers` (which includes a `# MISE alias="b"` line) still passes, so header parsing behavior is unchanged by this refactor. If `# MISE`/`// MISE`/`:: MISE` (space-separated, no brackets) should be accepted, that's a pre-existing matter independent of this PR and best handled separately to avoid scope creep here.

PR #10387, opened 2026-06-13
fix(task): join wrapped command lines in task output header

JamBalaya56562:
## Problem

When a task''s run script starts with a command wrapped across multiple physical lines with backslash continuations, the task header shows only the first physical line (ending in `\`) and then glues the task''s extra CLI args onto that dangling backslash. Reported in #10083:

```
[foo] $ echo long_command \ --extra args
```

## Root cause

`exec_script` (`src/task/task_executor.rs`) built the displayed command by taking only the first non-boilerplate **physical** line via `script.lines().find(...)`. For a backslash-wrapped command that line ends in `\`, and the header is then formatted as `format!("$ {display_script} {args_str}")`, gluing the extra CLI args right after the dangling backslash. The existing `trunc()` only ellipsizes the first line, so it can''t fix this.

## Fix

Extract `display_first_command()`, which:

- skips leading shebang / blank / `set ...` boilerplate (same as before), and
- joins backslash-continued lines into one logical line, so the header shows the full command (still truncated to terminal width by `trunc()`, and shown in full under CI / `task.show_full_cmd`).

Execution is unchanged — only the displayed `$ ...` line is affected.

### Literal trailing backslash

A trailing backslash with **no following line** is treated as literal data (it can''t be a continuation) and is kept verbatim, so e.g. a Windows path `echo C:\tmp\` is displayed as-is rather than being altered. Only genuine multi-line continuations are merged. (The remaining ambiguity — a literal `\` at the end of a line that *is* followed by another line — is still treated as a continuation, which is acceptable for a display-only string and noted in the doc comment.)

## Testing

- 8 unit tests for `display_first_command` (plain, boilerplate-skipping, multi-line join, join after boilerplate, literal trailing `\`, Windows-path trailing `\`, all-boilerplate, and a header-with-args case asserting no `\ ` sequence).
- `cargo fmt --all -- --check` ✅

Addresses #10083

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task command header display to correctly handle multi-line commands with continuations and edge cases such as Windows paths and shebangs.

* **Tests**
  * Added comprehensive unit tests to validate task header display behavior across various scenarios.

### DIFF
--- a/src/task/task_executor.rs
+++ b/src/task/task_executor.rs
@@ -68,6 +68,46 @@ fn resolve_task_sandbox_path(p: &Path, task_base: Option<&Path>) -> PathBuf {
     }
 }
 
+/// Build the single-line command shown in a task's header (the `$ ...` line).
+///
+/// Skips leading shebang/blank/`set ...` boilerplate so the first real command is
+/// shown, and joins backslash-continued lines into one logical line. Without the
+/// join, a command wrapped across physical lines would display only its first line
+/// ending in `\`, and any extra CLI args would be glued onto that dangling
+/// backslash (e.g. `$ echo foo \ --bar`). Returns the whole script if it contains
+/// only boilerplate.
+///
+/// A trailing backslash with no following line is left untouched, so a literal
+/// trailing backslash that is data rather than a continuation (e.g. a Windows path
+/// like `echo C:\tmp\`) is shown as-is. The remaining ambiguity — a literal `\` at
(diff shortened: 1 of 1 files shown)


iamkroot:
Umm, this completely disregards the `task.show_full_cmd` option. I don't want my huge command to get squished into a single line.

Though, looking at the diff, it seems like the issue was already present and not part of this PR.

Update: yeah, looks like my issue was introduced in #9844

PR #10402, opened 2026-06-13
fix(plugins): prompt before installing non-registry plugin URLs

risu729:
## Summary
- Tighten plugin trust checks to an explicit allowlist instead of trusting unknown plugin URLs by default.
- Continue allowing `github.com/mise-plugins/*` plugin remotes and official asdf/vfox plugin backend URLs from the registry, including explicit registry backend URLs.
- Add unit and e2e coverage for a safe untrusted config that pins `vfox:attacker/evil` and must show the community plugin prompt before clone.

## Trusted plugin handling
| Plugin source | Before this PR | After this PR |
| --- | --- | --- |
| `github.com/mise-plugins/*` | Trusted without prompt | Trusted without prompt |
| Registry shorthand with matching official plugin URL | Trusted without prompt | Trusted without prompt |
| Explicit asdf/vfox URL matching any registry tool backend, including non-`mise-plugins` owners like `echocat/vfox-mongod` | Trusted without prompt | Trusted without prompt |
| Non-registry asdf/vfox URL such as `vfox:attacker/evil` | Trusted without prompt because it was not a registry shorthand mismatch | Community plugin prompt before clone |
| Invalid or non-normalizable plugin remote | Could fall through as trusted when not treated as a shorthand mismatch | Untrusted; community plugin prompt path applies |

## Context
#10360 is not related to this PR's code changes, but it is why I thought about this attack shape: safe configs can now load plain `[tools]` pins without `mise trust`, so a plain pin like `"vfox:attacker/evil" = "latest"` should not silently clone a non-registry plugin. This keeps that safe-config UX while closing the silent plugin-clone gap.

This is also in the same spirit as #9913: do not implicitly trust install paths or sources just because they are reachable through normal install flows.

## Tests
- `cargo fmt --check`
- `mise x rust -- cargo test registry::tests::test_is_trusted_plugin`
- `MISE_E2E_BIN=/home/risu/.worktrunk/risu729/mise/mise.codex-20260614-004629-bd49fd/target/debug/mise mise run test:e2e e2e/config/test_trust_safe_config`

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened plugin trust checks to reject malformed or non-official plugin URLs and reliably recognize official plugin remotes.

* **Tests**
  * Expanded coverage for community plugin trust behavior, including prompts/output when trust is disabled.
  * Added tests confirming tasks defined in configuration run correctly without requiring trust markers.

### DIFF
--- a/e2e/config/test_trust_safe_config
+++ b/e2e/config/test_trust_safe_config
@@ -39,7 +39,11 @@ cat <<'EOF' >mise.toml
 [tools]
 "vfox:attacker/evil" = "latest"
 EOF
-output=$(MISE_YES=0 mise install 2>&1 || true)
+set +e
+output=$(MISE_YES=0 mise install 2>&1)
+status=$?
+set -e
+[[ $status -ne 0 ]] || fail "expected community plugin install to abort without confirmation"
 echo "$output" | grep -q "community-developed plugin" || fail "expected community plugin prompt, got: $output"
 echo "$output" | grep -q "github.com/attacker/evil" || fail "expected attacker plugin URL, got: $output"
 assert_no_trust_marker "community plugin prompt should not trust config"
--- a/e2e/config/test_trust_safe_config
+++ b/e2e/config/test_trust_safe_config
@@ -34,7 +34,26 @@ MISE_YES=0 mise install tiny
 assert_contains "MISE_YES=0 mise ls tiny" "3.1.0"
 assert_no_trust_marker "safe config should not be silently trusted"
 
+# safe config tool pins still prompt before installing non-registry plugin URLs
+cat <<'EOF' >mise.toml
+[tools]
+"vfox:attacker/evil" = "latest"
+EOF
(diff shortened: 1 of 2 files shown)


risu729:
@coderabbitai full review

jdx:
I don't think this makes anything safer, you could already have a template with exec() in the config or install something with github: backend that wouldn't have a prompt.

risu729:
I think this can prevent installing (and executing unsafe hooks) without explicit approval. `mise trust` is not needed for plain `vfox:<repo>` string tools, and `mise run` or other commands may install the vfox plugin and run hooks.
The difference between GitHub and other backends is that the vfox plugin can execute any commands, so even without calling the installed tool binary itself, it can be unsafe.

However, I'm not too sure how effective this is. Please feel free to close.

jdx:
ah yeah I guess so since we don't need trust for basic tools now

PR #10418, opened 2026-06-13
chore: update mise.lock

JamBalaya56562:
## What

Refresh `mise.lock` to the latest eligible versions for the tracked dev tools.

> Note: this PR originally also added `pitchfork` to `[tools]`. Per review, pitchfork is installed globally and `pitchfork.toml` is kept for docs, so the pitchfork changes were dropped — this is now purely a lockfile refresh.

## mise.lock bumps

| tool | old → new |
|---|---|
| aube | 1.16.1 → 1.19.0 |
| bun | 1.3.13 → 1.3.14 |
| cargo-binstall | 1.19.0 → 1.20.0 |
| cargo-insta | 1.47.2 → 1.48.0 |
| cargo:cargo-edit | 0.13.10 → 0.13.11 |
| communique | 1.1.2 → 1.1.3 |
| gh | 2.92.0 → 2.94.0 |
| hk | 1.45.0 → 1.48.0 |
| node | 24.15.0 → 24.16.0 |
| prettier | 3.8.3 → 3.8.4 |
| sops | 3.12.2 → 3.13.1 |
| stylua | 2.4.1 → 2.5.2 |

Versions respect the repo's `minimum_release_age` policy (e.g. aube resolves to 1.19.0 rather than the newer 1.20.0). Verified on Windows and Linux (Docker) that the locked versions install across platforms.

### DIFF
--- a/mise.lock
+++ b/mise.lock
@@ -99,186 +99,186 @@ url = "https://github.com/FiloSottile/age/releases/download/v1.3.1/age-v1.3.1-wi
 provenance = "github-attestations"
 
 [[tools.aube]]
-version = "1.16.1"
+version = "1.19.0"
 backend = "github:endevco/aube"
 
 [tools.aube."platforms.linux-arm64"]
-checksum = "sha256:5e1a207be4b83cd1e0186402764b56e5a702c0332f519959c16d02f4cee9e008"
-url = "https://github.com/jdx/aube/releases/download/v1.16.1/aube-v1.16.1-aarch64-unknown-linux-gnu.tar.gz"
-url_api = "https://api.github.com/repos/jdx/aube/releases/assets/433127937"
+checksum = "sha256:9e062f319809013c33d4ccb7eba7a36752e8735d08dabe5c8b227d83ae5de102"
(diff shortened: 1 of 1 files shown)


mise.toml
```diff
@@ -26,6 +26,7 @@ stylua = "latest"
 "npm:markdownlint-cli" = "latest"
 prettier = "3"
 "npm:ajv-cli" = "latest"
+pitchfork = "latest"
```
jdx: I don't think we need pitchfork

JamBalaya56562: Makes sense — I'll drop the `pitchfork = "latest"` line and its `mise.lock` entry, keeping just the lockfile refresh for the existing tools.

While I'm here: there's a `pitchfork.toml` at the repo root (a `[daemons.docs]` entry that runs `mise run docs`). If pitchfork isn't needed, is that file still used, or should I remove it as well?

jdx: it's just for docs I don't want to remove the pitchfork config, pitchfork also needs to be installed globally

JamBalaya56562: Done — removed `pitchfork = "latest"` and its `mise.lock` entry, and left `pitchfork.toml` in place. The PR is now just the `mise.lock` refresh for the existing tools.

PR #10425, opened 2026-06-14
chore: silence platform-conditional dead-code warnings

JamBalaya56562:
## Summary

Building the `mise` binary on Windows surfaced a handful of `dead_code` / `unused_import` warnings (one was also firing on macOS) coming from code that is only ever used on a subset of platforms. This PR gates each item to the platform(s) that actually use it, so the default `cargo check` / `cargo build` is warning-free on every target.

No behavior change — every item is still compiled on the platform that uses it; only the copies that are *unused on a given target* are removed.

## Changes

| File | Fix | Why |
|---|---|---|
| `src/system/launchd.rs` | `#[cfg(unix)]` on `current_uid_from` + its test | only called by the `#[cfg(unix)]` `current_uid()` (Windows uses a `cfg(not(unix))` fallback that returns 0) |
| `src/ui/progress_report.rs` | `#[cfg_attr(not(unix), allow(dead_code))]` on `ProgressIcon::Error` | the variant is matched in `Display`/`finish_with_icon` on all platforms, so it cannot be cfg-gated away, but it is only *constructed* by the `#[cfg(unix)]` brew package managers |
| `src/oci/layer.rs` | `#[cfg(unix)]` on `assert_layer_mode` / `assert_layer_entry_owner` | test helpers used only by `#[cfg(unix)]` permission-bit tests |
| `src/plugins/core/erlang.rs` | `#[cfg(linux)]` on `linux_precompiled_cache_name` | only called from the `#[cfg(linux)]` `install_precompiled` (also clears a pre-existing macOS warning, since the caller was already linux-gated) |
| `src/config/config_file/mise_toml.rs` | `#[cfg(unix)]` on `update_bootstrap_brew_tap` / `remove_bootstrap_brew_tap` | only called from the `#[cfg(unix)]` `mise system brew tap/untap` CLI |
| `src/backend/pkgx.rs` | `#[cfg(any(unix, test))]` on `shell_quote` | used by the `#[cfg(unix)]` wrapper writer and by a non-gated unit test (so `test` is needed for the Windows test build) |
| `src/cmd.rs` | drop the redundant `use std::os::windows::process::CommandExt` | `self.cmd` is a `tokio::process::Command`, whose `raw_arg` is an inherent method, so the trait import is unnecessary (`hooks.rs`/`path.rs` keep theirs because they use `std::process::Command`) |

## Verification

- `cargo fmt --all -- --check` ✅
- `cargo check --all-targets` on Windows (msvc): **0 warnings, 0 errors** (previously 8 warnings) ✅
- Each cfg gate was checked against every call site to confirm it does not exclude an item that any platform still uses.

## Release Notes

* **Bug Fixes**
  * Improved cross-platform compatibility by restricting Unix-specific helpers and behaviors so Windows builds no longer include unnecessary Unix-only logic.

* **Chores**
  * Further refined conditional compilation for platform-specific utilities and test helpers (Linux/Unix/Windows) to keep builds clean and consistent.

* **Documentation**
  * Updated command documentation to reflect how raw argument handling is provided by the command interface.

### DIFF
--- a/src/ui/progress_report.rs
+++ b/src/ui/progress_report.rs
@@ -18,6 +18,9 @@ pub enum ProgressIcon {
     Skipped,
     #[allow(dead_code)]
     Warning,
+    // Constructed only by the brew package managers (`#[cfg(unix)]`), so it reads
+    // as never-constructed on the windows build.
+    #[cfg_attr(windows, allow(dead_code))]
     Error,
 }
 
--- a/src/backend/pkgx.rs
+++ b/src/backend/pkgx.rs
@@ -931,6 +931,7 @@ fn pkgx_arch_for_target(target: &PlatformTarget) -> String {
     }
 }
 
+#[cfg(any(unix, test))]
 fn shell_quote(value: &str) -> String {
     format!("'{}'", value.replace('\'', "'\\''"))
 }
--- a/src/cmd.rs
+++ b/src/cmd.rs
@@ -506,7 +506,8 @@ impl<'a> CmdLineRunner<'a> {
     /// and discussion #9355.
     #[cfg(windows)]
     pub fn raw_arg<S: AsRef<OsStr>>(mut self, arg: S) -> Self {
-        use std::os::windows::process::CommandExt;
+        // tokio's `Command` exposes `raw_arg` as an inherent method, so the
+        // `std::os::windows::process::CommandExt` trait import is unnecessary.
         self.cmd.raw_arg(arg);
         self
     }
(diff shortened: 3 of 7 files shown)


src/ui/progress_report.rs
```diff
@@ -18,6 +18,9 @@ pub enum ProgressIcon {
     Skipped,
     #[allow(dead_code)]
     Warning,
+    // Constructed only by the brew package managers (`#[cfg(unix)]`), so it reads
+    // as never-constructed on non-unix builds.
+    #[cfg_attr(not(unix), allow(dead_code))]
```
jdx: i think it probably should be windows instead of not(unix)

jdx:
if you don't verify this in ci it's just going to break again

src/ui/progress_report.rs
JamBalaya56562: Good call — switched to `#[cfg_attr(windows, allow(dead_code))]`. mise's only non-unix target is windows, so this is equivalent here but more explicit about where the variant goes unconstructed. Updated in 42d34e6.

PR #10447, opened 2026-06-15
fix(brew-cask): skip unsupported artifact types with a warning 

VeryBadFrags:
When encountering an unsupported artifact in a brew cask, mise bails instead of installing the App artifact.
This change updates the behavior to warn and continue instead of bailing.
The logic to parse the App artifact ensures that mise will still bail if no App is found.

This fixes the following error for casks that include multiple artifacts:

```txt
$ mise bootstrap packages use brew-cask:ghostty
mise ~/.dotfiles/mise.toml: added brew-cask:ghostty
mise ERROR brew-cask:ghostty: unsupported artifact type manpage
mise ERROR Version: 2026.6.10 macos-arm64 (2026-06-14)
```

For reference, these are the artifacts produced by Ghostty:

```txt
==> Artifacts
Ghostty.app (App)
/Applications/Ghostty.app/Contents/Resources/man/man1/ghostty.1 (Manpage)
/Applications/Ghostty.app/Contents/Resources/man/man5/ghostty.5 (Manpage)
/Applications/Ghostty.app/Contents/Resources/bash-completion/completions/ghostty.bash (Bash Completion)
/Applications/Ghostty.app/Contents/Resources/fish/vendor_completions.d/ghostty.fish (Fish Completion)
/Applications/Ghostty.app/Contents/Resources/zsh/site-functions/_ghostty (Zsh Completion)
```

Tested on mise version `2026.6.10 macos-arm64 (2026-06-14)`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness during package installation by treating unsupported artifact types as non-fatal. The system now issues a warning (including the relevant token and detected type) and skips those artifacts so the overall process can continue. It still stops with an error when no valid app artifacts are found.


VeryBadFrags:
Re-reading the docs, this might lead to an incomplete installation, as mise does not delegate to brew for the other artifacts.

PR #10451, opened 2026-06-15
chore(exec): Also pass-through the `COLORTERM` environment variable

sschuberth:
The `COLORTERM` variable extends `TERM` to indicate support for modern 24-bit True Color.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed color output support in sandboxed environments with restricted environment variable access.


sschuberth:
Should I update all the docs manually as part of the PR, or will they somehow be auto-generated?

PR #9848, opened 2026-05-14
fix(shim): preserve optioned aliases during rebuilds

risu729:
Fixes https://github.com/jdx/mise/discussions/9798

## What was wrong

The regression came from the inline/backend option overlay work in #9306. That PR did two things that are individually useful but unsafe together:

1. It copied config-derived tool options onto the `BackendArg` for configured tools, so options like `asset_pattern` and `bin_path` are available to GitHub install and bin discovery.
2. It changed `backend::get()` so `BackendArg`s with explicit options bypass the global backend cache. That is correct for command-scoped inline options because caching them would leak one command's options into later commands.

The missing distinction was that config-scoped alias options are not just one-off inline options. They still describe active tools in the current toolset. Because optioned aliases bypassed the global backend cache, rebuild scans that only walked `backend::list()` could miss those active alias backends.

That meant `mise install` could download and extract the tool because the active `Toolset` held the transient alias backend directly. After install, runtime symlink and shim rebuild paths could still miss the alias backend, so alias `latest` links or executable shims were not reliably created.

## How this fixes it

This PR keeps command-scoped optioned backends out of the global backend cache, but makes active configured aliases visible to rebuild scans:

- `Toolset::list_cached_and_current_backends()` now merges active current-version backends with cached backends, preferring the active toolset backend and deduplicating by `installs_path`.
- `Toolset::list_installed_versions()` uses that helper and matches current versions by `(installs_path, version)` instead of `(backend_id, version)`, so the optioned alias backend is used for its own install directory.
- `runtime_symlinks::rebuild_for_toolset()` uses the same helper, so runtime symlink rebuilds and shim rebuilds see the same backend set.
- `rebuild_shims_and_runtime_symlinks()` rebuilds runtime symlinks before shims, so shim discovery sees the freshly rebuilt runtime install tree.
- `mise upgrade` now resolves the current toolset before its early runtime symlink rebuild so it can pass the active backend set through the same path.

With that backend included, shim generation sees the alias install directory, asks the alias backend for its configured bin paths, discovers the executable, and creates the shim normally.

## Regression coverage

The e2e coverage in `e2e/backend/test_github_tool_alias_asset_pattern` verifies that GitHub `tool_alias` entries with per-alias `asset_pattern`/`bin_path` options create:

- the alias-specific install directories,
- `latest` runtime symlinks for each alias, and
- the executable shim.

## Testing

- `cargo fmt --all`
- `git diff --check`
[...]

PR #10470, opened 2026-06-15
fix(install): rebuild symlinks after partial installs

risu729:
Fixes the follow-up reported in #9848 where successful tools from a partially failed `mise install` could be left without runtime version symlinks such as `installs/bun/latest`.

## What was wrong

`Toolset::install_all_versions()` already preserves partial success in `Error::InstallFailed { successful_installations, failed_installations }`. However, `mise install` used `?` on that result, so any failed tool skipped the later `rebuild_shims_and_runtime_symlinks()` call entirely. That meant tools that had already installed successfully during the same command could miss their runtime symlinks and shims until another successful install/rebuild path ran.

This is an install partial-failure problem, not a `mise reshim` problem. Runtime version symlinks remain owned by the install/runtime-symlink rebuild flow.

## How this fixes it

- Mirrors the existing `mise upgrade` pattern: extract `successful_installations` from `InstallFailed`, continue post-install rebuild work for those versions, then return the original install error.
- Applies that behavior to both explicit `mise install <tools>` and bare `mise install` from config.
- Extends `e2e/cli/test_install_parallel_failure` to assert that the successfully installed tool gets its `latest` runtime symlink even though another tool failed.

## Testing

- `/home/risu/.cargo/bin/cargo fmt --all --check`
- `git diff --check`
- `env -u MISE_PROJECT_ROOT -u MISE_ORIGINAL_CWD mise run test:e2e e2e/cli/test_install_parallel_failure`

Not run: full local test suite.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling during parallel installations to ensure shims and symlinks are properly rebuilt even when some tool installations fail.

* **Tests**
  * Enhanced end-to-end tests to verify correct installation symlinks after parallel install scenarios.

### DIFF
--- a/e2e/cli/test_install_parallel_failure
+++ b/e2e/cli/test_install_parallel_failure
@@ -26,6 +26,7 @@ assert_fail "mise install" "Failed to install asdf:dummy@other-dummy"
 
 # Verify tiny was installed successfully despite dummy failure
 assert_contains "mise ls --installed tiny" "3.1.0"
+assert "readlink $MISE_DATA_DIR/installs/tiny/latest" "./3.1.0"
 
 # Verify dummy failed to install the bad version
 assert_not_contains "mise ls --installed dummy" "other-dummy"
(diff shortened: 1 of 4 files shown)


risu729:

Addressing the Greptile review on PR description mismatch:

The title and description were updated on Jun 17 to reflect the actual change: this PR fixes partial-failure handling in `mise install`, not `mise reshim`. The reshim symlink rebuild concern from #9848 is handled by the install/runtime-symlink rebuild flow; `reshim.rs` is intentionally unchanged here.

Additional updates in this push:
- Rebased onto latest `main`
- Extracted `errors::split_install_result` so install and upgrade share the same `InstallFailed` extraction logic

PR #10472, opened 2026-06-15
fix(pipx): upgrade shared pip for release age installs

risu729:
## Summary
- run `pipx upgrade-shared` before pipx fallback installs that forward `minimum_release_age` to pip
- default modern pipx to the pip backend for these installs so `--uploaded-prior-to` is handled by pip, not uv

## Background

pipx forwards `pipx install` `--pip-args` into shared-library bootstrap (`pip install --upgrade pip>=23.1`), not just the package install. When mise passes `--pip-args=--uploaded-prior-to=...`, bootstrap pip (from ensurepip) may not understand that flag and shared pip never upgrades successfully.

This is the same chicken-and-egg class of bug as [pypa/pipx#544](https://github.com/pypa/pipx/issues/544) (install `pip_args` breaking shared pip upgrade with flags the bootstrap pip does not support). There is no open pipx issue specifically for `--uploaded-prior-to`; upstream discussion is mostly [pypa/pipx#1811](https://github.com/pypa/pipx/issues/1811) / [#1801](https://github.com/pypa/pipx/issues/1801).

Prior mise work on this area:
- [#9190](https://github.com/jdx/mise/pull/9190) — forward `install_before` / release-age cutoff to pipx and uv
- [#10138](https://github.com/jdx/mise/pull/10138) — titled as shared-pip upgrade support but merged as docs only; the bootstrap workaround was not implemented

Running `pipx upgrade-shared` first (without release-age flags) makes shared pip valid so the subsequent `pipx install` `shared_libs.create()` is a no-op and `--uploaded-prior-to` only applies to the package install.

Related reports: [discussion #10088](https://github.com/jdx/mise/discussions/10088), [discussion #10484](https://github.com/jdx/mise/discussions/10484).

## Reproduction

Reproduced with an isolated `PIPX_HOME` by downgrading pipx's shared pip to `pip==25.3`, then running:

```sh
pipx install pre-commit==4.6.0 --backend pip --pip-args=--uploaded-prior-to=2026-06-09T00:00:00Z
```

That failed with `no such option: --uploaded-prior-to`. Running `pipx upgrade-shared` first upgraded shared pip, and the same install then succeeded.

## Testing
- `mise x cargo -- cargo fmt --check`

Full local tests were not run.

## Discussion
- Addresses https://github.com/jdx/mise/discussions/10088

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `pipx` installs with release date constraints: when `minimum_release_age`/limits apply, mise now attempts to upgrade `pipx`’s shared pip environment before running `pipx install`, so pip options such as `--uploaded-prior-to` work as expected.
* **Documentation**
  * Updated `pipx` “minimum_release_age” guidance to reflect the upgrade-first behavior, removing the need for manual Python/pip compatibility prerequisites.
* **Tests**
  * Added an end-to-end test for the `pipx` minimum release age flow to verify the shared pip environment behavior.

### DIFF
--- a/docs/dev-tools/backends/pipx.md
+++ b/docs/dev-tools/backends/pipx.md
@@ -30,7 +30,6 @@ If you have `uv` installed, mise will use `uv tool install` under the hood and y
 mise forwards [`minimum_release_age`](/configuration/settings.html#minimum_release_age)
 to transitive Python dependency resolution during install. The uv install path uses uv's
 `--exclude-newer` flag, and the `pipx` fallback passes pip's `--uploaded-prior-to` flag.
-When using `minimum_release_age` with the `pipx` install path, mise upgrades pipx's shared pip environment first so pip supports `--uploaded-prior-to`.
 
 In case you need `pipx` for other reasons, you can install it with or without mise.
 Here is how to install `pipx` with mise:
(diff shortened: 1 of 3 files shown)


risu729:
@coderabbitai resume

PR #10473, opened 2026-06-15
fix(aqua): support checksum verification fields

risu729:
## Summary

- Parse aqua `checksum.replacements`, `checksum.minisign`, and `checksum.github_artifact_attestations`.
- Render checksum file assets/URLs with checksum-specific replacements, `AssetURL` for HTTP checksum metadata, and consistent OS/arch template context (`GOOS`/`GOARCH` included).
- Verify checksum-file minisign/GitHub attestations in the aqua provenance flow, then continue using the parsed checksum to verify the downloaded artifact.
- Bind checksum-scoped provenance to the checksum digest used for the artifact at both install time and lock time.
- Bump the compiled aqua-registry cache namespace from `v4` to `v5` so existing compiled caches are invalidated after the `AquaChecksum` schema change.

## Scope

Addresses the `checksum.*` slice of the [remaining aqua-registry schema gaps tracker](https://github.com/users/risu729/projects/3/views/1?pane=issue&itemId=183902266). This does not close the tracker item.

## Changes

**Type layer (`crates/aqua-registry`)**
- Add checksum-level `replacements`, `minisign`, and `github_artifact_attestations` fields to `AquaChecksum`.
- Add `template_ctx` / `effective_replacements` helpers for checksum-specific OS/arch replacement merging.
- Handle `replacements: null`, partial version-override merges, and shared YAML map deserialization.
- Re-export `AquaMinisign` and `AquaGithubArtifactAttestations` from the crate.

**Backend (`src/backend/aqua.rs`)**
- Fall back to checksum-scoped minisign/attestation/cosign configs when top-level configs are absent or disabled.
- Preserve checksum release asset names for minisign `{{.Asset}}` rendering instead of local download filenames.
- Compare verified checksum file contents against expected/lockfile digests after checksum-scoped provenance verification.
- Include checksum-scoped attestations in `security_info()` signer workflow lookup.

## Verification

- `mise run lint`
- `cargo clippy -- -D warnings`
- `mise x sccache -- cargo test -p aqua-registry checksum --lib`
- `mise x sccache -- cargo test backend::aqua::tests`

## Summary by CodeRabbit

* **New Features**
  * Added checksum-level `replacements` support to customize URL and template values.
  * Extended GitHub Artifact Attestations and Minisign provenance to derive from checksum-provided configuration when package settings are missing.
* **Improvements**
  * Refined template rendering and replacement precedence for OS/architecture-aware URL generation.
  * Expanded the publicly available registry API to include additional attestation and signature configuration types.
* **Bug Fixes**
  * Correctly handles `replacements: null` and applies overrides only when explicitly provided.
  * Added a safeguard to prevent provenance verification when verified checksums conflict.
* **Chores**
  * Bumped the compiled registry cache version, invalidating older compiled caches.

### DIFF
--- a/crates/aqua-registry/src/cache.rs
+++ b/crates/aqua-registry/src/cache.rs
@@ -7,7 +7,7 @@ use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::time::{Duration, SystemTime};
 
-const COMPILED_REGISTRY_CACHE_VERSION: &str = "v4";
+const COMPILED_REGISTRY_CACHE_VERSION: &str = "v5";
 
 #[derive(Debug, Clone)]
 pub struct RegistryCache {
--- a/crates/aqua-registry/src/lib.rs
+++ b/crates/aqua-registry/src/lib.rs
@@ -17,8 +17,8 @@ pub use cache::RegistryCache;
 pub use codec::{decode_package_rkyv, encode_package_rkyv};
 pub use compiled::{CompiledRegistry, ParsedRegistry};
 pub use types::{
-    AquaChecksum, AquaChecksumType, AquaCosign, AquaFile, AquaMinisignType, AquaPackage,
-    AquaPackageType, AquaVar, RegistryYaml,
+    AquaChecksum, AquaChecksumType, AquaCosign, AquaFile, AquaGithubArtifactAttestations,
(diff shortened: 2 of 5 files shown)


risu729:
@coderabbitai resume

PR #10506, opened 2026-06-17
fix(backend): add riscv64/loongarch64 archs, patterns

elerch:
I have a [spacemit k3 soc](https://www.spacemit.com/community/development-kit/k3-pico-itx) which is based on riscv64, and wanted to do some development on it using mise. However, installing `github:astral-sh/uv` resulted in mise attempting to download the s390x version. This change adds Riscv64 (and loongarch64) as architectures and adds the necessary regex patterns to create a real match.

The regex is likely a bit limited here...uv uses "riscv64gc" in its tarball name. That pattern is a bit dated, and I would expect most others to simply use "riscv64". The "gc" are a couple common extensions to the specification, but lately people have been dropping that. The full regex pattern for any set of extensions could be a bit unwieldy (but also unlikely to encounter in practice), so I thought it was best to defer additional work.

## Summary by CodeRabbit

* **New Features**
  * Added support for RISC-V 64-bit and Loongarch64 CPU architectures in asset platform detection, enabling proper recognition and selection of matching assets.
* **Bug Fixes**
  * Improved platform scoring to prevent RISC-V 64-bit assets (including `riscv64gc`) from being incorrectly shadowed by similarly scored architectures.
* **Tests**
  * Added a regression test to cover correct selection behavior for RISC-V 64-bit assets.

### DIFF
--- a/src/backend/asset_matcher.rs
+++ b/src/backend/asset_matcher.rs
@@ -40,6 +40,8 @@ pub enum AssetArch {
     Arm64,
     X86,
     Arm,
+    Riscv64,
+    Loongarch64,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq)]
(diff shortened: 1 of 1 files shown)


elerch:
rebased PR on b0de95f 2026-06-24

PR #10509, opened 2026-06-17
feat(http): resolve cross-platform lock checksums from published metadata

itochan:
## Summary

- Let the `http` backend fill in per-platform `url` + `checksum` in `mise.lock` for platforms other than the host, from a single machine, without downloading the artifacts.
- Read checksums from a tool's published metadata: individual checksum files, `SHASUMS`-style lists, or a JSON manifest.

Refs #10218.

## Details

New `http` backend options:

- `checksum_url` — an individual checksum file (`<artifact>.sha256`) or a `SHASUMS` list. The algorithm is detected from the file name (`*.sha512`, `SHA512SUMS`, `*.md5`, `*.b3`, else `sha256`).
- `checksum_expr` — an [expr-lang](https://expr-lang.org) expression to extract the hash from a JSON manifest. Variables: `body`, `version`, `os`, `arch`, `url`, `filename`. It must evaluate to a qualified `algo:hash` **string** (e.g. `sha256:<hash>`); build the prefix in the expression (`"sha256:" + h`, or `entry.algo + ":" + entry.hash` when it varies).

Supporting changes:

- `os()` / `arch()` / `os_family()` render for the target platform during `mise lock` (deferred at config load — scoped to the http backend's `url` and `checksum_url` options so the templates survive to lock time).
- A platform with no resolvable URL fails closed and is reported as skipped, rather than written as an empty entry counted as success.
- When a checksum source *is* configured but resolves to nothing for a target (manifest miss, `SHASUMS` filename mismatch, unreachable file), the URL is still locked but a warning is emitted, so dropping checksum verification is never silent.

The registry entries are left unchanged; adding `checksum_url` / `checksum_expr` to specific tools would be a follow-up `registry:` PR.

## Validation

- `mise run test:e2e test_http_lock_checksum test_http_lock_install_verify test_http_lock_skips_unresolved`
- `mise run test:unit`
- Manually verified against dart / neo4j (individual file), HashiCorp + OpenShift (SHASUMS), and julia / flutter (manifest) resolving every platform from one host.

## Summary by CodeRabbit

* **New Features**
  * HTTP backend now resolves target-specific checksums during `mise lock` without downloading artifacts, using `checksum_url` and manifest-based `checksum_expr`, with `checksum_algo` support.
  * Artifact URL and checksum lookup are now platform-aware, with better handling of unresolved platforms.
* **Documentation**
  * Documented `checksum_url`, `checksum_expr`, and `checksum_algo`, including supported checksum source formats and extraction behavior.
* **Bug Fixes**
  * More robust SHASUMS parsing (skips malformed lines, normalizes `*file` entries) and improved checksum algorithm detection/formatting.
* **Tests**
  * Added end-to-end HTTP lock/verification coverage, including checksum mismatch failures and skipping unresolved platforms.

### DIFF
--- a/docs/dev-tools/backends/http.md
+++ b/docs/dev-tools/backends/http.md
@@ -199,9 +199,24 @@ the literal key `"version"`.
 
 ### `checksum_algo`
 
-Algorithm for a **manifest** checksum (`checksum_expr`) when it returns a bare
-hash, since the algorithm isn't in a file name there. Defaults to `sha256`.
-Individual and SHASUMS checksum files detect the algorithm from their file name instead.
+Applies **only to the manifest path** (`checksum_expr`), and only when the
+expression returns a **bare hash** — there's no file name to infer the algorithm
+from, so this supplies it. Defaults to `sha256`.
+
+It is **not** consulted for:
+
+- **individual checksum files** and **SHASUMS** lists — the algorithm is detected
+  from the checksum file's name (`*.sha512`, `SHA512SUMS`, `*.md5`, `*.b3`,
+  else `sha256`);
+- a `checksum_expr` that returns an `algo:hash` string or a `{ algo, checksum }`
+  map — those carry their own algorithm.
+
+```toml
+# manifest of bare sha512 hashes with no algorithm in the file or the value
+checksum_url  = "https://example.com/versions.json"
+checksum_expr = 'filter(fromJSON(body).files, { #.url == url })[0].hash'
+checksum_algo = "sha512"
+```
 
 ### `size`
 
--- a/docs/dev-tools/backends/http.md
+++ b/docs/dev-tools/backends/http.md
@@ -169,18 +169,17 @@ evaluated with [expr-lang](https://expr-lang.org). The following variables are
 available: `body` (the raw manifest), `version`, `os`, `arch`, `url` (the
 resolved artifact URL for the target), and `filename`.
 
-The expression must evaluate to a **string**: either an already-qualified
-`sha256:<hash>` (or `sha512:<hash>`, etc.), or a bare hash, which is assumed to
-be `sha256`. If the manifest stores the algorithm in a separate field and it
-isn't sha256, build the prefix in the expression itself, e.g.
-`entry.algo + ":" + entry.hash`.
+The expression must evaluate to a qualified `algo:hash` **string** (e.g.
+`sha256:<hash>`, `sha512:<hash>`). Build the prefix in the expression: prepend a
+literal when the algorithm is fixed (`"sha256:" + entry.hash`), or read it from
+the manifest when it varies (`entry.algo + ":" + entry.hash`).
 
 ```toml
 [tools."http:my-tool"]
 version = "1.10.0"
 checksum_url = "https://example.com/versions.json"
-# Match the file whose url equals the resolved artifact url, return its sha256
-checksum_expr = 'filter(fromJSON(body)[version + ""].files, { #.url == url })[0].sha256'
+# Match the file whose url equals the resolved artifact url, return sha256:<hash>
+checksum_expr = '"sha256:" + filter(fromJSON(body)[version + ""].files, { #.url == url })[0].sha256'
 
 [tools."http:my-tool".platforms]
 linux-x64 = { url = "https://example.com/my-tool-{{ version }}-linux-x86_64.tar.gz" }
--- a/docs/dev-tools/backends/http.md
+++ b/docs/dev-tools/backends/http.md
@@ -138,14 +138,18 @@ single machine produce a complete, cross-platform lockfile.
 and is platform-specific via `platforms.<key>.checksum_url`). It may point at any
 of:
 
-- a single-checksum **sidecar** file (e.g. `<artifact>.sha256`), which may contain
+- an **individual checksum file** (e.g. `<artifact>.sha256`), which may contain
   just the hash or `<hash>  <filename>`;
 - a **SHASUMS**-style file listing `<hash>  <filename>` for many platforms (the row
   is matched by the artifact's filename);
 - a **manifest** (e.g. JSON), combined with `checksum_expr` below.
 
+For individual and SHASUMS checksum files, the algorithm is detected from the
+file's name (`*.sha512`, `SHA512SUMS`, `*.md5`, `*.b3`, defaulting to sha256), so
+`checksum_algo` is not needed.
+
 ```toml
-# Sidecar (one file per artifact)
+# Individual checksum file (one per artifact)
 [tools."http:my-tool"]
 version = "1.0.0"
 url = "https://example.com/releases/my-tool-{{ version }}-{{ os() }}-{{ arch() }}.tar.gz"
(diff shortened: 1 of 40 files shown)


itochan:
@greptileai thanks for the review — all three substantive findings were valid. Addressed in follow-up commits:

**1. `os_family()` not deferred (P1) — fixed** in `5742541fe`.
`get_tera_preserving_os_arch` now re-emits `os_family()` as a deferred fragment alongside `os()`/`arch()`, so it resolves against the lock target rather than the host. Added a round-trip unit test (`test_preserving_os_family_round_trips_through_target`) covering locking a windows target from a unix host.

**2. SHASUMS fallback picking a wrong hash (P1) — fixed** in `446effb53`.
This was the important one. `resolve_lock_checksum` now guards the individual-file fallback with `shasums_has_entries`: if the file parses as a SHASUMS list with entries but none match the artifact filename, the platform is reported unresolved instead of falling through to a first-hash scan that would silently lock another platform's checksum. Bare individual checksum files (no parseable entries) still fall back as before. Added an e2e case (a naming-mismatched SHASUMS) asserting a url-only entry is written and the stray hash is never locked.

**3. Double HTTP fetch for individual checksum files (P2) — fixed** in `8f8bc536b`.
`fetch_checksum_from_file` now uses `get_text_cached`, so the body already fetched by the SHASUMS probe is reused instead of re-fetched.

**4. `reemit_arg_literal` not escaping `"` (P2) — intentionally left as-is.**
Tera string literals don't support escape sequences, so a remap value can't contain a `"` to begin with, and os/arch remap values (`"darwin"`, `"amd64"`) never do. The existing doc comment already notes the no-escape behavior, so this is unreachable in practice rather than a latent bug.

src/tera.rs
```diff
@@ -413,6 +413,98 @@ pub fn get_tera(dir: Option<&Path>) -> Tera {
...
+            .iter()
+            .filter_map(|(k, v)| reemit_arg_literal(v).map(|lit| format!("{k}={lit}")))
+            .collect();
+        let rendered = if parts.is_empty() {
+            format!("{{{{ {name}() }}}}")
+        } else {
+            parts.sort();
+            format!("{{{{ {name}({}) }}}}", parts.join(", "))
+        };
+        Ok(Value::String(rendered))
```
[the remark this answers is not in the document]
itochan: Good catch — this limitation is real, but I'm intentionally not handling filtered `os()`/`arch()` expressions here, for the reasons below.

The deferral only re-emits the *function call* (`{{ os() }}`, `{{ arch() }}`, and now `{{ os_family() }}`), including its remap arguments, which `reemit_arg_literal` round-trips. So the canonical, supported cross-platform form keeps working:

```toml
url = 'https://example.com/{{ version }}/tool_{{ os(macos="darwin") }}_{{ arch(x64="amd64") }}.tar.gz'
```

A piped filter like `{{ arch() | replace(from="x64", to="amd64") }}` can't be preserved this way: the filter would have to apply to the *resolved target value*, but at config-load time `arch()` only yields the deferred literal, so the filter runs against `{{ arch() }}` and no-ops. Preserving it correctly would mean capturing and re-emitting the entire surrounding expression (filters, operators, nesting) — a much larger change than re-emitting a single function call, and not worth it given the remap-argument form already covers the real OS/arch-naming use cases.

So rather than add a regression test that locks in a broken behavior, I'll treat filter/pipe forms of `os()`/`arch()` in http `url`/`checksum_url` as unsupported and document the remap-argument form as the way to remap names for cross-platform locking. If demand for filtered forms shows up, full-expression preservation can be revisited.

docs/dev-tools/backends/http.md
```diff
@@ -127,6 +127,97 @@ linux-x64 = {
...
+Applies **only to the manifest path** (`checksum_expr`), and only when the
+expression returns a **bare hash** — there's no file name to infer the algorithm
+from, so this supplies it. Defaults to `sha256`.
+
+It is **not** consulted for:
+
+- **individual checksum files** and **SHASUMS** lists — the algorithm is detected
+  from the checksum file's name (`*.sha512`, `SHA512SUMS`, `*.md5`, `*.b3`,
+  else `sha256`);
+- a `checksum_expr` that returns an `algo:hash` string or a `{ algo, checksum }`
```
jdx: why not simplify this and always return algo:hash?

itochan: Thanks for the review! I'd implemented the map because it was part of the shape we settled on in the discussion (#10218), but agreed it's redundant: expr-lang can do `algo + ":" + hash` itself, and `algo:hash` is already the convention everywhere else. Two ways to go, both drop the map and `checksum_algo`.

A. Strict `algo:hash` only, exactly as you suggest. Most consistent; every expr writes the prefix:

```toml
checksum_expr = '"sha256:" + filter(...)[0].sha256'
```

B. Accept `algo:hash`, with a bare hash defaulting to sha256 (what `normalize_checksum` already does). Keeps the common case prefix-free:

```toml
checksum_expr = 'filter(...)[0].sha256'
checksum_expr = '"sha512:" + filter(...)[0].hash'
```

I lean B for the least friction on the common case, and it stays safe since `is_checksum_hex` rejects a wrong-length bare hash. A is cleaner and more consistent. Which do you prefer?

itochan: I went ahead and implemented this (commit 5e220850d): dropped the `{ algo, checksum }` map and the `checksum_algo` option. `checksum_expr` now returns a string, either an `algo:hash` used as-is or a bare hash treated as sha256, normalized to `algo:hash` before it's stored, matching how checksums are verified and stored elsewhere (`verify_checksum_str`, the lockfile's `blake3:`/`sha256:` entries).

I checked what the http-backend tools in the registry actually publish, and it backs the bare-to-sha256 default:

- All of them publish sha256: hashicorp/openshift via SHASUMS, neo4j/dart via individual files, flutter/claude via JSON manifests.
- The two manifest ones (flutter, claude) are the only ones that would use `checksum_expr`, and both expose a bare sha256 with no algorithm label. claude's field is literally `checksum`:

  ```json
  "linux-x64": { "checksum": "849e0072...1f9e", "size": 247469776 }
  ```

  So `checksum_expr = 'fromJSON(body).platforms[os + "-" + arch].checksum'` stays prefix-free with the bare-to-sha256 default. Strict `algo:hash` would force `"sha256:" +` on every real entry, for a benefit no current tool needs.

If you'd still rather it be strict `algo:hash` only (reject bare), that's a one-line flip and I'll redo it. Let me know which you prefer.

jdx: yeah I don't think 'bare hashes' should be done either, no reason to have more than 1 way to do things

itochan: Done. checksum_expr now takes only an `algo:hash` string, exactly as you suggested. Dropped the map, the `checksum_algo` option, and the bare-hash default, so there's one way to express a checksum. (37e963b57)

Also, per your note that unresolved platforms should be reported clearly: when a checksum source is set but resolves to nothing, the lock writes a url-only entry and now warns, instead of silently dropping verification. (cfab213ff)

PR #10514, opened 2026-06-17
fix(backend): resolve dependency executables on Windows

risu729:
## Summary
- centralize dependency executable resolution for PATH, active toolsets, and backend dependency toolsets
- apply configured Windows executable extensions to shared dependency lookup instead of only dependency warnings
- use resolved binary paths for uv, aube, and cargo-binstall install commands
- keep gpg verification unchanged because gpg is not managed by mise

## Behavior Changes
- `Backend::which` now uses `windows_executable_extensions` on Windows instead of the previous hardcoded `exe/cmd/bat/ps1` list.
- `dependency_which` now checks non-pristine PATH with Windows executable extensions before checking dependency toolsets.
- `warn_if_dependency_missing` relies on the shared dependency lookup instead of doing its own Windows extension retry.
- pipx selects uv when `uv`/`uv.exe` is available from the active toolset, dependency toolset, or PATH, and invokes the resolved uv path for `uv tool install`.
- npm/aube uses the shared resolved-path helper for aube detection and execution.
- cargo-binstall detection now uses the same active-toolset/dependency/PATH lookup and invokes the resolved cargo-binstall path.
- sops external-CLI mode resolves configured `sops` first and propagates configured tool resolution failures; when `sops` is not configured, it falls back to PATH with the same executable-extension handling.

## Root Cause
`warn_if_dependency_missing` retried Windows executable extensions, but direct dependency decisions such as uv and cargo-binstall detection did not. That meant `uv.exe`/`cargo-binstall.exe` style binaries could be missed, and command execution could fall back to bare program names even after a path was resolved elsewhere.

## Review Notes
- Addressed review feedback by preserving sops resolution errors for configured `sops` tools instead of silently falling back to PATH.
- Addressed review feedback around the Windows gate by making the target-platform check explicit with `cfg!(target_os = "windows")`.

## Validation
- `mise x cargo -- cargo fmt`
- `mise x cargo -- cargo check`
- `mise x cargo -- cargo test backend::tests`
- `mise x cargo -- cargo test backend::cargo::tests`
- `mise x cargo -- cargo test backend::npm::tests`
- `mise x cargo -- cargo test backend::pipx::tests`
- `git diff --check`

## Summary by CodeRabbit

* **Refactor**
  * Unified executable discovery for dependencies and tools, with consistent Windows handling for extensionless binaries.
  * Added install-time dependency path resolution that prefers toolset-aware lookup with fallback behavior.
* **Bug Fixes**
  * Improved install execution by using resolved program paths for Cargo Binstall, Aube, Npm, and Pipx, including more accurate “installed” detection.
  * Updated Sops command resolution to use the improved lookup for more reliable decrypts.
* **Tests**
  * Added Windows coverage for extension expansion and explicit-extension preservation.

### DIFF
--- a/src/backend/cargo.rs
+++ b/src/backend/cargo.rs
@@ -1,4 +1,5 @@
 use std::collections::BTreeMap;
+use std::path::PathBuf;
 use std::{fmt::Debug, sync::Arc};
 
 use async_trait::async_trait;
@@ -16,10 +17,9 @@ use crate::cli::args::BackendArg;
 use crate::cmd::CmdLineRunner;
 use crate::config::{Config, Settings};
 use crate::env::GITHUB_TOKEN;
-use crate::file;
 use crate::http::HTTP_FETCH;
 use crate::install_context::InstallContext;
-use crate::toolset::{ToolRequest, ToolVersion, ToolVersionOptions};
+use crate::toolset::{ToolRequest, ToolVersion, ToolVersionOptions, Toolset};
 
 #[derive(Debug)]
 pub struct CargoBackend {
@@ -84,7 +84,7 @@ impl<'a> CargoOptions<'a> {
 
 #[derive(Debug)]
 enum BinstallStatus {
-    Enabled,
+    Enabled(PathBuf),
     Disabled,
     Unavailable,
     UnsupportedOptions(Vec<&'static str>),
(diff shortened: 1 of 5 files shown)


risu729:
@coderabbit full review

PR #10552, opened 2026-06-21
registry: add checksums for http-backend tools

itochan:
## Summary

Adds published-checksum resolution to the registry entries that use the `http` backend, building on cross-platform lock checksum support (#10509). With these, `mise lock` records a verified `algo:hash` for each platform instead of locking the URL without verification.

Two checksum sources are used, matching each upstream's published format:

- **`checksum_url`** (SHASUMS / individual checksum file): `dart`, `neo4j`, `nomad-pack`, `sentinel`, `tfc-agent`, `oc`, `openshift-install`
- **`checksum_url` + `checksum_expr`** (sha256 extracted from a JSON manifest): `claude`, `flutter`, `julia`

The three manifest-based entries build an `algo:hash` string in the expression (e.g. `"sha256:" + ...`), as required by the resolver merged in #10509 — a bare hash is intentionally rejected rather than locked unverified.

## Verification

`mise lock` was run against the real upstream endpoints for all 10 tools across every platform they ship:

- 48/48 supported (tool, platform) entries resolved a real `sha256:`/`sha512:`; **zero** url-only (unverified) entries.
- The only skips are platforms a tool does not ship (e.g. `flutter` linux-arm64, `openshift-install` windows-x64).

## Tests

The `checksum_expr` lock path previously had only Rust unit tests, so this adds e2e coverage (synthetic local server, ~9s):

- **`test_http_lock_checksum_expr`** — the three manifest shapes (platform-keyed object / filename filter / version-keyed + url match), exercising the `os`/`arch`/`version`/`url`/`filename` template vars; fail-closed rejection of a bare-hash result; an algorithm taken from the manifest (sha512); and case-insensitive algo/hex normalization.
- **`test_http_lock_install_verify`** — extended so a manifest-resolved checksum is enforced at install and a tampered value fails the install.

## Summary by CodeRabbit

* **New Features**
  * Improved checksum verification across multiple tool backends (Claude, Dart, Flutter, Julia, Neo4j, Nomad Pack, OpenShift, Sentinel, TFC Agent) by adding `checksum_url` support and enabling `checksum_expr` where needed for platform- and version-specific manifest resolution.
  * Locking now canonicalizes checksums (normalized `algo:hash`) and fails closed when an invalid checksum format is returned.

* **Tests**
  * Added end-to-end coverage for `checksum_expr` resolution in the HTTP backend and for the full `mise lock` → `mise install --locked` verification loop, including tamper and error scenarios.

### DIFF
--- a/registry/claude.toml
+++ b/registry/claude.toml
@@ -12,6 +12,10 @@ full = "http:claude"
 
 [backends.options]
 bin = "claude"
+# Per-version manifest.json carries each platform's sha256 under platforms.<os>-<arch>
+# (os: darwin/linux/win32), so map mise's os/arch onto that key.
+checksum_expr = '"sha256:" + fromJSON(body).platforms[(os == "macos" ? "darwin" : (os == "windows" ? "win32" : "linux")) + "-" + arch].checksum'
+checksum_url = "https://storage.googleapis.com/claude-code-dist-86c565f3-f756-42ad-8dfa-d59b1c096819/claude-code-releases/{{ version }}/manifest.json"
 # NOTE: upstream `/stable` is often stale; prefer `/latest` for correct `mise latest claude`
 # See: https://github.com/jdx/mise/discussions/7329
 url = 'https://storage.googleapis.com/claude-code-dist-86c565f3-f756-42ad-8dfa-d59b1c096819/claude-code-releases/{{ version }}/{{ os(macos="darwin") }}-{{ arch() }}/claude'
--- a/e2e/backend/test_http_lock_checksum_expr
+++ b/e2e/backend/test_http_lock_checksum_expr
@@ -20,16 +20,22 @@ mkdir -p "$SRV"
 # Published sha256/sha512 values embedded in the served manifests. `mise lock`
 # only fetches the manifest (checksum_url) and evaluates the expression; it never
 # downloads the artifact, so these are manifest data, not artifact hashes.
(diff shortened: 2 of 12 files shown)


itochan:
@greptileai

registry/claude.toml
```diff
@@ -12,6 +12,10 @@ full = "http:claude"
 
 [backends.options]
 bin = "claude"
+# Per-version manifest.json carries each platform's sha256 under platforms.<os>-<arch>
+# (os: darwin/linux/win32), so map mise's os/arch onto that key.
+checksum_expr = '"sha256:" + fromJSON(body).platforms[(os == "macos" ? "darwin" : (os == "windows" ? "win32" : "linux")) + "-" + arch].checksum'
```
itochan, line 17: This ternary maps mise's os (`macos`/`linux`/`windows`) to claude-code's `manifest.json` keys (`darwin`/`win32`/`linux`) + `-<arch>`.

It's inline because `checksum_expr` runs through expr-lang, not Tera — so the `{{ os(macos="darwin") }}` helper (used in `url`/`checksum_url`) isn't available, and `checksum_expr` isn't resolved per `[platforms.*]` block today. If you'd prefer per-platform blocks, I can make `checksum_expr` target-aware (small `http` backend change) and split it — your call.

Verified with `mise lock` + `mise install --locked` on darwin-arm64/x64, linux-x64/arm64, win32-x64.

PR #10559, opened 2026-06-22
fix(aqua): don't cache transient filesystem state in list_bin_paths

devnulled:
## Problem
After `mise upgrade` with concurrent installs (the report uses python + aqua:uv +
npm:prettier), the `uv`/`uvx` shims are intermittently not created, and `mise reshim`
can't repair it. Fixes #6468.

## Root cause
`aqua list_bin_paths` memoized the result of a `.filter(|p| p.exists())` computed *inside*
the cache closure. During a concurrent upgrade it ran before uv's binaries were extracted,
cached `[]`, and every later shim generation / `mise reshim` then trusted that frozen empty
snapshot (it's "fresh" by the install-dir mtime) — which is why a separate `reshim` process
also observed `[]`.

## Fix
- **`list_bin_paths`**: cache only the registry-derived candidate dirs (a pure function of
  the package definition, extracted as `candidate_bin_paths_for_platform`); check existence
  **live** on each call, on the install_path basis, before mounting onto runtime_path. The
  cached value can no longer encode transient install state. The cache filename is bumped to
  `bin_paths_v2` so pre-fix poisoned caches are ignored. The shim layer (`list_tool_bins`)
  already filters existence live, so this loses no behavior. Also documents the
  `Backend::list_bin_paths` contract ("return candidates; don't cache existence; callers
  filter live") to prevent recurrence.
- **`get_version_tags`**: an adjacent same-class fix found during the work — guard with
  `get_or_try_init_async_if(.., |v| !v.is_empty())` (the #9616/#9444 pattern) so an empty
  tag list (valid repo, all tags filtered out for the platform) isn't cached.

## Tests
- Unit test pinning that the candidate derivation is filesystem-independent.
- A deterministic e2e (`test_aqua_bin_paths_stale_cache_slow`) that reproduces the
  cache-poisoning (install uv, drop cache, hide the binary dir, reshim, restore + freeze
  mtime, reshim) and asserts the shims survive.
- Note: as a `_slow` test it runs in the release lane (`TEST_ALL=1`), not on per-PR CI —
  consistent with the repo's other slow network-installing e2e tests.

*This PR was prepared with the help of an AI coding assistant.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `aqua list_bin_paths` caching so a temporary missing backend directory no longer causes empty cache entries and prevents later shim/bin discovery.
* **Tests**
  * Added an end-to-end regression test for stale-cache behavior when an install directory is temporarily absent.
  * Added coverage to ensure platform-specific bin-path candidates are derived consistently even if install paths aren’t present.
* **Documentation**
  * Clarified expected behavior of bin-path candidate discovery versus runtime existence checks.

### DIFF
--- a/e2e/backend/test_aqua_bin_paths_stale_cache_slow
+++ b/e2e/backend/test_aqua_bin_paths_stale_cache_slow
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+# Regression test for #6468: aqua list_bin_paths must not cache transient
+# filesystem state. The old code cached the result of `.filter(|p| p.exists())`
+# computed inside the cache closure; if the binary dir was not on disk yet
+# (mid-install during a concurrent `mise upgrade`), it cached an empty list and
+# every later reshim trusted that empty snapshot, so the uv/uvx shims were
+# never (re)created — and `mise reshim` could not repair it.
+#
+# This reproduces the poisoning deterministically (no concurrency race):
+#   1. install uv (shims present; cache holds the real bin dir)
+#   2. drop the bin_paths cache and move uv's binary dir aside so it "doesn't exist"
+#   3. reshim -> recomputes list_bin_paths with the dir absent
+#        buggy code: caches []  (the .exists() filter strips everything)
+#        fixed code: caches the registry-derived candidate dir (no .exists())
+#   4. restore the dir, then freeze the install dir's mtime into the past so the
+#      freshly-written cache stays "fresh" (a normal mtime bump would invalidate
+#      the cache and mask the bug)
+#   5. reshim again -> consumes the cached value
+#        buggy code: cached [] -> uv shim NOT created  (this test fails = RED)
+#        fixed code: cached candidate -> live .exists() now passes -> shim created (GREEN)
+
+export MISE_AQUA_GITHUB_ATTESTATIONS=0
+
+mise use -g "aqua:astral-sh/uv@0.8.21"
+mise reshim
+assert "test -f \"$MISE_DATA_DIR/shims/uv\""
+
+# Locate the install from the executable itself. The install short-name for the
+# explicit `aqua:astral-sh/uv` backend is `aqua-astral-sh-uv` (not `uv`), and the
+# binary lives in a per-platform subdir (e.g. uv-aarch64-apple-darwin/uv), so we
+# derive both the bin dir and the version dir (= tv.install_path()) from the exe.
+uv_exe="$(find "$MISE_DATA_DIR/installs" -type f -name uv | head -n1)"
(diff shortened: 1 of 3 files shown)


devnulled:
Thanks for the reviews! Addressed the convergent finding from @coderabbitai and @greptile-apps:

- **e2e path discovery**: added an `assert "test -n \"$uv_exe\""` guard right after the `find`, so if the `uv` executable isn't located the test fails clearly instead of letting `dirname` fall back to `.` and having the later `mv`/`touch` operate on the current directory. (In the isolated e2e `TEST_HOME` there's exactly one `uv` install, so this is defensive — but it removes the cwd footgun.)

On the other observation:

- **`get_version_tags` clone on cache hit** (@greptile-apps): keeping the owned return. It's a small `Vec<(String, String)>` cloned a handful of times per resolve (not in a hot loop), and the alternative "don't cache empty" idiom (`get_or_try_init_async` + `clear()`, the #9444 pattern) also requires the clone — the immutable borrow has to be released before the cache can be mutated, so there's no zero-clone way to guard emptiness here.

jdx:
Thanks for digging into this. I think the root cause and the core shape of the fix are right: `list_bin_paths` needs to stay memoized for the hot shell/env paths, but the `.exists()` check should not be part of the memoized value.

The part I would change is the `bin_paths_v2.msgpack.z` filename. That feels like a cache schema migration trail we probably do not want to start. Could we keep using `bin_paths.msgpack.z` instead?

Unless I am missing something, `CacheManagerBuilder` already puts `built_info::PKG_VERSION` into the base cache key, along with target/profile/features/platform. So once this ships in a new mise version, old poisoned `bin_paths` cache files should already be ignored by the key. I do not think we need an extra filename version just to bust this cache.

In other words, cache the pure registry-derived candidate dirs, then always do the filesystem existence filter live after reading/fetching the cached candidates. That keeps the perf win for normal `env`/`hook-env`/PATH flows, while avoiding stale `[]` from install/upgrade/reshim races.

devnulled:
Agreed — dropped the \`bin_paths_v2\` filename and went back to \`bin_paths.msgpack.z\` in f5577a07c.

Verified your read of the key: \`CacheManagerBuilder::new\` seeds \`cache_keys\` from \`BASE_CACHE_KEYS\` (which includes \`built_info::PKG_VERSION\`) plus target/profile/features/os/arch/libc, and \`build()\` hashes that into the on-disk filename suffix. So a pre-fix poisoned \`bin_paths\` cache lands under a different key once this ships in a new mise version — no manual filename version needed. The comment now spells that out so nobody re-adds a \`_v2\`.

The actual cache-poisoning fix is unchanged: cache the pure registry-derived candidate dirs, filter \`.exists()\` live on every call.

PR #10597, opened 2026-06-24
fix(config): treat symlinked global config paths as global

JamBalaya56562:
## Summary

Fixes [discussion #10483](https://github.com/jdx/mise/discussions/10483): on systems where the shell's working-directory prefix differs from mise's resolved `$HOME` via a symlink — notably **Fedora Atomic / ostree**, where `/home` is a symlink to `/var/home` — files in the global config dir (`~/.config/mise/conf.d/*.toml`, `config.toml`, etc.) are wrongly treated as **non-global**. Global-only settings get stripped and the user sees:

```
mise WARN  yes in non-global config /var/home/.../conf.d/externally-managed.toml is ignored for security reasons
```

## Root cause

The same file is produced in two string forms:
- Local discovery globs `.config/mise/conf.d/*.toml` relative to the shell PWD → `/var/home/<user>/.config/mise/conf.d/x.toml`.
- `global_config_files()` builds it from `$HOME` → `/home/<user>/.config/mise/conf.d/x.toml`.

`load_config_paths` already treats these as one file (it dedupes via `file::desymlink_path`), but the security gate `is_global_config`/`is_system_config` used **raw string equality** (`IndexSet::contains`). The PWD-derived path isn't byte-equal to the `$HOME`-derived entry, so the file is judged non-global and `strip_local_only_settings` removes the global-only keys.

## Fix

Make the global/system membership checks symlink-aware, reusing the existing `file::desymlink_path` helper that the dedupe path already trusts. A new private `config_set_contains` keeps a raw-equality **fast path** (no filesystem hit) and only canonicalizes on a miss:

```rust
fn config_set_contains(set: &IndexSet<PathBuf>, path: &Path) -> bool {
    if set.contains(path) {
        return true;
    }
    let target = file::desymlink_path(path);
    set.iter().any(|p| file::desymlink_path(p) == target)
}
```

`is_global_config`/`is_system_config` now call it. With no symlinked home (the common case) behavior is unchanged — the fast path matches and no extra filesystem calls are made. The change only makes genuinely-global files be recognized as global; it cannot classify an unrelated local file as global unless it canonicalizes to the exact same path as a real global config file.

## Testing

Pure, CI-runnable unit test (`#[cfg(unix)]`, models the `/home` → `/var/home` setup with a real tempdir symlink): `test_config_set_contains_matches_symlinked_prefix` in `src/config/mod.rs`. It asserts an exact match (fast path), a symlinked-prefix match (the bug — false before the fix), and a non-member. Fails on the unpatched tree, passes after.

`rustfmt --edition 2024 --check` passes.

> [!NOTE]
> The authoring sandbox can't build mise (insufficient memory), so compile/clippy/tests are left to CI. Manual check for a maintainer on an ostree box (`/home` → `/var/home`): put `yes = true` in `~/.config/mise/conf.d/x.toml`, run a mise command from `$HOME`, and confirm the "ignored for security reasons" warning is gone and the setting is honored; a genuinely local `mise.toml` with a global-only key should still warn/strip.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of global and system configuration files when reached via symlinked path prefixes.
  * Configuration matching now correctly recognizes the same config file through canonicalized/desymlinked paths, reducing missed valid configs.
* **Tests**
  * Added coverage to ensure config matching works for both exact paths and symlink-prefix aliases.

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -1369,11 +1369,26 @@ pub async fn load_config_hierarchy_from_dir(
 }
 
 pub fn is_global_config(path: &Path) -> bool {
-    global_config_files().contains(path) || is_system_config(path)
+    config_set_contains(&global_config_files(), path) || is_system_config(path)
 }
 
 pub fn is_system_config(path: &Path) -> bool {
-    system_config_files().contains(path)
+    config_set_contains(&system_config_files(), path)
+}
+
+/// Membership test that tolerates symlinked path prefixes (e.g. Fedora Atomic's
+/// `/home` -> `/var/home`, where the shell PWD and `$HOME` disagree on the
+/// prefix). The fast path is raw equality (no filesystem hit); only on a miss do
+/// we canonicalize via [`file::desymlink_path`], matching how `load_config_paths`
+/// already dedupes config paths. Without this, a global config discovered via the
+/// PWD prefix is not byte-equal to its `$HOME`-derived entry and is wrongly
+/// treated as a local config (stripping global-only settings).
+fn config_set_contains(set: &IndexSet<PathBuf>, path: &Path) -> bool {
+    if set.contains(path) {
+        return true;
+    }
+    let target = file::desymlink_path(path);
+    set.iter().any(|p| file::desymlink_path(p) == target)
 }
 
 /// Returns true if the path should be filtered out due to MISE_CONFIG_DIR override.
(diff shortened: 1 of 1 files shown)


JamBalaya56562:
Thanks @greptile-apps — addressed the out-of-diff note. `is_default_config_dir_override_filtered` was using the same raw `global_config_files().contains(path)` check, so it shared the identical symlinked-home bug class (under a `MISE_CONFIG_DIR` override). It now goes through the same `config_set_contains` helper, so a global config discovered via a symlinked PWD prefix is no longer incorrectly filtered out during traversal. Same root cause, same file, one-line change.

PR #10600, opened 2026-06-24
fix(aqua): support doggo v1.2.0 asset naming in vendored registry

jhult:
Doggo v1.2.0 changed its release asset naming from versioned (`doggo_{{version}}_{{OS}}_{{Arch}}.tar.gz`) to versionless (`doggo-{{os}}-{{arch}}.tar.gz`) with lowercase OS names and aarch64/i386 arch names.

Added a version override for >= 1.2.0 in the vendored aqua registry to match the new naming while the existing catch-all continues handling older versions.

## Summary by CodeRabbit

* **Chores**
  * Updated release artifact handling for a packaged download to better match operating systems, architectures, and archive formats.
  * Added checksum-based verification and improved platform-specific packaging, including a Windows zip download option.

### DIFF
--- a/vendor/aqua-registry/registry.yml
+++ b/vendor/aqua-registry/registry.yml
@@ -65696,6 +65696,23 @@ packages:
           type: github_release
           asset: doggo_{{trimV .Version}}_checksums.txt
           algorithm: sha256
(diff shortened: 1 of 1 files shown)


jhult:
Will fix upstream.

PR #10615, opened 2026-06-25
fix(hooks): set MISE_INSTALLED_TOOLS to [] on no-op install (keep running postinstall)

JamBalaya56562:
## Summary

Addresses discussion #10574. On a no-op `mise install` (every configured tool already installed), the global `postinstall` hook left `MISE_INSTALLED_TOOLS` **unset**, contradicting the docs which promise a JSON array of installed tools.

Per review, this keeps `postinstall` running on every `mise install` — no change to *when* the hook fires (it's commonly relied on for idempotent project setup once the toolchain is ensured, even on a fresh clone that already has every tool installed) — and instead makes `MISE_INSTALLED_TOOLS` **consistently valid JSON**:

- `[]` when nothing was installed (no-op, or all installs failed)
- `[{"name":..,"version":..}, …]` when tools were installed

Hooks that should only act on real installs can guard on `MISE_INSTALLED_TOOLS != "[]"`.

## Change

The only behavioral change is on the no-op path in `src/cli/install.rs`: it now passes an empty installed-tools list (`Some(&[])`) to the postinstall hook instead of `None`, so the env-setter emits `[]` rather than skipping. The earlier skip logic (in `install.rs` and `src/toolset/toolset_install.rs`) is reverted.

## Tests / docs

- `e2e/config/test_hooks_installed_tools`: adds a no-op-install case asserting `MISE_INSTALLED_TOOLS=[]`.
- `e2e/config/test_hooks`: confirms `postinstall` still fires on a no-op `mise install`.
- `docs/hooks.md`: documents that `postinstall` always runs and `MISE_INSTALLED_TOOLS` is `[]` when nothing was installed.

Ref: discussion #10574

## Summary by CodeRabbit

* **Documentation**
  * Clarified that install commands still run post-install hooks even when nothing new is installed.
  * Updated hook environment variable guidance to show it contains a JSON array, including `[]` for no-op installs.

* **Bug Fixes**
  * Ensured hooks receive an empty installed-tools list during no-op installs, so scripts can reliably detect real installs.
  * Expanded end-to-end checks to verify hook behavior for both no-op and normal install flows.

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -59,7 +59,9 @@ postinstall = { run = "echo installed", run_windows = "Write-Output installed" }
 
 For `preinstall` and `postinstall`, `script = ...` is a legacy alias for `run = ...`. If a `shell` is also set on a `script`/`scripts` hook, mise warns that the shell is ignored and still runs the script with the default inline shell. Use `run = ...` with `shell = "bash -c"` to choose the inline shell command. The `script` alias for install hooks is deprecated.
 
-The `postinstall` hook receives a `MISE_INSTALLED_TOOLS` environment variable containing a JSON array of the tools that were just installed:
+A `mise install` that finds nothing to install (all configured tools are already present) still runs the `postinstall` hook — it is not skipped on a no-op install.
+
+The `postinstall` hook receives a `MISE_INSTALLED_TOOLS` environment variable containing a JSON array of the tools that were just installed, or `[]` when nothing was installed (e.g. a no-op install). Hooks that should only act on real installs can guard on `MISE_INSTALLED_TOOLS != "[]"`:
 
 ```toml
 [hooks]
--- a/e2e/config/test_hooks
+++ b/e2e/config/test_hooks
@@ -12,8 +12,10 @@ preinstall = { run = 'echo PREINSTALL' }
 postinstall = { run = 'echo POSTINSTALL' }
 EOF
 
+# preinstall + postinstall hooks fire when installing
 assert_contains "mise i 2>&1" "PREINSTALL"
 assert_contains "mise i dummy@1 2>&1" "POSTINSTALL"
+# `mise i` with nothing missing still runs postinstall (MISE_INSTALLED_TOOLS=[], #10574)
 assert_contains "mise i 2>&1" "POSTINSTALL"
 assert_contains "mise i dummy@1 2>&1" "POSTINSTALL"
 eval "$(mise hook-env)"
--- a/e2e/config/test_hooks_installed_tools
+++ b/e2e/config/test_hooks_installed_tools
@@ -33,3 +33,14 @@ else
   echo "Output: $output"
   exit 1
 fi
+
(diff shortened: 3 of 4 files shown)


jdx:
I think the current behavior is correct

JamBalaya56562:
Thanks for taking a look, and that's a totally fair point — honestly I started from the same assumption that the current behavior was correct. What made me reconsider was comparing it against how other package managers treat "post-install" hooks:

- **npm** (dependency lifecycle scripts), **uv**, **pip**, **cargo** (`build.rs`), **Homebrew** (`post_install`), **dpkg** (`postinst`) all tie the post-install step to an *actual* install event and skip it on a no-op. Running `npm install`/`uv sync`/etc. when everything is already present doesn't re-fire those hooks.

So my worry isn't really that the current behavior is "wrong" — it's that it diverges from the convention people carry over from other tools, which makes it easy to misread. The name `postinstall` reads as "after something was installed," and mise already leans that way by passing `MISE_INSTALLED_TOOLS` (the docs describe it as "a JSON array of the tools that were just installed"). In the no-op case that variable is currently left unset rather than `[]`, which also hints the hook is conceptually about an install event.

To be fair, there *is* precedent for "run every time": npm's **root** `package.json` scripts do run on every `npm install`. So I can see the argument for keeping it — this is genuinely a judgment call and I'm happy to defer to you on it.

If you'd rather keep the current behavior, a couple of alternatives I'd be glad to do instead:
- just document it clearly (postinstall always runs; guard with `MISE_INSTALLED_TOOLS`), and fix the no-op case to pass `[]` instead of leaving the var unset, or
- make the skip opt-in via a setting so the default is unchanged.

Whatever you prefer — let me know and I'll adjust or close. Thanks!

jdx:
I don't want to merge this as-is because I consider it a breaking behavior change.

Global `postinstall` is commonly used for idempotent project setup after `mise install` has ensured the toolchain is available, not only after mise physically installs a new tool version. For example, a fresh clone may already have all configured tool versions installed from another project, but still rely on `mise install` to run setup tasks afterward. Skipping `postinstall` on that no-op path would silently break those workflows.

I agree there is a bug here, but I think the fix should be narrower: keep running `postinstall`, and make `MISE_INSTALLED_TOOLS` consistently contain JSON, using `[]` when nothing was installed. That gives users who want actual-install-only behavior a reliable guard without changing existing hook semantics.

JamBalaya56562:
Thanks, that makes sense — reworked to your suggestion.

`postinstall` now always runs (no change to *when* it fires), and `MISE_INSTALLED_TOOLS` is consistently a JSON array: `[]` when nothing was installed, `[{"name":..,"version":..}]` otherwise. Hooks that only want to act on real installs can guard on `MISE_INSTALLED_TOOLS != "[]"`.

The skip logic in both the `install.rs` no-op branch and `toolset_install.rs` is reverted — the only behavioral change now is that the no-op path passes an empty installed-tools list instead of `None`, so the env var is `[]` instead of unset. Updated the e2e tests and docs to match.

PR #10622, opened 2026-06-25
fix(aqua): download private release assets via the GitHub API asset endpoint

yacchi:
## Summary
- the `aqua` backend downloaded release assets from the browser download URL (`github.com/.../releases/download/...`), which 404s for **private** repositories even with a valid token, so private tools could not be installed
- keep the browser URL as the canonical URL (filename derivation, lockfile storage and version detection are unchanged for public repos) and carry the API asset URL (`api.github.com/.../releases/assets/{id}`) alongside it; at download time, probe the browser URL and fall back to the API endpoint on a `404` / HTML auth page
- extract the browser-URL-with-API-fallback logic that already existed inline in the `github` backend into a shared `github::pick_reachable_asset_url` helper used by both backends

## Tests
- `cargo test --bin mise backend::aqua` (56 passed)
- `cargo test --bin mise github` (76 passed, incl. new `pick_reachable_asset_url` tests via mockito: reachable / `404` / `text/html` / equal-URL skip)
- `cargo clippy --bin mise` (clean)
- manual end-to-end against a real **private** repo: browser URL `404` → fallback to API asset endpoint → correct filename → checksum (API digest) → extract → binary runs; public-repo installs unaffected

*This PR description was generated by an AI coding assistant.*

## Summary by CodeRabbit

* **New Features**
  * Improved GitHub/GitLab/Gitea release downloads by automatically selecting the first reachable asset endpoint (browser vs API) to avoid errors and HTML sign-in responses.
  * Enhanced locked and private-repo install flows by persisting both the resolved download URL and the API fallback for more reliable re-installs.
  * Updated provenance and auxiliary asset fetching to use the same reachable endpoint selection.

* **Bug Fixes**
  * Reduced failures when browser links return 404 or HTML pages.
  * Improved consistency between lockfile-referenced metadata and actual download endpoints.

* **Tests**
  * Added unit tests for reachable selection, 404 fallback, HTML detection (including header variations), and skipping probing when endpoints match.

### DIFF
--- a/src/backend/aqua.rs
+++ b/src/backend/aqua.rs
@@ -880,9 +880,7 @@ impl Backend for AquaBackend {
                 .resolve_slsa_url(&pkg, &v, target_os, target_arch)
                 .await
             {
-                // Store the canonical URL in the lockfile; the transport URL is only
-                // resolved at download time (and may differ for private repos).
-                Ok((canonical_url, _transport_url)) => {
+                Ok((canonical_url, _url_api)) => {
                     provenance = Some(ProvenanceType::Slsa {
                         url: Some(canonical_url),
                     });
@@ -1383,18 +1381,19 @@ impl AquaBackend {
         }
     }
 
-    /// Resolve the SLSA provenance `(canonical_url, transport_url)` for a target platform
-    /// without downloading. Uses cached GitHub release data or template-based URL
-    /// construction. `canonical_url` carries the real filename and is what gets stored as
-    /// the lockfile provenance URL; `transport_url` is where the file is actually fetched
-    /// from (see `github_release_asset_urls`).
+    /// Resolve the SLSA provenance canonical URL and optional API URL for a target
+    /// platform without downloading or probing. `canonical_url` carries the real
+    /// filename and is what gets stored as the lockfile provenance URL; `api_url`
+    /// (if present) is the GitHub API asset endpoint that can serve private-repo
+    /// assets — the actual reachability probe is deferred to download time (see
+    /// `run_slsa_check`).
     async fn resolve_slsa_url(
         &self,
         pkg: &AquaPackage,
         v: &str,
         target_os: &str,
         target_arch: &str,
-    ) -> Result<(String, String)> {
+    ) -> Result<(String, Option<String>)> {
         let slsa = pkg
             .slsa_provenance
             .as_ref()
@@ -1407,13 +1406,12 @@ impl AquaBackend {
         match slsa.r#type.as_deref().unwrap_or_default() {
             "github_release" => {
                 let asset_strs = slsa.asset_strs(&slsa_pkg, v, target_os, target_arch)?;
-                self.github_release_asset_urls(&slsa_pkg, v, asset_strs)
-                    .await
+                let (url, url_api, _) = self.github_release_asset(&slsa_pkg, v, asset_strs).await?;
+                Ok((url, url_api))
             }
-            // Non-release URLs have no separate transport endpoint.
             "http" => slsa
                 .url(&slsa_pkg, v, target_os, target_arch)
-                .map(|url| (url.clone(), url)),
+                .map(|url| (url, None)),
             t => Err(eyre!("unsupported slsa type: {t}")),
         }
     }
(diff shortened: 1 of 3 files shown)


jdx:
I think the direction here is right, but this is also where I think we may have intentionally avoided doing it earlier: the browser release URL and the API asset URL are not interchangeable.

The browser URL is the canonical asset identity: it has the real release asset filename and is what lockfile matching/version detection/templates should reason about. The API URL is a transport endpoint (`/releases/assets/{id}`) that only becomes useful at download time for private repos, with `Accept: application/octet-stream`.

So I would be cautious about returning a single “downloadable URL” from helper paths that are also used for filename derivation, checksum/minisign template context, or stored provenance URLs. If one of those auxiliary assets falls back to the API URL, `get_filename_from_url()` sees the numeric asset id rather than the release asset name. That may be harmless for a raw download, but it is not harmless everywhere.

I would prefer to keep the split explicit throughout aqua, something like: canonical/browser URL + optional API URL + selected transport URL. Use the browser URL for lockfile identity, filename derivation, template/matching/provenance metadata; use the selected transport URL only at the actual HTTP download call.

This is basically the same constraint the github backend already respects: do not send GitHub auth to `github.com/.../releases/download/...`; only use the authenticated API asset endpoint when the browser URL is not reachable.

yacchi:
Thanks — agreed, and this is exactly the part to be careful about. I've reworked the auxiliary asset paths so the browser/API split is explicit throughout aqua, the same way the github backend already handles it.

**What changed**

- Dropped `github_release_asset_download_url` (which collapsed both URLs into a single "downloadable URL") in favor of `github_release_asset_urls`, which returns `(canonical_url, transport_url)`.
- `canonical_url` is the browser download URL — used for filename derivation and for any URL persisted as metadata (the stored SLSA provenance URL).
- `transport_url` is used **only** at the actual `HTTP.download_file`/`get_text` call. For private repos it falls back to the API asset endpoint via the existing `github::pick_reachable_asset_url` helper.
- This means a private-repo fallback can no longer leak the numeric `releases/assets/{id}` into `get_filename_from_url()` or into the stored provenance URL. Applied across all auxiliary assets: SLSA provenance, cosign (key/signature/bundle), minisign, and checksum files.
- The main artifact path already kept this split, so it's unchanged.
- `pick_reachable_asset_url` stays shared between the github and aqua backends so both resolve private assets identically (no GitHub auth sent to `github.com/.../releases/download/...`; the API asset endpoint is used only when the browser URL isn't reachable).

This also resolves the CodeRabbit findings about `url_api` being discarded on the SLSA provenance and checksum paths — those now download via the transport URL too.

`cargo test backend::aqua` (56), `cargo test github` (77), `cargo clippy`, and `cargo fmt --check` are all green.

jdx:
Thanks — the rework addresses my concern properly. I traced the call sites: filenames and persisted metadata (lockfile `url`, stored SLSA provenance URL) always derive from the canonical browser URL, and the transport URL only ever feeds the actual HTTP fetch, so the numeric `/releases/assets/{id}` can no longer leak into filenames or lockfile identity. The case-insensitive content-type match in `pick_reachable_asset_url` is a nice fix folded into the extraction, and reusing the existing `url_api` lockfile field keeps this consistent with the github backend.

One thing I'd like fixed before merge:

- **Lock generation does a wasted HEAD probe.** `resolve_lock_options_for_target` → `resolve_slsa_url` → `github_release_asset_urls` resolves a transport URL that's immediately discarded (`Ok((canonical_url, _transport_url))`). During `mise lock` across N platforms that's a pointless network round-trip per platform. A resolve-only variant — return `(canonical, Option<api_url>)` and defer the probe to download time — would fix it cleanly.

Non-blocking observations, no action needed:

- Every fresh aqua GitHub-release install now pays a HEAD probe on the browser URL (plus one per auxiliary asset). That matches the github backend's existing behavior so I'm fine with it, but if it ever shows up in install latency, skipping the probe when no GitHub token is configured would be a cheap win (the API fallback can't serve private assets unauthenticated anyway).
- Old lockfiles for private tools (`url` present, no `url_api`) in `--locked` mode still fail with a raw 404; re-locking fixes it, so not worth handling.
- Trivial: `tv.lock_platforms.get(&platform_key)` is looked up twice back-to-back at the top of `install_version_`.

yacchi:
Thanks for the review! Addressed the blocking item:

**Lock generation wasted HEAD probe — fixed in c98e86e**

`resolve_slsa_url` now returns `(canonical_url, Option<api_url>)` without probing. The reachability probe is deferred to `run_slsa_check` where the transport URL is actually needed for downloading. This eliminates the wasted HEAD round-trip per platform during `mise lock`.

Changes:
- `resolve_slsa_url` calls `github_release_asset` directly instead of `github_release_asset_urls`, returning the raw `(url, url_api)` without the `pick_reachable_asset_url` probe
- `run_slsa_check` applies the probe at download time where the transport URL is actually consumed
- `"http"` SLSA type returns `None` for `api_url` (no separate transport endpoint)

All tests pass: `cargo test backend::aqua` (56), `cargo test github` (77), `cargo clippy`, `cargo fmt --check`.

yacchi:
Rebased onto main to resolve the conflict with #10703 (`fix(github): handle missing lockfile url_api`).

The `url_api.is_empty()` guard from #10703 is incorporated into the `pick_reachable_asset_url` call site — empty `url_api` skips the probe and uses the browser URL directly, matching the upstream behavior.

PR #10626, opened 2026-06-25
fix(brew-cask): handle raw binaries, $APPDIR paths, and app bundle copying

arthurh4:
Fixes issues raised in #10625 (discussion).

## Summary

Three related fixes for `brew-cask` cask installation failures:

**1. Raw executable binaries (e.g. `claude-code`)**

Casks like `claude-code` ship a raw precompiled binary rather than an archive. The extractor was bailing with `unsupported archive type` because `ExtractionFormat::Raw` doesn't pass `is_archive()`. Fix: copy the file directly into the extract directory using the original URL filename (last path segment of the cask URL), so `find_artifact` can match it against the `binary` stanza source name.

**2. `$APPDIR` binary sources (e.g. `cursor`)**

Some casks (e.g. `cursor`) use `$APPDIR/Cursor.app/Contents/Resources/app/bin/code` as the binary source. Previously `find_artifact` would search for the literal string `$APPDIR/...` and fail. Fix: resolve `$APPDIR` to `/Applications` (where `install_app` placed the bundle) and symlink into the installed app bundle — rather than copying the binary out — so the CLI wrapper script can trace back through symlinks to find the `.app`.

**3. App bundle copying uses `ditto` instead of plain recursive copy**

`file::copy_dir_all` is a plain file-by-file copy that strips resource forks and extended attributes. On macOS this causes the "damaged app" Gatekeeper dialog. Fix: use `ditto` (the same tool Homebrew uses) which preserves HFS+ metadata, resource forks, and extended attributes.

Also strips `com.apple.quarantine` from downloaded archives immediately after download, before extraction propagates it into the app bundle.

## Testing

Verified locally against both affected casks:
- `brew-cask:claude-code@latest` — installs and launches successfully
- `brew-cask:cursor` — installs, opens without "damaged" dialog, and `cursor` CLI works

*This PR was authored with the assistance of an AI coding assistant.*

## Summary by CodeRabbit

* **New Features**
  * Improved handling for raw executable downloads, including correct placement and launch permissions.
  * Added support for app sources that reference an app directory path, enabling cleaner app linking during installation.
* **Bug Fixes**
  * Reduced macOS quarantine issues by clearing quarantine attributes right after download and after app installation.
  * Improved app bundle installation reliability with safer copy-and-swap behavior and rollback on failure.
* **Other**
  * Better error reporting when an unsupported extraction format is encountered.

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -59,7 +59,9 @@ postinstall = { run = "echo installed", run_windows = "Write-Output installed" }
 
 For `preinstall` and `postinstall`, `script = ...` is a legacy alias for `run = ...`. If a `shell` is also set on a `script`/`scripts` hook, mise warns that the shell is ignored and still runs the script with the default inline shell. Use `run = ...` with `shell = "bash -c"` to choose the inline shell command. The `script` alias for install hooks is deprecated.
 
-The `postinstall` hook receives a `MISE_INSTALLED_TOOLS` environment variable containing a JSON array of the tools that were just installed:
+A `mise install` that finds nothing to install (all configured tools are already present) still runs the `postinstall` hook — it is not skipped on a no-op install.
+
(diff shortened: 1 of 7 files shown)


arthurh4:
@coderabbitai review

PR #10653, opened 2026-06-27
feat: Android asset detection

bltavares:
Since the achievement of having `mise` in Termux packages (https://github.com/jdx/mise/discussions/7026)
I've noticed it does not detect yet Android binaries.

This commit introduces Android asset detection. It expects the `android`
term to be present to pickup the asset. I've not added it as a `fn common_platform()` 
as most tools don't have `linux-android` releases and it would increase the lockfile for all
users with little gains. `Platform` will still accept a valid value as an argument, but it
will not add entries by default on lockfiles unless running on Android as `current`.

I've considered both how `goreleaser` (Go) and `cargo cross` generate
binaries for triplet detection. Rust ecosystem uses
`aarch64-linux-android`, which would trigger the Linux package regex as
well. That is why the Android matcher comes first on the `OS_PATTERNS`.

I've tested this with 2 different packages:

- github:bltavares/opkssh - Go raw binary with `goreleaser` typical GOOS and GOOARCH
- github:bltavares/rotz - Rust .zip with `cross` target triplets

Now `mise x` can downlod the right binary and `.zip`

## Summary by CodeRabbit

* **New Features**
  * Added Android to platform detection to improve asset selection and downloads.
  * Android is now supported across common artifact naming formats, including `android-<arch>` platform strings.

* **Bug Fixes**
  * Platform validation now accepts `android`, and unsupported-OS messaging reflects it.
  * Android-labeled assets are correctly prioritized for Android targets and not favored for non-Android targets.

* **Tests**
  * Expanded platform selection and scoring coverage to include Android for both naming schemes and additional conversion cases.

### DIFF
--- a/src/backend/asset_matcher.rs
+++ b/src/backend/asset_matcher.rs
@@ -1489,8 +1489,8 @@ abc123def456abc123def456abc123def456abc123def456abc123def456abcd  tool-1.0.0-dar
     }
 
     #[test]
-    fn test_asset_picker_functionality_rust_triplets() {
-        // rust triplet platofmrs contain linux OS for Android binaries
+    fn test_asset_picker_functionality_rust_triplets_with_android() {
+        // rust triplet platforms contain linux as segment in filname for Android binaries
         let assets = vec![
             "tool-1.0.0-aarch64-apple-darwin.zip".to_string(),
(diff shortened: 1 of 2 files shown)


bltavares:
@coderabbitai resume

PR #10683, opened 2026-07-01
fix(shims): don't point shims at a foreign `mise` wrapper on PATH

timqi:
## Summary

`reshim` picks the target that every tool shim is symlinked to via
`file::which_no_shims("mise")`. That helper excludes mise's **own** shim
directories (added in #9071 to fix the `mise -> mise` self-reference when a
`mise` shim exists in mise's shims dir), but it still trusts **any other**
program named `mise` that happens to appear earlier on `PATH`.

If a third-party shim/launcher directory on `PATH` contains a `mise` wrapper —
for example a tool that intercepts `mise exec …` by putting its own `mise`
shim first on `PATH` — then `which_no_shims("mise")` returns that wrapper, and
**every** generated tool shim gets symlinked to it. Running any managed tool
(`node`, `gh`, `lsd`, …) now re-enters the wrapper, which resolves the tool
name back to the same wrapper, and the shim loops until it hits a recursion
guard and errors out. All shims are silently broken, and because mise reshims
automatically (on install/use), it recurs right after any manual fix.

This is the same bug class as #9071, just from a directory that the
`which_no_shims` guard doesn't cover.

## Reproduction

```sh
# A foreign wrapper named `mise` earlier on PATH than the real binary:
mkdir -p /tmp/wrapper && ln -sf "$(command -v some-other-tool)" /tmp/wrapper/mise
export PATH="/tmp/wrapper:$PATH"

mise reshim --force
readlink ~/.local/share/mise/shims/node
#   -> /tmp/wrapper/mise      (should be the real mise binary)
node --version
#   -> loops / recursion-guard error instead of running node
```

Before this change every shim points at `/tmp/wrapper/mise`; after it, mise
warns about the shadowing `mise` and points shims at the running executable.

## Fix

Add `file::resolve_mise_bin()` and route both call sites through it (`reshim`
in `src/shims.rs`, and `analyze_shims` in `src/cli/doctor/mod.rs`, mirroring
#9071). It keeps the existing `which_no_shims("mise")` behavior but validates
the result: if the candidate does **not** symlink-resolve to a file actually
named `mise` (or `mise-shim`), it warns that a non-mise `mise` is shadowing us
and falls back to the running executable (`env::MISE_BIN`).

## Compatibility with #1222

Shims intentionally point at a **stable** `mise` on `PATH` rather than
`current_exe()`, so they survive in-place upgrades where `current_exe()` is a
versioned/transient path (Homebrew `.../Cellar/mise/<version>/bin/mise`,
`cargo install`, self-update). This is preserved: when the candidate passes the
check, the **un-canonicalized** `PATH` entry is returned. Canonicalization is
used **only** to verify identity, never as the shim target. Behavior changes
**only** in the previously-broken case (a foreign `mise` wrapper), where the
running executable is strictly better than a self-referential loop.

## Tests

- `test_canonical_is_mise` — pure discriminator: accepts real mise binaries at
  any install path/extension (incl. `mise.exe`, `mise-shim`), rejects foreign
  names (`vt`, `direnv`).
- `test_resolve_mise_bin_sees_through_foreign_wrapper_symlink` (unix) — a
  `PATH` entry literally named `mise` that is a symlink to a foreign binary is
  rejected after resolution, while a `mise` symlink pointing at a real file
  named `mise` (the Homebrew/self-update case) is still accepted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of the `mise` executable so shim diagnostics and reshimming use the correct binary more reliably.
  * Prevented misleading path resolution when a similarly named executable appears earlier on `PATH`, reducing incorrect shim behavior.
  * Added clearer fallback handling and warnings when the detected executable is not a valid `mise` binary.

### DIFF
--- a/src/cli/doctor/mod.rs
+++ b/src/cli/doctor/mod.rs
@@ -745,7 +745,7 @@ impl Doctor {
     }
 
     async fn analyze_shims(&mut self, config: &Arc<Config>, toolset: &Toolset) {
-        let mise_bin = file::which_no_shims("mise").unwrap_or(env::MISE_BIN.clone());
+        let mise_bin = file::resolve_mise_bin();
 
         if let Ok((missing, extra)) = shims::get_shim_diffs(config, mise_bin, toolset).await {
             let cmd = style::nyellow("mise reshim");
--- a/src/file.rs
+++ b/src/file.rs
@@ -924,6 +924,58 @@ pub fn which_no_shims<P: AsRef<Path>>(name: P) -> Option<PathBuf> {
     _which(name, &paths)
 }
 
+/// True if `real` (an already symlink-resolved path) is actually the `mise`
+/// binary rather than some foreign program. Split out from
+/// [`resolve_mise_bin`] so the discriminating logic is unit-testable without
+/// touching the filesystem or PATH. The extension is ignored via `file_stem`
+/// (so `mise.exe` counts), and mise's own Windows shim (`mise-shim`) is
+/// accepted.
+fn canonical_is_mise(real: &Path) -> bool {
+    matches!(
+        real.file_stem().and_then(|s| s.to_str()),
+        Some("mise") | Some("mise-shim")
+    )
+}
+
+/// Resolve which `mise` binary the generated shims should point at.
+///
+/// Shims are symlinked to a *stable* `mise` found on PATH rather than to the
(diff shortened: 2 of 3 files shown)


jdx:
do not shim mise

PR #10730, opened 2026-07-02
fix: extend match of Linux musl asset for Android

bltavares:
Previously we added Android as a valid platform for asset scoring by requiring packages to publish an `*-android-*` artifact. This is expected, as these binaries take in consideration differences between Linux and Android.

Not every package requires Android specific compilation to work on Termux tho. Often, `musl` binaries might work with Android/Termux environment if it's statically linked. Typically tho, they have tend to have network issues specially arround DNS resolution, yet it might work as is if there is no network.

An example, `github:starship/starship`, `github:sharkdp/bat` and `github:jdx/hk` which don't publish Android specific assets, but their `linux-musl` satic binaries can be executed on Termux.

This commit expands the initial Android asset matcher to also consider `linux-musl` as a valid target. For this to work, `OS_PATTERNS` for Android was expanded to allow `linux-musl`, which required some rework on the scoring logic for Linux. Now, the scoring matcher will not stop on the first asset and target_os combination, and fall through to the next matchers.

It will only bail if there was a regex matching the asset file name but no other item matched regex and target_os. This retain effectively the same logic as before, now allowing it the fall-through on similar regex for different OS. To avoid potential performance loss on more common patterns, the Android regex was moved to last of the list to be the last fall-through matcher.

Now that `linux-musl` are a valid target for 2 different platforms, there was an additional boost/demotion logic required, otherwise assets might tie. For Linux target OS, we actually demote any Android marked asset to avoid ties, and for Android targets we boost the score.

Typically, Android/Termux can work with static Linux musl, but will fail with SEGV on GNU glibc binaries due to shared libraries not matching the ABI. The Android boost score take it in consideration and increases the preference in the following order: `*android*`, `*musl*` and `*gnu*`.

Scoring for Linux remains the same, ignoring any Android asset and using the correct `MISE_LIBC` defined.

This will allow Android users to download static musl binaries as a fallback if an Android targeted asset are not published, but there is no guarantee that the upstream package will work as is. If there are issues running Linux musl binaries on Termux, it should be reported upstream. Mise will pickup automatically new releases when they start publishing Android specific assets.

Tested repositories on Termux with this commit:
  - `mise x github:bltavares/opkssh -- opkssh login`
  Downloads Android release from goreleaser and works as expected
  - `mise x github:bltavares/rotz -- rotz install`
  Downloads Android release from cross and works as expected
  - `mise x github:sharkdp/bat -- bat README.md`
  Downloads Linux musl release and works as expected
  - `mise x github:starship/starship -- starship prompt`
  Downloads Linux musl release and works as expected
  - `mise x github:jdx/usage -- usage help`
  Downloads Linux musl release and works as expected
  - `mise x github:jdx/hk -- hk --version`
  Downloads Linux musl release and run, except for network calls (typical Linux on Android issue which requires a targeted Android release)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved asset selection by correctly handling Android artifacts that also include Linux or musl indicators.
  * Refined platform scoring and OS token matching to boost accurate matches and apply stronger mismatch penalties.
  * Improved fallback behavior for Android targets and enhanced platform detection to avoid misclassifying Android artifacts as Linux.
* **Tests**
  * Added and updated test coverage for Android fallback, scoring order/penalties, and platform detection outcomes.

### DIFF
--- a/src/backend/asset_matcher.rs
+++ b/src/backend/asset_matcher.rs
@@ -122,10 +122,6 @@ impl DetectedPlatform {
 static OS_PATTERNS: LazyLock<Vec<(AssetOs, Regex)>> = LazyLock::new(|| {
     vec![
         (
-            AssetOs::Android,
-            Regex::new(r"(?i)(?:\b|_)(?:android)(?:\b|_)").unwrap(),
-        ),
-        (
             AssetOs::Linux,
             Regex::new(r"(?i)(?:\b|_)(?:linux|manylinux(?:[0-9_]+)?|musllinux(?:[0-9_]+)?|ubuntu|debian|fedora|centos|rhel|alpine|arch)(?:\b|_|32|64|-)")
                 .unwrap(),
@@ -137,6 +133,10 @@ static OS_PATTERNS: LazyLock<Vec<(AssetOs, Regex)>> = LazyLock::new(|| {
         (
             AssetOs::Windows,
             Regex::new(r"(?i)(?:\b|_)(?:mingw-w64|win(?:32|64|dows)?)(?:\b|_)").unwrap(),
+        ),
+        (
+            AssetOs::Android,
+            Regex::new(r"(?i)(?:\b|_)(?:android|linux|musllinux(?:[0-9_]+)?)(?:\b|_)").unwrap(),
         ),
     ]
 });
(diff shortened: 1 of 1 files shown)


bltavares:
>  The one gap is a combined test asserting that a native android asset beats a linux-musl fallback when both are present in the same release—this scenario exists in production but is untested, so a scoring regression there would be silent.

@greptileai This should be covered in `test_asset_picker_functionality_rust_triplets_with_android` already. Android and MUSL asset are listed and Android target OS prefers native Android asset

PR #10750, opened 2026-07-03
fix: accept scoped release tags in cached GitHub asset URLs

zeitlinger:
## Summary
- accept cached GitHub release download URLs whose tags span multiple path segments
- keep validating owner/repo and the trailing asset filename as before
- add coverage for Biome's scoped tag format (`@biomejs/biome@2.5.2`)

## Root cause
GitHub serves Biome download URLs like:
`/releases/download/%40biomejs/biome%402.5.2/biome-linux-x64`

The cached release validator in `mise` expected the tag to occupy exactly one path segment between `download/` and the asset filename. For scoped tags, the encoded `/` becomes a real path separator in the download URL, so the validator rejected otherwise-valid asset URLs and logged `outcome=invalid_asset_urls` before falling back to the GitHub API.

## Validation
- `cargo test test_valid_github_browser_download_url`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for GitHub release download links, including URLs with more complex tag paths and encoded characters.
  * Expanded support for browser download links that previously may not have been recognized correctly.


risu729:
Verified locally with an actual `mise` command.

Command used on both `main` and this PR branch, with isolated data/cache dirs:

```sh
rm -rf /tmp/mise-repro-10750-aqua /tmp/mise-repro-10750-aqua-cache
env \
  MISE_DATA_DIR=/tmp/mise-repro-10750-aqua \
  MISE_CACHE_DIR=/tmp/mise-repro-10750-aqua-cache \
  MISE_TRACE=1 \
  ./target/debug/mise install aqua:biomejs/biome@2.5.2
```

On `main`, the install succeeds, but the cached GitHub release metadata is rejected and mise falls back to the GitHub API:

```text
mise-versions endpoint=github_release repo=biomejs/biome tag=@biomejs/biome@2.5.2 outcome=invalid_asset_urls fallback=true
aqua:biomejs/biome@2.5.2 ✓ installed
```

On this PR branch, the same install succeeds without that fallback and uses the versions host release object:

```text
mise-versions endpoint=github_release repo=biomejs/biome tag=@biomejs/biome@2.5.2 outcome=success
got GitHub release biomejs/biome@@biomejs/biome@2.5.2 from mise-versions
aqua:biomejs/biome@2.5.2 ✓ installed
```

So this is not a hard install failure for Biome today; it fixes the cache-path behavior by avoiding the unnecessary GitHub API fallback for Biome-style scoped release tags.

PR #10752, opened 2026-07-03
fix: re-lock stale platform entries during upgrade

zeitlinger:
## Summary
- auto-lock stale tool versions that `mise upgrade --bump` rewrites in existing lockfiles
- snapshot pre-update platform keys from the whole resolved toolset, not only freshly installed tools
- add regression coverage for the bare-entry detection used by auto-lock

## Root cause
`mise upgrade --bump` can touch more lockfile entries than the tools it installs in the current run.

In the repro, only `node` and `npm:renovate` were freshly upgraded, but the command also rewrote other lockfile entries like `ruff`, `biome`, and `typos` to match already-bumped versions in `mise.toml`. `update_lockfiles` replaced those stale entries with bare version/backend records, while `auto_lock_new_versions` only revisited `new_versions`, so the rewritten entries were left without their cross-platform checksums and URLs.

## What changed
After `update_lockfiles`, auto-lock now scans the resolved toolset plus `new_versions`, filters to URL-locking backends, and re-locks any matching lockfile entry that is missing required platform metadata for the pre-update target platform set.

That keeps the existing lockfile platform scope intact while repairing stale entries that were rewritten during the upgrade.

## Validation
- `cargo test tool_version_needs_auto_lock`
- `cargo build`
- repro with patched binary:
  - clone `prometheus/client_java`
  - run `mise up -l`
  - verify rewritten entries like `ruff`, `biome`, `typos`, `editorconfig-checker`, and `zizmor` retain/populate platform data instead of being reduced to bare version-only entries

## Related
- discussion: #10563
- discussion: #10564

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved automatic lockfile updates to ensure tool versions are resolved correctly across all required platforms.
  - Enhanced detection of incomplete lockfile entries (e.g., missing download metadata) and fills them in before applying updates.
  - Reduced unnecessary lockfile processing by skipping upgrades when per-platform lock details are already complete.
- **Tests**
  - Added an end-to-end scenario to verify upgrades properly refresh and remove stale relock-target lock entries.


jdx:
could we have an e2e test?

PR #10773, opened 2026-07-05
fix(core): don't panic when stderr is unwritable

daandemeyer:
When the process that spawned mise (a shim, tool stub, or LSP client)
closes its end of the stderr pipe, every eprintln! panics with
"failed printing to stderr". The panic hook then hit the same panic
in its own eprintln!, and a panic while the hook is running escalates
to a process abort with SIGABRT and a core dump.

Use calm_io::stderrln! (already used by prefix_eprintln!) in the panic
hook, the logger, and VerboseReport so a broken stderr silently drops
log output instead of taking the process down.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CLI stability when standard error is unavailable or unwritable.
  * Error output, panic reporting, and progress/timing messages now safely handle stderr write failures without crashing.
* **Tests**
  * Added an end-to-end test that forces stderr write failures to confirm the CLI still completes successfully.

### DIFF
--- a/e2e/cli/test_broken_stderr
+++ b/e2e/cli/test_broken_stderr
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+# mise must not crash when stderr is unwritable, e.g. when the process that
+# spawned it (a shim, tool stub, or LSP client) closes its end of the pipe.
+# eprintln!-style macros panic on write errors, and a panic that reaches the
+# panic hook while stderr is broken used to escalate to SIGABRT.
+# /dev/full makes every write fail with ENOSPC, exercising the same code path.
+
+if [[ ! -w /dev/full ]]; then
+  # /dev/full is Linux-only
+  exit 0
+fi
+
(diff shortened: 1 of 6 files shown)


jdx:
I think this is directionally right, especially the panic-hook change: if we chain back to color-eyre's default hook, we can still hit panic-prone stderr writes when stderr is broken.

One small refactor I'd consider before merging: we already use `calm_io` in a few places (`src/output.rs`, `src/task/task_results_display.rs`, `src/ui/table.rs`, `src/cli/tasks/ls.rs`). Since this PR adds several more `let _ = calm_io::stderrln!(...)` call sites, it might be worth adding a tiny project-level stderr helper/macro in `src/output.rs` (parallel to `miseprintln!` / `miseprint!`) and using that here. That would make the intentional behavior clearer: write to stderr, but ignore broken-pipe/unwritable-stderr failures.

Something like `mise_eprintln!(...)` or `safe_eprintln!(...)` would keep the policy in one place and make these call sites read as deliberate rather than incidental `calm_io` usage.

PR #10778, opened 2026-07-05
fix(nushell): resolve str upcase deprecation warning

martinhiller:
Resolves a deprecation warning in the Nushell activation script by replacing `str upcase` with a case-insensitive regex match.

## Context
Nushell `0.114.0` deprecated the `str upcase` command in favor of `str uppercase` (and `str downcase` in favor of `str lowercase`). See the [Nushell 0.114.0 release blog](https://www.nushell.sh/blog/2026-07-04-nushell_v0_114_0.html#deprecate-str-upcase-str-downcase-and-add-str-uppercase-str-lowercase).

This PR resolves the resulting deprecation warning:
> `str upcase was deprecated in 0.114.0 and will be removed in a future release.`

### Why this approach?
* **Backwards Compatibility**: Nushell `0.114.0` renamed `str upcase` to `str uppercase`. Simply swapping the command names would break compatibility with older Nushell installations.
* **Static Parser Constraints**: We cannot dynamically check the Nushell version inside the Nushell script (e.g., using `if (version).minor >= 114`) to call the new command. Because Nushell parses commands statically, older versions of Nushell would fail to parse the script at startup due to `str uppercase` being unrecognized, regardless of whether that branch executes.
* **No Parser Warnings**: The regex operator `=~` and `(?i)` case-insensitivity are standard features in Nushell, avoiding the deprecation warning on newer versions while remaining fully functional on older versions.

## Validation
- Ran Rust unit tests to verify the generated script output matches the updated snapshot: `cargo test shell::nushell`
- Manually verified that the case-insensitive regex correctly evaluates to `true` on Nushell `0.114.0` for all casing variants of the PATH variable (`PATH`, `Path`, `path`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved environment activation so `PATH` is recognized correctly regardless of letter casing, making shell setup more reliable across different systems.

### DIFF
--- a/src/shell/nushell.rs
+++ b/src/shell/nushell.rs
@@ -67,7 +67,7 @@ impl Shell for Nushell {
           def --env "update-env" [] {{
             for $var in $in {{
               if $var.op == "set" {{
-                if ($var.name | str upcase) == 'PATH' {{
+                if ($var.name =~ '(?i)^path$') {{
                   $env.PATH = ($var.value | split row (char esep))
                 }} else {{
                   load-env {{($var.name): $var.value}}
--- a/src/shell/snapshots/mise__shell__nushell__tests__hook_init.snap
+++ b/src/shell/snapshots/mise__shell__nushell__tests__hook_init.snap
@@ -9,7 +9,7 @@ def "parse vars" [] {
 def --env "update-env" [] {
   for $var in $in {
     if $var.op == "set" {
-      if ($var.name | str upcase) == 'PATH' {
+      if ($var.name =~ '(?i)^path$') {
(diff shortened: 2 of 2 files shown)


saschabratton:
This fix looks good! thanks @martinhiller ❤️ 

The failing test is the `nightly` job with some upstream rustc compile issue.
It's not related to this PR and shouldn't be blocking anything.

@jdx could this be merged please? nushell warning about it is quite verbose and annoying.

martinhiller:
FYI: The `nightly` check, which failed for every PR from the last days, has straightened itself out by now.

PR #10818, opened 2026-07-07
fix(shim): support docker compose plugin

JamBalaya56562:
## Summary
- route `docker compose ...` through the mise-managed `docker-cli-plugin-docker-compose` binary when invoked via the `docker` shim
- preserve normal `docker ...` shim behavior when no compose plugin is active
- add unit coverage plus an e2e case that verifies no global `~/.docker/cli-plugins/docker-compose` symlink is created

## Why
Docker only discovers CLI plugins from its plugin directories, so installing `docker-cli` and `docker-compose` via mise left `docker compose` unable to find the mise-managed compose plugin. Handling this in the shim avoids global filesystem side effects and works with project-local toolsets.

## Related discussion
- Fixes https://github.com/jdx/mise/discussions/5950

## Validation
- `cargo +stable fmt --check`
- Started `cargo +stable test docker_compose`, but stopped before completion because this environment has limited memory and the command was triggering a large first build. CI should cover compile, clippy, and e2e validation.

## Summary by CodeRabbit

* **New Features**
  * Docker Compose commands now resolve through the app’s shimmed toolchain, so `docker compose` uses the expected installed version.

* **Bug Fixes**
  * Command resolution now correctly detects and routes Docker Compose invocations, including when command-line options precede the `compose` subcommand.
  * Validation no longer relies on the legacy Docker Compose plugin location.

* **Tests**
  * Added an end-to-end check that pins Docker/Compose tool versions, verifies the shimmed `PATH` output for `docker compose version`, and confirms the legacy `docker-compose` plugin is not present.

### DIFF
--- a/e2e/backend/test_aqua_docker_compose_shim
+++ b/e2e/backend/test_aqua_docker_compose_shim
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+cat >mise.toml <<EOF
+[tools]
+docker-cli = "28.3.3"
+docker-compose = "2.39.1"
+EOF
+
+mise install
+mise reshim
+
+assert_contains "PATH=\"$MISE_DATA_DIR/shims:\$PATH\" docker compose version" "Docker Compose version"
+assert_fail "test -e \"$HOME/.docker/cli-plugins/docker-compose\""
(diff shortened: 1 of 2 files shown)


jdx:
this isn't ok

PR #10827, opened 2026-07-07
fix(backend): reinstall rolling versions when outdated

zelch:
Rolling releases like `nightly` maintain the same version string, so checking if the install directory exists is insufficient to determine if an update is needed. Add detection for outdated rolling versions and force reinstallation when necessary.

- Check if a version is a rolling release and outdated before skipping installation
- Trigger uninstall and reinstall for outdated rolling versions
- Update dry-run and post-lock checks to account for rolling version status

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of rolling-version installs by detecting when rolling content is outdated rather than treating it as already up to date.
  * Reinstalls now occur for outdated rolling versions (including when install state changes while waiting), unless explicitly forced.
  * Dry-run output now distinguishes between “already installed” and “would install” based on rolling content freshness.

* **Tests**
  * Added an end-to-end test validating checksum-driven rolling install/reinstall behavior and corresponding dry-run messaging.


jdx:
Thanks for digging into this — the underlying bug is real (`mise outdated` flags a rolling version via the checksum comparison, but `mise upgrade` runs install with `force: false` and `install_version` early-returns "already installed", so the upgrade silently no-ops). A couple of substantive issues with the current approach though:

**As written, `rolling_reinstall` can never be true for any backend in the tree.** It requires both:

1. `is_rolling_channel(version)` — only zig implements this (`"master"`); the trait default returns `false`.
2. `is_rolling_version_outdated(...)` — requires `VersionInfo.rolling` + a checksum, which only the vfox backend populates (and only vfox writes stored checksums at install time).

Zig passes (1) but fails (2) since its version list never sets `rolling`/checksum — and zig doesn't need this anyway, because it resolves `master` to a concrete version (`0.17.0-dev.NNN`) so each nightly lands in its own install dir. Vfox passes (2) but fails (1). Rust's channels set `rolling: true` but implement neither of the other pieces. So the `is_rolling_channel` gate added in the last commit quietly made the whole change a no-op.

**Where the fix lives matters.** Putting this inside `install_version` changes `mise install` from an idempotent "ensure installed" into an auto-updater with a remote-versions fetch on the way. I'd rather keep that behavior in the upgrade path: `OutdatedInfo` already computes `rolling_outdated`, so `mise upgrade` could force-reinstall just those tools — no new network calls in install, no semantic change, and a much smaller diff.

**This needs an e2e test** — which would have immediately caught that the current version doesn't do anything. The behavior to prove: an already-installed rolling version is not skipped as "already installed" when the upstream checksum changes. A deterministic test could use a fake/local vfox plugin fixture rather than a real upstream nightly:

1. install a rolling version such as `tool@nightly` that produces build/content A and writes checksum A
2. change the fixture's remote metadata so the same version string now has checksum B and installs build/content B
3. run `mise upgrade tool`
4. assert the installed tool now reports build/content B
5. run the command again without changing the checksum and assert it does not reinstall

That would also catch the keying edge case here: the stale check compares against `tv.request.version()` while the checksum write-back in vfox uses `tv.version` — those happen to match for vfox rolling versions today, but if they ever diverge this either fails to reinstall or reinstalls every time.

zelch:
@jdx I think everything has been addressed, but please double check me.

I haven't exactly gotten a ton of sleep this week.

PR #10873, opened 2026-07-08
fix(s3): resolve cross-platform lockfile URLs for locked mode

icholy:
mise lock previously wrote no platform entries for s3 tools, so installs with --locked / MISE_LOCKED=1 always failed with "No lockfile URL found" and the suggested fix (running mise lock) could never satisfy it.

Implement resolve_lock_info for the s3 backend, mirroring the http backend: resolve the url for each target platform from the platforms table or base template, render os()/arch() for the target rather than the host, and copy a configured checksum into the lockfile. Platforms without a configured URL fail closed so lock runs report them as skipped. Config parsing now defers os()/arch() rendering of s3 url options, as it already did for http, so cross-platform URLs render correctly.

## Summary by CodeRabbit

* **New Features**
  * `mise lock` now resolves platform-specific S3 tool URLs (including per-OS/per-architecture variants) for the generated lockfile.
  * Locked installs use the resolved, platform-specific S3 URLs from `mise.lock`.
* **Bug Fixes**
  * Improved S3 URL/checksum template rendering so `{{ os() }}` / `{{ arch() }}` resolve at the right time for targeted installs.
  * Targets without a configured S3 URL are skipped cleanly during lock generation.
* **Tests**
  * Added end-to-end coverage for S3 lockfile resolution and locked-mode installation.


icholy:
@jdx the lint job flaked, can you re-run it?

PR #10886, opened 2026-07-09
fix(file): prefer bsdtar for sparse tar extraction fallback (#10674)

JamBalaya56562:
## Summary

Follow-up to #10821 for discussion #10674. That PR added a fallback to the system `tar` when the pure-Rust `tar` crate can't handle a PAX GNU-sparse entry, but it ran the **bare `tar` (whatever is first on `PATH`)**.

GNU tar mishandles some Go/GoReleaser-produced sparse archives (e.g. `github:smol-machines/smolvm` — a 20 GiB-logical `.ext4` in a ~31 MB `.tar.gz`), failing with `tar: Unexpected EOF in archive`, while **bsdtar/libarchive extracts them**. So on macOS with Homebrew GNU tar (`gnubin`) ahead of `/usr/bin/tar`, the fallback picked GNU tar and failed — reported after #10821 shipped in v2026.7.3 ([discussion comment](https://github.com/jdx/mise/discussions/10674#discussioncomment-17584044)); the reporter confirmed `PATH="/usr/bin:$PATH" mise install …` (bsdtar) works.

## Change

`extract_with_system_tar` now builds an ordered, de-duplicated candidate list preferring bsdtar, and tries each until one succeeds:

1. `bsdtar` on `PATH` (libarchive)
2. the platform's built-in bsdtar — macOS `/usr/bin/tar`, Windows `System32\tar.exe`
3. `tar` on `PATH`

Candidates are resolved **absolute** paths, run literally (a relative `PathBuf` would be dotified by duct and never PATH-searched). When none resolve, it falls back to the bare `"tar"` string (OS `PATH` search) — exactly the previous behavior. Each attempt is run with output captured; if all fail, the error lists what was tried and hints that installing **bsdtar/libarchive** enables GNU-sparse (PAX) extraction.

De-dup is by canonical path (`canonicalize_or_self`), so the common macOS case where `/usr/bin/tar` and `which("tar")` are the same binary doesn't run twice.

## Why it's a no-op where it should be

- **macOS default / Windows default:** the preferred candidate already resolves to the same bsdtar as before.
- The pure-Rust `tar` crate remains the primary path; this only affects the GNU-sparse fallback branch.

## Tests / verification

- New host-independent unit test `order_tar_candidates` (ordering, dedup, `None`-skip, empty→bare-`tar` fallback).
- Existing `#[cfg(unix)]` `test_extract_archive_falls_back_for_pax_sparse_tar` unchanged.
- `cargo fmt`/`clippy`/`check` clean.
- Verified locally that the candidate loop **prefers the platform bsdtar over a GNU tar earlier on `PATH`** (debug log shows the bsdtar path chosen), and that the improved error + hint is emitted when no candidate can extract. The reported macOS case is resolved because the fallback now prefers `/usr/bin/tar` (bsdtar), which the reporter confirmed extracts these archives.

> Draft: opening for early review; happy to adjust the candidate order / platform handling.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sparse-archive extraction reliability by trying multiple system tar options in a predictable order.
  * Added better cleanup between retries so failed attempts don’t leave behind partial output.
  * Error messages now include the candidates tried and the last failure details, making issues easier to diagnose.

* **Tests**
  * Added coverage to ensure tar candidate ordering is deterministic and duplicates are ignored.

### DIFF
--- a/src/file.rs
+++ b/src/file.rs
@@ -1252,22 +1252,141 @@ fn path_has_gnu_sparse_artifact(path: &Path) -> bool {
     })
 }
 
+/// Ordered, de-duplicated list of `tar` binaries to try for the sparse-archive
+/// fallback, preferring bsdtar/libarchive. GNU tar mishandles some PAX GNU-sparse
+/// archives — e.g. Go/GoReleaser-produced ones like `github:smol-machines/smolvm`,
+/// where it fails with "Unexpected EOF in archive" — while bsdtar extracts them
+/// (#10674). Windows' System32 `tar` and macOS's `/usr/bin/tar` are bsdtar, so
+/// the bare-`tar` fallback already resolves to bsdtar there.
+///
+/// Only resolved (absolute) paths go in the returned Vec: `cmd!` (duct) dotifies
+/// a *relative* `PathBuf` to `./tar` and never PATH-searches it, so a bare-name
+/// candidate can't be a `PathBuf`. When this list is empty,
+/// `extract_with_system_tar` falls back to the bare `"tar"` string (OS PATH
+/// search), preserving the original behavior.
+fn system_tar_candidates() -> Vec<PathBuf> {
+    let bsdtar = which("bsdtar");
+    let tar = which("tar");
+    order_tar_candidates(bsdtar, platform_system_tar(), tar)
+}
+
+/// The platform's built-in bsdtar/libarchive `tar`, if present. macOS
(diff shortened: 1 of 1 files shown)


jdx:
Closing this in favor of #11028, which replaces the sparse tar fallback with in-process extraction via jdx-tar.

Thanks for working on this and for the contribution.

PR #10914, opened 2026-07-10
perf(aqua): skip public probes for private releases

risu729:
## Summary

- preserve aqua's `private` package field in YAML and compiled registry data
- use the authenticated GitHub API asset URL immediately for declared-private release assets
- retain browser-URL probing and API fallback when `private` is absent or false
- bump the compiled aqua registry cache version from v5 to v6

## Why

Aqua registries can mark GitHub-hosted packages as private. Mise previously discarded that field, so private release downloads always made a browser-URL probe that is known to fail before selecting the authenticated API asset endpoint.

This keeps the fallback behavior for registries with missing or incorrect metadata while avoiding the unnecessary request for correctly declared private packages. The same selection applies to auxiliary release assets such as checksums, signatures, and provenance.

## Validation

- `cargo test -p aqua-registry private`
- `cargo test -p aqua-registry test_rkyv_package_roundtrip_preserves_var_default`
- `cargo test test_private_github_release_uses_api_url_without_probe`
- `cargo test test_github_release_without_api_url_uses_browser_url`

The full `cargo test -p aqua-registry` run passed 107 tests and hit one pre-existing platform-dependent failure in `test_override_envs_all_matches_any_platform` (`.exe` expectation). The repository-wide `mise run format` task could not complete because `ajv` is unavailable on PATH.

## Summary by CodeRabbit

* **New Features**
  * Registry packages can now be marked as private (with a default of public when omitted).
  * Private GitHub-hosted releases use direct authenticated API download URLs for package artifacts and related verification assets.
* **Bug Fixes**
  * Improved GitHub release asset URL resolution for private packages, avoiding reachability probing when an API URL is available.
  * Preserved package privacy settings during serialization/deserialization and version overrides.
  * Updated compiled registry cache version to refresh stored cache data.

### DIFF
--- a/crates/aqua-registry/src/cache.rs
+++ b/crates/aqua-registry/src/cache.rs
@@ -7,7 +7,7 @@ use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::time::{Duration, SystemTime};
 
-const COMPILED_REGISTRY_CACHE_VERSION: &str = "v5";
+const COMPILED_REGISTRY_CACHE_VERSION: &str = "v6";
 
 #[derive(Debug, Clone)]
 pub struct RegistryCache {
--- a/crates/aqua-registry/src/codec.rs
+++ b/crates/aqua-registry/src/codec.rs
@@ -30,6 +30,7 @@ mod tests {
         let mut package = AquaPackage::default();
         package.repo_owner = "owner".into();
         package.repo_name = "repo".into();
+        package.private = true;
         package.vars = vec![AquaVar {
             name: "channel".into(),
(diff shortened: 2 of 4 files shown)


risu729:
Addressed the Greptile concern with a regression test confirming that top-level `private: true` survives version override application.

I intentionally did not merge `private` from the selected `version_overrides` entry: upstream aqua’s `VersionOverride` type and generated registry schema do not define a `private` property. The supported behavior is for package-level privacy to remain inherited while version-specific fields are applied, which the new test now covers.

PR #10923, opened 2026-07-10
feat(cargo): disable cargo-quickinstall by default

risu729:
## Summary

- add `cargo.binstall_quickinstall`, defaulting to `false`
- pass `--disable-strategies quick-install` to external `cargo-binstall` unless users opt in
- document the trust boundary and add settings plus stubbed command coverage

## Motivation

`cargo-quickinstall` binaries are built and hosted by a third party. By default, mise should limit external `cargo-binstall` to crate-author releases or artifacts declared in `package.metadata.binstall`, then allow its normal compile fallback.

Users who accept the cargo-quickinstall trust boundary can restore that strategy with `cargo.binstall_quickinstall = true` or `MISE_CARGO_BINSTALL_QUICKINSTALL=1`.

This does not change mise native binstall behavior, cargo-binstall compile fallback behavior, compile-strategy ownership, or lockfile format. `cargo.binstall = false` remains the way to disable binstall entirely.

## Registry impact

`magika` is the only registry tool whose first backend is `cargo:`. `magika-cli` 1.1.0 has no `package.metadata.binstall`, and cargo-binstall currently probes cargo-quickinstall for it. With the new default, it will compile from source unless crate-author metadata or binaries become available; users can opt back into quickinstall.

## Validation

- `mise run render:schema`
- `mise run lint-fix`
- `cargo test --all-features cargo_binstall_quickinstall_setting`
- `mise run test:e2e e2e/backend/test_cargo_binstall_token`

## Summary by CodeRabbit

* **New Features**
  * Added `cargo.binstall_quickinstall` setting and `MISE_CARGO_BINSTALL_QUICKINSTALL` env var to control whether external `cargo-binstall` may use the quick-install strategy.
  * Quick-install is **disabled by default**; enabling it allows quick-install (distinct from the native `binstall_native` path).
* **Bug Fixes**
  * When the setting is off, mise now disables **only** the quick-install strategy during `cargo-binstall`.
* **Documentation**
  * Clarified behavior in Cargo-related documentation, including how it interacts with `cargo.binstall = false`.
* **Tests**
  * Added e2e coverage for default vs enabled quick-install behavior, plus a unit test validating the new setting.


jdx:
oh I didn't realize they were third-party, I think we should default this to false

risu729:
oh ok, I'll update the PR then

PR #10947, opened 2026-07-11
fix(copr): fix x86_64 builds with add nasm/cmake for Fedora and PREBUILT_NASM for RHEL/EPEL

bestagi:
## Problem

Every x86_64 COPR chroot fails while aarch64 on the same targets succeeds:

| Chroot | Result |
|---|---|
| `fedora-rawhide-x86_64` | ❌ failed |
| `fedora-44-x86_64` | ❌ failed |
| `fedora-43-x86_64` | ❌ failed |
| `epel-10-x86_64` | ❌ failed |
| `fedora-rawhide-aarch64` | ✅ succeeded |
| `fedora-44-aarch64` | ✅ succeeded |
| `fedora-43-aarch64` | ✅ succeeded |
| `epel-10-aarch64` | ✅ succeeded |

**Root cause:** `aws-lc-sys` requires **NASM** at build time to compile x86_64-optimized assembly crypto routines. `nasm` was not declared as a `BuildRequires`, so COPR's mock chroot does not install it and the `aws-lc-sys` build script panics with exit status 101. aarch64 is unaffected because `aws-lc-sys` has no NASM dependency on that architecture.

Failing build: https://copr.fedorainfracloud.org/coprs/jdxcode/mise/build/10703774/

## Fix

**Fedora** (`%{?fedora}`): add `BuildRequires: nasm` and `BuildRequires: cmake`. Both are available in standard Fedora repos on all active releases (43, 44, rawhide).

**RHEL/EPEL** (`%{?rhel}`): export `AWS_LC_SYS_NO_ASM=1` before `cargo build`. On EL9/EL10, `nasm` lives in CRB (CodeReady Builder) which is not enabled in COPR chroots by default. `AWS_LC_SYS_NO_ASM=1` disables the ASM path, falling back to portable C — performance impact is negligible for mise's usage patterns (tool downloads, not throughput-sensitive crypto).

```spec
BuildRequires:  cmake
%if 0%{?fedora}
BuildRequires:  nasm
%endif
```

```spec
%if 0%{?rhel}
export AWS_LC_SYS_NO_ASM=1
%endif
cargo build --profile __BUILD_PROFILE__ --frozen --bin mise
```

## Testing

This fix covers all 4 failing x86_64 chroots with a single conditional approach that requires no extra repo configuration in COPR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RPM build reliability across Fedora and RHEL by adjusting the generated spec’s build requirements.
  * Added an unconditional CMake build dependency to ensure consistent builds.
  * Enabled NASM support only for Fedora x86_64 builds, while RHEL x86_64 builds now use prebuilt NASM objects instead of requiring NASM in the COPR chroot.

### DIFF
--- a/packaging/copr/build-copr.sh
+++ b/packaging/copr/build-copr.sh
@@ -200,7 +200,9 @@ BuildRequires:  cmake
 # Available in standard Fedora repos; on RHEL/EPEL it lives in CRB
 # (not enabled in COPR chroots by default), so we use NO_ASM there.
 %if 0%{?fedora}
+%ifarch x86_64
 BuildRequires:  nasm
+%endif
 %endif
 
 %description
--- a/packaging/copr/build-copr.sh
+++ b/packaging/copr/build-copr.sh
@@ -195,6 +195,13 @@ BuildRequires:  cargo
 BuildRequires:  gcc
 BuildRequires:  git
 BuildRequires:  openssl-devel
+BuildRequires:  cmake
+# nasm provides optimized x86_64 ASM routines for aws-lc-sys.
+# Available in standard Fedora repos; on RHEL/EPEL it lives in CRB
+# (not enabled in COPR chroots by default), so we use NO_ASM there.
+%if 0%{?fedora}
+BuildRequires:  nasm
+%endif
 
 %description
 mise prepares your development environment before each command runs. It installs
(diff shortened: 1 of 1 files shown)


bestagi:
wait prebuilt_nasm only for windows, will fix it later.
or maybe turn on CRB in COPR project setting ? @jdx

bestagi:
oh wait, copr publish process are canceled and its still broke, i will check it again tomorrow  @jdx

PR #10955, opened 2026-07-12
fix(task): respect configured directory when editing tasks

risu729:
## Summary

- choose the first existing directory from the effective `task_config.includes` list when creating a file task
- preserve default include behavior, using the first default entry only when no default directory exists
- return an error instead of silently using a default directory when explicit includes contain no existing directory
- edit existing tasks at their known source without requiring a file-task creation directory

## Behavioral decision

Explicit include paths are not auto-created. An entry may represent a task TOML file or a directory, so mise must not infer the type of a path that does not exist. For example, with `includes = ["tasks"]`:

- if `tasks/` exists, a new file task is created there;
- if `tasks` does not exist, the command returns an error and does not create either `tasks/` or `mise-tasks/`;
- the user should create the intended directory before running `mise tasks edit` or `mise tasks add --file`.

This matches the clarification in [Discussion #10954](https://github.com/jdx/mise/discussions/10954#discussioncomment-17613384). Built-in default includes are known directory paths, so when none exist mise may still create the first effective default entry.

## Root cause

`Task::task_dir` fell back to `mise-tasks` whenever no include expanded to an existing directory. That ignored the replacement semantics of an explicit `task_config.includes` value and could create a task outside the configured include paths.

## Validation

- `mise run lint-fix`
- `mise run test:e2e e2e/tasks/test_task_edit_nested_names`
- `shellcheck -x e2e/tasks/test_task_edit_nested_names`

Addresses https://github.com/jdx/mise/discussions/10954

### DIFF
--- a/e2e/tasks/test_task_edit_nested_names
+++ b/e2e/tasks/test_task_edit_nested_names
@@ -91,3 +91,12 @@ EOF
 TASK_PATH4=$(mise tasks edit configured --path 2>&1)
 assert "printf '%s' \"$TASK_PATH4\"" "$PWD/tasks/configured"
 assert "test -f \"$TASK_PATH4\""
+
+# Explicit includes replace the default directories; do not fall back to mise-tasks.
+cat <<'EOF' >mise.toml
+[task_config]
+includes = ["tasks.toml"]
+EOF
+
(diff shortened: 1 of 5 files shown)


risu729:
@coderabbitai review

risu729:
@coderabbitai review

PR #10962, opened 2026-07-12
fix(prune): don't delete the active version selected by minimum_release_age

isaka1022:
## Problem

`mise prune` can delete the exact version that normal resolution selects under an active `minimum_release_age` policy, reported in [discussion #10944](https://github.com/jdx/mise/discussions/10944).

For a tracked `latest` request, prune resolves the same request through two different paths:

- **Normal (date-aware) resolution** honors the release-age cutoff and picks the older, age-eligible version.
- **Offline resolution during prune** (added in #9406 so prune never hits the network) has only the installed version strings — no release timestamps — so it falls back to the newest *installed* version, which may be a too-new version the age policy excludes.

When those two passes disagree, prune protects the too-new version and marks the age-eligible one — the version the policy actually requires — as prunable, then deletes it.

Reproduced exactly as in the discussion (isolated `HOME`/XDG/mise dirs, dry-run only):

```
resolving jq@latest (before_date=2020-01-01T23:59:59Z)
resolved: aqua:jqlang/jq@1.6            # normal, age-eligible
resolving jq@latest (before_date=2020-01-01T23:59:59Z, offline)
resolved: aqua:jqlang/jq@1.7.1          # offline, too new
jq@1.6 [dryrun] uninstall               # deletes the active version
```

## Fix

This follows the maintainer's preferred **option 1** from the discussion: make offline prune conservative by default. When a floating `latest` request carries an explicit `minimum_release_age` cutoff that offline resolution cannot pin safely, prune now protects **every installed version** of that backend instead of trusting the newest-installed guess, and prints a warning that precise cleanup was skipped.

This mirrors the over-protection already used for `sub-N:latest` in #9406 ("when offline resolution cannot identify the required installed version, over-protection is safer than deleting one that may still be active"). The check is scoped precisely:

- Only fires when the cutoff is **explicit** (a user-set `minimum_release_age`), never the built-in default — so `mise prune` still prunes stale versions in the common no-cutoff case.
- Only fires for backends that report release timestamps (where the two passes can actually diverge). Backends without dates resolve identically online and offline, so they are pruned normally.
- Leaves `resolve_version` untouched, so `hook-env` / `exec` / `completion` (which also resolve `latest` offline but want a concrete installed version to use) are unaffected.

The change is localized to `collect_needed_versions`, the helper both `mise prune` and stub tracking use to compute which installed versions to protect.

The opt-in `mise prune --remote` path the maintainer also described (network-aware precise cleanup) is a larger, separable feature; I've kept this PR to the safety fix that stops the data loss. Happy to follow up with `--remote` here or in a separate PR — whichever you prefer.

## How was this tested

Built before/after binaries and reproduced against the real `aqua:jqlang/jq` backend with a fixed calendar cutoff (`minimum_release_age = "2020-01-01"`; jq 1.6 is age-eligible, jq 1.7.1 is too new), in a fully isolated environment with dry-run only:

| | offline resolves | prune result |
| --- | --- | --- |
| before | `jq@1.7.1` | **`jq@1.6 [dryrun] uninstall`** — active version deleted, `ls --prunable` lists `jq 1.6` |
| after | `jq@1.7.1` | keeps both; warns; `ls --prunable` empty |

Regression check (no explicit cutoff): `mise prune` still deletes the stale version, confirming the fix does not make prune conservative in the normal case.

Added `e2e/cli/test_prune_minimum_release_age_slow` covering the jq scenario. It fails on `main` (deletes `jq@1.6`) and passes with this change. Existing `e2e/cli/test_prune`, the toolset unit tests, `cargo clippy`, and `cargo fmt --check` all pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `mise prune` safety when an explicit minimum release age is configured.
  * Prevents eligible and newly released versions from being removed when a floating `latest` request could resolve ambiguously, including in offline scenarios.
  * Emits a warning when cleanup is skipped to preserve installed versions under these conditions.
* **Tests**
  * Added an end-to-end regression test covering pruning with multiple installed versions and release-age restrictions.


isaka1022:
#10973 landed after this PR was opened and closes the discussion in #10944 from the other direction — making normal online resolution treat installed versions as eligible, rather than making offline prune conservative. I re-ran the repro on current `main` (126e7755c, which includes e39201389) to check whether this PR is still needed:

```
mise jq@1.6          uninstall
mise jq@1.6          remove ~/.local/share/mise/installs/jq/1.6
ERROR: expected '1.6' to be in 'jq  1.7.1'
```

With `MISE_MINIMUM_RELEASE_AGE=2020-01-01` and `jq = "latest"`, normal resolution installs jq@1.6 (the age-eligible latest). After also installing jq@1.7.1, `mise prune --tools` still removes 1.6 — the version the active policy requires. So the prune path isn't covered by #10973.

I also addressed the automated review comments (10b21f98d):

- gated `has_explicit_release_age_cutoff` on backends that report release timestamps, so the over-protection no longer fires for e.g. `asdf:*` with an explicit setting — this is what the PR description claimed but the code didn't enforce
- documented that a caller-provided cutoff (`BeforeDateSource::Provided`) isn't detected, since `prune` has no `--minimum-release-age` flag today

That new gate has no regression test of its own: the existing e2e covers jq, which resolves through `aqua:` and so passes the gate either way. Covering the negative case needs an e2e against a timestamp-less backend, which is slower and less reliable in CI — happy to add one if you'd rather have it locked down.

Happy to close this if you'd prefer to fix the prune path differently.

jdx:
Thanks for the thorough investigation and writeup here — the two-pass divergence you diagnosed was real.

The catch is timing: this overlaps with #10973, which merged 2026-07-13 and addresses the same discussion (#10944) at the resolution layer. #10973 makes `minimum_release_age` a **remote-discovery gate, not an installed-version gate** — once a version is installed it stays eligible, and only `mise lock` date-filters installed versions. That removes the divergence this PR works around, rather than protecting against it in prune.

I re-ran your exact reproduction on current `main` (which has #10973 but not this PR), jq 1.6 + 1.7.1 installed under `minimum_release_age = 2020-01-01`:

```
mise which jq        → .../installs/jq/latest/jq  (1.7.1)
mise exec -- jq      → jq-1.7.1
mise ls --current jq → jq 1.7.1  ... latest
mise prune --tools   → jq@1.6 uninstall   (1.7.1 kept)
```

Normal resolution now selects **1.7.1**, so `which`/`exec`/`prune` all agree it's the active version — the offline and online passes no longer disagree. That means 1.6 is genuinely unreferenced and pruning it is the intended behavior under #10973; the regression test here (asserting 1.6 survives) would actually lock in the pre-#10973 semantics.

So I'm going to close this as superseded by #10973 — not because the analysis was wrong, but because the underlying bug is already resolved upstream. Really appreciate the careful work regardless.

PR #10992, opened 2026-07-14
fix(bootstrap): support cask array position helpers

casparbreloh:
## Summary

- add Homebrew-compatible `Array#second`, `#third`, `#fourth`, and `#fifth` helpers to the cask lifecycle shim
- cover CSV cask versions with OrbStack's exact `version.csv.first` / `version.csv.second` URL pattern

## Problem

OrbStack's cask evaluates:

```ruby
version.csv.second
```

The mise cask shim returns a normal Ruby array from `Version#csv`, but does not provide the positional helpers from Homebrew's `extend/array.rb`. The lifecycle hook therefore fails before `postflight` runs:

```text
undefined method `second` for an instance of Array (NoMethodError)
```

Homebrew defines the four helpers added here with the same index semantics.

## Validation

- `cargo test cask_shim_supports_csv_version_array_helpers`
- `cargo test cask_shim_supports_language_and_system_conditionals`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Homebrew cask version handling when versions are provided in CSV-style formats.
  * Ensured cask shims correctly resolve individual version components.

* **Tests**
  * Added coverage for CSV-formatted version arrays.
  * Updated existing checks to validate version propagation through the cask shim.

### DIFF
--- a/src/system/packages/brew/cask.rs
+++ b/src/system/packages/brew/cask.rs
@@ -1412,13 +1412,14 @@ mod tests {
         shim: &Path,
         cask: &Path,
         staged_path: &Path,
+        version: &str,
     ) -> std::io::Result<std::process::Output> {
         std::process::Command::new(ruby)
             .arg(shim)
             .env("LANG", "zz_ZZ.UTF-8")
             .env("MISE_BREW_CASK_FILE", cask)
             .env("MISE_BREW_CASK_TOKEN", "example")
-            .env("MISE_BREW_CASK_VERSION", "1.0.0")
+            .env("MISE_BREW_CASK_VERSION", version)
             .env("MISE_BREW_CASK_STAGED_PATH", staged_path)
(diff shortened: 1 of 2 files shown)


casparbreloh:
The only failing check is the `benchmark` job. Could a maintainer rerun it?

The failure is isolated to `mise ls`: 26.4 +/- 3.9 ms for this branch versus 20.0 +/- 1.3 ms for stable, with a relative result of 1.32 +/- 0.21. The lower uncertainty edge only narrowly exceeds the 1.10 threshold (1.11). The other benchmarks are equal or faster. This patch only extends the Ruby cask shim and its unit-test helper, neither of which is executed by `mise ls`, so this looks like benchmark noise rather than a related regression. Fork contributors cannot rerun the job.

PR #11038, opened 2026-07-16
chore(ci): tighten required CI guards

risu729:
## Summary

- make `test-ci` and `registry-ci` run only when their required-check invariant is violated
- reject `failure`, `cancelled`, and unexpected `skipped` results from blocking test jobs
- strictly preserve each documented conditional skip in the registry workflow
- remove the intentionally non-blocking `nightly` job from `test-ci.needs` without changing its execution
- keep the existing required check names

## Why this is better

The aggregate jobs currently allocate a runner on every successful workflow only to inspect `needs`. The new job-level conditions run their single failing step only when an invariant is violated.

`test-ci` has no conditionally skipped blocking jobs, so `failure`, `cancelled`, and `skipped` are all invalid. `registry-ci` has intentional skips, so its condition checks the expected result for each event/output path: `list-changed-tools`, `validate-new-tools`, and `test-tool` may be skipped only when their own workflow conditions say they should be. This also catches an unexpected skip instead of allowing GitHub to treat it as a successful required check.

The leading `!cancelled()` respects whole-workflow cancellation from `cancel-in-progress`. Direct `needs.*.result` checks still catch a dependency that independently fails, times out, or is cancelled.

`failure() && cancelled()` cannot express this because it requires both status functions to be true. Even `failure() || cancelled()` did not run after a reusable dependency timed out in the live test, while checking its direct `needs` result did.

## Live test

The workflow and runs are in [risu729/gha-guard-semantics](https://github.com/risu729/gha-guard-semantics). It covers success, failure, intentional skip, dependency timeout/cancellation, and whole-workflow cancellation. In the dependency-timeout run, the dependency concluded `cancelled`, the status-function variants were skipped, and the direct-result guard ran and failed: [run](https://github.com/risu729/gha-guard-semantics/actions/runs/29473609173).

## Nightly

`nightly` remains non-blocking and keeps `continue-on-error: true` so ordinary nightly failures do not fail the workflow. Removing it from `test-ci.needs` also prevents its timeout (`cancelled`) from failing the required aggregate. The job itself still runs normally.

## Validation

- `mise run lint-fix`
- `mise run lint`
- exhaustive comparison of the strict `registry-ci` predicate against 9,216 registry-changing result combinations (zero mismatches)

## Summary by CodeRabbit

* **Chores**
  * Improved CI workflow handling for failed, cancelled, and skipped validation steps.
  * Simplified automated checks to provide more consistent pass/fail outcomes.
  * Updated test workflow dependencies and streamlined status evaluation.


jdx:
I’m concerned this weakens the required-check invariant by treating every upstream `skipped` result as acceptable.

GitHub considers a conditionally skipped required job successful for merge purposes. With this predicate, if a blocking job such as `lint`, `e2e`, `windows-e2e`, or `test-tool` is unexpectedly skipped:

1. there is no `failure` or `cancelled` result,
2. the aggregate guard is itself skipped, and
3. the required `test-ci`/`registry-ci` check is therefore satisfied.

The existing guards deliberately reject skipped blocking jobs while allowing only specific documented registry skips. This change replaces that distinction with blanket acceptance of skips.

The failure-only optimization could still work for `test-ci` by treating `skipped` as an error too:

```yaml
if: ${{ !cancelled() && (
  contains(needs.*.result, 'failure') ||
  contains(needs.*.result, 'cancelled') ||
  contains(needs.*.result, 'skipped')
) }}
```

For `registry-ci`, the condition would need to preserve the exact legitimate skip cases (no registry changes, deletion-only `test-tool`, etc.); otherwise I think the existing explicit guard is safer.

Removing the intentionally non-blocking `nightly` job from `needs` seems fine. My concern is specifically that unexpected skips in blocking jobs can now pass the required aggregate.

GitHub documents this behavior here: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks

risu729:
Thanks for the review! I assumed the correct conditions, but I agree that we should check the misconfigured skips. I fixed them.

jdx:
Thanks for updating the test guard. Could you also strengthen `registry-ci` rather than preserving the old guard’s permissive cases?

In particular, the aggregate should enforce the upstream jobs’ actual intended conditions:

- For a registry-changing PR, require `list-changed-tools` to conclude `success`; allow `skipped` only for the non-PR push/workflow-dispatch path where that job is intentionally skipped.
- When `validate-new-tools` is expected to run (PR author is not `jdx` and `new_tools` is non-empty), require `success`, so an unexpected `skipped` result cannot pass.
- Reject `validate-new-tools == 'cancelled'` as well as `failure`.
- When `skip-test-tool == 'true'`, require `test-tool` to be `skipped`; otherwise require `success`.

The revised predicate is equivalent to the old shell guard, but that guard has an inherited hole: it blanket-allows `list-changed-tools == 'skipped'` and only rejects `validate-new-tools == 'failure'`. Since this PR is redesigning the required-check policy, I think it’s the right place to make those intentional skip cases explicit and close the hole.

risu729:
I think it's better, but I applied it, and it's now really complex. If this is acceptable, I guess it's ready now.

PR #11057, opened 2026-07-16
feat(task): skip dependencies with empty templated names

risu729:
## Summary

- skip `depends`, `depends_post`, and `wait_for` entries when a templated task name renders empty or whitespace-only
- apply the behavior during both initial rendering and deferred `usage.*` rendering
- preserve the existing error for literal empty task names

## Root cause

`TaskDep::render` rendered dependency task names in place, but callers always retained the dependency. A conditional name that rendered to an empty string therefore reached task matching and failed with `task not found:`.

The renderer now reports whether the dependency should be retained, and the shared list helper filters only names that contained template syntax and rendered empty. Literal empty names remain in the list and continue through the existing validation path.

## Real-world example

Biwa previously needed a hidden no-op task to make schema generation conditional in CI:

```toml
["util:noop"]
run = ":"
hide = true

["schema:tombi"]
depends = [
  "{% if env.CI is defined %}util:noop{% else %}render:schema{% endif %}",
]
```

With this change, the dependency itself can be omitted:

```toml
["schema:tombi"]
depends = [
  "{% if env.CI is defined %}{% else %}render:schema{% endif %}",
]
```

When `CI` is defined, the rendered task name is empty and skipped. Outside CI, `render:schema` still runs first. [risu729/biwa#868](https://github.com/risu729/biwa/pull/868) removes the workaround using this behavior.

## User impact

Tasks can use Tera conditions to make complete dependency entries optional, matching the conditional structured-argument behavior added in #11055.

## Validation

- `mise run format`
- `cargo test task_dep::tests`
- `mise run test:e2e test_task_dep_args`

## Summary by CodeRabbit

* **Bug Fixes**
  * Dependency entries in `depends`, `depends_post`, and `wait_for` that resolve to an empty task name from templates are now omitted.
  * Usage-conditional dependency templates are now evaluated according to the selected mode.
  * A literal empty dependency name is still treated as invalid and fails with a clear “task not found” message.
* **Tests**
  * Added end-to-end coverage for optional/usage-gated empty templated dependencies and `wait_for` execution order.
  * Expanded unit tests for templated `TaskDep` rendering, including empty/whitespace-only behavior.

### DIFF
--- a/e2e/tasks/test_task_dep_args
+++ b/e2e/tasks/test_task_dep_args
@@ -46,6 +46,61 @@ EOF
 assert_contains "mise run lint" "fix=false"
 assert_contains "mise run lint --fix" "fix=true"
 
+# Skip dependency entries whose task name template renders empty, both during
+# initial rendering and deferred usage rendering.
+cat <<'EOF' >mise.toml
+[tasks.initial-dependency]
+run = 'echo "initial dependency ran"'
+
+[tasks.usage-dependency]
+run = 'echo "usage dependency ran"'
+
+[tasks.usage-post-dependency]
+run = 'echo "usage post-dependency ran"'
+
+[tasks.usage-wait-task]
+run = 'sleep 0.3; echo wait >> wait-order.txt; echo "usage wait task ran"'
+
+[tasks.optional]
+usage = 'flag "--enabled" default=#false'
+depends = [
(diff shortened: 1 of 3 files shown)


risu729:
@coderabbitai full review

risu729:
@coderabbitai full review

PR #11103, opened 2026-07-17
fix(task): preserve inline overlay precedence

risu729:
## Summary

- apply only the highest-precedence inline task definition to a same-named script task
- keep config-file precedence first-wins instead of layering competing TOML definitions through a script base
- preserve each selected script's template and task-directory context across local and monorepo loaders
- keep later `conf.d` fragments at their documented higher precedence

### Lower-precedence metadata leaked into script tasks

When a script task and multiple `[tasks.<name>]` definitions shared a name, the first inline definition was merged into the script but the result remained file-backed. Every lower-precedence inline definition therefore also appeared eligible for overlay. Lower scalar values could replace higher ones, while additive fields such as aliases and environment directives accumulated from configs that should have lost.

For example, given a script at `.mise/tasks/build` and these competing definitions:

```toml
# .config/mise.toml (lower precedence)
[tasks.build]
description = "lower"
alias = "low"
env = { LOWER = "set" }
```

```toml
# .mise/config.toml (higher precedence)
[tasks.build]
description = "higher"
alias = "high"
env = { HIGHER = "set" }
```

```bash
#!/usr/bin/env bash
# .mise/tasks/build
echo "HIGHER=${HIGHER:-unset}"
echo "LOWER=${LOWER:-unset}"
```

Before this fix, the task could end up with the lower description, both aliases, and both environment values:

```text
$ mise tasks
build  lower
$ mise run high
HIGHER=set
LOWER=set
$ mise run low
HIGHER=set
LOWER=set
```

After this fix, only the higher inline definition decorates the script:

```text
$ mise tasks
build  higher
$ mise run high
HIGHER=set
LOWER=unset
$ mise run low
no task low found
```

The same failure mode affected `conf.d`: after selecting a later override fragment, an earlier base fragment could still leak metadata into the file-backed result.

### Loader order discarded the winning config context

Some loaders merged one config at a time. A lower config could rediscover a script and cause its template variables or `task_config.dir` to replace the context selected from a higher config. Out-of-hierarchy monorepo roots exposed the same problem because their configs were finalized separately instead of as one precedence-ordered batch.

For example, a monorepo can define the same script task from two config locations:

```toml
# projects/app/.config/mise.toml (lower precedence)
[vars]
task_context = "lower"

[task_config]
dir = "/repo/projects/app/low-dir"

[tasks.build]
env = { TASK_CONTEXT = "{{vars.task_context}}", LOWER = "set" }
```

```toml
# projects/app/.mise/config.toml (higher precedence)
[vars]
task_context = "higher"

[task_config]
dir = "/repo/projects/app/high-dir"

[tasks.build]
env = { TASK_CONTEXT = "{{vars.task_context}}", HIGHER = "set" }
```

```bash
#!/usr/bin/env bash
# projects/app/.mise/tasks/build
echo "TASK_CONTEXT=$TASK_CONTEXT"
echo "TASK_DIR=$PWD"
```

Before this fix, processing the lower config separately could replace the selected script's context:

```text
$ mise run //projects/app:build
TASK_CONTEXT=lower
TASK_DIR=/repo/projects/app/low-dir
```

After this fix, the configs are merged as one precedence-ordered batch and the script retains the higher context:

```text
$ mise run //projects/app:build
TASK_CONTEXT=higher
TASK_DIR=/repo/projects/app/high-dir
```

## Fix

- collect raw file and inline task sources before final merging
- select only the first, highest-precedence inline definition for each task name
- overlay that definition once onto the selected script task
- preserve each config's own root, variables, and task directory while loading
- batch all configs for a discovered monorepo root before merging

## Impact

Local, `conf.d`, and monorepo task definitions now follow the same first-wins precedence model. Script tasks retain the command from the selected script and metadata only from the winning inline definition.

## Tests

- `mise run test:e2e e2e/tasks/test_task_config_precedence`
- `mise run test:e2e e2e/tasks/test_task_monorepo_config_precedence`
- `mise run lint-fix`
- `git diff --check`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved task precedence handling across configuration files, fragments, inline definitions, and file-backed tasks.
  - Ensured higher-precedence settings correctly override lower-precedence descriptions, aliases, directories, and environment variables.
  - Improved task discovery in symlinked paths and monorepo subdirectories.
  - Prevented duplicate loading of global tasks and preserved correct task naming and ordering.

### DIFF
--- a/e2e/tasks/test_task_config_precedence
+++ b/e2e/tasks/test_task_config_precedence
@@ -1,7 +1,5 @@
 #!/usr/bin/env bash
 
-export MISE_EXPERIMENTAL=1
-
 mkdir -p .config/mise/conf.d .mise/tasks
 
 cat <<'EOF' >.config/mise.toml
@@ -11,27 +9,23 @@ run = "echo .config/a"
 
 [tasks.b]
 description = ".config/b"
+alias = "low"
 env = { LOWER_PRECEDENCE = "applied" }
 EOF
 
 cat <<'EOF' >.mise/config.toml
-monorepo_root = true
-
 [tasks.a]
 description = ".mise/a"
 run = "echo .mise/a"
 
 [tasks.b]
 description = ".mise/b"
+alias = "high"
 env = { HIGHER_PRECEDENCE = "applied" }
-
-[monorepo]
-config_roots = ["projects/*"]
 EOF
 
 cat <<'EOF' >.mise/tasks/b
 #!/usr/bin/env bash
-echo file/b
 echo "HIGHER_PRECEDENCE=${HIGHER_PRECEDENCE:-<unset>}"
 echo "LOWER_PRECEDENCE=${LOWER_PRECEDENCE:-<unset>}"
 EOF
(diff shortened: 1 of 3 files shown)


risu729:
@coderabbitai resume

PR #11107, opened 2026-07-18
fix(bootstrap): support auto-updating cask metadata

casparbreloh:
## Summary

- accept Homebrew casks that declare `auto_updates`
- treat it as metadata in the bootstrap cask shim
- cover the current OrbStack cask shape in the existing regression test

## Test

- `cargo test cask_shim_supports_csv_version_array_helpers`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Homebrew cask compatibility by recognizing the `auto_updates` setting in cask definitions.
  * Casks that use this setting can now be processed without triggering unsupported lifecycle errors.
  * Existing cask version handling remains unchanged.

### DIFF
--- a/src/system/packages/brew/cask.rs
+++ b/src/system/packages/brew/cask.rs
@@ -1516,6 +1516,7 @@ end
             r#"cask "example" do
   version "2.2.1,20628"
   url "https://example.com/OrbStack_v#{version.csv.first}_#{version.csv.second}.dmg"
+  auto_updates true
   preflight do
     File.write staged_path/"result", version.csv.second
   end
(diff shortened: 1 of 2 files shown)


davidpoblador:
Two more affected casks with slightly different failure points, in case they're useful for test coverage: `docker-desktop` fails at **postflight** and `vlc` at **preflight**, both with `Error: cask uses `auto_updates`, which mise's cask shim does not support` (mise 2026.7.7, macOS arm64). With docker-desktop the app bundle had already been swapped when the postflight failed, so the install was left half-applied.

*Drafted with AI assistance; reviewed by a human before posting.*

PR #11132, opened 2026-07-20
feat(oci): built-in registry push client, drop skopeo/crane dependency

jdx:
## Summary

Removes the external-tool dependency from `mise oci`: pushing now uses a built-in OCI Distribution Spec client, and `mise oci run --engine docker` streams the image into `docker load` instead of requiring skopeo. Related context: discussion #11121 (reusing unchanged layers from the previously pushed image) — the native client gets the upload-side half of that for free via per-blob HEAD checks.

### `mise oci push`
- **Native push client** ([src/oci/registry.rs](../blob/claude/mise-discussion-11121-3e9391/src/oci/registry.rs)): HEAD existence check per blob → only new blobs are uploaded (repeat pushes of a mostly-unchanged toolset transfer almost nothing), POST/PUT monolithic upload streamed from disk, manifest PUT under tag or digest.
- **Credential resolution** ([src/oci/auth.rs](../blob/claude/mise-discussion-11121-3e9391/src/oci/auth.rs)) mirrors docker/podman: `$REGISTRY_AUTH_FILE` → `$XDG_RUNTIME_DIR/containers/auth.json` → `~/.config/containers/auth.json` → `~/.docker/config.json`, with inline `auths`, `credsStore`/`credHelpers` credential helpers, and Docker Hub identity tokens (OAuth2 refresh flow). `docker login` / `podman login` is all the setup needed.
- **Challenge-based auth**: probes `/v2/` and answers the `WWW-Authenticate` challenge (Bearer token fetch or Basic) instead of the previous hardcoded realm list — as a side effect, `--from` base image pulls now work against any registry and support private images when logged in.
- Loopback registries (`localhost:5000`) use plain HTTP, matching docker's insecure-by-default convention.
- `--tool` is removed. Escape hatch stays structural: `mise oci build -o ./img && skopeo copy oci:./img docker://…`.

### `mise oci run --engine docker`
- New [src/oci/docker_archive.rs](../blob/claude/mise-discussion-11121-3e9391/src/oci/docker_archive.rs) converts the OCI layout to a docker-archive on the fly (layers decompressed via magic-byte sniffing — gzip/zstd/plain — to match `rootfs.diff_ids`) and streams it into `docker load` stdin. No skopeo needed; podman path unchanged.

## Verification

Ran live against docker 29.5.1 with **no skopeo/crane installed**:

- `mise oci push localhost:5000/test/devenv:v1` against `registry:2` — full push, then idempotent re-push with `SOURCE_DATE_EPOCH` reporting `0 blob(s) uploaded, 4 already present`; served manifest digest verified byte-identical via the registry API.
- Same against an htpasswd-protected `registry:2`: anonymous push fails with an actionable `docker login` hint; with `REGISTRY_AUTH_FILE` the Basic-auth push succeeds.
- `mise oci run --engine docker --no-mise -- jq --version` → `jq-1.8.1` (docker-load path).
- Base image pulls from docker.io exercised the new challenge-based token flow.

Tests: 1770 unit tests pass (new coverage for auth-file parsing, challenge parsing, loopback detection, docker-archive structure); `test_oci_run_push_errors_slow` (updated: `--tool` cases replaced with an unreachable-registry native-push case) and `test_oci_build_slow` pass. Docs regenerated with `mise run render`.

---

> [!NOTE]
> **Medium Risk**
> Adds network registry push/pull/auth and credential-helper execution; behavior changes for oci push/run and base-image pulls, though scope is limited to experimental OCI commands.
> 
> **Overview**
> **`mise oci push`** no longer shells out to skopeo or crane. It uploads through an in-tree OCI Distribution v2 client (`registry::push_image`), skipping blobs the registry already has (HEAD per digest) and reporting uploaded vs skipped counts. Registry auth is resolved like docker/podman (`src/oci/auth.rs`: auth files, inline `auths`, credential helpers, Docker Hub identity tokens). The **`--tool`** flag is removed.
> 
> **`mise oci run --engine docker`** loads the layout by streaming a docker-archive into **`docker load`** (`docker_archive::load_into_docker`) instead of **`skopeo copy`**. Only **podman** or **docker** is required (no docker+skopeo).
> 
> **`src/oci/registry.rs`** pull path is reworked: challenge-driven **`AuthSession`** replaces hardcoded anonymous token realms, so **`--from`** can use logged-in credentials for private base images. Loopback registries use HTTP.
> 
> **Hardening:** `read_blob` validates digests before path use; push streams blobs with **`tokio-util`**. Docs and e2e updated (unreachable-registry push test replaces `--tool` cases).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7bcf8ef5adeb8d62336d43cd8e763d0a49abdd45. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * `mise oci push` now uses mise’s built-in registry client (no external push tools) and uploads only missing blobs on repeat pushes.
  * Registry authentication is resolved automatically from Docker/Podman auth locations and credential helpers.
  * `mise oci run` loads the built OCI image into Docker via `docker load` (Podman loads natively); engine requirement is now `docker` or `podman`.
* **Bug Fixes**
  * Improved registry connection/auth error messages, including clearer failures for unreachable registries.
* **Documentation**
  * Updated manuals and CLI docs for `oci push`/`oci run`, including revised authentication guidance and removal of the prior forced push tool option.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1670,7 +1670,7 @@ version = "3.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "faf9468729b8cbcea668e36183cb69d317348c2e08e994829fb56ebfdfbaac34"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -2368,7 +2368,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -2588,7 +2588,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -4982,7 +4982,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "160f2eade097f30263b548aae5deb12ad349c909baa710fa24b92c9090b2e006"
 dependencies = [
  "scopeguard",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -5002,7 +5002,7 @@ version = "1.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7a1886916523694cd6ea3d175f03a1e5010699a2a4cc13696d83d7bea1d80638"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -5515,6 +5515,7 @@ dependencies = [
  "test-log",
  "thiserror 2.0.19",
  "tokio",
+ "tokio-util",
  "toml 1.1.3+spec-1.1.0",
  "toml_edit 0.25.13+spec-1.1.0",
  "ubi",
@@ -5762,7 +5763,7 @@ version = "0.50.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7957b9740744892f114936ab4a57b3f487491bbeafaf8083688b16841a4240e5"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -5972,7 +5973,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7d8fae84b431384b68627d0f9b3b1245fcf9f46f6c0e3dc902e9dce64edd1967"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -6609,7 +6610,7 @@ dependencies = [
  "once_cell",
  "socket2",
  "tracing",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -7631,7 +7632,7 @@ dependencies = [
  "errno 0.3.14",
  "libc",
  "linux-raw-sys 0.12.1",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -7690,7 +7691,7 @@ dependencies = [
  "security-framework",
  "security-framework-sys",
  "webpki-root-certs",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -8907,10 +8908,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "32497e9a4c7b38532efcdebeef879707aa9f794296a4f0244f6f69e9bc8574bd"
 dependencies = [
  "fastrand",
- "getrandom 0.4.3",
+ "getrandom 0.3.4",
  "once_cell",
  "rustix 1.1.4",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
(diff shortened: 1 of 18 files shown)


jdx:
Addressed both Cursor Bugbot findings in 2137a67:

**1. (Medium) 200 on `/v2/` skips auth** — auth is now challenge-driven on the actual operation rather than assumed from the `/v2/` probe. A new `AuthSession` keeps the probe as an upfront optimization but re-negotiates from any request's `WWW-Authenticate` when it returns 401 and retries once. This covers both push (HEAD/POST/PUT) and pull (manifest fetches), so a registry that answers 200 on `/v2/` while guarding manifest/blob/upload requests now works.

**2. (Low) HEAD failure aborts push** — `blob_exists` now treats only 200 as "present"; any other status (404, or a 405 from a proxy that doesn't implement blob HEAD) is treated as "not present" so the upload proceeds instead of aborting.

Verified against a local registry that returns 200 on `GET /v2/` but 401s (Basic challenge) on operations: the push succeeds by answering the per-operation challenge, and an idempotent re-push reports `0 uploaded / 4 present`, confirming the HEAD dedup path carries the negotiated credentials. The anonymous and probe-401 Basic-auth paths still pass.

jdx:
Addressed the second round of review feedback (CodeRabbit, Greptile, Cursor) in 1d11e74:

- **Auth resolution no longer aborts on a single bad source** — a `credHelpers` helper that exits non-zero because it has no entry for the host (docker's "credentials not found" convention) and an unreadable auth file are now treated as misses, falling through to inline `auths`, `credsStore`, and later files (`~/.docker/config.json`) instead of failing the push.
- **Token refresh on pull** — config/layer blob downloads now go through the same challenge-driven `AuthSession::send` path as push and manifest fetches, so a bearer token that expires partway through a large multi-layer pull is refreshed on a 401 rather than failing. (Push already re-negotiated on 401 after the first round.)
- **`WWW-Authenticate` parsing** — parameters are now parsed honoring quoted values instead of splitting on bare commas, so a realm URL with a query string or an echoed `scope="…:pull,push"` isn't truncated at an interior comma.
- **`upload_blob` file-open error** — the blob file is opened up front so an unreadable/missing file returns a clear I/O error instead of surfacing as a confusing empty-body PUT / registry digest error.
- Dropped the now-unused `http::get_bytes_with_headers` (its only caller moved to the challenge-aware path).

New unit tests cover the quoted-comma and unquoted/spaced challenge cases. Re-verified live after the refactor: anonymous push, the 200-on-`/v2/`-but-401-on-operations scenario (challenge-driven auth plus HEAD-dedup re-push), and a cold Docker Hub base-image pull exercising a real bearer challenge. 1776 unit tests pass; lint clean.

jdx:
Addressed the remaining review finding (Greptile P2 on `docker_archive.rs`) in a362d35:

`load_into_docker` previously wrote the entire docker-archive to `docker load`'s stdin *before* draining stdout+stderr via `wait_with_output`. On a large image, a fatal mid-load daemon error (disk full, overlay failure) could fill docker's stderr pipe buffer (~64 KB) and stop it reading stdin, while mise stayed blocked writing — deadlocking `mise oci run` indefinitely. The archive is now written on a separate thread so the parent drains stdout+stderr concurrently. A writer error other than the expected broken pipe (e.g. a layer that fails to decompress) is appended to docker's failure message so a truncated-archive error can't mask the real cause.

Verified `mise oci run --engine docker -- jq --version` still loads and runs. 1776 unit tests pass; lint clean.

The other bot comments on this round were re-flags of items already fixed in earlier commits (the `WWW-Authenticate` comma parsing and the `upload_blob` file-open validation, both from 1d11e74).

jdx:
Addressed the latest review finding (Cursor) in cf63ac8:

A corrupt matching inline `auths` entry (bad base64/UTF-8/shape) made `credential_from_file` return an error, which `resolve_credential` treated as "skip this whole file" — so a valid `credsStore` in the same file (or a usable entry in a later file) was never consulted. The malformed entry is now warned about and skipped, falling through to `credsStore` and subsequent files. New unit test covers it.

jdx:
Addressed the latest review finding (Cursor) in 7bcf8ef:

The `--image-dir` push and `docker load` paths read blobs using digests from a layout's `index.json` / manifest without the `validate_sha256_digest` guard that `pull_base_image` applies, so a crafted layout could use `sha256:../../../etc/passwd` to escape the blobs directory via `read_blob` / `blob_path`.

`ImageLayout::read_blob` now validates the digest before building the path (defense in depth for all callers), and the docker-load path validates layer digests up front since it reads them through `blob_path` rather than `read_blob`. The push upload loop already validated each blob descriptor. Added a `read_blob_rejects_traversal_digest` test; verified a normal push and `oci run --engine docker` still work with validation in place.

PR #11141, opened 2026-07-20
feat(oci): chunked uploads, retries, cross-repo mounts, progress bars

jdx:
## Summary

Follow-up to #11132 (item 3 of the follow-up list: push-protocol robustness), **stacked on #11140** — the base will retarget when that merges; only the last commit is new here.

- **Chunked blob upload**: blobs over 64 MiB transfer as `PATCH` chunks + a finalizing `PUT` instead of one monolithic `PUT` — keeps request bodies under the limits some registries/CDNs impose (e.g. 100 MB behind Cloudflare) and bounds the cost of a mid-upload failure.
- **Transient-failure retries**: blob uploads and pull-side downloads retry connect/timeout/body errors and 5xx/408/429 with mise's standard backoff (`http_retries`) — closing the gap noted on #11132 where OCI blob transfer bypassed mise's HTTP retry wrapper.
- **Cross-repository blob mounts**: `mise oci build` now records the base image in the standard `org.opencontainers.image.base.name` manifest annotation; when the base lives on the destination registry, push attempts `POST ?mount=<digest>&from=<repo>` so base layers are linked server-side with **zero bytes transferred** (202 fallback to regular upload per spec). Token requests include the mount source's `pull` scope.
- **Progress bars**: uploads and base-layer pulls report byte progress through `MultiProgressReport`; the streaming request body resets its progress position on auth-retry so re-sends don't double-count.

## Verification (live)

- 100 MiB incompressible layer chunk-uploaded to both `registry:2` and crane's in-memory registry — both pass `crane validate --remote` (deep digest/size validation), proving chunk reassembly is byte-perfect.
- Mount path proven against `registry:2` (which uses per-repo blob links, unlike crane's shared store): pushing an app image whose base lives on the same registry reports `1 blob(s) uploaded, 0 already present, 3 mounted from base repo`, and the mounted image passes `crane validate`.
- e2e `test_oci_push_registry` (from #11140) passes against this branch; 1789 unit tests pass; lint clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Changes core OCI registry HTTP (auth scopes, uploads, retries) and new settings for insecure registries; well-covered by e2e and unit tests but affects network-heavy push/pull paths.
> 
> **Overview**
[...]

PR #11140, opened 2026-07-20
test(oci): e2e happy-path push coverage against a real registry

jdx:
## Summary

Follow-up to #11132 (item 5 of the follow-up list: CI coverage).

The oci e2e suite covered only error paths — CI had no registry to push to, so happy-path regressions in the native push client had to be caught by hand. This adds real coverage with zero CI-infrastructure changes:

- **New `e2e/oci/test_oci_push_registry`** (fast lane, runs in every PR tranche): starts `crane registry serve` — an in-memory OCI Distribution registry — on a loopback port inside the test's own environment (same pattern as `test_s3_minio_slow`), then uses crane as an *independent* client to verify what mise pushed:
  - build (scratch base, offline) + push
  - idempotent re-push reports `0 blob(s) uploaded`
  - `crane digest` == local manifest digest byte-for-byte
  - `crane validate --remote` deep-validates manifest/config/layer digests & sizes
  - tool-layer annotations survive the round-trip
  - push by digest reference
  - build+push in one shot; reproducible rebuild uploads nothing

- **New `oci.insecure_registries` setting** (docker's `insecure-registries` equivalent): non-loopback plain-HTTP registries (`registry.lan:5000`) can now be pushed to by opting in. Loopback stays insecure-by-default, and the anonymous-push warning is downgraded to debug for insecure registries where anonymous is the normal case there.

## Verification

Test passes locally in ~5s (after tool downloads): all 7 assertions green, including crane's deep validation (`PASS`). 1780 unit tests pass (new coverage for the insecure-registries matcher and host parsing); lint clean; `mise run render:schema` applied.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Changes how registry base URLs and TLS vs HTTP are chosen for pushes; misconfiguration could send traffic to the wrong scheme, though opt-in only affects listed hosts.
> 
> **Overview**
> Adds **CI happy-path coverage** for `mise oci push` via `e2e/oci/test_oci_push_registry`, which runs an in-memory `crane registry serve` on loopback and checks build/push, idempotent re-push (`0 blob(s) uploaded`), digest parity, `crane validate --remote`, annotations, digest refs, and build+push in one shot.
> 
[...]

PR #11142, opened 2026-07-20
feat(oci): reuse unchanged tool layers from the previously pushed image

jdx:
## Summary

Implements discussion #11121 — **stacked on #11141** (base retargets when it merges; only the last commit is new here).

`mise oci push` now fetches the destination ref's manifest (or `--cache-from REF`) and reuses tool layers whose cache key matches, **skipping the tar/gzip build entirely**. Reused tools don't even need to be installed locally — in CI, only tools whose version actually changed get installed and packaged.

Addressing the two caveats from the discussion:
- **Cache-key completeness**: the key is (tool short, version, in-image prefix, file owner). Tool layer descriptors now carry `dev.mise.layer.prefix` and `dev.mise.layer.owner` annotations alongside the existing short/version; images pushed by older mise versions just aren't reusable until pushed once more.
- **Trust model**: `--no-cache` forces a full local rebuild (docker-style escape hatch). Reused descriptors + config diff_ids are taken verbatim from the remote manifest/config.

Design notes:
- Default cache source is the destination ref itself; `--cache-from` supports per-commit-tag CI (`mise oci push --cache-from ghcr.io/me/dev:latest ghcr.io/me/dev:$SHA`) and is validated to be in the **same repository** — reused blobs are never re-uploaded, so they must already live where we push.
- Cache lookup failures are soft (warn + full rebuild); a broken cache can't fail a push.
- `mise oci build` / `oci run` never reuse — their output layouts must contain every blob.
- Caveat documented: `exec_env` derivation runs against local installs, so an uninstalled reused tool with an exotic backend could contribute incomplete env vars; `--no-cache` with the tool installed is the escape hatch.

## Verification

New e2e `test_oci_push_layer_reuse` (runs in every PR tranche, ~7s) proves the full story against a live registry:
1. first push → no reuse
2. **`mise uninstall jq` → same-tag re-push succeeds**: `1 tool layer(s) reused from previous image`, while `mise oci build` without cache correctly fails
3. `--cache-from` to a new tag → reuse works, jq layer digest identical across tags, `crane validate --remote` passes (blobs really are all present)
4. cross-repo `--cache-from` rejected with a clear error
5. `--no-cache` + uninstalled tool → clear "run mise install" failure
6. reinstall + `--no-cache` → byte-identical image (`0 blob(s) uploaded`)

Unit test covers the reuse-index keying (missing annotations → not reusable; owner mismatch → miss). 1790 unit tests pass; lint clean; CLI docs regenerated.

---

> [!NOTE]
> **Medium Risk**
> Reused layers trust remote manifest annotations and omit local blobs (same-repo push only); wrong cache or exotic `exec_env` on uninstalled tools could yield incomplete image config until `--no-cache`.
> 
> **Overview**
> **`mise oci push`** can now skip rebuilding tool layers when they match a previously pushed image: same tool, version, in-image prefix, and file owner. Matching layers are taken from the registry manifest (no local tar/gzip, and the tool does not need to be installed). Push output reports how many layers were reused.
> 
> New flags: **`--cache-from REF`** (reuse from another tag in the **same repository**, e.g. `latest` → per-commit SHA), **`--no-cache`** (force full local rebuild). Cache lookup defaults to the destination ref; failures are treated as a miss with a warning so pushes still succeed. **`--cache-from`** across repos is rejected because reused blobs are not re-uploaded.
> 
> The builder adds layer annotations **`dev.mise.layer.prefix`** and **`dev.mise.layer.owner`** for the cache key; **`oci build`** and **`oci run`** always set `reuse_from: None` so on-disk layouts stay complete. Docs, usage KDL, man page, unit test for reuse indexing, and e2e **`test_oci_push_layer_reuse`** cover the behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1e8df93378b13f3e8e7373e9cfcf28ad4cd00590. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Enhanced `mise oci push` with tool-layer reuse caching, reusing unchanged tool layers from the previously pushed image or from `--cache-from`.
  * Added `--no-cache` to force a full local rebuild.
  * Push output now reports how many tool layers were reused.

* **Documentation**
  * Updated `mise oci push` docs (including the man page) with clearer layer reuse rules, `--cache-from`/`--no-cache` behavior, and reuse caveats.

* **Tests**
  * Added an end-to-end test validating layer reuse, `--cache-from` behavior, and `--no-cache` rebuilds.

### DIFF
--- a/docs/cli/oci/push.md
+++ b/docs/cli/oci/push.md
@@ -11,6 +11,11 @@ required. If `--image-dir` is not passed, builds fresh from the current
 mise.toml first. Only blobs the registry doesn't already have are
 uploaded, so repeat pushes of mostly-unchanged toolsets are cheap.
 
+Tool layers whose tool, version, mount point, and file owner match the
+previously pushed image (or `--cache-from`) are reused without being
+rebuilt — those tools don't even need to be installed locally. Pass
+`--no-cache` to force a full local rebuild.
+
 Credentials are read from the same places docker and podman use:
 `$REGISTRY_AUTH_FILE`, `$XDG_RUNTIME_DIR/containers/auth.json`,
 `~/.config/containers/auth.json`, and `~/.docker/config.json`
@@ -26,6 +31,12 @@ Requires `mise settings experimental=true` (or `MISE_EXPERIMENTAL=1`).
 Destination registry reference (e.g. `ghcr.io/me/devenv:latest`)
 
 ## Flags
+
+### `--cache-from <REF>`
+
+Reuse unchanged tool layers from this image instead of the destination ref
+
+Must live in the same repository as the destination. Useful when each push gets a unique tag (e.g. per-commit tags in CI): `--cache-from ghcr.io/me/dev:latest ghcr.io/me/dev:$SHA`.
 
 ### `--from <FROM>`
 
@@ -44,6 +55,10 @@ See `mise oci build --help` for details.
 ### `--mount-point <MOUNT_POINT>`
 
 Override in-image mount point (ignored with --image-dir)
+
+### `--no-cache`
+
+Don't reuse tool layers from the previously pushed image
 
 ### `--no-mise`
 
--- a/docs/dev-tools/mise-oci.md
+++ b/docs/dev-tools/mise-oci.md
@@ -156,6 +156,35 @@ cross-repository mounted instead of re-uploaded (no bytes transferred).
 Large layers upload in chunks with progress bars, and transient network
 failures are retried with backoff (`http_retries` controls attempts).
 
+### Layer reuse
+
+Tool layers whose cache key (tool, version, in-image prefix, and file
+owner) matches the previously pushed image are **reused from the
+registry instead of rebuilt** — skipping the tar/gzip work entirely.
+Reused tools don't even need to be installed locally, which makes CI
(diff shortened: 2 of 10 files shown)


jdx:
Addressed the review finding (Greptile) in a follow-up commit: the layer-reuse log said "destination image", but with `--cache-from` the reused layer comes from the cache ref. Reworded to "cache image", which is accurate in both the default and `--cache-from` cases.

jdx:
Rebased onto main (now that #11140 and #11141 have merged) and addressed the review:

- **CodeRabbit** (host-arch warning miscounts reused layers): the "building on <host> — N tool layer(s) contain <host> binaries" warning used `versions.len()`, counting tools whose layers are reused from the cache image and never built locally. On a CI re-push where every tool is reused, it falsely warned even though nothing was built on this host. It now counts only locally-built (non-reused) layers and stays silent when there are none.
- The earlier `lint` failure was against the pre-rebase commit; `mise run lint` (fmt, prettier, clippy, shellcheck, shfmt) is clean on the rebased branch.

55 OCI unit tests pass; `test_oci_push_layer_reuse` green.

PR #11144, opened 2026-07-20
feat(oci): multi-arch image index support (--update-index)

jdx:
## Summary

Follow-up to #11132 (item 4: multi-arch), **stacked on #11142** (base retargets when it merges; only the last commit is new here).

A single host builds one platform, so single-manifest push couldn't produce a multi-arch image. `mise oci push --update-index` fixes that: it pushes the built platform manifest by digest, then points the tag at an OCI **image index** carrying one entry per platform — preserving entries other architectures pushed. One runner per arch assembles a multi-arch tag.

- **Upsert**: re-pushing the same platform replaces its entry (no duplicates); the manifest list is sorted deterministically so an unchanged platform set yields byte-identical index bytes and digest.
- **Single-arch upgrade**: a tag that was previously a plain manifest is wrapped as an index entry (platform read from its config blob) so that platform survives the upgrade.
- **Reuse composes**: `fetch_remote_image` now descends an index into the build-platform entry, so layer reuse (#11142) keeps working against `--update-index` tags.
- **Race documented**: the index update is read-modify-write (Distribution has no conditional manifest PUT); docs show sequencing per-platform CI jobs to avoid clobbering.

## Verification

New e2e `test_oci_push_update_index` (runs every tranche, ~7s):
1. `--update-index` on a fresh tag → index media type, one entry, correct host platform, `crane validate --remote` passes
2. re-push → still one entry, **identical index digest** (deterministic)
3. reuse-through-index: `uninstall jq` then re-push → `1 tool layer(s) reused`
4. single-arch tag → `--update-index` upgrades it to an index without losing the platform

Unit tests cover upsert replace/preserve and deterministic ordering. All three OCI e2e tests pass together; 1792 unit tests pass; lint clean; docs + CLI reference regenerated.

This is the last of the four follow-ups from #11132 (CI coverage #11140, push robustness #11141, layer reuse #11142, multi-arch here).

---

> [!NOTE]
> **Medium Risk**
> Changes registry push and remote cache behavior for tagged refs; index updates are read-modify-write and can race under concurrent pushes, though documented and tested.
> 
> **Overview**
> Adds **`--update-index`** to `mise oci push`: the platform manifest is pushed by digest, then the tag is updated to an OCI **image index** that **upserts** this platform and keeps other architectures’ entries. Re-pushing the same platform replaces rather than duplicates; a plain single-arch tag can be **wrapped** into the index on first use. The CLI prints **`updated image index: sha256:…`** when applicable.
> 
> Registry push logic gains **read-modify-write** index maintenance (`update_tag_index`, platform identity normalization including arm64/v8). **`fetch_remote_image`** now accepts index media types, **descends** into the host platform’s manifest so **layer reuse** still works when the cache tag is an index.
> 
> Docs (CLI, man, usage KDL, mise-oci) describe CI sequencing to avoid index update races. New **e2e** `test_oci_push_update_index` and **unit tests** for upsert/deterministic ordering.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c3e09866804991928a728c3badb4acc142a34f08. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added the `--update-index` option to `mise oci push` to keep tags as multi-architecture OCI image indexes by merging per-platform entries while preserving previously published platforms.
  * Re-pushing the same platform updates (upserts) its entry without increasing the index entry count; single-platform tags can be upgraded to multi-arch indexes.
  * Improves cross-platform layer/config reuse through the index.
* **Documentation**
  * Expanded CLI help and added “Multi-arch images” guidance, including recommended per-platform sequencing to avoid update races.
* **Tests**
  * Added an end-to-end test covering index creation, upsert/upgrade behavior, and remote validation.

### DIFF
--- a/docs/cli/oci/push.md
+++ b/docs/cli/oci/push.md
@@ -70,6 +70,12 @@ UID[:GID] to assign to every tar entry when building (conflicts with --image-dir
 
 Overrides [oci].user_id / [oci].group_id. Defaults to 0:0. If GID is omitted, it defaults to UID. This affects file ownership only; [oci].user controls the image USER directive.
 
+### `--update-index`
+
+Maintain the tag as a multi-arch image index
+
+Pushes this build's manifest by digest and points the tag at an OCI image index containing one entry per platform, preserving entries other architectures pushed. Run `mise oci push --update-index` from one runner per platform to assemble a multi-arch tag.
+
 Examples:
 
 ```
--- a/docs/dev-tools/mise-oci.md
+++ b/docs/dev-tools/mise-oci.md
@@ -414,6 +414,33 @@ For a working image, run `mise oci build` on a linux host (or inside a
 linux container — `docker run -v $PWD:/src -w /src debian mise oci build`
 works). mise prints a warning when this mismatch is detected.
 
+### Multi-arch images
+
+A single host builds a single platform, but `mise oci push
+--update-index` lets one runner per architecture assemble a multi-arch
+tag: each push uploads its platform manifest by digest and points the
+tag at an OCI **image index** that preserves the entries other
+platforms pushed.
+
+```yaml
+# CI sketch: one job per arch, same tag
+jobs:
+  push-amd64: # runs-on: ubuntu-24.04
+    run: mise oci push --update-index ghcr.io/me/dev:latest
+  push-arm64: # runs-on: ubuntu-24.04-arm
+    needs: push-amd64 # sequence to avoid a read-modify-write race
+    run: mise oci push --update-index ghcr.io/me/dev:latest
+```
+
+Re-pushing the same platform replaces its entry (no duplicates), and a
+previously single-arch tag is upgraded to an index without losing the
+existing platform. Layer reuse works through indexes — the cache
+resolves to the entry matching the build platform.
+
+Note the index update is read-modify-write (the Distribution API has no
+conditional writes), so concurrent pushes to the same tag from
+different runners can race — sequence them as above.
+
 ## Known limitations (v1)
 
 - `asdf` / `vfox` backends are rejected (see above).
--- a/e2e/oci/test_oci_push_update_index
+++ b/e2e/oci/test_oci_push_update_index
@@ -1,0 +1,83 @@
+#!/usr/bin/env bash
+# Tests `mise oci push --update-index`: the tag becomes an OCI image index
+# with one entry per platform. A single host can only build one platform, so
+# this covers: creating the index, idempotent re-push (same platform upserts
+# rather than duplicates), and upgrading a previously single-arch tag.
+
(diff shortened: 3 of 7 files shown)


jdx:
Addressed the review findings (Cursor, Greptile) in follow-up commits:

- **Index Accept header**: `fetch_remote_image` now accepts index media types, so an `--update-index` tag (which is an image index) is served by strict registries like GHCR — previously the index-descent path for layer reuse could 404 there (crane's lenient in-memory registry masked it).
- **Platform normalization**: `platform_from_config` / `wrap_single_manifest` normalize architecture/os (`normalize_arch`/`normalize_os`), so a config carrying `x86_64`/`aarch64` no longer breaks reuse or produces duplicate index entries after `--update-index`.
- **os.version/features**: preserved (were dropped), and `upsert_platform_manifest`'s platform identity now includes `os_version` so two Windows platforms differing only by `os.version` don't collide. No-op for the linux case, but removes a latent data-loss path.

Unit test added for the os.version distinction; `test_oci_push_update_index` still passes end-to-end.

jdx:
Addressed the remaining review findings (Cursor) in a follow-up commit. The index-descent in `fetch_remote_image` and `upsert_platform_manifest` now share one canonical `platform_identity_parts`, so they agree on which entry represents the host platform. It:

- normalizes arch/os (so a config using `x86_64`/`aarch64` matches the normalized host and doesn't create a duplicate entry after `--update-index`), and
- fills arm64's implied `v8` variant, so mise's variantless arm64 and a buildx-style `arm64/v8` entry are treated as the same platform (re-push replaces rather than duplicates); `os.version` still distinguishes Windows platforms.

The earlier normalization comment (registry.rs:1071) was already fixed in the prior commit. Unit tests added for the arm64-variant equivalence and arch normalization; `test_oci_push_update_index` and `test_oci_push_layer_reuse` pass.

jdx:
Rebased onto main (now that #11142 has merged) and addressed the failing check + review:

- **Failing `lint`**: the real failure was `cargo clippy -- -D warnings` flagging `clippy::unnecessary_sort_by` on the index-entry ordering in `upsert_platform_manifest`. Switched to `sort_by_key` (ordering unchanged; `upsert_is_deterministically_ordered` still passes). `test-ci` was aggregating the lint failure.
- **CodeRabbit nitpick**: extracted the duplicated `sha256:`-digest computation (in `update_tag_index` and `wrap_single_manifest`) into a small `sha256_digest` helper.

Also regenerated `man/man1/mise.1` for the `--update-index` flag (the render-diff check). Builds clean; OCI unit tests pass.

PR #11148, opened 2026-07-20
feat(core): verify node/swift signatures in-process with rpgp, drop external gpg

jdx:
## Summary

Users frequently hit problems with the external `gpg` binary (missing, misconfigured keyrings, GnuPG version quirks) when mise verifies Node.js and Swift downloads. This replaces shelling out to `gpg` with the pure-Rust [rPGP](https://github.com/rpgp/rpgp) crate (`pgp`), doing detached OpenPGP signature verification entirely in-process.

mise's use of gpg was always narrow — **detached signature verification against a bundled keyring**, nothing else (no signing, encryption, or web-of-trust; both call sites already used `--trust-model always`). That's well within scope for a library.

## What changed

- New `src/gpg.rs`: `verify_node` / `verify_swift` verify detached signatures against the bundled release-signing keyrings (`src/assets/gpg/*.asc`) in-process.
- Keys are matched by the signature's issuer key-id/fingerprint on the fast path, so the signed content is hashed at most once. Falls back to trying all trusted keys only when a signature carries no issuer hint.
- Handles the real-world encodings: node ships a **binary** `.sig` with **multiple** signature packets; the parser accepts both binary and ASCII-armored signatures, and splits the concatenated multi-block keyrings (node bundles 29 key blocks — rPGP's `from_armor_*` only reads the first block, so we split them).
- The Swift tarball (hundreds of MB) is **streamed from disk**, matching mise's existing chunked hashing — it is never buffered whole in memory.
- Removes the external `gpg` binary dependency, the `gpg --import` step, and the bundled-key import helpers.

## Behavior change (please review)

Previously, if the `gpg` binary was **not installed** and the setting was unset, verification was **silently skipped**. Since no external binary is now required, **verification always runs** when enabled. This is a security improvement, but it removes that escape hatch:

- Opt out explicitly with `gpg_verify = false` (or `MISE_NODE_GPG_VERIFY=false` / `MISE_SWIFT_GPG_VERIFY=false`).
- The node "signature not found (404) → skip" path is preserved (that's about availability, not a downgrade).
- Verification uses the **same bundled keyrings** as before, so anything that verified with the old code verifies now.

## Testing

- Unit tests in `src/gpg.rs` (parse all bundled node/swift keys, reject empty/garbage signatures).
- **Real end-to-end**: installed `node@20.11.0` with `MISE_NODE_GPG_VERIFY=true` using the built binary — real `SHASUMS256.txt.sig` verified and node runs.
- **Negative test**: confirmed a tampered `SHASUMS256.txt` is *rejected* against the genuine signature (temporary test, removed before commit).
- `cargo clippy` clean, `cargo fmt` applied, `render:schema` regenerated.

## rPGP dependency rationale

- Pure Rust, permissively licensed (MIT/Apache-2.0), RFC 9580, actively maintained.
- It's the OpenPGP implementation **rustup and Delta Chat depend on** — battle-tested for exactly this "verify a release signature against a bundled keyring" use case.
- Chosen over `sequoia-openpgp`, which by default pulls the `nettle` C crypto backend (system lib + bindgen) — the opposite of the goal here.

---

> [!NOTE]
> **Medium Risk**
> Touches supply-chain signature verification and changes default behavior for environments that relied on missing gpg to skip checks; mitigated by same bundled keys and explicit `gpg_verify = false`.
> 
> **Overview**
> **Replaces shelling out to `gpg`** with built-in detached OpenPGP verification using the `pgp` (rPGP) crate. Node and Swift installs now verify against the same bundled keyrings in `src/assets/gpg/*.asc` without importing keys or invoking an external binary.
> 
> The new `src/gpg.rs` handles multi-block keyrings, binary and armored signatures, issuer-based key matching, and **streams Swift tarballs from disk** during verification. Node/Swift plugin paths call `gpg::verify_node` / `gpg::verify_swift` instead of `gpg --verify`.
> 
> **Behavior change:** when `gpg` was missing and settings were unset, verification used to be **skipped silently**. It now runs whenever `gpg_verify` is not explicitly `false` (global `gpg_verify` still propagates to node/swift). Node still skips verification if the `.sig` download 404s.
> 
> Docs, `settings.toml`, and `schema/mise.json` describe built-in verification; e2e comments clarify `MISE_NODE_VERIFY=0`. Unit tests include real Node v20.11.0 SHASUMS fixtures.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1183a39a53503a5bafc933f164f76a3d6aca8c9f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added built-in OpenPGP (detached) signature verification for Node.js and Swift downloads.
  * Verification no longer requires an external `gpg` installation (including on Linux).
  * Supports both armored and binary signature formats using bundled trusted signing keys.
* **Documentation**
  * Updated security documentation and OpenPGP setting descriptions to reflect built-in verification and that setting `false` disables verification.
* **Tests**
  * Expanded verification and parsing test coverage for valid, invalid, and tampered inputs.


jdx:
Thanks for the reviews — addressed in d74d40a:

**Fallback re-tries issuer-bearing signatures (CodeRabbit + Greptile P1):** Fixed. The fallback loop now filters to `!has_issuer` signatures only, matching the stated intent — a signature naming an untrusted issuer is genuinely unverifiable and no longer triggers re-reads of the (large) content against every key.

**No automated crypto-correctness test (Greptile P2):** Added a real, checked-in test vector — the genuine Node.js v20.11.0 `SHASUMS256.txt` plus its binary detached `.sig` (a multi-packet bundle) — verified against the actual bundled keyring. Also added tamper-rejection and wrong-keyring-rejection tests. (Note: the existing `e2e/core/test_node_slow` already installs node@24 with verification on, so the happy path also has CI coverage; the unit vector adds a fast, offline regression guard.)

**Skip missing Swift `.sig` on 404 (CodeRabbit Major):** Intentionally keeping the hard-fail here, with a code comment added to explain. The cited gap (Windows toolchains lacking signatures) doesn't apply to this path — mise only runs Swift GPG verification on Linux (`cfg!(target_os = "linux")`), where swift.org publishes a detached signature for every release tarball. For a security feature, failing closed on an unexpectedly-missing Linux signature is preferable to silently downgrading, and this matches the pre-existing behavior.

**e2e `MISE_NODE_VERIFY=0` comment (CodeRabbit Minor):** `MISE_NODE_VERIFY=0` disables `node.verify`, which gates the entire checksum path — and GPG verification runs *inside* that path (`get_checksum`), so setting it does skip signature verification (by skipping the download+checksum+verify step wholesale). The updated comment ("skip checksum/signature verification") reflects that.

PR #11154, opened 2026-07-21
fix(ruby): bypass versions host for custom precompiled releases

risu729:
## Summary

- let callers of the shared GitHub release helper opt out of `mise-versions`
- keep `mise-versions` release and attestation lookups for the default `jdx/ruby` precompiled source
- fetch release metadata directly from GitHub for custom `ruby.precompiled_url = "owner/repo"` sources
- keep hosted and direct release responses in separate cache entries

## Root cause

Ruby's precompiled installer treated an arbitrary `ruby.precompiled_url = "owner/repo"` as though it were a registry-backed release source. Exact-tag and fallback lookups therefore queried the restricted `mise-versions` endpoint first, producing a 403 warning for custom repositories the service cannot allowlist.

The default `jdx/ruby` source is known and intentionally supported by `mise-versions`. With jdx/mise-versions#273 allowing its release metadata, this PR now branches by source instead of bypassing the service for every precompiled repository.

## Routing after this change

- default `jdx/ruby`: use `mise-versions` first for release metadata and attestations, with the existing GitHub fallback
- custom `owner/repo`: query GitHub directly for release metadata and attestations
- `oneclick/rubyinstaller2`: unchanged; its separate Windows path continues using `mise-versions`

The source-aware flag is propagated through locked build-revision lookups and exact-release fallbacks, so custom repositories avoid the rejected service request on every path.

## Validation

- `mise run lint-fix`
- `cargo test --all-features test_versions_host_flag_splits_release_cache`
- `cargo test --all-features test_ruby_precompiled_versions_host_only_for_default_source`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved release lookup reliability across supported release sources, including cases where different “versions host” handling applies.
  * Fixed fallback behavior when matching build revisions are unavailable, ensuring the correct release metadata is used.
  * Improved detection and retrieval of precompiled Ruby assets for both locked and unlocked builds.
  * Ensured consistent release information usage and improved verification behavior for precompiled artifact attestations.

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -305,12 +305,20 @@ async fn list_tags_with_dates_(api_url: &str, repo: &str) -> Result<Vec<GithubTa
 }
 
 pub async fn get_release(repo: &str, tag: &str) -> Result<GithubRelease> {
-    let key = release_cache_key(API_URL, repo, tag, true);
+    get_release_with_versions_host(repo, tag, true).await
+}
+
+pub async fn get_release_with_versions_host(
+    repo: &str,
+    tag: &str,
+    use_versions_host: bool,
+) -> Result<GithubRelease> {
+    let key = release_cache_key(API_URL, repo, tag, use_versions_host);
     let cache = get_release_cache(&key).await;
     let cache = cache.get(&key).unwrap();
     cache
         .get_or_try_init_async_if(
-            async || get_release_with_options(API_URL, repo, tag, true).await,
+            async || get_release_with_options(API_URL, repo, tag, use_versions_host).await,
             should_cache_release,
         )
         .await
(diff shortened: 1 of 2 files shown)


jdx:
is this still needed?

risu729:
> is this still needed?

Yes, for users with custom `ruby.precompiled_url = "owner/repo"`. Without this PR, mise-versions is always used for the repo even if it's not `jdx/ruby`.

jdx:
ok but we definitely want to use it for jdx/ruby and oneclick/rubyinstaller2, I think this regresses that behavior

risu729:
It doesn't regress, as we didn't use `mise-versions` for `jdx/ruby` (it was only configured for the attestations), but I agree we should use it. I opened a PR for mise-versions. https://github.com/jdx/mise-versions/pull/273

jdx:
Verified the routing in this PR against the code:

- **Default `jdx/ruby`** — `use_versions_host_for_precompiled_source` returns `true`, so release metadata + attestations still go through mise-versions first (with the existing GitHub fallback). Unchanged for default `core:ruby` users.
- **Custom `ruby.precompiled_url = "owner/repo"`** — flag is `false`, so these now query GitHub directly and skip the versions host, avoiding the 403 on every path (locked build-revision, exact-tag fallback, and attestation verify).
- **`oneclick/rubyinstaller2` (Windows)** — untouched. That path goes through `resolve_rubyinstaller_lock_info` → plain `github::get_release(...)`, which after this change delegates to `get_release_with_versions_host(repo, tag, true)` (hardcoded `true`), so it continues using mise-versions exactly as before.

The title ("bypass versions host for custom precompiled releases") accurately describes the net behavioral change — keeping the versions host for the default source is pre-existing behavior, and the new bit is the bypass for custom repos.

PR #11172, opened 2026-07-21
fix(sync): clear stale incomplete markers for external links

risu729:
## Summary

- serialize external-link state updates with the tool-version install lock
- clear a stale incomplete marker when an existing external link is verified healthy
- clear the marker after creating a healthy external link
- cover the uv-to-mise sync path with an end-to-end regression test

## Bug

An interrupted or failed mise install can leave an `incomplete` marker for a version. If that version is later provided by uv, nvm, pyenv, nodenv, or Homebrew, sync creates or confirms the external link but leaves the stale marker behind. Mise therefore continues treating the healthy external version as incomplete.

For example:

```sh
uv python install 3.11.1
mkdir -p "$MISE_CACHE_DIR/python/3.11.1"
touch "$MISE_CACHE_DIR/python/3.11.1/incomplete"
mise sync python --uv
mise where python@3.11.1
```

Before this fix, the final command fails even though the uv installation exists and mise linked it. After this fix, sync clears the marker once it confirms the external target exists, and `mise where` returns the linked installation.

## Split follow-ups

- #11682 reconciles links by provider ownership and preserves managed installs, runtime aliases, and other providers' links
- #11683 treats Windows junctions like symlinks when syncing mise Python installs to uv
- #11686 serializes provider reconciliation with install, uninstall, and link operations one version at a time
- #11687 continues reconciliation after marker cleanup failures and reports the errors after rebuilding runtime links

## Testing

- `mise run lint-fix`
- `mise run test:e2e e2e/sync/test_sync_python_uv`

### DIFF
--- a/e2e/sync/test_sync_python_uv
+++ b/e2e/sync/test_sync_python_uv
@@ -5,7 +5,11 @@ export MISE_PYTHON_GITHUB_ATTESTATIONS=0
 # (astral-sh/uv#19278). Bump once 0.11.10 ships with attestations restored.
 assert "mise use -g uv@0.11.8 python@3.11.3"
 assert "mise x -- uv python install 3.11.1"
+mkdir -p "$MISE_CACHE_DIR/python/3.11.1"
+touch "$MISE_CACHE_DIR/python/3.11.1/incomplete"
 export UV_PYTHON_DOWNLOADS=never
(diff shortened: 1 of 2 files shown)


risu729:
@coderabbitai resume

PR #11204, opened 2026-07-22
fix(task): deduplicate equivalent task sources

risu729:
## Summary

- compare task sources using resolved physical paths
- deduplicate global and monorepo configs reached through aliases or overlapping config roles
- compare safe deferred task file paths with an empty Tera engine without executing mise template helpers
- reuse rendered file tasks across global config precedence units without changing include-order precedence
- preserve unresolved path components when canonicalization cannot safely resolve them

### Equivalent sources were loaded more than once

The same physical config or task file can be reached through a real path, a symlink, or both user-global and system config settings. For example, a user config may include `~/.config/mise/tasks` while a system config includes a symlink to that directory. A task header such as:

```bash
#MISE description="{{ exec(command='echo rendered >> /tmp/count') }}"
```

could be rendered twice even though only one task appeared in the final list.

The loader now identifies physical sources before reusing rendered script and TOML tasks across global config precedence units. Reuse is deliberately delayed until the current config is complete, so `includes = ["first", "second"]` still lets `second` win.

### Source comparison could execute task templates

A task may intentionally defer its file path:

```toml
[tasks.build]
file = "{{ exec(command='printf /repo/tasks/build; touch /tmp/ran') }}"
```

`Task::render` leaves this `file` template deferred. Before this PR, merely comparing two task sources used the normal Tera engine, so a read-only command such as `mise tasks` could execute the command and create `/tmp/ran` even though it only needed to list tasks.

Source comparison now uses an empty Tera engine with only the source-stable `config_root` context. Mise helpers such as `exec()`, `read_file()`, `hash_file`, and `arch()` are unavailable. If a `file` expression needs one of them, comparison returns “unknown” and conservatively does not deduplicate or apply a source-based overlay.

This does not disable these task definitions. When the task file path is actually needed—for example, by `mise run build`—`Task::file_path()` uses the normal Tera engine and `exec()` runs then. The tradeoff is an intentional false negative: equivalent sources are not deduplicated when proving equivalence would itself require executing code.

### The active config context could create a false source match

Suppose the active config defines `vars.script = "/tasks/a"`, while a task defines its own `vars.script = "/tasks/b"` and uses `file = "{{ vars.script }}"`. Resolving that task with the active config context incorrectly identifies it as `/tasks/a`, potentially applying an unrelated lower-precedence overlay.

Source identity now exposes only the stable `config_root` value. Task-local or active config vars cannot manufacture a false match.

### Unresolved parents after symlinks could compare incorrectly

When canonicalization failed, lexical `..` collapsing could equate distinct paths. For example, if `link` points outside the current directory, `link/../tasks/build` must resolve relative to the symlink target, not the directory containing `link`. Likewise, `missing/../target` must not be collapsed when `missing` does not exist.

Path fallback now canonicalizes the existing prefix one component at a time and preserves the unresolved suffix verbatim after the first failure.

## Tests

- `cargo test --all-features test_desymlink_path`
- `mise run test:e2e e2e/tasks/test_task_global_config_dedup`
- `mise run test:e2e e2e/tasks/test_task_monorepo_config_dedup`
- `mise run test:e2e e2e/tasks/test_task_source_comparison_no_exec`
- `mise run test:e2e e2e/tasks/test_task_ls_global`
- `mise run test:e2e e2e/tasks/test_task_config_precedence`
- `mise run test:e2e e2e/tasks/test_task_monorepo_nested_root`
- `mise run lint-fix`
- `git diff --check`

Rebased onto the latest `main` after #11103 merged.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved task/config discovery and de-duplication to be symlink-aware (including relative/broken targets) to prevent duplicate task execution and template rendering.
  - Kept include precedence (“later include wins”) consistent across global, local, and monorepo setups, including when includes are reached via symlinks.
  - Prevented deferred templates from running during task source comparison.
- **Tests**
  - Added/expanded end-to-end coverage for global overlays, monorepo de-duplication, symlink-aware resolution, include precedence, and deferred-template safety.


risu729:
@coderabbitai resume

PR #11222, opened 2026-07-23
fix(task): preserve HTTP task snapshots

risu729:
## Summary

- give individually fetched no-cache HTTP task files unique, command-owned artifacts
- reuse one command-scoped snapshot when dependency resolution materializes the same direct remote task again, including across `Config` resets, while keeping separate task definitions independent
- keep temporary task files alive through execution by retaining their artifacts in a command-scoped map
- establish HTTP artifact ownership before the download awaits, so cancellation cannot leave a late atomic persist outside the owned directory
- release artifacts through a refcounted command RAII guard after normal completion, returned errors, and requested exit statuses
- retry only narrow transient teardown races while preserving strict file-or-directory cleanup

## Root cause

No-cache HTTP task providers reused URL-keyed destinations. When the same mutable source was resolved more than once, a later fetch could overwrite the path retained by an earlier task, so mise could parse one body and execute another. Dependency resolution may also materialize the same logical task again after a config reset.

The command-scoped direct-task map now single-flights repeated materializations by config source, task name, and remote source. Different task definitions that intentionally reference the same URL still receive independent snapshots.

The artifact directory is created and owned before the first cancellable await. Both the downloader temporary file and final destination remain inside it, and the final active command scope releases the artifact after execution.

## Scope

This is the shared artifact-ownership and HTTP layer extracted from the former combined snapshot PR. Git task files and repository/ref task-include reuse are intentionally excluded into a dependent Git layer.

Trust provenance, passive-fetch authorization, remote alias/visibility resolution, and metadata side-effect checks remain deferred to #11113. Cached publication, locking, retention, platform cache hardening, and CLI no-cache propagation remain deferred to #11203.

## Merge ordering

#11440 is merged. Its deferred process termination lets the command-scoped guard unwind before process exit.

## Validation

- `mise run lint-fix`
- `cargo test --locked --bin mise task::task_fetcher::tests`
- `cargo test --locked --bin mise task::task_file_providers`
- `mise run test:e2e -- test_task_remote_snapshots`


risu729:
Rebased this PR onto the latest `main` and resolved the remote Git provider conflict.

Post-rebase review fixes:
- scope direct `TaskFetcher` artifacts with RAII so unit/direct callers do not leak snapshots
- serialize last-scope cleanup to avoid clearing a newly active scope
- retry transient Windows sharing violations during artifact cleanup
- assert one Git upload-pack operation per repo/ref/command in the e2e test

Validation:
- `mise run lint-fix`
- `cargo test task::task_file_providers`
- `cargo test task::task_fetcher::tests`
- `cargo test file::tests::test_retry_remove_all`
- `mise run test:e2e -- test_task_remote_git_includes test_task_remote_snapshots`
- three independent subagent reviews; no remaining findings

risu729:
Follow-up subagent review fixes are pushed in 2c45c6b0a.

- unified direct-task and Git-include artifact cleanup under the final active command scope, with nested-scope regression coverage
- retained the securely-created Git checkout directory through clone instead of deleting and recreating it
- localized strict file-or-directory teardown retries to temporary task artifacts, including transient Windows sharing failures
- kept trust/provenance (#11113) and cached publication/locking/retention (#11203) work out of this foundation PR

Validated with `mise run lint-fix`, focused task-fetcher/provider unit tests, both remote snapshot e2Es, and the Git include e2E under `MISE_GIX=1`.

PR #11234, opened 2026-07-24
fix(python): deprecate virtualenv tool option in favor of _.python.venv

JamBalaya56562:
## Summary

Per @jdx's review, this PR now **deprecates** the legacy python `virtualenv` tool option instead of improving it.

- Emits a `deprecated_at!` warning when the `virtualenv` tool option is used, steering users to the `_.python.venv` env directive (same activate-an-existing-venv behavior).
- **warn immediately** (`2026.7.0`), **remove in ~12 months** (`2027.7.0`).
- Docs (`docs/plugins.md`, `docs/lang/python.md`) now carry a deprecation banner.
- The earlier path-resolution change (and its helper/tests) has been reverted; `get_virtualenv` keeps its original behavior aside from the deprecation warning.

This resolves [discussion #6019](https://github.com/jdx/mise/discussions/6019) by migration (users move to `_.python.venv`, which resolves relative paths against the declaring config file) rather than by touching the legacy resolution path.

### Commits
1. `5a73042` — original path-resolution fix (superseded)
2. `d378f87` — pivot to deprecation per review

(Kept as additive commits so the maintainer's review history stays intact; happy to squash on merge.)

---

Refs #6019

*This pull request was created by an AI coding assistant.*

## Summary by CodeRabbit

* **Documentation**
  * Updated the Python “Tool Options” documentation to mark the legacy `python.virtualenv` option as deprecated, advising users to use `_.python.venv` in the `[env]` section instead.
* **User Experience**
  * Added a deprecation warning when the legacy `python.virtualenv` option is used, pointing users to `_.python.venv` for future-compatible configuration.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -123,8 +123,12 @@ of tools. One example of this is virtualenv on python runtimes:
 
 ```toml
 [tools]
-python = { version='3.11', virtualenv='.venv' }
+python = { version='3.11', virtualenv='.venv' } # a relative path resolves against this file's config root
 ```
+
+A relative `virtualenv` path is resolved against the config root of the file that declares it
+(matching `_.python.venv`), not the current project root — so a `virtualenv` set in the global
+config won't be looked up inside unrelated projects.
 
 This will be passed to all plugin scripts as `MISE_TOOL_OPTS__VIRTUALENV=.venv`. The user can specify
 any option, and it will be passed to the plugin in that format.
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -123,12 +123,13 @@ of tools. One example of this is virtualenv on python runtimes:
 
 ```toml
 [tools]
-python = { version='3.11', virtualenv='.venv' } # a relative path resolves against this file's config root
+python = { version='3.11', virtualenv='.venv' }
 ```
 
-A relative `virtualenv` path is resolved against the config root of the file that declares it
-(matching `_.python.venv`), not the current project root — so a `virtualenv` set in the global
-config won't be looked up inside unrelated projects.
+::: warning
+The python `virtualenv` tool option is deprecated and will be removed in a future release.
+Use [`_.python.venv`](/lang/python.html#automatic-virtualenv-activation) in the `[env]` section instead.
+:::
 
 This will be passed to all plugin scripts as `MISE_TOOL_OPTS__VIRTUALENV=.venv`. The user can specify
 any option, and it will be passed to the plugin in that format.
--- a/src/plugins/core/python.rs
+++ b/src/plugins/core/python.rs
@@ -6,7 +6,6 @@ use crate::build_time::built_info;
 use crate::cache::{CacheManager, CacheManagerBuilder};
 use crate::cli::args::BackendArg;
 use crate::cmd::CmdLineRunner;
-use crate::config::config_file::config_root;
 use crate::config::{Config, Settings};
 use crate::file::{ExtractOptions, ExtractionFormat, display_path};
 use crate::git::{CloneOptions, Git};
@@ -14,7 +13,7 @@ use crate::http::{HTTP, HTTP_FETCH};
 use crate::install_context::InstallContext;
 use crate::lockfile::{PlatformInfo, ProvenanceType};
 use crate::platform::Platform;
-use crate::toolset::{ToolRequest, ToolSource, ToolVersion, ToolVersionOptions, Toolset};
+use crate::toolset::{ToolRequest, ToolVersion, ToolVersionOptions, Toolset};
 use crate::ui::progress_report::SingleReport;
 use crate::{Result, lock_file::LockFile};
 use crate::{dirs, file, plugins, sysconfig};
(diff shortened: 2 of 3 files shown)


src/plugins/core/python.rs
```diff
@@ -529,13 +530,12 @@ impl PythonPlugin {
...
-            let mut virtualenv: PathBuf = file::replace_path(Path::new(virtualenv));
-            if !virtualenv.is_absolute() {
-                // TODO: use the path of the config file that specified python, not the top one like this
-                if let Some(project_root) = &config.project_root {
-                    virtualenv = project_root.join(virtualenv);
-                }
-            }
+            let virtualenv: PathBuf = file::replace_path(Path::new(virtualenv));
+            let virtualenv = resolve_virtualenv_path(
+                virtualenv,
```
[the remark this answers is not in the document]
JamBalaya56562: Thanks — this is a real code path (`ToolsetBuilder::load_runtime_args` copies the config's options onto the CLI-override request via `apply_arg_options`/`set_options` while keeping `ToolSource::Argument`), so a relative `virtualenv` copied onto a `mise exec python@x` override does fall back to `project_root` here.

I've intentionally left this out of the current PR for two reasons:

1. **It is not a regression.** Before this change, *all* cases resolved against `config.project_root`, so the CLI-override path behaves exactly as it does today — this PR strictly improves the config-declared case without changing the override case.
2. **A correct fix belongs in the builder, not this resolver.** The resolver only receives the copied options; the declaring config's `ToolSource`/path is discarded at the copy site. Properly fixing it means preserving provenance when options are carried over in `load_runtime_args`, which affects every backend (source display, lockfile identity, "command argument" messaging), so it's too broad to fold into a focused python venv fix.

Happy to open a follow-up that threads the declaring source through the arg-option copy if a maintainer wants that behavior for CLI overrides.

*This reply was generated by an AI coding assistant.*

src/plugins/core/python.rs
```diff
@@ -1237,6 +1271,52 @@ mod tests {
...
+        assert_eq!(out, PathBuf::from("/tmp/mise-fallback-proj/.venv"));
+    }
+
+    #[test]
+    #[cfg(unix)]
+    fn resolve_virtualenv_path_relative_no_base_uses_cwd() {
+        // No source path and no project_root -> resolve under CWD.
+        let out = resolve_virtualenv_path(PathBuf::from(".venv"), &ToolSource::Argument, None);
+        assert!(out.ends_with(".venv"));
+    }
```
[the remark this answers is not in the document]
JamBalaya56562: Thanks — applied all three in the amended commit (`5a73042`):

- **`file::replace_path` docstring:** removed the "env vars are expanded" claim; it only expands a leading `~/`, so the doc now says just that.
- **"config root" terminology:** updated the helper docstring and both docs (`docs/plugins.md`, `docs/lang/python.md`) to say "config root" consistently, since `config_root()` isn't always the literal config-file directory.
- **CWD fallback test:** now asserts `out == dirs::CWD.as_ref().unwrap().join(".venv")` instead of the weaker `ends_with`.

*This reply was generated by an AI coding assistant.*

docs/plugins.md
```diff
@@ -123,9 +123,13 @@ of tools. One example of this is virtualenv on python runtimes:
 
 ```toml
 [tools]
-python = { version='3.11', virtualenv='.venv' }
+python = { version='3.11', virtualenv='.venv' } # a relative path resolves against this file's config root
```
jdx: is this option not already deprecated? it should be. We definitely shouldn't be improving it anymore.

JamBalaya56562: You're right — it isn't currently deprecated: there's no `deprecated_at!` on the `virtualenv` tool option and the docs don't mark it as such.

Happy to pivot this PR in that direction instead of improving it: drop the path-resolution change and instead deprecate the `virtualenv` tool option via `deprecated_at!`, pointing users to `_.python.venv` (which covers the same activate-an-existing-venv use case). That would resolve #6019 by migration rather than by touching the legacy resolution path.

If that works for you, what `warn_at` / `remove_at` versions would you like — e.g. warn starting at the current release? I'll push it as a follow-up commit.

jdx: yes let's use `deprecated_at!` and get rid of it instead. We can warn immediately, I think we could wait 12 months to get rid of it since it isn't really hurting anything being there.

JamBalaya56562: Done — pushed as a follow-up commit (`d378f87`):

- Added a `deprecated_at!("2026.7.0", "2027.7.0", "python.virtualenv", …)` warning that fires when the `virtualenv` tool option is used, pointing to `_.python.venv`. Warns immediately, `debug_assert` removal reminder at ~12 months.
- Reverted the earlier path-resolution change (helper + tests); `get_virtualenv` is back to its original behavior aside from the warning.
- Docs now carry a deprecation banner instead of the resolution note.

Kept it as an additive commit rather than force-pushing so your review history stays intact — feel free to squash on merge. PR title/body updated to reflect the deprecation.

src/plugins/core/python.rs
```diff
@@ -529,9 +529,14 @@ impl PythonPlugin {
...
+                "2027.7.0",
+                "python.virtualenv",
+                "the python `virtualenv` tool option is deprecated. Use `_.python.venv` in the `[env]` section instead: https://mise.en.dev/lang/python.html#automatic-virtualenv-activation"
+            );
             let mut virtualenv: PathBuf = file::replace_path(Path::new(virtualenv));
             if !virtualenv.is_absolute() {
-                // TODO: use the path of the config file that specified python, not the top one like this
                 if let Some(project_root) = &config.project_root {
                     virtualenv = project_root.join(virtualenv);
                 }
```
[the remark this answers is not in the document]
JamBalaya56562: This is intentional. Per the maintainer's decision on this PR ([review comment](https://github.com/jdx/mise/pull/11234#discussion_r3642703455)), the `virtualenv` tool option is being **deprecated rather than improved**, so the earlier `config_root`-based resolution fix was deliberately reverted and `get_virtualenv` keeps its original `project_root` behavior.

The residual behavior you describe (a relative `virtualenv` in the global config resolving against the active project) is exactly why the option is now deprecated: the new `deprecated_at!` warning steers users to `[env] _.python.venv`, which does resolve relative paths against the declaring config file. So #6019 is addressed by migration, not by changing this path — reintroducing the resolution fix here would undo the maintainer's requested direction.

*This reply was generated by an AI coding assistant.*

PR #11286, opened 2026-07-25
registry: fix danger-swift installation

Marukome0743:
## Summary

- add an explicit `install_command` option for Swift Package Manager source installs
- use Danger Swift official installer so the executable, `libDanger`, and Swift modules are installed together
- document the generic behavior and add regression coverage for Danger Swift local execution

## Root cause

The SPM backend only discovered executable products and ran `swift build --product`. Danger Swift also needs its dynamic `libDanger` library and Swift modules at runtime, and those artifacts are only placed correctly by its official install target.

## Implementation

- validate `install_command` as a non-empty string and reject its use with `filter_bins`
- run the command from the checked-out package with the configured inline shell, `install_env`, Swift dependency `PATH`, and fixed `PREFIX` / `MISE_TOOL_INSTALL_PATH`
- register it as an install-time option so option changes invalidate stale install metadata
- keep artifact bundle installs unchanged and never auto-run package Makefiles
- configure the existing `danger-swift` registry entry to force a source install and run `make install PREFIX="$MISE_TOOL_INSTALL_PATH"`

## Testing

- `shellcheck -x e2e/core/test_swift_slow`
- `shfmt -d -s e2e/core/test_swift_slow`
- `markdownlint docs/dev-tools/backends/spm.md`
- `taplo fmt --check` / `taplo check` for `registry/danger-swift.toml`
- `cargo test backend::spm::tests` (23 passed)
- `mise run test:e2e e2e/core/test_swift_slow` on macOS, including SwiftFormat and `danger-swift local` with Danger JS 13.0.10
- built the current source as a linux-x64 ELF in `ghcr.io/jdx/mise:e2e`; the full Docker E2E is left to the native CI runner because local arm64 Docker required slow QEMU emulation

## Popularity

- GitHub: approximately 1.1k stars and 151 forks
- Latest release: 3.22.1
- Active development: repository activity observed on 2026-07-23

Addresses https://github.com/jdx/mise/discussions/3540

*This pull request was generated with assistance from an AI coding assistant.*

## Summary by CodeRabbit

* **New Features**
  * Added support for custom installation commands when installing Swift packages from source.
  * Custom commands run with the configured environment, dependency tools, and installation paths.
  * Added documentation and an example for configuring custom SPM install commands.

* **Bug Fixes**
  * Improved Danger Swift installation to ensure its CLI, libraries, and Swift modules are correctly available across platforms.
  * Added validation for invalid or conflicting installation options.

### DIFF
--- a/docs/dev-tools/backends/spm.md
+++ b/docs/dev-tools/backends/spm.md
@@ -143,6 +143,32 @@ product in the package, installation fails with a clear error.
 "spm:swiftlang/swiftly" = { version = "latest", filter_bins = "swiftly" }
 ```
 
+### `install_command`
+
+Run an explicit command from the checked-out package directory instead of discovering executable
+products and running `swift build --product`. The command uses mise's default inline shell and
+inherits [`install_env`](#install_env) plus the `PATH` for the Swift dependency. `PREFIX` and
+`MISE_TOOL_INSTALL_PATH` are both set to the tool's installation directory.
+
+This option only applies to source installs and cannot be combined with `filter_bins`. Mise never
+automatically runs a package's Makefile or other installation scripts; the command must be
+configured explicitly.
+
+For example, Danger Swift's official `make install` installs both its executable and the dynamic
+library required by `danger-swift local`:
+
+```toml
+[tools]
+"spm:danger/swift" = { version = "3.22.1", artifactbundle = false, install_command = "make install PREFIX=\"$MISE_TOOL_INSTALL_PATH\"" }
+```
+
+The `danger-swift` registry shorthand already supplies these options, so this is sufficient:
+
+```toml
+[tools]
+danger-swift = "3.22.1"
+```
+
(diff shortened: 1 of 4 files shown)


jdx:
Reverted in https://github.com/jdx/mise/commit/4c32d0fac because this change blocks the release CI on Ubuntu 26.04.

The new Danger Swift integration invokes SwiftPM using Swift's Ubuntu 24.04 toolchain. Ubuntu 26.04 provides `libxml2.so.16`, while that toolchain requires `libxml2.so.2`, so `swift-package` and `swift-build` cannot start. Danger Swift is then reported as installed even though its executable was never built, and `core/test_swift_slow` fails on both CI attempts.

We should revisit this once there is a native installation path compatible with Ubuntu 26.04 (for example, an upstream Swift 26.04 toolchain or suitable prebuilt Danger Swift artifacts).

PR #11293, opened 2026-07-25
security(config): restrict default shell args to global config

jdx:
## Summary
- mark the Unix and Windows default file/inline shell argument settings as global-only
- add regression coverage proving local config values are stripped while global config values remain supported

## Rationale
Settings from local project configuration are preloaded before trust evaluation. Allowing those files to override the default shell interpreter arguments lets an untrusted repository influence how commands from trusted sources are executed. Applying the existing `global_only` policy to all four settings closes that gap while preserving global configuration and environment-variable customization.

Reported by @arpitjain099.

## Validation
- `cargo test config::settings::tests` (45 passed)
- `mise run lint-fix` (includes formatting, schema validation, shell checks, and `cargo check --all-features`)

---

> [!NOTE]
> **High Risk**
> Security fix for command execution: local repos could previously steer default shell invocation; behavior change for anyone relying on project-level overrides of these four settings.
> 
> **Overview**
> Closes a security gap where **untrusted project `mise.toml` files could override Unix/Windows default file and inline shell arguments** before config trust is evaluated, influencing how mise runs inline and file-based commands.
> 
> The four settings (`unix_default_*` and `windows_default_*` shell args) are now **`global_only`**, so `strip_local_only_settings` drops them from non-global config while **global config and `MISE_*` env vars still apply**. Unit tests assert local stripping vs global preservation.
> 
> The **watch-files shell e2e** no longer sets `unix_default_inline_shell_args` in generated `mise.toml`; it uses **`MISE_UNIX_DEFAULT_INLINE_SHELL_ARGS`** for the fallback-shell case and keeps per-`watch_files` `shell` only in project config.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 67bcea5ff4c637c83a634af70df40e0ee9b00a5d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **Bug Fixes**
  - Restricted default Unix and Windows shell-argument settings to global configuration only.
  - Prevented local configuration from overriding these default shell arguments.
  - Kept the settings intact when loaded from global configuration.
- **Tests**
  - Expanded coverage to verify local vs. global stripping behavior for default shell-argument settings.
- **E2E Tests**
  - Updated watch-file shell argument checks to use an environment-provided default instead of embedding it in generated config.

### DIFF
--- a/e2e/config/test_watch_files_shell
+++ b/e2e/config/test_watch_files_shell
@@ -20,7 +20,7 @@ exec bash -c "$2"
 SCRIPT
 chmod +x "$HOME/watch-explicit-shell"
 
-export MISE_UNIX_DEFAULT_INLINE_SHELL_ARGS="$HOME/watch-fallback-shell -c"
+export MISE_UNIX_DEFAULT_INLINE_SHELL_ARGS="\"$HOME/watch-fallback-shell\" -c"
 
 cat <<EOF >mise.toml
 [[watch_files]]
@@ -30,7 +30,7 @@ run = "echo WATCH_FALLBACK_RUN"
 [[watch_files]]
 patterns = ["explicit.txt"]
 run = "echo WATCH_EXPLICIT_RUN"
-shell = "$HOME/watch-explicit-shell -c"
+shell = '"$HOME/watch-explicit-shell" -c'
 EOF
 
 touch fallback.txt explicit.txt
--- a/e2e/config/test_watch_files_shell
+++ b/e2e/config/test_watch_files_shell
@@ -20,10 +20,9 @@ exec bash -c "$2"
 SCRIPT
 chmod +x "$HOME/watch-explicit-shell"
 
-cat <<EOF >mise.toml
-[settings]
-unix_default_inline_shell_args = "$HOME/watch-fallback-shell -c"
+export MISE_UNIX_DEFAULT_INLINE_SHELL_ARGS="$HOME/watch-fallback-shell -c"
 
+cat <<EOF >mise.toml
 [[watch_files]]
 patterns = ["fallback.txt"]
 run = "echo WATCH_FALLBACK_RUN"
--- a/settings.toml
+++ b/settings.toml
@@ -2921,12 +2921,14 @@ type = "ListPath"
 default = "sh"
 description = "Default shell arguments for Unix to be used for file commands. For example, `sh` for sh."
 env = "MISE_UNIX_DEFAULT_FILE_SHELL_ARGS"
+global_only = true
 type = "String"
 
 [unix_default_inline_shell_args]
 default = "sh -c -o errexit"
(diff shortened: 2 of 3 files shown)


sargunv:
With my large cross-platform monorepo, I used this option to standardize the shell to `bash` on all platforms, specifically Git Bash on Windows:

https://github.com/maplibre/maplibre-native-ffi/blob/09cb6f503461d168fcd1294f080b5edc9ab119df/mise.toml#L22-L25

This is a pretty significant breaking change; is there a new recommended way to standardize the shell for tasks across a project?

Global only doesn't make a lot of sense to me for this setting; if I set a particular shell in my global mise settings, any mise-managed project that expects the default shell will break.

> Settings from local project configuration are preloaded before trust evaluation. 

The security angle here doesn't make sense to me either; if I `mise trust` a project, and am running `mise run` tasks within it, of course that project can steer what code gets run. 

What am I missing?

jdx:
You can see the vulnerability here: https://github.com/jdx/mise/security/advisories/GHSA-g74g-rg72-j2p3
I’m open to ideas for working around it but the problem is you don’t need to trust the config.

sargunv:
ah yeah, looks like the core issue is it was applied _before_ trusting the repo, and applied even to global tasks outside the project that defined the setting

I'd recommend, first, highlight the breaking change in the 2026.7.14 release notes.

Then, one of these two:

Partially restore original behavior:

- follow up with restoring the setting but only load it in trusted projects, documenting which places it doesn't apply (where evaluation happens before trust, like tera templates exec I think?)
- ideally, scope the setting to the project it's in, not applying it outside (as that'll likely just break global commands if say, a project chooses an incompatible shell from a global task)
- and when the setting is set globally it should probably only apply to global configs

or, if you'd like to take this breaking change as an opportunity to redesign the setting, I'd suggest:

- recommend affected users migrate to task templates
- that task template can set `shell = `
- other call sites the user would like to go through that shell (like postinstall, custom deps) can defer to a task to go through the template
- add a new setting for "default task template" (or a special named task template like "default") so projects with many tasks and config roots, that want to just default to bash, don't need to remember to set it everywhere

jdx:
@sargunv thanks for the thoughtful suggestion. I agree that restoring project settings only after trust evaluation is not a clean fit for the current settings architecture, and trust alone would not prevent a project shell default from affecting unrelated global tasks, hooks, or tool installation.

I opened #11354 with a task-scoped alternative: `task_config.shell`. It provides a default shell for inline and included TOML tasks owned by the same config root, while task-local and task-template `shell` values take precedence. Child config roots do not inherit it, so the default does not escape the project scope that defined it.

The PR also adds a prominent migration warning to the 2026.7.14 release notes.

PR #11297, opened 2026-07-25
perf(cli): stop rebuilding the clap tree twice on every invocation

jdx:
## Problem

Every `mise` command builds the clap command tree **three times**. One is needed
for argument parsing; the other two come from `env.rs`.

[#11190](https://github.com/jdx/mise/pull/11190) replaced a string scan in
`env.rs` with a call through `crate::cli::first_non_global_arg_idx`, which takes
a `&clap::Command`:

```rust
fn first_non_global_arg_idx(args: &[String]) -> Option<usize> {
    crate::cli::first_non_global_arg_idx(
        &<crate::cli::Cli as clap::CommandFactory>::command(),  // ← builds the whole tree
        args,
    )
}
```

It is called twice per invocation — once directly in `prefer_offline`, once via
`prefer_offline_command` → `command_arg`. Both run from `Lazy` statics during
startup, so essentially every command pays.

Callgrind on 2026.7.13 shows the extra pair:

```
2x  mise::env::PREFER_OFFLINE::{closure#0}  ->  Cli::augment_args
1x  mise::cli::Cli::run::{closure#0}        ->  CommandFactory::command -> augment_args
```

One tree construction measures ~3.1M instructions, so this cost ~6.3M on every
command. Found by bisecting instruction counts between v2026.7.0 and HEAD; the
direct parent→child A/B across #11190 confirms it in isolation.

## Fix

`first_non_global_arg_idx` only ever consults the top-level flags that take a
following value. That list is now a `const`, and `env.rs` uses it instead of
building anything:

```rust
pub(crate) const GLOBAL_FLAGS_WITH_VALUES: &[&str] = &[
    "--cd", "-C", "--env", "-E", "--jobs", "-j", "--profile", "-P",
    "--shell", "-s", "--tool", "-t", "--log-level", "--output",
];
```

Hardcoding it would normally be a drift hazard, which is what #11190 was fixing,
so `test_global_flags_with_values_matches_clap` asserts the const equals what
clap reports. Adding a value-taking flag to `Cli` without updating the list
fails CI rather than silently mis-parsing arguments.

Callers that already have a `Command` (argument parsing) keep using its real flag
set — the two paths share the scan via a predicate, so behaviour is identical.

## Effect

Instruction counts via cachegrind, release feature set
(`--no-default-features --features "rustls-native-roots,self_update,vfox/vendored-lua,openssl/vendored"`),
measured in a Debian container with a warmed cache:

| command | before | after | | pre-#11190 |
|---|---:|---:|---|---:|
| `ls` | 39,451,978 | **33,165,753** | −15.9% | 33,212,050 |
| `env` | 39,121,809 | **32,821,732** | −16.1% | 32,840,298 |
| `hook-env -s bash` | 30,800,771 | **24,587,056** | −20.2% | 24,536,438 |
| `--version` | 10,493,720 | 10,461,486 | −0.3% | 10,463,168 |

The right-hand column is `d34aa29b4`, the commit before #11190 landed. The fixed
numbers land within 0.2% of it, so the regression is fully recovered rather than
merely reduced.

## Testing

- **1917 unit tests pass** (1916 + the new drift guard)
- `cargo fmt --check` and `cargo clippy` clean for both changed files
- `prefer_offline` keeps its original shape, so the existing
  `test_prefer_offline_command_skips_global_option_values` and
  `test_remote_fetch_command_skips_global_option_values` cover the behaviour
  unchanged

---

---

> [!NOTE]
> **Low Risk**
> Performance refactor with behavior preserved via shared scan logic and a clap parity test; main risk is stale `GLOBAL_FLAGS_WITH_VALUES` if someone bypasses CI.
> 
> **Overview**
> **Startup no longer builds the full clap command tree twice** when `env.rs` resolves the subcommand for `PREFER_OFFLINE` and `REMOTE_FETCH_COMMAND`. Those paths now use a hardcoded `GLOBAL_FLAGS_WITH_VALUES` list and `first_non_global_arg_idx_cached`, avoiding ~6.3M instructions per run that came from constructing `Cli::command()` inside `Lazy` statics.
> 
> The scan logic is shared via `first_non_global_arg_idx_with`; callers that already have a `&Command` still derive flags from clap. **`test_global_flags_with_values_matches_clap`** keeps the const in sync with `Cli` so new value-taking global flags fail CI instead of mis-parsing argv.
> 
> `env.rs` also pulls prefer-offline and remote-fetch subcommand names into **`PREFER_OFFLINE_COMMANDS`** / **`REMOTE_FETCH_COMMANDS`** and routes command detection through **`is_command`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bf88a00ccbef6de1457baaecdf6240f2bfb1bcae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup handling of global command-line options, including options that require values.
  * Improved detection of subcommands when global options appear before the command.
  * Ensured offline and remote version-fetch behavior is applied consistently across supported commands.

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -419,8 +419,55 @@ fn escape_args_after_separator(args: &[String], separator_idx: usize) -> Vec<Str
     result
 }
 
+/// Long and short forms of the top-level flags that consume a following argument.
+///
+/// Hardcoded rather than derived because `env.rs` needs it from `Lazy` statics
+/// during startup — before anything has parsed arguments — and deriving it means
+/// building the entire clap tree, which costs ~3.1M instructions. Doing that
+/// there is what made every mise command ~6.3M instructions more expensive.
+///
+/// `test_global_flags_with_values_matches_clap` asserts this equals what clap
+/// reports, so adding a value-taking flag to [`Cli`] without updating this list
+/// fails CI rather than silently mis-parsing arguments.
+pub(crate) const GLOBAL_FLAGS_WITH_VALUES: &[&str] = &[
+    "--cd",
+    "-C",
+    "--env",
+    "-E",
+    "--jobs",
+    "-j",
+    "--profile",
+    "-P",
+    "--shell",
+    "-s",
+    "--tool",
+    "-t",
+    "--log-level",
+    "--output",
+];
+
+/// Index of the first argument that is not a global flag or one of its values.
+///
+/// Takes a `&Command` so callers that have already built one (argument parsing)
+/// use its real flag set. Callers without one want
+/// [`first_non_global_arg_idx_cached`].
 pub(crate) fn first_non_global_arg_idx(cmd: &clap::Command, args: &[String]) -> Option<usize> {
-    let (flags_with_values, _) = get_global_flags(cmd);
+    let flags = get_global_flags(cmd).0;
+    first_non_global_arg_idx_with(|f| flags.iter().any(|x| x == f), args)
+}
+
+/// As [`first_non_global_arg_idx`], against [`GLOBAL_FLAGS_WITH_VALUES`].
+///
+/// For callers with no `Command` to hand, which would otherwise build the whole
+/// tree just to read its top-level arguments.
+pub(crate) fn first_non_global_arg_idx_cached(args: &[String]) -> Option<usize> {
+    first_non_global_arg_idx_with(|f| GLOBAL_FLAGS_WITH_VALUES.contains(&f), args)
+}
+
+fn first_non_global_arg_idx_with(
+    takes_value: impl Fn(&str) -> bool,
+    args: &[String],
+) -> Option<usize> {
     let mut i = 1;
     while i < args.len() {
         let arg = &args[i];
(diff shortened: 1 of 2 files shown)


jdx:
Additional data: the numbers in the description are from the `release` profile. Repeating on `serious` — the profile releases are actually built with (`scripts/build-tarball.sh` uses `--profile=serious`, plus PGO) — shows the regression was larger there, and hit `mise --version` too:

`mise --version`, warm cache, cachegrind, `serious` profile:

| build | Ir |
|---|---:|
| v2026.7.0 | 12,459,066 |
| 2026.7.13 (before this PR) | 15,956,465 |
| **this PR** | **10,211,865** |

That is **+28% introduced** and **−36% recovered**, ending 18% below v2026.7.0.

Worth noting because the two profiles disagree about `--version`: on `release` it is flat across #11190, on `serious` it pays a full extra tree build. Codegen differences change which `Lazy` statics get initialised on that path, so the LTO build touches `PREFER_OFFLINE` where the non-LTO one does not. The shipped binaries are the LTO ones, so this was user-visible.

Full picture, both profiles, all versus v2026.7.0:

| profile | command | v2026.7.0 | main | this PR |
|---|---|---:|---:|---:|
| release | `ls` | 35,615,272 | 39,451,978 | 33,165,753 |
| release | `env` | 35,355,001 | 39,121,809 | 32,821,732 |
| release | `hook-env` | 28,241,695 | 30,800,771 | 24,587,056 |
| release | `--version` | 13,583,409 | 10,493,720 | 10,461,486 |
| serious | `--version` | 12,459,066 | 15,956,465 | 10,211,865 |

Every command ends up at or below its v2026.7.0 cost.

---

PR #11299, opened 2026-07-25
fix(swift): record the target distro in lockfile entries

jdx:
Reported in https://github.com/jdx/mise/discussions/11295.

## The problem

Swift publishes a different Linux tarball per distro (`ubuntu24.04`, `fedora39`, `ubi9`, `amazonlinux2`, …), but nothing about the distro reached the lockfile — entries were keyed by a bare `linux-x64` with no options. Two bugs followed:

1. **Cross-distro checksum failures.** An entry written on Ubuntu matched exactly on Fedora (same key, both option-free), so the Ubuntu checksum was verified against the Fedora tarball. The error named the downloaded filename and two digests with nothing to suggest a distro mismatch.

2. **`mise lock` reported success while writing nothing.** `core:swift` implements no `resolve_lock_info`, so it fell through to `resolve_lock_info_fallback`, which returns an all-`None` `PlatformInfo`. That's `Ok`, so the run counted it as an update, but `set_platform_info` drops empty info — `✓ Updated 2 platform entries (0 skipped)` over an untouched lockfile. With an entry already present, the empty info merged as \"keep everything\", silently preserving the previous distro's checksum. Changing `swift.platform` therefore left a project where `mise lock` wouldn't re-lock and `mise install` rejected the artifact the setting selected.

## The fix

`core:erlang` already had this problem — its precompiled builds are per-distro — and solved it with a `precompiled_os` lockfile option. Swift now does the same:

- **`swift_platform` option** on `core:swift` lock entries. Entries are matched on options exactly, so a Fedora machine no longer matches the Ubuntu entry, and changing `swift.platform` produces a new entry instead of reusing a checksum for a different artifact.
- **`resolve_lock_info`** records the URL each entry describes, so `mise lock` writes something meaningful. swift.org publishes no checksum sidecar (only a detached `.sig`), so a checksum still can't be resolved without downloading the ~1 GB toolchain — it's filled in at install. Locking a musl target now errors rather than recording a URL that doesn't exist.
- **Target-aware URL construction**, so cross-platform locking stops labeling another machine's platform with the host's distro. A non-current target falls back to Ubuntu, the only distro with Linux arm64 builds.
- **Legacy entries** (written before the distro was recorded) keep their version pin but have their artifact data ignored, so upgrading neither drops an existing pin nor asserts a checksum for an unknown distro. They're also excluded from option rekeying, which would otherwise stamp the local distro's name onto a foreign artifact's checksum.
- **`mise lock` accounting**: a resolution that wrote no entry counts as skipped, not as an update. This applies to any backend that returns the empty fallback.
- **Checksum mismatch errors** name the lock entry the expectation came from — platform key, options, and locked URL:

  ```
  Failed to install aqua:mikefarah/yq@4.44.3: lockfile entry for aqua:mikefarah/yq@4.44.3 on linux-x64 locks https://github.com/mikefarah/yq/releases/download/v4.44.3/yq_linux_amd64: Checksum mismatch for file …
  ```

## Before / after

The discussion's second repro (`swift.platform = "ubi9"`, `mise lock swift --platform linux-x64,linux-arm64`) previously produced a lockfile with no platform entries at all. Now:

```toml
[[tools.swift]]
version = "6.3.1"
backend = "core:swift"

[tools.swift.options]
swift_platform = "ubi9"

[tools.swift."platforms.linux-arm64"]
url = "https://download.swift.org/swift-6.3.1-release/ubi9/swift-6.3.1-RELEASE/swift-6.3.1-RELEASE-ubi9-aarch64.tar.gz"

[tools.swift."platforms.linux-x64"]
url = "https://download.swift.org/swift-6.3.1-release/ubi9/swift-6.3.1-RELEASE/swift-6.3.1-RELEASE-ubi9.tar.gz"
```

Starting from an Ubuntu-written lockfile and pinning `ubi9`, `mise lock` now creates a separate `ubi9` entry with the right URL and leaves the old entry untouched, instead of silently keeping the `ubuntu24.04` checksum.

## Testing

- 6 new unit tests in `src/plugins/core/swift.rs` (distro options per target, empty options off Linux, URL resolution for a foreign target, musl rejection, Ubuntu fallback).
- New `e2e/lockfile/test_lockfile_swift_distro` — covers the distro option, re-locking after changing `swift.platform`, and a legacy entry keeping its version pin. Locking swift resolves URLs only, so the test never downloads the toolchain.
- Full unit suite (1922 tests) and all 44 non-slow `e2e/lockfile` tests pass; `mise run lint` clean.

## Note

Migration is one-directional but recoverable: an existing option-free swift entry keeps pinning its version, and the first install on each machine writes a distro-labeled entry with the correct checksum. The old entry is preserved rather than rewritten, since there's no way to tell which distro it came from.

---

> [!NOTE]
> **Medium Risk**
> Changes lockfile matching, install checksum verification, and Swift download URL resolution across platforms; behavior is well-tested but affects reproducible installs for Swift on Linux.
> 
> **Overview**
> Fixes Swift lockfile behavior on Linux where every distro shared the same `linux-x64` key, so checksums from one distro were verified against another distro’s tarball.
> 
> **Swift lock entries** now carry a `swift_platform` option (e.g. `ubuntu24.04`, `fedora39`, `ubi9`) and are matched on options exactly, so each distro gets its own entry when `swift.platform` changes. **`resolve_lock_info`** records distro-specific download URLs (HEAD-checked; missing pairs like ubi9 arm64 are skipped; musl fails instead of locking a bogus URL). URL building is **target-aware** for cross-platform `mise lock`, with a Ubuntu fallback for foreign Linux targets; `swift.platform` no longer leaks into macOS/Windows URLs.
> 
> **Legacy option-free Swift entries** still pin the version but drop checksum/URL via `without_artifact_data()` and skip legacy option rekeying (`lockfile_options_are_host_specific`). **`mise lock`** counts only entries that actually wrote data (`apply_lock_result` returns `bool`); checksum errors name the lock entry (options + URL).
> 
> Docs and an e2e test cover multi-entry lockfiles and distro changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5e04e13e7f2ab2f3c983b5d6e5a9c5f682e52f49. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Swift lockfiles now distinguish Linux distributions, selecting the correct platform-specific artifact when switching between distributions.
  * Existing lockfiles can be upgraded while preserving pinned Swift versions and refreshing distribution-specific artifact data.
  * Lock entries now provide clearer artifact identity and checksum-related error details.

* **Documentation**
  * Added guidance and examples for tools with multiple lockfile entries for the same version but different artifact options.

* **Bug Fixes**
  * Improved lock progress reporting to reflect entries actually written rather than attempted operations.

### DIFF
--- a/docs/dev-tools/mise-lock.md
+++ b/docs/dev-tools/mise-lock.md
@@ -82,6 +82,26 @@ Each tool entry (`[[tools.name]]`) can contain:
 - **`options`** (optional): Backend-specific options that identify the artifact (e.g., `{exe = "rg", matching = "musl"}`)
 - **`platforms`** (optional): Platform-specific metadata (checksums, URLs, sizes)
 
+A tool can have several entries for the same version when its artifact identity
+depends on more than the platform key. Swift, for example, publishes a different
+Linux tarball per distro, so its entries record which one they describe:
+
+```toml
+[[tools.swift]]
+version = "6.3.1"
+backend = "core:swift"
+options = { swift_platform = "ubuntu24.04" }
+
+[[tools.swift]]
+version = "6.3.1"
+backend = "core:swift"
+options = { swift_platform = "fedora39" }
+```
+
+Entries are matched on options exactly, so a machine only verifies against the
+entry written for its own distro. Pin `swift.platform` to make every machine
+resolve the same artifact, and commit the entry it produces.
+
 ### Platform Keys
 
 The platform key format is generally `os-arch` but can be customized by backends:
--- a/docs/dev-tools/mise-lock.md
+++ b/docs/dev-tools/mise-lock.md
@@ -99,8 +99,10 @@ options = { swift_platform = "fedora39" }
 ```
 
 Entries are matched on options exactly, so a machine only verifies against the
-entry written for its own distro. Pin `swift.platform` to make every machine
-resolve the same artifact, and commit the entry it produces.
+entry written for its own distro. Pin `swift.platform` to make every Linux
+machine resolve the same artifact, and commit the entry it produces. A platform
+whose artifact the tool doesn't publish — `ubi9` has no arm64 build, for
+instance — is reported as skipped rather than locked.
 
 ### Platform Keys
 
--- a/e2e/lockfile/test_lockfile_swift_distro
+++ b/e2e/lockfile/test_lockfile_swift_distro
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+
+# Regression test for https://github.com/jdx/mise/discussions/11295
+# Swift's Linux tarballs differ per distro (ubuntu24.04, fedora39, ubi9, …) but
+# the lock platform key is just linux-x64, so an entry written on one distro used
+# to be matched on another and its checksum verified against a different
+# tarball. Entries now record the distro in their options.
+#
+# Locking swift resolves URLs only — it never downloads the ~1GB toolchain.
+
+if [[ "$(uname -s)" != Linux ]]; then
+  echo "skipping: swift artifacts are only distro-specific on linux"
+  exit 0
+fi
+
+export MISE_LOCKFILE=1
+
+write_config() {
+  cat <<EOF >mise.toml
+[settings]
+swift.platform = "$1"
+[tools]
+"core:swift" = "6.3.1"
+EOF
+}
+
+write_config "ubi9"
+mise lock swift --platform linux-x64
+assert_contains "cat mise.lock" 'swift_platform = "ubi9"'
+assert_contains "cat mise.lock" 'swift-6.3.1-RELEASE-ubi9.tar.gz'
+
+# Changing the pinned distro must produce its own entry rather than silently
+# keeping the previous distro's artifact data.
+write_config "ubuntu24.04"
+mise lock swift --platform linux-x64
+assert_contains "cat mise.lock" 'swift_platform = "ubuntu24.04"'
+assert_contains "cat mise.lock" 'swift-6.3.1-RELEASE-ubuntu24.04.tar.gz'
+
+# An entry written before the distro was recorded still pins the version, but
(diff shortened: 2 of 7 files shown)


jdx:
Addressed the review feedback in 5e04e13 — all three bots flagged the same real bug, thanks.

**`swift.platform` leaking into non-Linux targets.** Correct, and it was worse than cosmetic: because the distro option is only recorded for Linux, a bogus `.../ubi9/swift-6.3.1-RELEASE-ubi9.pkg` would land in an option-free macOS entry that every macOS host matches. `platform()` now inspects the target OS first and only consults the setting for Linux targets — the same shape as CodeRabbit's suggested diff. Covered by a new `pinned_distro_does_not_apply_off_linux` unit test.

**Distro/arch pairs that aren't published** (Greptile's second half — `ubi9` has no aarch64 build). Which pairs swift.org publishes shifts per release, so rather than encode a matrix that would go stale, lock resolution now HEADs the artifact and reports the platform as skipped if it isn't there. Verified against the real service: a followed HEAD returns 404 for `swift-6.3.1-RELEASE-ubi9-aarch64.tar.gz` and 200 for the x64 build.

The discussion's repro with `lockfile_platforms = ["linux-x64", "linux-arm64"]` and `swift.platform = "ubi9"` now reports `✓ Updated 1 platform entries (1 skipped)` and records only the x64 URL, instead of writing an arm64 URL that doesn't resolve. Also asserted in the e2e test.

I looked at recording `size` from the HEAD while I was there, but `PlatformInfo::size` is `#[serde(skip_serializing)]` by design ("preserved from existing lockfiles but not written"), so the HEAD is used purely as an existence check.

Unit suite 1923 passing, `e2e/lockfile` green, `mise run lint` clean.

PR #11305, opened 2026-07-25
fix(npm): add allow_low_downloads backend option

jdx:
aube refuses a package whose weekly download count falls below `lowDownloadThreshold` (1000 by default), and mise currently has **no way to opt out** — `AddToProjectOptions` exposes no `allow_low_downloads` field, so the embedded installer always hits the gate. Two curated registry tools are uninstallable as a result:

```
refusing to add bibtex-tidy: only 930 weekly downloads (threshold: 1000)
refusing to add purty: only 845 weekly downloads (threshold: 1000)
```

Users have no workaround short of `npm.shell_out=true`.

## Approach

New `allow_low_downloads` option, written to the aube install `.npmrc` as `allowedUnpopularPackages=<package>` — naming **only the requested package**:

```toml
[tools]
"npm:some-tool" = { version = "latest", allow_low_downloads = true }
```

I deliberately did not set `lowDownloadThreshold = 0`, which is the other thing aube's error message suggests. That disables the gate for the whole install, so an unpopular *transitive* dependency would sail through too. Scoping to the named package keeps the threshold intact for everything the user didn't explicitly ask for. aube gates the directly-requested set, which for mise is always exactly one package.

Enabled on the two affected registry entries. For a registry tool the entry is itself the popularity decision, which is what the download gate approximates.

## Worth a second opinion

Both are under mise's own registry bar, and `purty`'s last release was **2021-03-02** — over five years ago:

| | weekly downloads | latest release |
|---|---|---|
| bibtex-tidy | 930 | 2024-09-07 |
| purty | 845 | 2021-03-02 |

Deleting the entries is a defensible alternative, `purty` especially. I kept them because removal breaks anyone using the shorthand today, and the option is worth having regardless — but the call is yours, and I'm happy to drop either entry from this PR.

## Verification

- `cargo test --all-features npm` — 59 passed, 0 failed, incl. 3 new tests covering the scoped `.npmrc` value, the default-off path, and non-boolean rejection
- `mise install bibtex-tidy@latest` → installed
- `mise install purty@latest` → installed

---

> [!NOTE]
> **Low Risk**
> Changes are limited to aube embed install `.npmrc` generation and opt-in registry entries; supply-chain gates remain for transitive dependencies unless explicitly exempted elsewhere.
> 
> **Overview**
> Adds **`allow_low_downloads`** so embedded **aube** installs can bypass the weekly-download gate for the **requested** npm package only, instead of failing when popularity is below `lowDownloadThreshold` (default 1000).
> 
> When enabled, mise validates the option up front, then writes `allowedUnpopularPackages=<tool>` into the synthetic install `.npmrc` (not a global threshold change), so transitive deps still hit the gate. The option is documented alongside other aube-only tool options and registered as install-time metadata.
> 
> **`bibtex-tidy`** and **`purty`** registry backends are updated to set `allow_low_downloads = true` so curated shorthand installs work again without `npm.shell_out`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 89d7558938c3ada3f74e1cda415af499caa38d80. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added an `allow_low_downloads` option for npm-based installations, allowing specified packages below the weekly download threshold to install.
  * Kept the exemption limited to the requested package; transitive dependencies remain subject to existing checks.
  * Added registry support for enabling the option on selected packages.

* **Documentation**
  * Documented configuration, defaults, scope, and supported package managers for the new option.

* **Bug Fixes**
  * Added validation for accepted option values and ensured the allowance is omitted when disabled.

### DIFF
--- a/docs/dev-tools/backends/npm.md
+++ b/docs/dev-tools/backends/npm.md
@@ -296,6 +296,28 @@ To exempt only selected versions, use aube's package-version pattern syntax:
 `trust_policy_excludes` is written to the aube install `.npmrc` as `trustPolicyExclude`. It does
 not affect `npm`, `pnpm`, or `bun` installs.
 
+### `allow_low_downloads`
+
+Allows the requested package to install even though its weekly download count falls below aube's
+`lowDownloadThreshold` (1000 by default). Without it, aube refuses:
+
+```
+refusing to add some-tool: only 930 weekly downloads (threshold: 1000).
+```
+
+```toml
+[tools]
+"npm:some-tool" = { version = "latest", allow_low_downloads = true }
+```
+
+The exemption is scoped to the package you asked for, written to the aube install `.npmrc` as
+`allowedUnpopularPackages=<package>`. Transitive dependencies stay gated, and the threshold itself
+is left alone — so this cannot silently admit an unpopular dependency you did not choose.
+
+Download count is a popularity signal, not a safety one: a low count means few others have vetted
+the package, so prefer confirming you trust the publisher over reaching for this. It does not affect
+`npm`, `pnpm`, or `bun` installs.
+
 ### Investigating trust downgrades
 
 A `trustPolicy=no-downgrade` failure is a supply-chain signal, not an ordinary inability to find a
--- a/registry/bibtex-tidy.toml
+++ b/registry/bibtex-tidy.toml
@@ -1,3 +1,10 @@
-backends = ["npm:bibtex-tidy"]
 description = "Cleaner and Formatter for BibTeX files"
 test = { cmd = "bibtex-tidy --version", expected = "v{{version}}" }
+
+[[backends]]
+full = "npm:bibtex-tidy"
+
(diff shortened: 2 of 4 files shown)


jdx:
> Pre-validate `options.allow_low_downloads()` at the start of `write_aube_embed_project()`, before creating or writing `package.json`

Good catch — fixed in 89d7558. Both fallible options are now resolved before any write, so a malformed value fails cleanly instead of leaving a `package.json` with no `.npmrc` beside it.

I included `trust_policy_excludes` in the same move: it had the identical ordering problem already, and leaving one of the two validated late would just invite the bug back. Added `test_write_aube_embed_project_writes_nothing_when_an_option_is_malformed` to pin it — 60 npm tests passing.

PR #11306, opened 2026-07-25
feat(usage): declare what each command does to the world

jdx:
usage 3.6 shipped `effect=` (jdx/usage#739), which lets a spec say whether a command only inspects state, changes it, or destroys something. This declares it for all **172** mise commands.

```
- **Usage**: `mise uninstall [FLAGS] [INSTALLED_TOOL@VERSION]…`
- **Effect**: destructive — may delete or irreversibly overwrite
```

**89 read · 71 write · 7 destructive · 10 deliberately unclassified**

The seven destructive: `implode`, `prune`, `uninstall`, `unuse`, `deps remove`, `plugins uninstall`, `bootstrap packages prune`.

## Why one table

clap has no way to express this, so the classification lives in `src/cli/command_effects.rs` and is applied to the derived spec. That's also the right shape for review — a safety classification is far easier to audit as a single ordered list than as annotations spread across sixty command files.

## Unclassified is a real answer

Ten commands run code the user supplies — `run`, `exec`, `watch`, `mcp`, `en`, `tasks run`, `test-tool`, `tool-stub`, `oci run`, `direnv exec`, `asdf`. Their effect is whatever that code does. An unset effect means "unknown, ask", which is both honest and the safe default; labelling them `read` would be actively dangerous. Each carries its reason inline.

## Tests keep it honest

Three of them, in the same spirit as the round-trip guard in jdx/usage#740:

1. every command is classified,
2. no entry refers to a command that no longer exists,
3. nothing is classified twice.

The coverage test walks **hidden** commands too, which caught a real mistake: I initially skipped them, and `bootstrap systemd apply`, `bootstrap macos-defaults apply`, `bootstrap launchd apply`, `mise global`, `mise local` and `hook-not-found` are all hidden yet all change state. `hook-not-found` in particular installs a tool when `not_found_auto_install` is set.

`bootstrap packages brew` is `#[cfg(unix)]`, so its entries sit in a `cfg`-gated table — otherwise the stale-entry test fails on Windows.

## Two changes this depends on

- **`min_usage_version` 3.5 → 3.6.** Not optional: `usage` 3.5.x rejects the spec outright with `unsupported cmd prop effect`.
- **`is starting_with(...)` in `mise-extra.usage.kdl` now passes its argument by name.**

That second one is worth flagging on its own. **mise's docs render is already broken on usage >= 3.5.7, before this PR.** usage picked up tera 2.0 → 2.1 in v3.5.7, and tera 2.1 rejects positional test arguments:

```
× error: Found string but expected identifier.
  --> __tera_one_off:14:32
   | {%- elif path is starting_with("bootstrap/packages/brew") -%}
```

I confirmed this by rendering the *unmodified* committed spec with a 3.5.7 build. The two spellings are mutually exclusive — `starting_with("x")` works only on <= 3.5.6, `starting_with(pat="x")` only on >= 3.5.7 — so there is no template that satisfies both. Requiring 3.6 here makes the named form unambiguously correct for mise, but any other project with `is <test>(...)` in a usage template broke silently in that patch release.

## Verified

- `cargo build --all-features`, `cargo clippy --all-features --all-targets`, `cargo fmt --check` all clean
- the three coverage tests pass
- full `render` pipeline run end to end against usage 3.6.0: `mise.usage.kdl`, `docs/cli/**`, man page, four completions, `llms.txt`

---

> [!NOTE]
> **Low Risk**
> Mostly generated docs, lockfile bumps, and completion cache paths; behavioral risk is low aside from requiring usage 4.x for spec rendering and completions.
> 
> **Overview**
> This PR wires **usage 4.0** (`usage-lib` 3 → 4) so every command in the generated spec carries an **`effect`** (`read`, `write`, or `destructive`), driven from a central classification (e.g. `command_effects`) applied when emitting `mise usage`. **`min_usage_version`** is raised to **4.0** because older `usage` CLIs reject `effect=` on commands.
> 
> Regenerated artifacts pick up the change: **`mise.usage.kdl`**, all **`docs/cli/**`** pages now show a **`- **Effect**:`** line (read-only / modifies state / destructive), plus small doc fixes (heredoc `<<EOF`, Tera `starting_with(pat="...")` for usage’s template parser). **`mise.lock`** pins **usage 4.0.0** with updated download URLs and checksums.
> 
> **Shell completions** (bash, fish, zsh) no longer stash the usage spec in **`$TMPDIR`**; they use **`$XDG_CACHE_HOME/usage`** (mode **700**), prune specs older than 30 days when refreshing, then run **`mise usage`** into a versioned spec file.
> 
> E2E **`test_task_completion`** documents the usage ≥ 4 requirement and sets **`MISE_MINIMUM_RELEASE_AGE=0s`** so `usage@latest` resolves during tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bc362701e57b31287aad53a0334a59d2b869db24. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Shell completions (Bash/Fish) now generate and reuse the required usage specification from a persistent per-user cache (with restrictive permissions) and automatically prune cached entries older than 30 days.

- **Documentation**
  - Command reference pages now consistently annotate each command’s effect (read-only, modifies state, or destructive).
  - Fixed rendered heredoc examples and improved source-link routing for some bootstrap-related commands.

- **Compatibility**
  - The generated usage specification now requires version **4.0**.


jdx:
All three findings were real, and all three were wrong in the dangerous direction — claiming read-only for something that isn't. Fixed.

**`mise deps` mislabeled read** — correct. `self.command.unwrap_or(Commands::Install(self.install))`, so a bare `mise deps` runs install steps; its own help even says `# Install all project dependencies`. Now `write`.

I audited the other seven parents that default to a subcommand — `backends`, `config`, `direnv`, `plugins`, `shell-alias`, `tasks`, `tool-alias` — and they all default to `Ls` or `Activate`, so `deps` was the only one wrong.

**`mise settings` write path** — also correct. `Settings::run` routes `SETTING` + `VALUE` and `KEY=VALUE` to `SettingsSet`, writing config without touching the `settings set` subcommand. Now `write`.

Worth being clear that this one is a *floor*, not the truth: a bare `mise settings` only lists. Since `effect` is per-command, `write` is the honest conservative answer — it keeps the command out of a read-only allowlist, which is the distinction that matters. jdx/usage#742 (in usage 4.0) makes args able to raise the effect, so this becomes `read` on the command plus `write` on the value arg. I'll make it precise in the usage-4 pass.

**Debug-only `render-help`** — the best of the three. It's `#[cfg(debug_assertions)]`, so it doesn't exist in release builds while the table still claimed it, which would have failed `no_classification_refers_to_a_missing_command` in a release build. My own test would have caught it and I only ever ran it in debug. `PLATFORM_EFFECTS` now gates it alongside the unix-only brew commands, and I ran the coverage tests under `--release` as well as debug to confirm.

Totals are now 87 read / 73 write / 7 destructive.

---

Still to come on this PR, once the usage CLI 4.0.0 is indexed in mise-versions (currently still serving 3.6.0): bump `usage-lib` 3 → 4, migrate the seven struct literals that usage 4's `#[non_exhaustive]` requires, `min_usage_version "4.0"`, `mise.lock` → 4.0.0, and make `settings` precise per above.

jdx:
Declining the `mise env` finding — `env` stays `read`.

The reasoning generalises badly. `activate`, `exec`, `hook-env`, `shell`, `ls` and most of the rest can all install a declared version as a side effect of resolving the toolset, because that is what mise is for. If an incidental install makes a command `write`, then nearly every mise command is `write` and the field carries no signal at all — which defeats the point of adding it.

The line that matters is whether a command changes something **the user manages**: their config, their files, their toolset as a deliberate act. Installing a version the config already asks for is idempotent, mise recreates it on demand, and it changes nothing the user authored. `install` and `use` are `write` because installing is the point, not a precondition.

I have written that into the module docs rather than just replying here, so the next reviewer — human or bot — gets the answer from the source instead of re-raising it.

Two consequences worth flagging, since I would rather be consistent than merely defend the one case:

**`shell` was wrong and is now `read`.** Its only stated reason for `write` was "may install the requested version before emitting the shell code" — the exact reasoning being rejected here. It emits shell code and its session change is ephemeral.

**`hook-not-found` stays `write`, and the comment now says why it differs.** With `not_found_auto_install` set it resolves an *unknown command name* against the registry and installs it, so a typo can put a new tool on disk. That is not a declared version being realised, which is the distinction the principle turns on.

Totals move to 88 read / 72 write / 7 destructive.

jdx:
**Completion cache never refreshes** — declining this one too; the premise does not hold, and the change is not mine.

The cache key embeds the mise version:

```bash
local spec_file="$spec_dir/usage__usage_spec_mise_2026_7_13.spec"
```

So a mise upgrade produces a *different path* and regenerates. An older cached spec cannot be served to a newer binary — it is simply never looked up again. The stated failure mode, "completions can stay out of date even though the installed binary changed", is what the version in the filename exists to prevent.

The `if [[ ! -f "$spec_file" ]]` caching itself is unchanged and pre-existing on main. What this PR changes is only what usage 4.0.0's generator emits: the cache moves from `$TMPDIR` to `$XDG_CACHE_HOME/usage` with mode 700, and a `find -mtime +30 -delete` sweep is added — which is upstream *fixing* stale-file accumulation, the opposite of the concern raised. These lines come from regenerating completions with the newer usage, not from anything I wrote.

**`mise env`** was re-raised on the new commit; already declined above, and the reasoning is now in the module docs so it resolves from the source rather than needing a reply each push.

jdx:
**`direnv envrc` labeled read-only** — accepted, this one is a real miss. Its doc comment says it "writes an envrc file" and it does: `File::create` under `MISE_TMP_DIR/direnv`. Now `write`.

Worth separating it from the `env` finding I declined earlier, since they look similar and resolve differently. The line the module docs draw is whether the effect *is the command's purpose* or machinery on the way to something else. `env` prints, and an install is incidental. `direnv envrc` exists to produce a file — the temp location does not change that.

PR #11316, opened 2026-07-26
fix(task): skip output mtime guard when source_freshness_hash_contents=true

rabadin:
## Summary

\`task.source_freshness_hash_contents\` is documented as "Use content hashing (blake3) instead of metadata for source freshness", but mtime was still used as the final freshness gate even in hash mode. This makes the setting ineffective in CI cache-restore scenarios (e.g. GitHub Actions) where all mtimes are reset to the restore time.

**Root cause:** \`sources_are_fresh\` always fell through to the mtime comparison, even when a matching content-hash baseline existed.

**Fixes:**
- In hash mode, when the source content hash matches the stored baseline, freshness is determined by output hash only — mtime is never consulted.
- The UNIX_EPOCH tarball-mtime guard is also gated on \`!use_content_hash\`.
- **Pre-existing bug:** with multiple outputs, \`get_last_modified(...).is_some()\` returned true as long as any output existed, so a partial output set was treated as fresh. Outputs are now blake3-hashed after each successful run and compared on subsequent checks. Directory outputs are walked recursively — both file contents and subdirectory entries are recorded, so deletions of files or empty nested directories inside a declared output directory are detected.

**Tests:** three new e2e tests (no \`_slow\` suffix, each ≤16s):
- \`test_task_source_freshness_hash_contents\` — mtime changes don't trigger rebuild; content changes do; missing output triggers rebuild
- \`test_task_source_freshness_hash_contents_multi_output\` — partial output set is stale
- \`test_task_source_freshness_hash_contents_dir_output\` — file deleted inside directory is stale; empty nested subdirectory deleted is stale

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task freshness evaluation when content-based source hashing is enabled, avoiding timestamp-based staleness and making freshness driven by output stability.
  * Freshness state now better reflects output completeness, ensuring rebuilds occur when outputs are missing, incomplete, or only partially restored (including directory and multi-output cases).
* **Tests**
  * Added end-to-end coverage for content-hash freshness across single-output, multi-output, and directory-output tasks, including cases with modified mtimes and output deletions/restores.

### DIFF
--- a/e2e/tasks/test_task_source_freshness_hash_contents
+++ b/e2e/tasks/test_task_source_freshness_hash_contents
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+
+# When source_freshness_hash_contents=true, mtime must NOT be the final
+# freshness gate — the content hash alone determines freshness. This fixes
+# CI cache-restore scenarios (e.g. GitHub Actions) where all files get their
+# mtime reset to the restore time, making source mtimes appear newer than
+# output mtimes even though nothing has changed.
+
+cat <<EOF >mise.toml
+[settings.task]
+source_freshness_hash_contents = true
+
+[tasks.build]
+run = 'touch out.txt && echo built'
+sources = ['src.txt']
+outputs = ['out.txt']
+EOF
+
+echo "hello" >src.txt
+
+# First run: no stored hash, no output → stale → runs
+assert "mise run -q build" "built"
+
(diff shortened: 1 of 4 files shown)


rabadin:
@greptile review

rabadin:
Closing in favour of a fresh PR — this one accumulated too many stale automated review comments across iterations.

PR #11371, opened 2026-07-27
fix(brew): maintain linked-keg compatibility records

benjaminwestern:
> I do not normally write Rust. This is largely the work of GPT-5.6 Sol on XHigh, with multiple rounds of my poking to try to align it with the existing architectural patterns in mise and aube.

## Why

I want to use mise for all of my bootstrapping needs, including Homebrew-compatible formulae. I have hit a "disconnected brain" state several times on my machine where a keg and its executable remained present, but either mise's `opt` record or Homebrew's linked-keg record was missing. Mise and Homebrew could then disagree about whether the same formula was linked.

`pkgconf` was used for the end-to-end validation because it was one of the packages that de-linked on my machine.

Mise already writes Homebrew-compatible receipts, but non-keg-only installs did not maintain `<prefix>/var/homebrew/linked/<name>`. Nested relative links could also be misclassified during ownership checks because only their lexical path was considered.

## What Changed

- Non-keg-only bottle and source installs now create Homebrew's linked-keg record within the existing link and rollback flow.
- Configured legacy roots can restore one missing or mise-owned dangling active record without repouring the keg or replacing public links.
- Status reports this narrow condition as `needs repair`.
- New keg-only kegs carry a private marker so public-link inference does not synthesise linked-keg records for them.
- Nested relative links are resolved through their parent before Cellar ownership checks.
- Prune removes a matching linked-keg record with the keg.

## Scope

This change is intentionally limited to active records for configured formulae. Legacy linked-record repair requires existing Homebrew-style public links that match the keg layout, which avoids treating the Linux runtime-loader link as proof that keg-only `glibc` is linked. Foreign records and two valid records that disagree are left unchanged.

During investigation, we prototyped recursive dependency repair, persisted link-mode metadata, unified import inventory, stronger concurrent prune transactions, and broader package reporting changes. Those changes are not required for this compatibility fix, so they are not included here. Existing dependency kegs are not migrated unless they are configured directly or installed again.

## Validation

- `mise run lint` passed.
- `mise run test:unit` passed all 2,029 tests.
- `mise run test:e2e` passed locally. The added Linux bottle and source assertions remain platform-gated for suitable CI runners.
- The focused Homebrew suite passed all 129 tests.
- A disposable `pkgconf` prefix verified initial record creation and both repair directions while preserving the keg inode, receipt metadata, and public-link inode.

## Summary by CodeRabbit

* **New Features**
  * Homebrew installs/upgrades can now repair broken or missing linked-keg records without repouring, including keg-only installs.
  * System status and package JSON output now report a clear **“needs repair”** state.
* **Bug Fixes**
  * Pruning removes obsolete linked-keg records and related symlinks.
  * Repair and linked-record inference now only occur when the existing public link layout matches expected keg structure.
* **Documentation**
  * Updated guidance on coexistence with real Homebrew and refined prune/“pouring” semantics.
* **Tests**
  * Strengthened linked-keg repair, upgrade reporting, and filesystem/symlink assertions (including inode/timestamp stability).

*AI-assisted — Tool: OpenCode; model: openai/gpt-5.6-sol; version: unavailable.*


benjaminwestern:
The review findings are addressed in `8f781876a`:

- Mise-owned dangling records are repairable, while foreign records remain untouched.
- The source E2E assertion now verifies a live symlink target.
- Version mismatch takes precedence over record repair.
- New keg-only kegs use a marker to avoid linked-record inference, while their mandatory `opt` record remains repairable.
- Repair uses the atomic symlink helper.

`mise run lint`, `mise run test:unit` with 2,029 tests, `mise run test:e2e`, and the 129 focused Homebrew tests pass. A fresh disposable `pkgconf` prefix also passed initial linked-record creation and both repair directions without changing the keg, receipt, or public-link metadata.

*AI-assisted — Tool: OpenCode; model: openai/gpt-5.6-sol; version: unavailable.*

benjaminwestern:
The follow-up finding is addressed in `2e06d2174`. `repair_records` now repairs and filters only requests classified as `NeedsRepair`; a mismatched pinned request remains on the normal version-handling path and leaves the record untouched. The manager regression test covers that sequence.

The commit also addresses the all-target Clippy failure by documenting the test's intentional process-global environment lock across async manager calls.

Validation passed with 2,029 unit tests, the full E2E suite, `mise run lint`, and all-feature/all-target Clippy. Local Rust 1.97 required allowing two unrelated lint groups that CI's Rust 1.94 does not emit.

*AI-assisted — Tool: OpenCode; model: openai/gpt-5.6-sol; version: unavailable.*

PR #11378, opened 2026-07-27
registry: add micro (aqua:micro-editor/micro)

Jai-JAP:
Added configuration for the micro text editor with supported platforms and backends.

## Summary by CodeRabbit

* **New Features**
  * Added a registry manifest for the **micro** terminal text editor, including descriptive metadata.
  * Configured a backend entry for the **micro** editor source.
  * Added a version check that runs `micro -version` and verifies the reported `Version` value.

### DIFF
--- a/registry/micro.toml
+++ b/registry/micro.toml
@@ -4,11 +4,26 @@ test = { cmd = "micro -version", expected = "Version: {{version}}" }
 
 [[backends]]
 full = "aqua:micro-editor/micro"
-platforms = ["linux-x64", "linux-arm64", "macos-x64", "macos-arm64", "windows-x64"]
+platforms = [
+  "linux-x64",
+  "linux-arm64",
+  "macos-x64",
+  "macos-arm64",
+  "windows-x64",
+]
 
 [[backends]]
 full = "github:micro-editor/micro"
-platforms = ["linux-x64", "linux-x86", "linux-arm64", "macos-x64", "macos-arm64", "windows-x64", "windows-x86", "windows-arm64"]
+platforms = [
+  "linux-x64",
+  "linux-x86",
+  "linux-arm64",
+  "macos-x64",
+  "macos-arm64",
+  "windows-x64",
+  "windows-x86",
+  "windows-arm64",
+]
 
 [backends.options]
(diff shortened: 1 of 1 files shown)


registry/micro.toml
```diff
@@ -0,0 +1,15 @@
+description = "A modern and intuitive terminal-based text editor"
+os = ["linux", "macos", "windows"]
+test = { cmd = "micro -version", expected = "Version: {{version}}" }
+
+[[backends]]
+full = "aqua:micro-editor/micro"
+platforms = ["linux-x64", "linux-arm64", "macos-x64", "macos-arm64", "windows-x64"]
```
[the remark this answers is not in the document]
Jai-JAP: Micro doesn't provide linux-arm64 builds for 1.4.1 or earlier.

registry/micro.toml
```diff
@@ -0,0 +1,30 @@
+description = "A modern and intuitive terminal-based text editor"
+os = ["linux", "macos", "windows"]
+test = { cmd = "micro -version", expected = "Version: {{version}}" }
+
+[[backends]]
+full = "aqua:micro-editor/micro"
+platforms = [
+  "linux-x64",
```
jdx: why are you defining all of the platforms?

Jai-JAP: Not all platforms are supported yet on the aqua package which need fallback to github, for example windows-x86 and windows-arm64

jdx: remove the platform field, if aqua needs to be fixed that should be done upstream

Jai-JAP: Removed the platform field from both backends

registry/micro.toml
```diff
@@ -0,0 +1,13 @@
...
+test = { cmd = "micro -version", expected = "Version: {{version}}" }
+
+[[backends]]
+full = "aqua:micro-editor/micro"
+
+[[backends]]
+full = "github:micro-editor/micro"
+
+[backends.options]
+asset_pattern = "micro-{{ version }}-{% if os() == 'linux' %}{% if arch() == 'x64' %}linux64{% elif arch() == 'x86' %}linux32{% elif arch() == 'arm64' %}linux-arm64{% endif %}{% elif os() == 'macos' %}{% if arch() == 'x64' %}osx{% elif arch() == 'arm64' %}macos-arm64{% endif %}{% elif os() == 'windows' %}{% if arch() == 'x64' %}win64{% elif arch() == 'x86' %}win32{% elif arch() == 'arm64' %}win-arm64{% endif %}{% endif %}{% if os() == 'windows' %}.zip{% else %}.tar.gz{% endif %}"
```
jdx: is this necessary? I'd rather not have it at all even if it is

registry/micro.toml
```diff
@@ -0,0 +1,13 @@
+description = "A modern and intuitive terminal-based text editor"
+os = ["linux", "macos", "windows"]
```
jdx: we don't do this

registry/micro.toml
Jai-JAP: It falls back to incorrect architectures in certain cases, like linux-x64 resolves to linux32 instead of linux64.

I'll remove it if the asset_pattern is not preferred.

jdx: just don't use github backend, it's not worth it if we need to have something like this

Jai-JAP:
@jdx 
Could you please approve this again. I had to push linter fixes.

Sorry for the noisy history.

PR #11387, opened 2026-07-27
perf: replace the hyperfine job with instruction counting

jdx:
Replaces `hyperfine.yml`. Same shape now running in [aube](https://github.com/jdx/aube), [hk](https://github.com/jdx/hk/pull/1125), [fnox](https://github.com/jdx/fnox/pull/656), [usage](https://github.com/jdx/usage/pull/748) and [pitchfork](https://github.com/jdx/pitchfork/pull/674).

## Why replace it

`hyperfine.yml` gated on wall clock, and most of its logic was about why it could not decide: a 10% threshold, plus separate escape hatches for statistical outliers, for a relative uncertainty overlapping the threshold, and for output it could not parse.

`xtasks/test/perf` says it outright:

> Empirical noise floor on the CI runners is roughly ±10% on the small means we measure here, so a 10% threshold flagged near-every PR. Bump to 15% so a "regression" warning corresponds to something larger than baseline drift.

A threshold raised to 15% *because 10% failed everything* is not catching much.

## What the noise partly was

Three of the four commands hyperfine measured reach the network. Timed against a dead proxy:

| command | offline | |
|---|---:|---|
| `mise --version` | 3681ms | self-update check |
| `mise env` | 4498ms | HTTP warnings |
| `mise ls` | 2931ms | HTTP warnings |
| **`mise registry`** | **15ms** | no network |

No threshold rescues a measurement whose subject phones home.

## The subjects

| bench | what it measures |
|---|---|
| `startup` | `mise --help` — the fixed cost every invocation pays |
| `registry` | `mise registry` — 983 tools, ~75KB, the largest pure-CPU path that needs no network and no installed tools |

`startup` is the control: reading `registry` against it separates "mise got slower to start" from "the registry path got slower". `registry` also grows as tools are added, which makes it an early warning for anything scaling with registry size.

Both checked against a dead proxy, both byte-identical run to run, both run here against a real release build — `registry` 12.46ms, `startup` 10.41ms.

## Why 1% is a real threshold

Instruction counts reproduce to ~0.02% run to run. From aube's first run, on a pull request that changed **one YAML file and no Rust**:

```
instructions   +0.02%  -0.01%  -0.01%  -0.01%
wall clock    -30.21% -19.17% -10.39% -17.89%
```

## Notes

- **`--release`, not `--profile serious`.** LTO roughly triples the build for a number whose absolute value does not matter; what matters is that every measurement in a series is built the same way, and a series is partitioned on runner class regardless.
- **`xtasks/test/perf` is left in place.** Still useful by hand; simply no longer wired into CI. Worth deleting later if nobody runs it.
- **Nothing in the PR workflow writes to `refs/notes/tak`.** Measurements are recorded locally and discarded with the runner; only main contributes to the history.
- The gate has nothing to compare against until `perf.yml` has recorded on main a few times. Until then tak says "nothing was compared, and so nothing was gated" and passes — visible, not a silent green.
- actionlint and zizmor at `--persona=pedantic` are clean.

---

---

> [!NOTE]
> **Medium Risk**
> CI behavior changes (new failing checks, git-notes writes on main) and PR jobs execute untrusted code with valgrind; token scoping and cache isolation are intentional mitigations.
> 
> **Overview**
> **Replaces** the Hyperfine-based `hyperfine.yml` workflow with **tak** (valgrind/cachegrind) instruction-count benchmarking and tighter CI gates.
> 
> **`perf.yml`** records measurements on every **main** push into git notes `refs/notes/tak` via `tak push`, using a dedicated runner cache tag (`mise-perf-main`) so it does not share volumes with PR jobs.
> 
> **`perf-pr.yml`** measures the PR head against the merge base with `tak compare`, posts/updates a sticky PR comment from a separate **report** job (write token only there), and **fails the check** when instruction counts exceed the gate.
> 
> Benchmark wiring adds pinned **`github:jdx/tak`**, **`tak.toml`** benches (`startup`, `ls`, `env`, `hook-env`, `registry`), **`benchmarks/perf-fixture`** for offline deterministic config, and **`perf` / `perf:record`** mise tasks (`cargo build --release`, `MISE_OFFLINE=1`). **`test:perf`** is no longer invoked from CI.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 32a3e8824daa71f0a9aa675142a184facd6ce125. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added PR performance comparisons that enforce an instruction-count regression gate.
  - Added post-merge performance measurement on the main branch with published history and step summaries.
- **Documentation**
  - Documented the new `perf` and `perf:record` tasks.
- **Chores**
  - Pinned the benchmarking tool for more consistent results and introduced a deterministic offline benchmark fixture.
  - Removed the older Hyperfine-based benchmarking workflow and updated runner cache labeling to avoid conflicts.


jdx:
Fixed, plus two failures of my own that the reviews did not catch but CI did.

**The gate reported a regression that was never measured.** A missing status file was packaged as `1`, so `Fail on a regression` said "an instruction count rose beyond the gate" for a run where nothing was compared. fnox demonstrated it exactly: the build failed for a missing `libudev`, and the check blamed a regression. Missing now packages as `2` and the two get different messages.

**fnox: I broke my own libudev fix.** I added it to both workflows, then overwrote `perf-pr.yml` with a template that predated the fix and only re-committed `perf.yml` — so the dependency was present in the workflow that did not need it yet and absent from the one failing. Restored.

**hk: `taplo` formatting.** hk aligns its TOML and the lint job rejected the unaligned `tak.toml`. Ran `taplo format`.

## Not a bug: "baseline notes are never fetched"

`tak compare` fetches them itself — `notes::read` calls `fetch(remote)` before reading, and `cmd_compare` passes `Some(remote)` for the base. The workflow deliberately adds no notes refspec; hiding that seam is what tak is for.

Confirmed in production rather than by reading. aube's run produced real base numbers and ten-point sparklines, which cannot happen without the baseline.

## Not fixable: "artifact status bypasses the gate"

Correct as stated, and I do not think it can be closed. The measuring job runs code from the pull request, so anything it produces — the status file, the report, the numbers themselves — is under that code's control. Moving the verdict into the trusted job does not help, because the trusted job would still be deriving it from output the untrusted job produced.

This is the same boundary as the backfill validator in aube: **a subject that lies about its own performance cannot be caught, because the subject is the measurement.** What the job split does buy is that a compromised measurement cannot reach a token and write elsewhere in the repository. For a same-repo pull request the author already has push access, so the gate is a guard against mistakes, not against its own author.

Worth stating in the workflow rather than leaving implied — I will add a comment saying so.

---

jdx:
Three things addressed, and the offline question @jdx raised turned into the most useful change here.

## `MISE_OFFLINE=1` — yes, and it earns three benchmarks back

Three of the four commands hyperfine measured reach the network. With `MISE_OFFLINE=1` they stop:

| | without | with |
|---|---|---|
| `ls` | 25ms, 1 HTTP warning | **9ms, none** |
| `env` | 15ms, 1 HTTP warning | **9ms, none** |
| `hook-env` | — | **10ms, none** |
| `x -- echo` | 15ms, 1 warning | **fails** — cannot resolve a tool offline |

So `ls`, `env` and `hook-env` are benchmarks now rather than casualties, and their output is byte-identical run to run. Offline covers less of mise than a real invocation does — but the part it leaves out is exactly the part no benchmark can measure repeatably.

They read `benchmarks/perf-fixture`, a committed config pinning exact versions; a floating version would make the measurement depend on what the registry said that day. Five subjects now, all verified against a real release build:

```
startup   9.52ms    registry  12.53ms
env      20.62ms    hook-env  20.73ms    ls  21.58ms
```

## Cache poisoning across trust levels

Correct, and I had it wrong. `perf.yml` holds a write token and pushes notes; it was sharing `mise-pr-rust-linux` with `perf-pr.yml`, which builds pull-request code. Namespace shares a volume per repository and profile by default, so that is a real path from untrusted code into a privileged job's cache. `perf.yml` now uses `mise-perf-main`.

## `mise run render`

Not run, so `tasks.md` did not list the new tasks and lint rejected the diff. Fixed.

## Two self-inflicted rounds worth admitting

Adding the new cache tag broke actionlint twice — first because the label was not in `.github/actionlint.yaml`, then because I fixed that at the wrong indentation and made the config unparseable. Both times I pushed without reading actionlint's output, which is precisely what the allowlist exists to prevent. Verified now: the perf workflows are clean, and actionlint over every workflow reports only a pre-existing style nit in `registry.yml`.

---

PR #11388, opened 2026-07-27
fix(dotfiles): prune stale symlink-each links, add exclude patterns

jdx:
Fixes #11386.

## Problem

`symlink-each` only ever walked *source → target*, so nothing on the target side was ever inspected. Deleting a source file left its symlink dangling in the target forever, and `mise dotfiles status` still reported the entry as `applied`, so the drift wasn't even visible:

```
src/            tgt/                    after `rm src/a src/sub/b`, apply:
  a               a -> …/src/a            a     -> …/src/a      (DANGLING, kept)
  sub/b           sub/b -> …/src/sub/b    sub/b -> …/src/sub/b  (DANGLING, kept)
                  keepme (unmanaged)      keepme
```

Separately, declining the trust prompt puts the config on the ignore list, after which `mise dotfiles apply` reports `no dotfiles configured in [dotfiles]` — which reads as a config mistake when the real cause is that the file was never loaded, and gives no hint that `mise trust` is the fix.

## Fix

`stale_links()` walks the target directory and collects symlinks that point **into this entry's source** and whose destination no longer exists. Those are links mise itself created for this entry, so `apply` removes them, plus any directory they emptied out. Anything else — regular files, and links pointing outside the source — is left alone, which keeps `symlink-each`'s documented promise that the target directory can also hold unmanaged content. On Windows file symlinks are applied as copies, so there's nothing there to distinguish a leftover from a user's file and that platform prunes nothing.

The state is threaded through the rest of the entry's surface so it isn't silently repaired:

- `mise dotfiles status` → `differs (N stale link(s))`, and `--missing` exits 1
- `mise dotfiles apply --dry-run` → `ln -sf … (1 files), rm 2 stale link(s)`, with `--verbose` listing each path

For the trust case, `mise dotfiles apply`/`status` now check whether any ignored config actually declares a `[dotfiles]` table (a plain TOML parse — nothing rendered or executed) and warn with the paths and `mise trust` before the empty-table message.

## Verification

Reproduced the issue exactly as reported, then confirmed the fix:

```
$ mise dotfiles status
/tmp/dtest/tgt  symlink-each  /tmp/dtest/src  differs (2 stale link(s))
$ mise dotfiles apply --dry-run --verbose
ln -sf /tmp/dtest/src/* into /tmp/dtest/tgt/ (1 files), rm 2 stale link(s)
  stale link to remove: /tmp/dtest/tgt/a
  stale link to remove: /tmp/dtest/tgt/sub/b
$ mise dotfiles apply -y   # stale links + emptied `sub/` gone, `keepme` kept
$ mise dotfiles status
/tmp/dtest/tgt  symlink-each  /tmp/dtest/src  applied
```

`e2e/cli/test_dotfiles_files` gains coverage for pruning (including the nested case, an emptied directory, an unmanaged file, and a user-created dangling link that must survive) and for the untrusted-config message. `test_bootstrap`, `test_bootstrap_shell_activate`, `test_dotfiles_edits`, unit tests, and `mise run lint` all pass.

---

> [!NOTE]
> **Medium Risk**
> Changes how dotfiles apply/remove files under home directories; pruning is scoped to entry-owned symlinks but mistakes could remove links users expected to keep.
> 
> **Overview**
> **`symlink-each`** now treats dangling mise-managed symlinks as drift: `status` reports `differs (N stale link(s))`, apply/dry-run can remove them and collapse emptied directories, while unmanaged files and user links stay untouched (Windows still skips pruning where symlinks are copies).
> 
> Directory-walking modes gain an **`exclude`** list of glob patterns on whole-file entries (`symlink-each` and directory **`copy`**), with schema/docs and invalid-pattern warnings; newly excluded paths are reclaimed like deleted sources. **`mise dotfiles apply`** / **`status`** warn when untrusted ignored configs declare `[dotfiles]` instead of only saying nothing is configured (hint on stderr for `--json`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 781b9bbf52dc724a3e1fe251f658da8abdaace71. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * `symlink-each` now detects and cleans up stale mise-managed symlinks (and emptied managed directories) after source files are deleted, while preserving unmanaged items.
  * Added `exclude` glob patterns to skip source-relative paths for directory-walking dotfiles modes.
  * Dotfiles commands now warn when an ignored config declares `[dotfiles]`, prompting users to run `mise trust`.
* **Documentation**
  * Updated “Excluding files” docs and clarified `symlink-each` stale-link cleanup behavior.
* **Bug Fixes**
  * “No dotfiles configured” output now also surfaces the untrusted-config warning (including for `--json`).
* **Tests**
  * Expanded e2e coverage for stale-link cleanup, `exclude` semantics (anchored/unanchored, copies, invalid patterns), and trust messaging.


jdx:
Addressed the review feedback in 5b3c9b1.

**Ownership was too loose** (@greptile-apps, Cursor Bugbot). Using the destination *prefix* as proof mise wrote the link was wrong: a user's own dangling link pointing anywhere inside the source directory matched it, so apply would have deleted it — exactly contradicting what the docs promise. Ownership is now the exact link the entry writes, `<target>/rel -> <source>/rel`, mirroring `walk_source_files`. A link at any other path survives even when it dangles into the source, and the only case that still gets pruned is one that's byte-for-byte what mise itself would have created — indistinguishable from a leftover. New e2e case covers it:

```bash
ln -s "$PWD/dotfiles/bin/two" ~/.local/mybin/my-alias   # user's own, non-mirrored path
rm dotfiles/bin/two                                     # now dangling, inside the source
mise dotfiles apply --yes
assert "readlink ~/.local/mybin/my-alias" "$PWD/dotfiles/bin/two"   # survives
```

**`read_link` propagated its error** (@coderabbitai). Now logged at debug and skipped, matching how traversal errors were already handled — a link that vanishes mid-walk shouldn't fail the whole apply.

**The "unrelated ignored config" assertion tested nothing** (@coderabbitai). Correct, and worse than described: `unrelated.toml` isn't a default config filename, so it could never appear in `IGNORED_CONFIG_FILES` at all. Replaced with an ignored `mise.toml` that has no `[dotfiles]` table, which actually exercises the filter in `ignored_configs_with_dotfiles()`.

`test_dotfiles_files` (108 assertions), `test_bootstrap`, `test_dotfiles_edits`, unit tests, and `mise run lint` pass.

jdx:
Addressed the remaining review points in 781b9bb, and picked up the follow-up request from the discussion.

**Windows still pruned** (Cursor Bugbot). Right — the doc comment claimed the platform prunes nothing but no code enforced it. `stale_links` now returns early on Windows. That's the consistent choice: `link_path` writes copies there, so a leftover is a regular file indistinguishable from the user's, and `check_symlink` already treats those targets as copies. Without the guard, `check_symlink_each` could report a link stale that its own per-file check considers a copy.

**`--json` status skipped the hint** (Cursor Bugbot). Moved the call ahead of the output branch so both paths get it. It writes to stderr, so `--json` stdout stays parseable — covered by an `assert_json` on the empty result alongside the stderr assertion.

**New: `exclude` patterns.** Jelenkee [asked in the discussion](https://github.com/jdx/mise/discussions/11386#discussioncomment-17800585) for a stow-style ignore list — their `source = "."` entry sweeps up `mise.toml` itself, and listing every file individually isn't practical. Directory-walking modes (`symlink-each`, and `copy` with a directory source) now take one:

```toml
[dotfiles]
"~" = { source = ".", mode = "symlink-each", exclude = ["mise.toml", "*.md", ".git"] }
```

A pattern without `/` matches any single path component, so it applies at any depth; a pattern containing `/` is anchored to the source root; either kind matching a directory takes everything under it. It rides on the pruning work above — excluding something already applied reclaims the link on the next apply, so adding a pattern converges instead of stranding what's there. Invalid patterns warn against the entry that wrote them and are dropped rather than failing it.

Docs, JSON schema, five unit tests for the matching rules, and e2e coverage (both modes, bare vs. anchored patterns, reclaim, invalid pattern) are in. `test_dotfiles_files` is at 128 assertions; it plus `test_dotfiles_edits`, `test_bootstrap`, unit tests, and `mise run lint` pass.

Happy to split the `exclude` feature into its own PR if you'd rather keep this one to the bug fix.

PR #11391, opened 2026-07-27
fix(backend): report version listing fetch failures instead of blaming filters

jdx:
## Problem

A network failure while listing remote versions was reported as a date-filter mismatch:

```
mise WARN  mise-versions endpoint=version_list tool=claude outcome=failed status=0 fallback=false error="error sending request"
mise WARN  Remote versions cannot be fetched: error sending request
mise WARN  No versions found for claude
... same three lines, 7 times ...
mise ERROR Failed to install aqua:anthropics/claude-code@latest: no versions found for claude matching date filter
```

No filter ever ran. `AquaBackend::_list_remote_versions` degrades a failed fetch into `Ok(vec![])` and drops the error, so the empty list falls through to the generic "no versions found" path in `ToolVersion::no_versions_found`, which blames whatever cutoff happens to be configured. Users go looking for a `minimum_release_age` problem that isn't there.

The seven identical blocks were not just log spam — they were seven real refetches. `list_remote_versions_with_refresh` deliberately avoids writing an empty list to the disk cache, but `remote_versions.clear()` also destroys the in-memory memoization, so every call site that resolved the tool re-ran the whole failing fetch, HTTP retries and backoff included.

## Fix

- Backends that swallow a fetch failure record the cause; `no_versions_found` reports that instead of the filter. A genuinely empty listing keeps the existing wording.
- A recorded failure short-circuits later listings for the same backend in the same process. `--refresh` still bypasses it, and a successful non-empty listing clears the record.
- The three warnings that repeated per attempt became `warn_once!`.

After:

```
mise WARN  Remote versions cannot be fetched for anthropics/claude-code: HTTP host https://api.github.com:443 is unavailable after an earlier connection failure: error sending request
mise WARN  No versions found for claude
mise WARN  Failed to resolve tool version list for claude: [--runtime] claude@latest: unable to fetch versions for claude: HTTP host https://api.github.com:443 is unavailable after an earlier connection failure: error sending request
```

## Measurements

Reproduced with a dead HTTP proxy and `minimum_release_age = "7d"`, cold cache:

| | before | after |
|---|---|---|
| `mise x claude@latest` warning lines | 14 | 3 |
| `mise install claude@latest` warning lines | 86 | 39 |
| `mise install` HTTP attempts | 54 | 30 |
| `mise install` wall clock | 57s | 34s |

## Testing

- New e2e test `e2e/backend/test_aqua_version_listing_fetch_failure`, confirmed to fail on the unpatched build rather than pass vacuously.
- `cargo test --bin mise`: 2027 passed, 0 failed.
- `mise run lint` clean; clippy clean on the changed lines.
- Existing `no versions found for fuzzytest` assertions still hold, so an empty listing is not misattributed to a fetch failure.

## Not included

- On `mise install` most of the remaining 39 lines are HTTP retry warnings from the latest-version fast path (`releases/latest`), which resolves outside the version-listing cache and so is not covered by the memoization here. Same class of bug, different code path.
- `strict_metadata()` still gates whether aqua's tag fetch hard-errors. The lenient default is unchanged — making a network failure fail the command outright seems worth deciding separately.

---

> [!NOTE]
> **Medium Risk**
> Changes core version-resolution and remote-listing behavior for all backends that record failures; wrong short-circuiting could hide recoverable network blips until a new process, but scope is limited to empty-list failure degradation paths.
> 
> **Overview**
> When remote version listing fails but backends return an empty list, mise now **records the real error** (keyed by `BackendArg::full()`) and uses it in install/resolve failures as `unable to fetch versions for …` instead of misleading **date-filter** or generic **no versions found** messages.
> 
> **`list_remote_versions_with_refresh`** short-circuits further listing attempts for that tool in the same process after a recorded failure, cutting duplicate HTTP retries across multiple resolve call sites. Related user-facing warnings move to **`warn_once!`**, and the generic **No versions found** warning is suppressed when a fetch failure was already recorded.
> 
> The aqua backend wires both registry and tag/metadata fetch failures into this path. A small **versions host** logging tweak marks non-404/429 errors as `fallback=true` to match upstream retry behavior.
> 
> Adds e2e coverage **`test_aqua_version_listing_fetch_failure`** (blocked hosts + `MISE_MINIMUM_RELEASE_AGE`) asserting correct error text, no date-filter blame, and exactly one fetch-failure warning.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 883bf9b56cde6e768aaac0a190db14848bfcea35. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved and de-duplicated warnings when remote version listings fail.
  * Prevented misleading “matching date filter” output when versions can’t be fetched.
  * Avoided repeatedly retrying version-list requests after a known failure.
  * Refined fallback-warning semantics when the versions host returns non-success responses.
* **Tests**
  * Added end-to-end coverage validating correct failure behavior and messaging when remote version listing cannot be fetched.

### DIFF
--- a/e2e/backend/test_aqua_version_listing_fetch_failure
+++ b/e2e/backend/test_aqua_version_listing_fetch_failure
@@ -14,13 +14,23 @@ export MISE_HTTP_RETRIES=0
 export MISE_MINIMUM_RELEASE_AGE=7d
 export RUST_BACKTRACE=0
 
-out="$(mise install "aqua:suzuki-shunsuke/tfcmt@latest" 2>&1 || true)"
+status=0
+out="$(mise install "aqua:suzuki-shunsuke/tfcmt@latest" 2>&1)" || status=$?
+
+if [[ $status -eq 0 ]]; then
+  fail "expected the install to fail, but it exited 0"
+fi
+ok "install failed with status $status"
 
 assert_contains_text "$out" "unable to fetch versions for aqua:suzuki-shunsuke/tfcmt"
 assert_not_contains_text "$out" "matching date filter"
+# The specific cause is warned instead of the generic "no versions" message.
+assert_not_contains_text "$out" "No versions found for"
 
+# Exactly one: zero would mean the warning was dropped and the regression this
+# test guards (one warning per call site that resolved the tool) went unnoticed.
 warns="$(grep -c "Remote versions cannot be fetched" <<<"$out" || true)"
-if [[ $warns -gt 1 ]]; then
-  fail "expected at most 1 fetch-failure warning, got $warns"
+if [[ $warns -ne 1 ]]; then
+  fail "expected exactly 1 fetch-failure warning, got $warns"
 fi
-ok "fetch failure warned $warns time(s)"
+ok "fetch failure warned exactly once"
--- a/e2e/backend/test_aqua_version_listing_fetch_failure
+++ b/e2e/backend/test_aqua_version_listing_fetch_failure
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+# When a remote version listing fails, say so. Previously the failure degraded
+# into an empty version list and the generic "no versions found" path blamed
+# whatever filter happened to be configured -- e.g. "no versions found for X
+# matching date filter" -- even though no filter ever ran. The same fetch was
+# also retried, and warned about, once per call site that resolved the tool.
+
+# Point every host the aqua backend lists versions from at a closed port so the
+# listing fails for a reason that has nothing to do with a date filter.
+export MISE_URL_REPLACEMENTS='{"https://api.github.com":"http://127.0.0.1:1","https://github.com":"http://127.0.0.1:1","https://mise-versions.jdx.dev":"http://127.0.0.1:1"}'
+export MISE_USE_VERSIONS_HOST=0
+export MISE_HTTP_RETRIES=0
+# A date cutoff is what made the old message misleading, so keep one active.
+export MISE_MINIMUM_RELEASE_AGE=7d
+export RUST_BACKTRACE=0
+
+out="$(mise install "aqua:suzuki-shunsuke/tfcmt@latest" 2>&1 || true)"
+
+assert_contains_text "$out" "unable to fetch versions for aqua:suzuki-shunsuke/tfcmt"
+assert_not_contains_text "$out" "matching date filter"
+
+warns="$(grep -c "Remote versions cannot be fetched" <<<"$out" || true)"
+if [[ $warns -gt 1 ]]; then
+  fail "expected at most 1 fetch-failure warning, got $warns"
+fi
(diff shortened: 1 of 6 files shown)


jdx:
Addressed in 883bf9b5. Triage of the bot findings:

**Fixed**

- **Canonical key** (CodeRabbit, Major) — legitimate. The failure record was keyed by `ba.short` while `get_remote_version_cache` keys by `ba.full()`, so `foo` and `foo[api_url=...]` shared an entry and could inherit each other's failures. Both helpers now take `&BackendArg` and key on `ba.full()`, so there is one canonical key at every call site.
- **Generic warning on top of the specific one** (CodeRabbit) — `No versions found for {id}` is now suppressed when a fetch failure was recorded. The offline exec path is down to 2 warning lines from 14 on `main`.
- **e2e masked the exit status** (CodeRabbit) — `|| true` replaced with a captured status and a non-zero assertion.
- **e2e allowed zero warnings** (CodeRabbit) — `-gt 1` was satisfied by deleting the warning entirely; now `-ne 1`.

**Not applicable**

- **"Versions host skips failure clear"** (Cursor, Medium) — reviewed against `a0abb15f`, the first commit, before `0b083940` moved the failure check ahead of both fetch branches. The scenario it describes (a later `--refresh` repopulating the cache while the entry remains) can no longer happen.
- **"Successful empty refresh stays failed"** (Greptile, P1) and **"Clear recovered failures even when the refreshed listing is empty"** (CodeRabbit) — same root claim, and it does not hold against current `main`-of-this-branch. Once an entry exists, `list_remote_versions_with_refresh` returns `vec![]` without calling `fetch` at all, `refresh` included:

  ```rust
  let versions = if version_listing_failure(&ba).is_some() {
      vec![]                                              // refresh included
  } else if refresh {
      remote_versions.refresh_async(fetch).await?
  } else {
      remote_versions.get_or_try_init_async(fetch).await?.clone()
  };
  ```

  So there is no "same-process refresh later succeeds" to clear. Following that through, `clear_version_listing_failure` was only reachable when nothing had been recorded — dead code that implied a recovery path the control flow cannot reach. Rather than make it more elaborate, I removed it and documented the invariant, including the one constraint that keeps it sound: a recorder must return an empty list alongside the record, or it would suppress its own later listings.

Verification: 2027 unit tests pass, `mise run lint` clean, and the e2e set covering this area passes — `test_aqua_version_listing_fetch_failure`, `test_github_ls_remote_inline_opts` (the inline-opts-vs-dead-`api_url` case the canonical-key fix is about), `test_install_prefer_offline_no_refresh`, `test_fuzzy_versions_do_not_install_literal_dirs`, `test_lock_latest`, `test_reshim_stale_install`.

PR #11457, opened 2026-07-29
fix(aqua): resolve versions by semantic precedence

pepicrft:
## What changed

- Added an opt-in Aqua backend option, `version_sort = "semver"`, that orders matching releases by [semantic versioning](https://semver.org/) precedence.
- Configured Tuist's Aqua registry entry to use semantic version ordering.
- Documented the option and its fallback behavior.

## Why

`mise latest tuist` should resolve to the highest stable Tuist version even when an older release line receives a newer backport.

## Root cause

The Aqua backend normally preserves the order returned by the upstream release source. When GitHub's marked latest release belongs to another package in the Tuist repository, mise falls back to release chronology. Tuist `4.197.3` was published after `4.202.6`, so the chronological fallback selected the lower version.

## Approach

The new behavior is opt-in because Aqua packages do not all use semantic versioning. When enabled, mise skips the marked-latest shortcut, collects matching release candidates, and orders valid semantic versions by precedence. If any candidate is not a valid semantic version, mise preserves the upstream order.

## Impact

Tuist now resolves `latest` and version prefixes by semantic version precedence. Existing Aqua tools keep their current ordering unless they explicitly enable the option.

## Validation

- `/Users/pepicrft/.cargo/bin/cargo fmt --all -- --check`
- `/Users/pepicrft/.cargo/bin/cargo test backend::aqua::tests::` with 52 tests passed
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added optional `version_sort="semver"` behavior for Aqua-managed tools, using semantic version precedence for `latest` and version-prefix matching.
  * When semver parsing succeeds, matching results are semver-sorted; when parsing fails, the original upstream ordering is preserved.
  * Enabled semantic version sorting for the Tuist Aqua backend in the registry configuration.

* **Documentation**
  * Documented the Aqua `version_sort` option, including the semver-based resolution and fallback behavior.

### DIFF
--- a/docs/dev-tools/backends/aqua.md
+++ b/docs/dev-tools/backends/aqua.md
@@ -136,6 +136,21 @@ Set them via tool options using either top-level keys or a nested `vars` table:
 Vars with defaults are filled automatically. Vars marked as required in the aqua registry must be set
 unless the registry also provides a default.
 
+### `version_sort`
+
+Aqua normally preserves the order returned by the upstream version source. For repositories that
+publish backports after newer releases, set `version_sort = "semver"` to resolve `latest` and
+version prefixes using [semantic version precedence](https://semver.org/):
+
+```toml
+[tools]
+"aqua:owner/tool" = { version = "latest", version_sort = "semver" }
+```
+
+If any matching version is not a valid semantic version, mise preserves the upstream order. The
+option is therefore intended only for tools whose release tags consistently follow semantic
+versioning.
+
 ### `prerelease`
 
 By default, releases flagged `prerelease: true` on GitHub are excluded from `mise ls-remote` and from `latest` resolution. Set `prerelease = true` to include them:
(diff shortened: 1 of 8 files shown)


risu729:
Thanks for identifying and demonstrating the Tuist ordering bug here. I opened [#11481](https://github.com/jdx/mise/pull/11481) as the generic follow-up that supersedes this implementation. It preserves the opt-in SemVer approach, but moves `version_order` into mise's shared resolution layer so it also covers the pnpm case from the source discussion, bypasses all backend latest shortcuts consistently, and leaves `mise-versions`/`ls-remote` data in source order. The new PR links and credits this one in its rationale.

jdx:
Closing as superseded by #11774, which merged the shared `version_order` implementation and incorporates the backend-specific boundary explored here.

*AI-assisted — Tool: Codex; model: OpenAI/GPT-5.6 Sol; version: unavailable.*

pepicrft:
Thanks @jdx

PR #11460, opened 2026-07-29
docs: add trailing slash to node.mirror_url npmmirror example

Bartok9:
The `node.mirror_url` help example used `https://npmmirror.com/mirrors/node` without a trailing slash. `Url::join` then drops the last path segment, so downloads resolve to `/mirrors/v...` (404) instead of `/mirrors/node/v...`. Add the slash to match `DEFAULT_NODE_MIRROR_URL` (`.../dist/`).

Proof: join without slash -> `.../mirrors/v22.17.1/SHASUMS256.txt` (404); with slash -> `.../mirrors/node/v22.17.1/SHASUMS256.txt` (200).

---

**Agent-Owner:** `sera` · **Platform:** hermes · **Claim:** `sera` · **Claim-TTL:** 72h

### DIFF
--- a/docs/cli/settings.md
+++ b/docs/cli/settings.md
@@ -65,6 +65,6 @@ $ mise settings always_keep_download
 # set the value of the setting "always_keep_download" to "true"
 $ mise settings always_keep_download=true
 
-# set the value of the setting "node.mirror_url" to "https://npmmirror.com/mirrors/node"
-$ mise settings node.mirror_url https://npmmirror.com/mirrors/node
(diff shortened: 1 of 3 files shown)


Marukome0743:
It looks like a Ready for review slot is available now, so this PR should no longer be blocked by the repository’s two-PR limit. If it is complete on your side, you can mark it as Ready for review through the GitHub UI.

Bartok9:
Thanks @Marukome0743 — marked ready for review. Docs-only trailing-slash fix on the `node.mirror_url` npmmirror example; CI is green.

PR #11516, opened 2026-07-30
chore(clippy): fix warnings surfaced by clippy 0.1.97

JamBalaya56562:
## What

25 clippy warnings across 17 files:

| lint | count |
|---|---|
| `clippy::useless_borrows_in_formatting` | 22 |
| `clippy::collapsible_if` | 2 |
| `clippy::manual_filter` | 1 |

All semantics-preserving. The edits are `cargo clippy --fix` output, reviewed rather than transcribed by hand — the only non-mechanical-looking ones are the two let-chain conversions and the `manual_filter`:

```diff
-    if shell_body_mode {
-        if let (Some(prog), [.., last]) = (program.to_str(), args.as_slice()) {
+    if shell_body_mode && let (Some(prog), [.., last]) = (program.to_str(), args.as_slice()) {
```

```diff
-        .and_then(|v| if v.trim().is_empty() { None } else { Some(v) })
+        .filter(|v| !v.trim().is_empty())
```

## Why these are not already caught

The `lint` job runs `cargo clippy -- -D warnings` on a Linux runner. Nothing runs clippy on Windows or macOS, so I ran `cargo clippy --all-targets` on `windows-latest` and `macos-latest` to see what was there. Two categories came back, and it is worth separating them:

**Three are genuinely platform-gated** — invisible to CI because the code is never compiled there:

- `src/cli/exec.rs` and `src/tokens.rs` — `collapsible_if` inside `#[cfg(windows)]` blocks
- `src/config/config_file/mise_toml.rs` — a test that is not compiled on Windows

This is the same drift #10149 ("clean up accumulated Windows-only clippy warnings") and #10425 cleaned up before.

**The other 22 are ordinary cross-platform code**, not behind any `cfg`. They are invisible for a different reason: the runners I used had **clippy 0.1.97 (rustc 1.97.1)**, which is newer than what the Linux `lint` runner is using. So these are not a platform problem at all — they are queued up to start failing the `lint` job the moment that toolchain moves. Fixing them now means that update is a non-event.

## Out of scope

`crates/*` warnings are left alone. Those crates sit outside the `lint` job's scope entirely (`cargo clippy` lints default members; `windows-unit`'s check is `-p mise`), and the findings there raise a separate question rather than a mechanical fix — `crates/mise-shim` calls `std::process::exit` five times, which the repo's own `clippy.toml` disallows, but a shim binary's `main` is exactly the outermost boundary that rule carves out. That deserves its own PR and its own discussion.

Also out of scope, but worth saying out loud: nothing here stops the drift recurring. Adding clippy to the existing `windows-unit` job (and a macOS equivalent) would, at the cost of build time. Happy to open that separately if you want it.

## Verification

The `lint` job here covers Linux as usual. For the two targets CI does not lint, I am re-running `cargo clippy --all-targets` on `windows-latest` and `macos-latest` against this branch and will post the result below — that re-run is the actual proof that the list is now empty, so I would rather report it than assert it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows command execution handling without changing existing command behavior.
  * Continued to ignore empty or whitespace-only authentication tokens.
  * Preserved tool, plugin, configuration, and task formatting and path-generation behavior.

* **Refactor**
  * Simplified internal string formatting and conditional logic across tool management, plugins, task handling, and CLI workflows.

### DIFF
--- a/src/backend/dotnet.rs
+++ b/src/backend/dotnet.rs
@@ -65,7 +65,7 @@ impl Backend for DotnetBackend {
                 // version arrays of packages that are visible.
                 "{}?q={}&packageType=dotnettool&take=1&prerelease={}&semVerLevel=2.0.0",
                 feed_url,
-                &self.tool_name(),
+                self.tool_name(),
                 true
             ))
             .await?;
--- a/src/cli/doctor/mod.rs
+++ b/src/cli/doctor/mod.rs
@@ -426,7 +426,7 @@ impl Doctor {
                 && !plugin.is_installed()
             {
                 self.errors
-                    .push(format!("plugin {} is not installed", &plugin.name()));
+                    .push(format!("plugin {} is not installed", plugin.name()));
                 continue;
             }
         }
@@ -823,7 +823,7 @@ impl Doctor {
 
             if is_core && matches!(plugin_type, Some(PluginType::Asdf | PluginType::Vfox)) {
                 self.warnings
-                    .push(format!("plugin {} overrides a core plugin", &plugin.id()));
+                    .push(format!("plugin {} overrides a core plugin", plugin.id()));
             }
         }
     }
--- a/src/cli/edit.rs
+++ b/src/cli/edit.rs
@@ -207,7 +207,7 @@ impl Edit {
     async fn interactive(&self, path: &Path) -> Result<()> {
         use crate::ui::progress_report::ProgressReport;
 
-        let title = format!("mise {} by @jdx", &*VERSION_PLAIN);
+        let title = format!("mise {} by @jdx", *VERSION_PLAIN);
 
         // Show loading spinner while setting up
         let pr = ProgressReport::new("edit".into());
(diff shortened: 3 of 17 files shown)


JamBalaya56562:
Verification promised above, now done.

Re-ran `cargo clippy --all-targets` on `windows-latest` and `macos-latest` against this branch — same invocations that produced the original 25 findings, including the Windows-specific feature set (`--no-default-features --features rustls-native-roots,vfox/vendored-lua,self_update`, since `--all-features` cannot build on Windows).

Both platforms are clean. The strongest signal is the `clippy --fix` patch each job emits: it was 11,773 bytes (Windows) and 10,311 bytes (macOS) before, and is **0 bytes on both now** — clippy has nothing left to change.

```
clippy-fix-windows.patch: 0 bytes
clippy-fix-macos.patch:   0 bytes
```

For completeness: the two `warning:` lines still in the run log are the Actions runner's own Node deprecation notice about `url.parse()`, not clippy output.

Nothing about that probe workflow is part of this PR — it lives on a throwaway fork branch and is deleted once this lands.

PR #3038, opened 2024-11-15
fix: mise use -p . panics

roele:
Fixes #3033

Issue #3033, opened 2024-11-14
`mise use -p .` panics


otaconix:
**Describe the bug**
Whenever using `mise use -p . <tool-here>`, mise panics.

The commit that seems to have introduced this (as found through a `git bisect`) is b36acc3288d5b087ef126593037d7c10232635ab

**To Reproduce**

```
$ mise use -p . java@temurin-21                                                                                                                  at 20:48:53
The application panicked (crashed).
Message:  called `Option::unwrap()` on a `None` value
Location: src/config/config_file/mod.rs:329

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```

**Expected behavior**
I'd expect mise to find or create a new config file in the PWD.

**`mise doctor` output**

```text
version: 2024.11.12 macos-arm64 (2024-11-14)
activated: yes
shims_on_path: no

build_info:
  Target: aarch64-apple-darwin
  Features: DEFAULT, NATIVE_TLS
  Built: Thu, 14 Nov 2024 08:03:00 +0000
  Rust Version: rustc 1.82.0 (f6e511eec 2024-10-15) (Homebrew)
  Profile: release

shell:
  /opt/homebrew/bin/fish
  fish, version 3.7.1

dirs:
  data: ~/.local/share/mise
  config: ~/.config/mise
  cache: ~/Library/Caches/mise
  state: ~/.local/state/mise
  shims: ~/.local/share/mise/shims

config_files:
  ~/.config/mise/config.toml
  ~/Projects/.mise.toml

backends:
  aqua
  asdf
  cargo
  core
  go
  npm
  pipx
  spm
  ubi
  vfox
  unknown

plugins:
  maven    https://github.com/halcyon/asdf-maven.git#83fb4db
  pnpm     https://github.com/jonathanmorley/asdf-pnpm.git#305baff
  pulumi   https://github.com/canha/asdf-pulumi.git#99d675b
  quarkus  https://github.com/asdf-community/asdf-quarkus.git#d00eb32
  ruff     https://github.com/simhem/asdf-ruff#97b4f67
  usage    https://github.com/jdx/mise-usage.git#fe3888a
  uv       https://github.com/asdf-community/asdf-uv.git#54b1c75
  yarn     https://github.com/mise-plugins/asdf-yarn.git#74ea3b9

toolset:
  asdf:mise-plugins/asdf-maven@3.9.9
[...]

PR #11575, opened 2026-08-01
fix(config): make --path <dir> target that directory

JamBalaya56562:
## Summary

`mise use --path <dir>` and `mise unuse --path <dir>` write to the **cwd's** config, not to one under `<dir>`. When the cwd has a config in scope, `--path` is discarded entirely. `unuse` is the destructive case — it removes the tool from the wrong file.

Measured on v2026.7.18 (Linux and Windows alike), with `here/mise.toml` and `other/mise.toml`, run from `here`:

| command | wrote to |
|---|---|
| `mise use --path ../other` | `here/mise.toml` |
| `mise use --path ./sub` | `here/mise.toml` |
| `mise unuse --path ../other uv` | **removed `uv` from `here/mise.toml`** |
| `mise use --path ../other` from a dir whose walk finds nothing | `other/mise.toml` (correct) |
| `mise use --path ../other/mise.toml` (a *file*) | `other/mise.toml` (correct) |

So `--path <dir>` only works when the cwd has no config in scope, which is why this has gone unnoticed.

## Why

`config_file_from_dir(p)` never searches `p`. It walks `all_dirs()` — the ancestors of `env::current_dir()`. `p` is used only by the `!p.is_dir()` early return and the trailing `p.join(...)` fallback.

That is a regression from #3273 ("feat: fragmented configs", 2024-11-29), which replaced

```rust
for p in FindUp::new(p, &filenames) { ... }     // walks up from `p`
```

with `file::all_dirs()` (walks up from the cwd) as collateral to the `conf.d` work; the PR body does not mention the function.

The `if from_dir.starts_with(&cwd) { from_dir } else { p.clone() }` guard in `use.rs`/`unuse.rs` came from #2818, where it kept that argument-based walk from escaping the tree. Once the walk started at the cwd instead, the guard became dead — and worse than dead: `from_dir` *is* the cwd's own config, so `starts_with` is true and the guard is what admits it. `.absolutize()` (#3038, fixing #3033) exists only to make that comparison well-defined.

## What changed

`config_file_in_dir(dir)` looks only inside the given directory — lowest-precedence non-global config, else `<dir>/mise.toml` (or `.tool-versions` under `asdf_compat`). That is the documented contract:

> If a directory is specified, it will look for a config file **in that directory** following the rules above.

("the rules above" being the file-precedence rules — `mise.toml` over `mise.local.toml`.) It also matches what #3033's reporter asked for: *"I'd expect mise to find or create a new config file in the PWD."*

- `use.rs` drops its bespoke `--path` branch and goes through `resolve_target_config_path` like everything else.
- `unuse.rs` drops the same copy of the guard.
- `resolve_target_config_path` absolutizes `opts.path` once, so every arm returns an absolute path. A relative one still resolved fine for reads, but it silently disabled the monorepo lockfile redirection in `lockfile_path_for_config`, which compares against an absolute root.
- **`config_file_from_dir` is left alone.** It is now only ever asked about the cwd, so nothing that omits `--path` changes behaviour.

This also fixes the six other commands that document `--path`/`--file` as accepting a directory and route through the same `path.is_dir()` arm: `set`, `unset`, `dotfiles add`, `system use`, `system import`, `system brew tap`, `system brew untap`.

## Behaviour change worth flagging

`mise use -p .` in a directory with **no** config no longer walks up to the project root — it creates `./mise.toml`. That is the documented behaviour and what #3033 asked for, but it differs from today, so it is worth a maintainer's eye.

## Tests

No existing test pointed `--path` at a directory outside the cwd. `e2e/cli/test_use:66` uses `--path .`, `test_unuse:20` uses a file, and `test_set_use_consistency` passes only because its cwd walk finds nothing. Added to `e2e/cli/test_use` and `e2e/cli/test_unuse`:

- `--path ../to` (relative) and the absolute form — target gets the tool, cwd's config is byte-identical
- a directory whose only config is `.tool-versions` keeps using it instead of growing a `mise.toml`
- `--path .` still means "here"
- `mise set --file ../to` — the `resolve_target_config_path` half
- `unuse --no-prune --path ../to` — the destructive case, asserting the cwd's config survives

Verified on Windows against the released v2026.7.18 binary. Everything below is byte-identical before and after: `--path <file>`, `--path .`, `use` with no `--path` (config-less deep dir / ancestor `mise.toml` / `mise.toml` + `mise.local.toml`), `use -g`, and `mise set` (nested dirs, lowest-precedence file, `MISE_IGNORED_CONFIG_PATHS`).

## Not in this PR

`mise use --path <directory that does not exist>` panics in `config_file::init` (`Unknown config file type`). I measured it on v2026.7.18 and on this branch — unchanged either way, so it is a separate pre-existing issue rather than something this change introduces or should quietly absorb.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved path handling for `mise use` and `mise unuse`, including relative and absolute paths.
  - Configuration changes now apply strictly within the specified directory and no longer modify parent or current-directory configuration files unexpectedly.
  - `mise set --file` reliably writes only to the requested location.
  - `.tool-versions` updates no longer create an unintended `mise.toml` file.
  - Added safeguards for emptying target configuration files when their only tool is removed.

### DIFF
--- a/e2e/cli/test_unuse
+++ b/e2e/cli/test_unuse
@@ -90,3 +90,18 @@ EOF
 assert "mise unuse --no-prune --path .tool-versions dummy@1.0.0"
 assert "cat .tool-versions" "dummy 2.0.0 # keep me
 tiny  3.1.0"
+
+# --path <dir> removes from a config inside that directory and leaves the cwd's alone. This used
+# to resolve from the cwd, so `unuse --path <elsewhere>` deleted the tool from the wrong file.
+mkdir -p "$HOME/unusepath/from" "$HOME/unusepath/to"
+printf '[tools]\ndummy = "1.0.0"\n' >"$HOME/unusepath/from/mise.toml"
+printf '[tools]\ndummy = "1.0.0"\n' >"$HOME/unusepath/to/mise.toml"
+cd "$HOME/unusepath/from" || exit 1
+
+assert "mise unuse --no-prune --path ../to dummy@1.0.0"
+assert "cat mise.toml" '[tools]
+dummy = "1.0.0"'
+# removing the only tool empties the file, as it does everywhere else
+assert "cat ../to/mise.toml" ""
+
+cd "$HOME/workdir" || exit 1
--- a/e2e/cli/test_use
+++ b/e2e/cli/test_use
@@ -163,6 +163,44 @@ assert_contains "mise use --path mise.path.toml dummy@1" "dummy@1."
 assert "cat mise.path.toml" '[tools]
 dummy = "1"'
 
+# --path <dir> writes inside that directory, never the cwd's config. This used to resolve from
+# the cwd and silently discard --path entirely.
+mkdir -p "$HOME/workdir/pathtarget/from" "$HOME/workdir/pathtarget/to"
+printf '[tools]\ndummy = "1"\n' >"$HOME/workdir/pathtarget/from/mise.toml"
+cd "$HOME/workdir/pathtarget/from" || exit 1
+
+mise use --path ../to dummy@2
+assert "cat ../to/mise.toml" '[tools]
+dummy = "2"'
+assert "cat mise.toml" '[tools]
+dummy = "1"'
+
+# and with an absolute path
+mise use --path "$HOME/workdir/pathtarget/to" dummy@3
+assert "cat ../to/mise.toml" '[tools]
+dummy = "3"'
+assert "cat mise.toml" '[tools]
+dummy = "1"'
+
+# a directory whose only config is .tool-versions keeps using it rather than growing a mise.toml
+mkdir -p "$HOME/workdir/pathtarget/tv"
+printf 'dummy 1.0.0\n' >"$HOME/workdir/pathtarget/tv/.tool-versions"
+mise use --path ../tv dummy@2.0.0
+assert "cat ../tv/.tool-versions" "dummy 2.0.0"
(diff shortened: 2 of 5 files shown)


JamBalaya56562:
Thanks — I looked at this one, and I'm leaving `unuse.rs` as it is. The premise doesn't hold: `--global` and `--path` can't both take effect, and `--global` isn't the one that wins.

`overrides_with_all` makes the **last flag on the command line** win, so clap has already collapsed the pair before the ladder runs. Measured on this branch, with `target.toml` in the cwd and a global config both declaring `uv`:

```console
$ mise unuse --no-prune -g -p target.toml uv
# -> removed from target.toml   (--path came last)

$ mise unuse --no-prune -p target.toml -g uv
# -> removed from the global config   (--global came last)
```

`mise use` behaves identically, and did so before this PR too. So the ladder order in `unuse` is not observable, and its `--path`/`--global` behaviour already matches the shared resolver.

Routing the whole ladder through `resolve_target_config_path` isn't a drop-in, either: `unuse`'s default arm searches the *loaded* configs for the one that declares the tool and returns early via `config_file::parse` (`unuse.rs:170-180`), which the shared resolver has no way to express. It would have to stay behind a guard, which is more code for no behaviour change — and this PR is deliberately scoped to the `--path` resolution bug.

Unifying the two ladders is worth doing on its own, though; the duplication is real.

PR #11616, opened 2026-08-02
feat(config): accept --file on `mise unuse` and --path on `mise unset`

JamBalaya56562:
Follow-up to #11577, which left the pair half-finished:

| | |
|---|---|
| `mise use --file` | works (#11577) |
| `mise unuse --file` | **does not** |
| `mise set --path` | works (#11577) |
| `mise unset --path` | **does not** |

So the command that adds a tool takes both spellings and the command that removes it does not. This adds `--file` as a visible alias for `mise unuse --path`, and `--path` for `mise unset --file` — the same one-line change #11577 made, applied to the two inverse commands.

The alias is accurate, not just a spelling: both commands accept a directory as well as a file. `unset` resolves through `resolve_target_config_path`, which has a `path.is_dir()` branch, and `unuse` calls `config::config_file_in_dir` when the argument is a directory.

## Background

#4881 asks for `--file`/`--path` as synonyms on the config-target commands. #11116 implemented it across eleven commands and was closed the next day. #11577 narrowed it to the two the discussion actually names, and that merged — so I am walking the rest in small steps rather than re-proposing the wide change.

Remaining after this one: `config get`, `config set`, `dotfiles add`, and the four `bootstrap packages` commands (`use`, `import`, `brew tap`, `brew untap`).

## Generated files

`mise.usage.kdl` and `man/man1/mise.1` carry the four expected lines. They were written by hand from the shape #11577's merged output already has in-tree (`mise set` at man:3295, `mise use` at man:4468), so `mise run render` should be a no-op — the `lint` job will say so either way.

## Tests

`e2e/cli/test_config_target_aliases` — the file #11577 added — gains the inverse half: `unuse --file` and `unset --path` remove what `use --file` and `set --path` wrote, and the canonical spellings still work afterwards.

## Summary by CodeRabbit

* **New Features**
  * Added `--path` as an alternative to `--file` for `mise unset`.
  * Added `--file` as an alternative to `--path` for `mise unuse`.
  * Existing option names remain supported.
  * File options now accept file or directory paths; directory targets use or create `mise.toml`.

* **Documentation**
  * Clarified file and directory path handling.

* **Tests**
  * Added end-to-end coverage for canonical options, aliases, and directory paths.

### DIFF
--- a/docs/cli/unset.md
+++ b/docs/cli/unset.md
@@ -22,6 +22,8 @@ e.g.: NODE_ENV
 
 Specify a file to use instead of `mise.toml`
 
+Can be a file path or directory. If a directory is provided, will create/use mise.toml in that directory.
+
 Defaults to [`MISE_DEFAULT_CONFIG_FILENAME`](https://mise.jdx.dev/configuration.html#mise_default_config_filename) environment variable, or `mise.toml`. Use [`MISE_GLOBAL_CONFIG_FILE`](https://mise.jdx.dev/configuration.html#mise_global_config_file) to choose a different global config path.
 
 ### `-g --global`
--- a/e2e/cli/test_config_target_aliases
+++ b/e2e/cli/test_config_target_aliases
@@ -23,3 +23,34 @@ mise use --path alias.toml dummy@2
 assert_contains "cat alias.toml" 'dummy = "2"'
 mise set --file alias.toml ALIAS_VAR=set-via-file
 assert_contains "cat alias.toml" 'ALIAS_VAR = "set-via-file"'
+
+# the commands that undo those two take the same pair of names
+# `mise unuse --file` is `mise unuse --path`
+mise unuse --no-prune --file alias.toml dummy@2
+assert_not_contains "cat alias.toml" "dummy"
+
+# `mise unset --path` is `mise unset --file`
+mise unset --path alias.toml ALIAS_VAR
+assert_not_contains "cat alias.toml" "ALIAS_VAR"
+
+# and their canonical names are unaffected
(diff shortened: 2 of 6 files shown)


JamBalaya56562:
Both addressed, thanks — the first one was a real gap.

**`mise unset` not documenting directory targets.** Correct, and it is the only one of the four that did not: `use --path` and `unuse --path` both say "Specify a path to a config file or directory", and `set --file` says "Can be a file path or directory…". `unset --file` said only "Specify a file", while resolving through the same `resolve_target_config_path` with its `path.is_dir()` branch. Advertising `--path` as a spelling for it made that worse, so it is fixed here.

Fixed at the source (`src/cli/unset.rs` doc comment) rather than in `man/man1/mise.1` — that file is generated, so editing it directly would be reverted by the next `mise run render`. `mise.usage.kdl`, `man/man1/mise.1` and `docs/cli/unset.md` are updated to match; the `lint` job's render check will confirm.

I did **not** rename the placeholder from `<FILE>` to `<PATH>`. `mise set --file` — the counterpart this PR is pairing with, merged in #11577 — renders `<FILE>` and documents directory support in prose, so matching that keeps the two halves consistent. Renaming only `unset`'s would make it the odd one out again, in the other direction.

**Directory coverage in the e2e.** Added: the test now exercises a directory target through all four commands under the aliased names, which is also what makes the new help text checkable rather than just asserted —

```bash
mise use --file subdir dummy@1
mise set --path subdir SUBDIR_VAR=via-dir
mise unuse --no-prune --file subdir dummy@1
mise unset --path subdir SUBDIR_VAR
```

PR #11643, opened 2026-08-02
feat(env): warn when PATH is long enough for cmd.exe to ignore

JamBalaya56562:
Draft — the detection is straightforward, but where it should fire and how loudly is your call. #5830's fifth suggestion, and the only one of that thread's eight that does not need a design decision first.

## What breaks, measured

Past **8191 characters** `cmd.exe` does not truncate an inherited variable — it drops it. Nothing on PATH resolves, so `npm`, `npx` and batch scripts report that *every* command is unrecognised, with nothing naming PATH, cmd.exe or mise.

Windows 11 26200 / cmd 10.0.26100.8875, same PATH for both columns, `Git\cmd` appended last:

| Path length | `where.exe where` | `git --version` |
|---|---|---|
| 2024 | OK | OK |
| 8184 | OK | OK |
| 8239 | OK | **not recognized** |
| 30019 | OK | **not recognized** |

(The `where.exe` column is there because it is the probe `docs/troubleshooting.md` currently recommends, and it never discriminates — `where.exe` is in System32, which cmd.exe resolves while ignoring PATH. #11642 fixes that separately.)

[KB 830473](https://learn.microsoft.com/en-us/troubleshoot/windows-client/shell-experience/command-line-string-limitation) documents the behaviour, but its "Applies to" stops at Windows 7 / Server 2008 R2 / 2012 R2 and hedges with "as appropriate to the operating system" — hence the measurement above rather than a citation alone.

## Where it fires

Two surfaces can reach that length, and they do not share a code path:

- **`PathEnv::join`** — every environment mise computes: `mise x`/`run` children, `mise env`, backends, core plugins.
- **`hook_env.rs`** — assembles the shell's PATH by hand with `join_paths(pre + user_paths + tool_paths + post + post_user)`, so `PathEnv::join`'s check never sees it. This is the copy a tool started *directly* from an activated shell inherits, which is the reported symptom (`npm install` from a normal prompt), so leaving it out would miss the main case.

`warn_once!` rather than `hint!`: hints require `console::user_attended()` and so never appear in scripts or CI, and are suppressed permanently after one appearance. A broken environment should be reported every process.

## Correctness details

- The predicate is pure and takes a length, so the boundary is unit-tested on every platform, including the `windows-unit` job, and `cfg!(windows)` inside it means it can never fire elsewhere.
- Length is counted in **UTF-16 code units**, which is what Windows counts. `OsStr::len` would return WTF-8 bytes and treat a three-byte character as three, warning about a PATH cmd.exe accepts. Pinned by a test.

**This only reports the condition.** It does not shorten anything — a large enough toolset still hits the limit, and the structural proposals in #5830 (shorter install dirs, hashed paths, a per-project shim dir) are untouched.

## Open questions — why this is a draft

1. **Suppression.** Someone sitting near the limit sees this on every process. Keep a plain warning, or add a `disable_hints`-style setting?
2. **hook-env.** It is the surface that matters most and also the noisiest — re-evaluated on every prompt, where per-invocation suppression buys little.
3. **A second threshold at 32767?** Past the Win32 per-variable limit, `CreateProcess` fails outright with "not enough memory resources" — also measured. Louder failure, equally opaque cause. Worth catching here too, or out of scope?
4. **Tests.** Unit tests cover the boundary. An end-to-end check belongs in `e2e-win/run.ps1` (build a long PATH, run `mise x`, assert the warning) — happy to add it once the above are settled, rather than write it against a shape that may change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added warnings when commands launched from an activated shell may not receive generated PATH updates.
  * Added detection for Windows Command Prompt PATH values exceeding supported length limits.
  * Improved PATH length measurement for Unicode characters.
  * Added checks for externally configured PATH values.
  * Clarified when PATH settings may be ignored or require adjustment.

### DIFF
--- a/src/cli/hook_env.rs
+++ b/src/cli/hook_env.rs
@@ -442,6 +442,10 @@ impl HookEnv {
         )?
         .to_string_lossy()
         .into_owned();
+        // This PATH goes to the user's own shell rather than to a computed child env, so
+        // `PathEnv::join`'s check never sees it — and it is the copy a tool started directly
+        // from an activated shell inherits.
+        crate::path_env::warn_if_cmd_ignores_path_str(&new_path);
         let mut ops = vec![EnvDiffOperation::Add(PATH_KEY.to_string(), new_path)];
 
         // For DIRENV_DIFF, we need to include both filtered user_paths and filtered tool_paths
--- a/src/path_env.rs
+++ b/src/path_env.rs
@@ -2,11 +2,12 @@ use crate::config::Settings;
 use crate::dirs;
 use std::collections::HashSet;
 use std::env::{join_paths, split_paths};
-use std::ffi::OsString;
+use std::ffi::{OsStr, OsString};
 use std::fmt;
 use std::fmt::{Display, Formatter};
 use std::path::PathBuf;
 use std::str::FromStr;
+use std::sync::atomic::{AtomicBool, Ordering};
 
 pub struct PathEnv {
     pre: Vec<PathBuf>,
@@ -68,7 +69,9 @@ impl PathEnv {
     }
 
     pub fn join(&self) -> OsString {
-        join_paths(self.to_vec()).unwrap()
+        let joined = join_paths(self.to_vec()).unwrap();
+        warn_if_cmd_ignores_path(&joined);
+        joined
     }
 
     /// [`Self::join`] over the verbatim projection — see [`Self::to_vec_verbatim`].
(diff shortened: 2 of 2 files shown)


JamBalaya56562:
Both halves taken.

**The double warning is real, and it is the common path rather than an edge case.** One `mise hook-env` run measures two PATHs: `ts.env_with_path_and_split` goes through `PathEnv::join`, and then hook-env assembles the shell's own PATH by hand. Those are different lengths, so keying `warn_once!` on the formatted message reports the same problem twice on the surface that is already the noisiest. Now keyed on the condition:

```rust
static WARNED_CMD_PATH_LIMIT: AtomicBool = AtomicBool::new(false);
...
if WARNED_CMD_PATH_LIMIT.swap(true, Ordering::Relaxed) {
    return;
}
```

That removes the first of the open questions in the description, so I have edited it out.

**"once per process" was accurate and reassuring in the wrong way**, since under `mise activate` the process is a fresh one on every prompt. Reworded to say what actually happens rather than leaving the reader to work it out:

> At most once per **invocation**, which is not once per user-visible event: under `mise activate`, `hook-env` is a fresh process on every prompt, so a PATH left over the limit warns on every prompt. Whether that is the right cadence is an open question on this PR.

That cadence is still the thing I most want a decision on — the guard above makes it one warning per prompt instead of two, which is an improvement rather than an answer.

While in there I also corrected the constant's doc comment, which claimed "nothing on PATH resolves and every command reports is not recognized". Programs in the system directory keep working, because `cmd.exe` finds those without consulting PATH — the same overclaim reviewers caught in #11642, and the reason the probe documented there never failed.

PR #11648, opened 2026-08-02
feat(config): give `mise set` the short -f its neighbours already have

JamBalaya56562:
Follow-up to #5501, whose second observation was:

> `mise use` has an option `-p .` to create the config file when it does not exist, but `mise set` seems to miss such an option

The long-name half of that landed in #11577 — `mise set --path` exists now. This closes the rest.

## `mise set` was the only one without a short flag

Now that every command naming a config file takes both `--file` and `--path` (#11577, #11616, #11631, #11640), the short flags line up like this:

| command | short |
|---|---|
| `use`, `unuse` | `-p` |
| `unset`, `config get`, `config set` | `-f` |
| `dotfiles add`, the four `bootstrap packages` commands | `-p` |
| **`set`** | **none** |

The letter follows the canonical name, so `set` — whose canonical name is `--file` — gets `-f`, matching `unset` and the two `config` commands.

`-f` is free on subcommands: the root `--force` is `#[clap(long, short, hide = true)]` without `global = true`, so it does not propagate, and the globals that do are `-C -E -j -P -q -v -y`. `mise set` itself only uses `-E` and `-g`. The three commands above already prove it.

## Not adding `-p`

#5501 asks for `-p` specifically, and `visible_short_alias = 'p'` would give exact parity with `mise use -p .`. I left it out because no command in this group carries a short alias — adding one to `set` alone makes it the only command with four spellings, which trades the current asymmetry for a different one. `--path` already gives the reporter the same words on both commands; happy to add the short alias if you would rather have full parity.

## Tests

`test_config_target_options_accept_both_names` in `src/cli/mod.rs` already walks all eleven commands asserting the visible alias. It now also asserts a short flag exists — existence only, since the letter legitimately differs by canonical name. That is exactly the invariant `set` was violating.

`e2e/cli/test_config_target_aliases` exercises `mise set -f` alongside the existing `--file` and `--path` cases.

No behavior changes beyond the new spelling.

## Summary by CodeRabbit

* **New Features**
  * Added `-f` as a short alias for the `mise set --file` option.
  * The alias works alongside the existing `--file` and `--path` options.

* **Documentation**
  * Updated CLI documentation, generated help, and the manual to include the new `-f` alias.

* **Tests**
  * Added coverage confirming `mise set -f` writes variables to the specified configuration file.

### DIFF
--- a/docs/cli/set.md
+++ b/docs/cli/set.md
@@ -27,6 +27,14 @@ e.g.: NODE_ENV=production
 
 Create/modify an environment-specific config file like .mise.&lt;env>.toml
 
+### `-f --file <FILE>`
+
+The TOML file to update
+
+Can be a file path or directory. If a directory is provided, will create/use mise.toml in that directory.
+Defaults to [`MISE_DEFAULT_CONFIG_FILENAME`](https://mise.jdx.dev/configuration.html#mise_default_config_filename) environment variable, or `mise.toml`.
+Use [`MISE_GLOBAL_CONFIG_FILE`](https://mise.jdx.dev/configuration.html#mise_global_config_file) to choose a different global config path.
+
 ### `-g --global`
 
 Set the environment variable in the global config file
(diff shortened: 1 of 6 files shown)


jdx:
not sure about this, it could make users think `--force`

PR #11662, opened 2026-08-03
fix(pipx): prefer RFC3339 upload_time_iso_8601 for release-age gating

Guria:
## Problem

`minimum_release_age` over-gates freshly-released pipx packages. A release uploaded at `10:05 UTC` is recorded as `23:59:59 UTC` the same day, so it appears up to ~24h younger than it really is and the "eligible" time is pushed into the future. The cause: the pipx backend reads PyPI's **timezone-naive** `upload_time`, which `parse_into_timestamp` parses as a `civil::Date` (discarding the time-of-day) and then substitutes end-of-day UTC.

Sorry for omitting Discord discussion step. It is blocked at country of my location. I hope this fails into exception:
> unless it is something obvious

## Reproduction

`pipx:notebooklm-py` 0.8.0 was uploaded to PyPI at `2026-08-03T10:05:14Z` (confirmed via the JSON API):

```sh
$ curl -s https://pypi.org/pypi/notebooklm-py/json | jq -r '.releases["0.8.0"][].upload_time_iso_8601'
2026-08-03T10:05:14.723989Z
2026-08-03T10:05:16.496081Z
```

With `minimum_release_age = "4h"`:

```toml
[tools]
"pipx:notebooklm-py" = { version = "latest", minimum_release_age = "4h" }
```

`mise upgrade` reports:

```text
mise WARN  newer pipx:notebooklm-py release 0.8.0 (released 2026-08-03, eligible 2026-08-04 03:59 UTC) ignored by minimum_release_age (4h); latest eligible release is 0.7.3
```

`eligible = created_at + age`, so `2026-08-04 03:59 UTC − 4h = 2026-08-03 23:59:59 UTC` is the `created_at` mise holds. The real upload at `10:05:14` would make it eligible at `14:05 UTC` the same day — so the gate waits until ~04:00 the next day instead. For contrast, the aqua backend reads a real RFC3339 timestamp from the GitHub releases API and lands on the precise instant.

## Root cause

`src/backend/pipx.rs` deserialized PyPI's naive `upload_time`:

```rust
struct PypiRelease {
    upload_time: Option<String>,   // "2026-08-03T10:05:14" — no offset
    ...
}
```

PyPI provides two fields per file:

- `upload_time` — `"2026-08-03T10:05:14"` — **timezone-naive, no `Z`**
- `upload_time_iso_8601` — `"2026-08-03T10:05:14.723989Z"` — RFC3339, precise

`parse_into_timestamp` (`src/duration.rs`) then fails the RFC3339 and Zoned branches (no offset/zone) and falls through to the date-only branch:

```rust
if let Ok(civil_date) = s.parse::<jiff::civil::Date>() {        // succeeds, drops 10:05:14
    let datetime = civil_date.at(23, 59, 59, 0);                // the 23:59:59
    ...
}
```

jiff's temporal parser parses a `civil::Date` from anything containing a date — per [the jiff temporal docs](https://docs.rs/jiff/0.2/jiff/fmt/temporal/index.html):

> "Smaller types can generally be parsed from strings representing a bigger type. For example, a `civil::Date` can be parsed from `2020-08-21T02:21:58`."

So `"2026-08-03T10:05:14"` parses to `Date(2026-08-03)`, discarding `10:05:14`, and the `.at(23, 59, 59, 0)` fallback substitutes end-of-day UTC.

## Fix

Prefer `upload_time_iso_8601` (the `…Z` field), so it parses as RFC3339 on the first branch of `parse_into_timestamp`. Keep `upload_time` as a fallback for custom indexes that don't provide the ISO field. Preserves the existing `.min()`-across-files semantics and yanked-file filtering — only the timestamp *source* changes.

## Tests

Adds `test_versions_from_pypi_package_preserves_time_of_day_from_iso_field`, asserting:

- both fields present → `created_at_timestamp()` is the real instant (`2024-01-02T10:05:14.723989Z`), not end-of-day;
- ISO field absent → naive fallback still yields a timestamp (`2024-01-03T23:59:59Z`), so release-age gating degrades gracefully on indexes without the ISO field.

All 17 existing + new pipx backend tests pass.

## Scope / impact

Any pipx package whose `upload_time` carries a time-of-day (essentially all of PyPI) is gated more strictly than configured while it is less than ~24h old. Once a release is a day old the date is correct and the cooldown behaves as intended, so the effect is confined to the release edge. `upload_time_iso_8601` is a [documented field](https://docs.pypi.org/api-index-api/json/) on PyPI's JSON API.

---

Tested locally against `2026.8.0`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PyPI release date accuracy by prioritizing precise upload timestamps.
  * Added reliable fallback handling when precise timestamp information is unavailable.
  * Corrected release ordering across timestamps with different time zone offsets.
  * Limited date selection to valid, non-yanked release files for more dependable results.
  * Expanded validation and test coverage for precise timestamps, fallback scenarios, and updated release data.

### DIFF
--- a/src/backend/pipx.rs
+++ b/src/backend/pipx.rs
@@ -6,6 +6,7 @@ use crate::cache::{CacheManager, CacheManagerBuilder};
 use crate::cli::args::BackendArg;
 use crate::cmd::CmdLineRunner;
 use crate::config::{Config, Settings};
+use crate::duration::parse_into_timestamp;
 #[cfg(unix)]
 use crate::env;
 #[cfg(unix)]
@@ -461,22 +462,28 @@ impl PIPXBackend {
             .filter(|(_, files)| files.iter().any(|f| !f.yanked))
             .sorted_by_cached_key(|(v, _)| Versioning::new(v))
             .map(|(version, files)| {
+                // Prefer the RFC3339 `upload_time_iso_8601` over the
+                // timezone-naive `upload_time`: the latter has no offset, so
+                // `parse_into_timestamp` parses it as a `civil::Date` and
+                // substitutes end-of-day UTC, dropping the real time-of-day
+                // and inflating `minimum_release_age` for same-day releases by
+                // up to ~24h. Custom indexes without the ISO field fall back to
+                // `upload_time` as before.
                 let created_at = files
                     .iter()
                     .filter(|f| !f.yanked)
-                    // Prefer the RFC3339 `upload_time_iso_8601` over the
-                    // timezone-naive `upload_time`: the latter has no offset,
-                    // so `parse_into_timestamp` parses it as a `civil::Date`
-                    // and substitutes end-of-day UTC, dropping the real
-                    // time-of-day and inflating `minimum_release_age` for
-                    // same-day releases by up to ~24h. Custom indexes without
-                    // the ISO field fall back to `upload_time` as before.
                     .filter_map(|f| {
                         f.upload_time_iso_8601
                             .clone()
                             .or_else(|| f.upload_time.clone())
                     })
-                    .min();
+                    // Pick the earliest upload as a parsed instant, not by
+                    // lexicographic string order: RFC3339 strings with
+                    // different offsets (`...00:00-05:00` vs `...04:00Z`) do
+                    // not sort chronologically, so a naive `.min()` on the raw
+                    // strings can select a later instant and over-gate.
+                    .min_by_key(|s| parse_into_timestamp(s).unwrap_or(Timestamp::MAX));
+
                 VersionInfo {
                     version,
                     created_at,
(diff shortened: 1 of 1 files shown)


Guria:
Both review comments (CodeRabbit @ 479, Codex P2) flagged the same issue: the earlier commit selected the earliest upload with `.min()` over the raw RFC3339 strings, which compares lexicographically rather than as UTC instants. RFC3339 strings with differing offsets do not sort chronologically — e.g. `"2024-01-02T00:00:00-05:00"` (= `05:00Z`) sorts before `"2024-01-02T04:30:00Z"` by string, but is 30 min later in UTC — so a custom index returning mixed offsets could pick a later instant and over-apply `minimum_release_age`.

Addressed in 3166eb451: candidates are now parsed to `jiff::Timestamp` and the earliest instant is selected (`min_by_key(|s| parse_into_timestamp(s).unwrap_or(Timestamp::MAX)`); unparseable values sort last). Added `test_versions_from_pypi_package_picks_earliest_instant_not_lexical_min`. On real PyPI (all-`Z`, same-second uploads) selection is unchanged; 18 pipx tests pass locally.

PR #11682, opened 2026-08-03
fix(sync): reconcile external provider links

risu729:
## Summary

- move external-provider reconciliation out of `Backend` and into the `sync` CLI
- preserve managed installs, runtime aliases, and links owned by another provider
- serialize every per-version inspection and mutation with the install-state lock
- identify Homebrew ownership through the immediate `opt` target, including dangling entries
- remove the legacy all-links-first cleanup helper replaced by reconciliation

The backend layer remains provider-agnostic; provider discovery and ownership policy stay in the CLI commands that know about Homebrew, nvm, nodenv, pyenv, and uv.

## Stack

This PR now sits directly on `main`. It builds on the filesystem-safe link ownership and removal primitives merged in #11715.

#11686 is superseded because per-version locking is required for reconciliation to be safe and is included here.

## Bug

The old sync flow removed every link matching a provider prefix before rebuilding desired links. It could delete links owned by another source and exposed a race where a concurrent install or `mise link --force` changed an entry after classification but before deletion or replacement.

Homebrew adds another edge case: mise links to stable entries under `opt`, which resolve into `Cellar` while installed but become dangling under `opt` after unlink or uninstall. Direct user links into Cellar must remain untouched.

### Concrete multi-provider failure

Suppose nvm has Node 23, nodenv no longer has Node 23, and mise still has a dangling link left by nodenv:

```text
$NVM_DIR/versions/node/v23.0.0                 # valid nvm install
$NODENV_ROOT/versions/23.0.0                   # absent
$MISE_DATA_DIR/installs/node/23.0.0
  -> $NODENV_ROOT/versions/23.0.0               # stale, dangling nodenv link
```

Running `mise sync node --nvm --nodenv` previously processed the providers sequentially:

1. nvm saw an existing link owned by another provider and preserved it instead of linking Node 23 from nvm.
2. nodenv then recognized its own link as stale and removed it.
3. The command succeeded but left no mise link for Node 23, even though nvm had a valid install.

This PR gathers the desired links and ownership namespaces for all selected providers before mutating anything. Under the per-version lock, the stale link is known to belong to a selected provider, nvm wins the documented provider precedence, and the link is replaced with `$NVM_DIR/versions/node/v23.0.0`. Managed installs, runtime aliases, and links from providers that were not selected remain untouched.

The final block of `e2e/sync/test_sync_nvm` reproduces this exact dangling-nodenv-link scenario and verifies that the resulting link points to nvm. `e2e/sync/test_sync_python_uv` covers the equivalent pyenv/uv case.

## Testing

- `mise run lint-fix`
- `cargo test --all-features --bin mise cli::sync::reconcile::tests`
- `mise run test:e2e e2e/sync/test_sync_nvm e2e/sync/test_sync_python_uv`
- full GitHub Actions test matrix, including Linux e2e, Windows build/unit/e2e, macOS unit, Clippy, MSRV, and generated-file checks

## Summary by CodeRabbit

* **New Features**
  * Improved runtime synchronization for Node, Python, and Ruby across multiple runtime managers.
  * Preserves aliases, managed installations, and externally maintained links.
  * Applies configured version ordering and reports synchronized versions.
  * Improves handling of incomplete or unrecognized installation directories.
  * Supports more reliable executable detection, including Windows command resolution.

* **Bug Fixes**
  * Prevents valid links and external installations from being overwritten or removed.
  * Cleans up stale provider-owned links.
  * Improves repeated synchronization after interrupted installations.

* **Documentation**
  * Clarified synchronization behavior and remote bootstrap options.

### DIFF
--- a/docs/cli/sync/node.md
+++ b/docs/cli/sync/node.md
@@ -9,7 +9,7 @@ Symlinks all tool versions from an external tool into mise
 
 For example, use this to import all Homebrew node installs into mise
 
-This won't overwrite any existing installs but will overwrite any existing symlinks
+This won't overwrite managed installs, runtime aliases, or links from other providers.
 
 ## Flags
 
--- a/docs/cli/sync/python.md
+++ b/docs/cli/sync/python.md
@@ -9,7 +9,7 @@ Symlinks all tool versions from an external tool into mise
 
 For example, use this to import all pyenv installs into mise
 
-This won't overwrite any existing installs but will overwrite any existing symlinks
+This won't overwrite managed installs, runtime aliases, or links from other providers.
 
 ## Flags
 
--- a/e2e/sync/test_sync_nvm
+++ b/e2e/sync/test_sync_nvm
@@ -17,7 +17,7 @@ ln -s ./18.0.0 "$MISE_DATA_DIR/installs/node/provider-runtime-alias"
 install_fake_node "$NVM_DIR/versions/node/v20.0.0"
 install_fake_node "$NVM_DIR/versions/node/v21.0.0"
 install_fake_node "$NVM_DIR/versions/node/v22.0.0"
-install_fake_node "$MISE_DATA_DIR/installs/node/20.0.0/bin"
+install_fake_node "$MISE_DATA_DIR/installs/node/20.0.0"
 install_fake_node "$PWD/other-provider/21.0.0"
 ln -s "$PWD/other-provider/21.0.0" "$MISE_DATA_DIR/installs/node/21.0.0"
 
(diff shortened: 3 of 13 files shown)


risu729:
@coderabbitai resume

risu729:
Reviewed the latest AI feedback against `upstream/main...HEAD`:

- CodeRabbit's `is_spawnable` directory check is not in this PR's diff; that function came from current `main` in `619854b468` / #12178. It may be a valid mainline follow-up, but changing it here would mix unrelated system-dependency behavior into this sync PR.
- The `trust --untrust` argument wording is also unchanged from current `main` and absent from this PR's diff. It should likewise be handled separately rather than adding unrelated generated-help changes here.
- Greptile's latest review reports no blocking finding and a 5/5 confidence score.

No sync code change is needed for those findings. The PR description now includes a concrete nvm/nodenv failure sequence. The existing `e2e/sync/test_sync_nvm` regression reproduces that exact dangling-link/provider-precedence case, and `e2e/sync/test_sync_python_uv` covers the equivalent Python case.

risu729:
@coderabbitai resume

PR #11686, opened 2026-08-03
fix(sync): lock provider reconciliation per version

risu729:
## Summary

- acquire the existing tool-version transaction lock before classifying or mutating each provider-linked version
- release each lock before moving to the next version instead of holding a whole batch of locks

## Bug

Provider reconciliation in #11682 checks who owns an install path and then may remove or replace it. Without the same version lock used by install, uninstall, and `mise link`, another process can change that path after the ownership check but before sync mutates it.

For example:

```text
process A: mise sync node --nvm sees node@22 as an nvm-owned link
process B: mise install --force node@22 replaces that path
process A: sync continues using its stale ownership decision
```

Before this fix, sync can fail or mutate a path whose ownership changed concurrently. After this fix, both operations serialize on the `node@22` transaction lock. Other versions remain independent and are not locked as a batch.

## Stack

This PR is stacked on #11682, which introduces provider-owned reconciliation. Review the final commit for the locking change; the parent diff will disappear after #11682 merges and this branch is rebased onto `main`.

#11687 builds on this guarantee to aggregate marker cleanup errors without losing successfully reconciled versions.

## Testing

- `mise run lint-fix`
- `mise run test:e2e e2e/sync/test_sync_nvm`

### DIFF
--- a/e2e/sync/test_sync_nvm
+++ b/e2e/sync/test_sync_nvm
@@ -12,13 +12,27 @@ install_fake_node "$NVM_DIR/versions/node/v18.0.0"
 mise sync node --nvm
 mise ls
 assert_contains "mise ls node --json | jq -r '.[] | .symlinked_to | select( . != null)'" "node/18.0.0"
+ln -s ./18.0.0 "$MISE_DATA_DIR/installs/node/provider-runtime-alias"
 
 install_fake_node "$NVM_DIR/versions/node/v20.0.0"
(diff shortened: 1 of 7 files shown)


risu729:
Superseded by #11682: reconciliation now acquires the per-version install lock before reclassification and holds it through every mutation and marker update. Keeping the lock in the reconciliation PR avoids an unsafe intermediate state.\n\n*AI-assisted — Tool: Codex; model: unavailable; version: unavailable.*

PR #11720, opened 2026-08-05
fix(backend): resolve remote versions with request options

risu729:
## What

Use the options from the active `ToolRequest` when selecting remote version candidates. This covers normal, prefix, sub-version, latest, date-filtered, upgrade, and outdated remote resolution paths.

Configured and inline options are layered before `latest` and `ls-remote` resolve `sub-N` requests. Obsolete request-unaware remote resolution wrappers are removed after their callers and tests are migrated.

## Why

Backend resolution previously re-read options from `Config`, whose backend-wide lookup can select the first configured request. In a multi-version declaration such as:

```toml
[tools]
tool = [
  { version = "1", prerelease = false },
  { version = "2", prerelease = true },
]
```

the second request could silently use the first request selection policy.

This is a pre-existing bug found while preparing #11481.

## Stack

Stacked on #12060, which preserves request-option provenance while backend, config, request, and inline options are layered. This PR consumes those preserved effective options when selecting remote versions.

The two PRs both modify `ToolRequest` and `ToolVersion` and conflict if applied independently. Review and merge #12060 first; this branch must then be rebased onto it while retaining both #12060's explicit-overlay tracking and this PR's request-aware selection.

## Scope

This PR is intentionally limited to remote and sub-request option propagation. Installed-version hardening, upgrade/outdated selector behavior, and lockfile validation remain separate changes.

## Test

- `mise run format`
- `cargo check --locked --tests`
- `mise run test:unit -- test_new_opts_applies_options_to_sub_requests`

## Summary by CodeRabbit

* **Enhancements**
  * Version resolution now consistently respects request-specific options across remote, latest, exact-match, prefix, and date-filtered lookups.
  * Improved handling for stable and prerelease versions, aliases, rolling channels, and sub-version requests.
  * Java listings more accurately select GA or EA metadata.
  * Remote caching, refresh behavior, registry isolation, and offline listing remain separated by selection options.

* **Bug Fixes**
  * Request options are now preserved for sub-, path-, and system-based requests.

### DIFF
--- a/src/backend/conda.rs
+++ b/src/backend/conda.rs
@@ -703,17 +703,17 @@ impl Backend for CondaBackend {
     }
 
     /// Override to bypass the shared remote_versions cache since conda's
-    /// channel option affects which versions are available. The override is
-    /// on `_with_refresh` so it applies to both cached and refresh-enabled
-    /// resolution paths; conda always queries the channel directly so the
-    /// `_refresh` flag is irrelevant.
-    async fn list_remote_versions_with_info_with_refresh(
+    /// channel option affects which versions are available. Conda always
+    /// queries the channel directly, so the `_refresh` flag is irrelevant.
+    async fn list_remote_versions_with_info_and_options(
         &self,
         config: &Arc<Config>,
+        _listing_opts: &ToolVersionOptions,
+        selection_opts: &ToolVersionOptions,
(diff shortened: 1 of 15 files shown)


risu729:
@coderabbitai resume

PR #11764, opened 2026-08-07
fix(aqua): validate registry file names against path traversal

bunlongheng:
the aqua backend joins registry-supplied file src/name and link values straight onto the install path without checking them, in install() (bin_names -> install_path.join) and file_link_for_version() (src join plus the link/name dst join that feeds create_file_link). a registry entry with files: [{name: "../../evil"}] or a link of ../../evil would copy, symlink, or hard-link outside the install dir. the http backend already guards the same kind of input with ensure_safe_relative_bin_path, so this applies that existing helper to the aqua values before any join. legit forms like src: bin/tool still pass since the check only rejects absolute paths and parent components. added a unit test covering both the src and link cases.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Aqua backend installation security by rejecting unsafe binary names and file or link paths.
  * Prevented path traversal through registry source and link values.
  * Added regression coverage to verify invalid paths are rejected.

### DIFF
--- a/src/backend/aqua.rs
+++ b/src/backend/aqua.rs
@@ -3,7 +3,7 @@ use crate::backend::backend_type::BackendType;
 use crate::backend::options::BackendOptions;
 
 use crate::backend::platform_target::PlatformTarget;
-use crate::backend::static_helpers::get_filename_from_url;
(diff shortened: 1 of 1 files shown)


jdx:
I don't think this is necessary

PR #11768, opened 2026-08-07
fix(task): handle azure devops remote git includes

cheesemans:
Azure Devops uses a different URL scheme for git repositories than most other git servers. It looks something like https://dev.azure/organisation/project/_git/repository, which is typically not suffixed by `.git` (unless the literal repo name ends with `.git`). Mise currently expects the url to end with `.git` so naturally these repositories do not work as remote git sources for tasks. I wrote #11660 where I tried demonstrating this.

I opted for adding a second regex for matching the azure devops url pattern, felt easier to get it right than modifying the existing one to match both patterns.

I've built and tested this against my company instance of Azure Devops. I still have some issues getting the git include feature to fully work. One issue regarding pack files when using ssh (issue has been patched in next version of gitoxide, whenever that drops) and some issues with https auth which I have not figured out yet (likely also some issue with gitoxide though). 

## Summary by CodeRabbit

* **New Features**
  * Added support for Azure DevOps Git repositories using SSH and HTTPS URLs.
  * Recognizes organization, project, repository, branch, credentials, ports, and task-file paths.
  * Supports common URL variations, including additional query parameters and optional SSH users.

* **Bug Fixes**
  * Improved Git source parsing for Azure DevOps repositories.
  * Prevents invalid or incomplete repository and task-file paths from being accepted.
  * Ensures repository caching distinguishes branches while reusing entries across task files.

### DIFF
--- a/src/remote_source.rs
+++ b/src/remote_source.rs
@@ -146,7 +146,7 @@ mod tests {
     }
 
     #[test]
-    fn parses_alt_git_ssh_sources() {
+    fn parses_azure_devops_git_ssh_sources() {
         let source = RemoteSource::parse_git(
             "git::ssh://git@dev.azure/myorg/myproj/_git/example//terraform/myfile?ref=master",
         )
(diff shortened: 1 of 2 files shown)


jdx:
wouldn't be microsoft if it wasn't both different and worse

PR #11814, opened 2026-08-09
feat(task): align task output prefixes to a common width

JamBalaya56562:
Closes #4397.

## The problem

Running tasks concurrently prefixes each output line with the task name, but the prefixes are not padded — so output the tasks themselves have carefully aligned stops lining up as soon as the names differ in length. Measured on **v2026.8.2**:

```console
$ mise run all
[a] $ echo one
[averylongtaskname] $ echo three
[a] one
[averylongtaskname] three
[a] $ echo two
```

`task.output` offers `prefix`, `interleave`, `keep-order`, `replacing` and `timed`, and none of them pads; there is no setting for it either.

## The change

Right-align each prefix within the run's widest:

```console
              [a] $ echo one
[averylongtaskname] $ echo three
              [a] one
[averylongtaskname] three
```

The width is computed in `cli/run.rs` where the whole dependency graph is already in scope — four lines below the same `tasks.all()` iteration that decides whether to spawn the timed-output printer — and rides along on `OutputHandler`, so the executor, the progress reporter and the failure summary all use the same number rather than each deriving one.

Ordering matters here and works out: `Deps::mark_ambiguous_prefixes()`, which can append args to a prefix to disambiguate `[test-docker 4.1]` from `[test-docker 4.2]`, runs while the graph is built, before the width is measured.

## Why the padding goes *before* the bracket

This is the part worth reviewing. I first implemented it the conventional way — left-aligned name, trailing spaces — and it broke three of mise's own e2e tests:

- `tasks/test_task_delegation_dedup` filters the command echo with `grep -v '^\[setup\] \$'`
- `tasks/test_task_deps` asserts on `"[hello:1] Hello from 1"`
- `tasks/test_task_nested_quiet` asserts on `"[dep1] dep1"`

All three break for the same reason: trailing padding lands *between* `[name]` and the line it prefixes. If mise's own test suite depends on `[name] output` being contiguous in three places, user scripts and log pipelines plausibly do too.

Right-aligning avoids that entirely. `[a] one` remains a contiguous substring, every existing matcher keeps working, and the output still lines up — which was the actual request. No e2e test needed changing.

The cost is a ragged left edge. If you would rather have the conventional look and accept the output-format break, the change is a one-line swap in `estyled_prefix_padded` plus updating those three tests.

## Other properties

- **A single-task run is unchanged**, because the widest prefix is its own. Padding only appears when more than one task runs — which is the only time it helps.
- **Padding carries no color**: it sits outside the styled span.
- **The width is bounded**: `Task::prefix` already truncates to 40 characters, so the pad can never exceed 42 columns.
- Width is measured with `console::measure_text_width` rather than counting characters, so a task named in CJK lines up with the rest.

## No new setting

Deliberately: adding one would mean regenerating `schema/mise.json` (and `mise.usage.kdl` / `man` / `docs/cli` if a flag came with it), as #11788 did. This change is confined to `src/`. If you would rather have it behind a setting I am happy to add one — just say which default you want.

## Tests

Four unit tests in `src/task/mod.rs`:

- the short prefix pads to exactly the run width, the pad is leading, and stripping it reproduces the unpadded form
- padding never truncates, and a single-task width is a no-op
- `max_prefix_width` picks the widest and returns 0 for an empty run
- the width is display width, not character count (CJK)

`Task::estyled_prefix()` is gone — `estyled_prefix_padded(0)` is the same thing, and keeping both left the old one dead in non-test builds.

No e2e: alignment assertions would be tied to ANSI handling and terminal width, which makes for a brittle test.

## Summary by CodeRabbit

* **Enhancements**
  * Improved task output alignment by padding task prefixes to a consistent width.
  * Error messages, progress updates, and failure summaries now line up more clearly.
  * Alignment accounts for display widths of truncated and non-Latin task names, ensuring consistent formatting across varied task names.
  * Improved readability when monitoring multiple tasks or reviewing failures.

### DIFF
--- a/src/cli/run.rs
+++ b/src/cli/run.rs
@@ -1070,7 +1070,12 @@ impl Run {
                     Error::get_exit_status(err)
                 };
                 if !interrupted && !this.is_stopping() && (panicked || status.is_none()) {
-                    let prefix = task.estyled_prefix();
+                    let prefix = task.estyled_prefix_padded(
+                        this.output_handler
+                            .as_ref()
+                            .map(|h| h.prefix_width())
+                            .unwrap_or(0),
+                    );
                     if Settings::get().verbose {
                         this.eprint(&task, &prefix, &format!("{} {err:?}", style::ered("ERROR")));
                     } else {
@@ -1169,6 +1174,9 @@ impl Run {
             raw: self.raw,
             is_linear: self.is_linear,
             jobs: self.jobs,
+            // Computed once here, where the whole run is known, so every task's
+            // prefix can be padded to the same width (discussion #4397).
+            prefix_width: crate::task::max_prefix_width(tasks.all()),
         };
         self.output_handler = Some(OutputHandler::new(output_config));
 
--- a/src/task/mod.rs
+++ b/src/task/mod.rs
@@ -2080,8 +2080,24 @@ impl Task {
         Ok(ctx.render_spec()?)
     }
 
-    pub fn estyled_prefix(&self) -> String {
-        style::prefix(self.prefix(), &self.display_name, true)
+    /// The task's styled `[name]` prefix, right-aligned within `width` so that
+    /// the output of every task in a run starts at the same column
+    /// (discussion #4397).
+    ///
+    /// Padding goes *before* the bracket, not after it. That keeps `[name] ` and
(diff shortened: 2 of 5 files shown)


jdx:
doesn't look like an improvement to me

PR #11851, opened 2026-08-10
fix(http): honor system install destinations

Marukome0743:
## Summary

- honor the destination selected by `mise install --system` and `mise install --shared` in the HTTP backend
- extract explicit system, shared, and `install-into` destinations directly instead of linking them to a persistent cache
- preserve normal user installs' existing cross-tool cache and symlink behavior
- resolve relative shared paths before installation and keep content-derived `latest` aliases inside the selected destination

## Root cause

The shared install pipeline supplied the requested destination through `ToolVersion.install_path`, but the HTTP backend rebuilt its install path from the primary user directory. It therefore left the pre-created system version directory empty and created the actual cache symlink under the user's install tree.

The initial destination-aware fix moved that cache beside each destination tool. Review identified that this still left three problems: relative `--shared` paths produced relative symlink targets, uninstalling a version left the hidden extraction cache behind, and `install-into` still depended on the invoking user's data directory.

## Behavior

Explicit system, shared, and `install-into` destinations now contain real extracted files. Extraction uses a temporary sibling directory followed by an atomic rename, so failed installs do not expose partial destinations. This removes the destination-local cache leak and makes those installations independent of the installing user's home directory.

Normal user installs retain the existing shared cache in `$MISE_DATA_DIR/http-tarballs` and continue to use symlinks for cross-tool deduplication. A normal forced install remains a user install even when an existing shared copy is discoverable, and it does not alter that shared copy.

Relative `--shared` paths are absolutized before they are assigned to tool versions. `latest` system installs keep a real content-derived directory plus the normal `latest` alias within the system tree. Concrete-version lookup continues to use the common shared/system discovery path.

Exact `install-into` destinations also bypass fuzzy runtime-path remapping during binary discovery. An existing normal installation and its runtime alias therefore cannot redirect validation or execution away from the newly populated destination.

## Verification

- `mise exec -- cargo test --all-features backend::http::tests` — 17 passed
- `mise exec -- cargo test --all-features test_runtime_path_for_exact_install_path_skips_runtime_alias` — passed
- `mise run test:e2e e2e/backend/test_http_system_install` — passed
  - covers archive and raw-file system installs, shared installs, relative `--shared`, normal force with an existing shared install, `latest --system`, and self-contained `install-into` in the presence of an existing fuzzy runtime alias
  - asserts exact JSON `install_path` values rather than matching tool names
- `mise exec -- cargo check --all-features` — passed
- `mise exec -- cargo clippy --workspace --all-features --all-targets -- -D warnings` — passed
- Rustfmt, shfmt, and ShellCheck passed individually

`mise run format` and `mise run lint` cannot run their hk wrapper in this Sapling working copy because hk requires a Git worktree. The relevant format and lint checks above were run directly.

Addresses https://github.com/jdx/mise/discussions/9475

### DIFF
--- a/docs/dev-tools/backends/http.md
+++ b/docs/dev-tools/backends/http.md
@@ -461,10 +461,17 @@ The HTTP backend implements an intelligent caching system to optimize disk usage
 
 ### Cache Location
 
-Downloaded and extracted files are cached in `$MISE_CACHE_DIR/http-tarballs/` instead of being stored separately for each tool installation. By default:
+For normal user installations, downloaded and extracted files are cached in
+`$MISE_DATA_DIR/http-tarballs/` instead of being stored separately for each tool
+installation. By default:
 
-- **Linux**: `~/.cache/mise/http-tarballs/`
-- **macOS**: `~/Library/Caches/mise/http-tarballs/`
+- **Linux**: `~/.local/share/mise/http-tarballs/`
+- **macOS**: `~/.local/share/mise/http-tarballs/`
+
+Explicit `mise install --system`, `mise install --shared`, and `mise
+install-into` installations are extracted directly into their destination.
+They do not use this persistent extraction cache, so the resulting installation
+is self-contained and does not link to the installing user's home directory.
 
 ### Cache Key Generation
 
@@ -476,30 +483,33 @@ Cache keys are generated based on the file content to ensure identical downloads
 Example cache directory structure:
 
 ```
-~/.cache/mise/http-tarballs/
+~/.local/share/mise/http-tarballs/
 ├── 71f774faa03daf1a58cc3339f8c73e6557348c8e0a2f3fb8148cc26e26bad83f/
-│   ├── extracted/
-│   │   └── bin/my-tool
+│   ├── bin/my-tool
 │   └── metadata.json
 └── 1c2af379bdf1fed266bc44b49271e2df5b0dafae09f1cc744b3505ec50c84719_strip_1/
-    ├── extracted/
-    │   └── my-tool
+    ├── my-tool
     └── metadata.json
 ```
 
 ### Symlinked Installations
 
-Tool installations are symlinks to the cached extracted content:
+Normal user installations are symlinks to the cached extracted content:
 
 ```bash
-~/.local/share/mise/installs/http-my-tool/1.0.0 → ~/.cache/mise/http-tarballs/71f774.../extracted
+~/.local/share/mise/installs/http-my-tool/1.0.0 → ~/.local/share/mise/http-tarballs/71f774...
 ```
 
 This approach provides several benefits:
 
-- **Space efficiency**: Multiple tools using the same tarball share a single cached copy
+- **Space efficiency**: Normal user installs share identical tarballs across tools
 - **Faster installations**: Cache hits avoid re-downloading and re-extracting files
 - **Consistency**: Identical file content always uses the same cache entry
+
+System, shared, and install-into destinations contain real files rather than
+these symlinks. This avoids leaving hidden cache entries behind after an
+uninstall and keeps shared installations independent of a specific user's data
+directory.
 
 ### Cache Metadata
 
@@ -517,9 +527,10 @@ Each cache entry includes a `metadata.json` file with information about the cach
 
 ### Cache Management
 
-The HTTP backend cache follows mise's standard cache management:
+Normal HTTP installations store their cache in
+`$MISE_DATA_DIR/http-tarballs/`. It is intentionally outside `MISE_CACHE_DIR`,
+so `mise cache clear` does not remove content that installed symlinks still
+reference.
 
-- Cache entries can be cleared with `mise cache clear`
-- The cache directory respects the `MISE_CACHE_DIR` environment variable
-- **Autopruner**: mise automatically cleans up unused cache entries after 30 days of inactivity
-- Manual cleanup is available with `mise cache clear` if needed
+System, shared, and install-into destinations do not create a persistent HTTP
+extraction cache.
--- a/src/backend/http.rs
+++ b/src/backend/http.rs
@@ -22,7 +22,7 @@ use crate::toolset::ToolRequest;
 use crate::toolset::ToolVersion;
 use crate::toolset::ToolVersionOptions;
 use crate::ui::progress_report::SingleReport;
-use crate::{dirs, file, hash};
+use crate::{dirs, env, file, hash};
 use async_trait::async_trait;
 use eyre::Result;
 use serde::{Deserialize, Serialize};
@@ -71,6 +71,11 @@ struct CachePlan {
     key: String,
     file_info: FileInfo,
     strip_components: usize,
+}
+
+struct CacheTarget<'a> {
+    dir: &'a Path,
+    plan: &'a CachePlan,
 }
 
 impl FileInfo {
@@ -240,24 +245,40 @@ impl HttpBackend {
     // Cache path helpers
     // -------------------------------------------------------------------------
 
-    /// Get the http-tarballs directory in DATA (survives `mise cache clear`)
+    /// Get the shared extraction cache used by normal user installs.
     fn tarballs_dir() -> PathBuf {
         dirs::DATA.join(HTTP_TARBALLS_DIR)
     }
 
     /// Get the path to a specific cache entry
-    fn cache_path(&self, cache_key: &str) -> PathBuf {
-        Self::tarballs_dir().join(cache_key)
+    fn cache_path(cache_dir: &Path, cache_key: &str) -> PathBuf {
+        cache_dir.join(cache_key)
+    }
+
+    /// Remove an install entry without following an existing symlink. This is
+    /// needed when migrating a system/shared install created by an older mise
+    /// version from a cache symlink to a real directory.
+    fn remove_install_path(path: &Path) -> Result<()> {
+        if path
+            .symlink_metadata()
+            .is_ok_and(|metadata| metadata.file_type().is_symlink())
+        {
+            std::fs::remove_file(path)?;
+        } else {
+            file::remove_all(path)?;
+        }
+        Ok(())
     }
 
     /// Get the path to the metadata file for a cache entry
-    fn metadata_path(&self, cache_key: &str) -> PathBuf {
-        self.cache_path(cache_key).join(METADATA_FILE)
+    fn metadata_path(cache_dir: &Path, cache_key: &str) -> PathBuf {
+        Self::cache_path(cache_dir, cache_key).join(METADATA_FILE)
     }
 
     /// Check if a cache entry exists and is valid
-    fn is_cached(&self, cache_key: &str) -> bool {
-        self.cache_path(cache_key).exists() && self.metadata_path(cache_key).exists()
+    fn is_cached(cache_dir: &Path, cache_key: &str) -> bool {
+        Self::cache_path(cache_dir, cache_key).exists()
+            && Self::metadata_path(cache_dir, cache_key).exists()
     }
 
     // -------------------------------------------------------------------------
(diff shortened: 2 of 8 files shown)


e2e/backend/test_http_system_install_slow
```diff
@@ -0,0 +1,113 @@
...
+EOF
+
+mise install --shared "$SHARED_INSTALLS"
+
+SHARED_ARCHIVE="$SHARED_INSTALLS/http-shared-archive/1.0.0"
+assert "test -L '$SHARED_ARCHIVE' && echo symlink" symlink
+assert_directory_exists "$SHARED_INSTALLS/http-shared-archive/.http-tarballs"
+assert "test ! -e '$USER_INSTALLS/http-shared-archive/1.0.0' && test ! -L '$USER_INSTALLS/http-shared-archive/1.0.0' && echo missing" missing
+assert "realpath '$SHARED_ARCHIVE' | grep '^$SHARED_INSTALLS/http-shared-archive/.http-tarballs/'" "$(realpath "$SHARED_ARCHIVE")"
+assert_contains "mise x -- system-archive" system-archive-ok
```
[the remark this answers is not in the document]
Marukome0743: Thanks for catching this. The latest commit now invokes shared-archive and checks for shared-archive-ok, so the assertion verifies the shared installation rather than repeating the system-install check.

docs/dev-tools/backends/http.md
```diff
@@ -461,10 +461,17 @@ The HTTP backend implements an intelligent caching system to optimize disk usage
 
 ### Cache Location
 
-Downloaded and extracted files are cached in `$MISE_CACHE_DIR/http-tarballs/` instead of being stored separately for each tool installation. By default:
```
jdx, line 464: why are you changing this? this was intentionally using cache

Marukome0743: This updates stale documentation rather than changing the normal install behavior in this PR. The HTTP backend has used `$MISE_DATA_DIR/http-tarballs` since #7273 (`fix(http): move http-tarballs from cache to data directory`) so that `mise cache clear` does not break installed tools whose symlinks point at the extracted content. The previous docs still described the pre-#7273 `$MISE_CACHE_DIR` layout.

This PR keeps that existing normal-install behavior and only adds destination-local storage for system/shared installs so they do not depend on the installing user’s data directory.

jdx (commented):
Reviewed by reading the full diff, running the unit tests and the new e2e test, and building **both the base commit and this PR** to diff behavior empirically.

## Verdict

The core fix is real and the direction is right — I reproduced the bug on the base commit and confirmed the fix works. But there is **one new regression**, **one unclaimed leak**, **a design choice worth pushing back on**, and **a test that will not run in CI plus two assertions that assert nothing**.

Base-commit behavior I reproduced (the problem statement checks out):

- `mise install --shared <dir>` for an `http:` tool → `<dir>/http-tool/1.0.0` is an **empty dir**, the real install lands in `$MISE_DATA_DIR/installs/`. `--shared` silently ignored. Bug confirmed.
- `mise install-into http:… <dest>` → `<dest>` is an **empty dir**, the install lands in `$MISE_DATA_DIR/installs/http-iinto/1.0.0`. So this PR **also fixes a completely broken `install-into` for the http backend**. The PR body describes that as "preserve the exact destination" / "keeps its established cache", which undersells it — it's a fix, and it's the strongest justification for the `install_path_is_exact` flag.

## Blocking

### 1. Relative `--shared` path now produces a dangling symlink (new regression)

`tarballs_dir` (`src/backend/http.rs:248`) derives the cache dir from `tv.install_path`, which is never absolutized (`src/cli/install.rs:390` → `src/toolset/toolset_install.rs:559`). With `mise install --shared relshared`:

```
relshared/http-probe3/1.0.0 -> relshared/http-probe3/.http-tarballs/8e28031c…
```

The target is relative, so it resolves relative to the link's own directory → dangling. `mise install` still prints `✓ installed`, `mise ls` reports `1.0.0 (missing)`, and the tool is unusable. On the base commit this same input produced a *working* tool (in the wrong place), so it is strictly worse for that input.

Fix: absolutize in `install_opts()` / `install_single_tool`, or absolutize `cache_path` before `file::make_symlink`. Please add a relative-`--shared` case to the e2e.

### 2. The destination-local cache is never cleaned up by anything

After `mise uninstall http:probe2@1.0.0` on a `--system` install:

```
<sys>/installs/http-probe2/.http-tarballs/1e4fdbd0…   ← full extracted payload, still there
<sys>/installs/http-probe2/.mise.backend.toml
```

The version symlink goes, the payload stays, and the tool dir is never removed — the "remove parent installs dir if empty" logic in `src/backend/mod.rs:3093` only looks at the *primary* `installs_path`, and `file::dir_subdirs` counts `.http-tarballs` anyway. `mise cache clear` cannot reach it (this PR's own docs now say so), nothing prunes `http-tarballs` at all (`HTTP_TARBALLS_DIR` appears only in `http.rs`), and under `/usr/local` cleanup needs root. That is a new class of leak in a root-owned location. At minimum `uninstall_version_` should drop the destination cache entry, and the docs should say how to clean it manually.

## Design pushback

### 3. Why per-tool `<installs>/<tool>/.http-tarballs` instead of per-root `<installs>/.http-tarballs`?

The per-tool choice costs the cross-tool dedup that is the entire stated point of this cache — the docs edit concedes it ("system/shared installs reuse them within each tool") — and it puts a non-version entry inside a directory that several call sites enumerate as versions. It works today only because each of them happens to filter `.`-prefixed names (`install_state.rs:252`, `install_state.rs:362`, `backend/mod.rs:2439`, `runtime_symlinks.rs:229`) — and `backend/mod.rs:3095` does **not**.

`install_path.parent().parent()` is already used two lines away for `.mise-installs.toml` (`backend/mod.rs:2760`). Putting the cache there keeps dedup, keeps the versions dir clean, and gives one place to clean up.

Cleaner still for system/shared installs: extract directly into `<installs>/<tool>/<version>` as a real directory and skip the symlink and cache entirely. A handful of system installs do not need dedup, and it makes the install genuinely self-contained.

### 4. Two competing definitions of "not the primary destination" in the same flow

`http.rs:252` uses `!install_path.starts_with(&tv.ba().installs_path)`; `backend/mod.rs:2710` / `:2757` use `env::install_path_category(...) != Local` and `starts_with(*dirs::INSTALLS)`. They already disagree (`--shared /tmp/x` is non-primary to http, `Local` to `install_path_category`).

More importantly, `ba().installs_path` is not a constant: `cli/args/backend_arg.rs:103` rewrites it to a *shared* tool dir from install state, and several registry tools resolve to the http backend (`julia`, `grok`, `openshift-install`). If such a `BackendArg` ever reaches an install, `starts_with` returns true and the cache silently falls back to the user's data dir — i.e. this exact bug returns, quietly. I could not reach it from the CLI today (http requests carry a config/CLI-derived `ba`), so it is latent rather than broken, but using `dirs::INSTALLS` or a shared helper removes the fragility for free.

### 5. `install-into` now symlinks into the invoking user's data dir

`<dest> -> $MISE_DATA_DIR/http-tarballs/<key>` (verified; the e2e locks it in at line 128). For a command documented as "building a tool to a directory for use outside of mise", that is precisely the home-directory coupling this PR is fixing for `--system`. Copying instead of symlinking would be the consistent answer. Worth deciding now, since the new test enshrines the current behavior.

## Test issues

### 6. `_slow` means this never runs in CI

`e2e/run_all_tests:73` skips `*_slow` unless `TEST_ALL=1`. This test takes **4–7s** against a localhost server (the threshold is 20s, `e2e/run_test:160`). Rename to `test_http_system_install` so the regression is actually guarded on PRs.

### 7. Two assertions are vacuous

Lines 70 and 110: `assert_contains "mise ls http:system-archive" system` / `… shared`. `mise ls` emits no system/shared label. With a neutrally-named tool installed via `--system`:

```
http:neutral  1.0.0 (symlink)  ~/workdir/mise.toml  1.0.0     ← no "system" anywhere
```

They pass only because the tools are *named* `system-archive` / `shared-archive` (and the paths contain those words). Assert on `mise ls --json`'s `install_path`, which does carry the destination.

### 8. Missing coverage

- The `mise install --force` case (line 116) never checks that the pre-existing shared install survived. It does survive — `backend/mod.rs:2707` redirects — but that safeguard is exactly what this test should pin down.
- No relative-`--shared` case (finding 1).
- No `version = "latest"` + `--system` case, which is the trickiest touched branch (`create_version_alias_symlink` + `install_path_for`) and is unit-tested only.

### 9. Test style

`assert "realpath … | grep '^$PREFIX'" "$(realpath …)"` computes its own expected value from the thing under test; `assert_contains "realpath …" "$PREFIX"` says the same thing plainly. Also `$SYSTEM_INSTALLS` is interpolated unescaped into a grep regex.

## Code nits

- `http.rs:740`: the `debug_assert_eq!` can never fire — `content_path` comes from `install_path_for`, which for the non-exact branch *is* `tool_dir.join(content_version_name(cache_key))`. Delete it, and the `.expect("…must have a parent")` along with it, by computing `tool_dir` once.
- `http.rs:449`: `extract_to_cache` recomputes `Self::tarballs_dir(tv)` while its caller (`http.rs:1088`) computes and threads `cache_dir` into every sibling function. Pass it in — two sources of truth for the same path is how these drift.
- `tool_version.rs:49`: `pub(crate) install_path_is_exact` sits next to `pub install_path`; also its contract ("backends must not rewrite this") is honored by exactly one backend, which is worth saying in the doc comment.

## Verification run locally

- `cargo test --all-features --bin mise backend::http::` → 17/17 pass, including the 4 new tests.
- `TEST_ALL=1 ./e2e/run_test e2e/backend/test_http_system_install_slow` → passes, 7s.
- `cargo clippy --workspace --all-features --all-targets` → clean, no exclusions added.
- Doc changes check out against the code: content *is* extracted flat into the cache entry (no `extracted/` subdir), the cache *is* in `DATA` not `CACHE`, and nothing prunes it — so removing the stale `MISE_CACHE_DIR` / autopruner bullets is a genuine correction, not just a rewrite.

*AI-assisted — Tool: Claude Code; model: anthropic/claude-opus-5; version: 2.1.220.*

Marukome0743:
Thanks for the detailed review. I updated the implementation so explicit system, shared, and install-into destinations are extracted directly as real directories instead of symlinking to a destination-local cache. This fixes relative `--shared` paths, avoids persistent root-owned cache entries, and keeps these installs independent of the invoking user’s data directory, while normal user installs continue using the existing shared extraction cache.

I also separated exact install-into destinations from explicit system/shared destinations, renamed the E2E so it runs in the normal suite, replaced the vacuous assertions with `install_path` checks, and added coverage for relative shared paths, forced reinstalls, `latest` system installs, and exact install-into behavior. The HTTP cache documentation now reflects the existing `$MISE_DATA_DIR` behavior and the direct-extraction behavior for explicit destinations.

src/backend/http.rs
```diff
@@ -1045,35 +1130,41 @@ impl Backend for HttpBackend {
...
+                    },
+                    &file_path,
+                    &url,
+                    &opts,
+                    Some(ctx.pr.as_ref()),
+                )?
+            };
+            self.create_install_symlink(&tv, &cache_dir, &cache_plan.key, &extraction_type, &opts)?;
+        }
         self.create_version_alias_symlink(&tv, &cache_plan.key)?;
```
jdx, line 1168: **[P1] Remove the content directory behind the `latest` alias**

For explicit installs whose resolved version is literally `latest`, `extract_to_install_path` writes the real payload under the content-derived directory, and this call creates `latest` as a separate symlink. After toolset re-resolution, generic uninstall removes only `tv.install_path()`—the `latest` symlink—so the real payload remains, including under a root-owned system directory. If the artifact changes, forced reinstalls also leave each previous content-derived directory behind.

Before replacing or uninstalling this alias, resolve its target and remove it only after verifying that it is a real directory inside the same tool directory. Please add E2E coverage that changes the `latest` artifact, force-reinstalls, and then uninstalls, asserting that neither old nor current content directories survive.

jdx (changes requested):
Requesting changes for one remaining lifecycle bug in explicit `latest` installs. The earlier destination, cache, path-remapping, and test-coverage findings have otherwise been addressed.

The new direct-extraction design needs to remove the content-derived payload when replacing or uninstalling its `latest` alias; otherwise system/shared updates accumulate stale payloads and uninstall leaves the current payload behind.

jdx:
on second thought, I don't know if it's worth fixing that

PR #11857, opened 2026-08-11
fix(install): run install hooks from project root

jdx:
## Summary

- run inline `preinstall` and `postinstall` hooks from the owning project root
- preserve invocation-directory access through `MISE_ORIGINAL_CWD`
- apply the behavior to the Windows `cmd.exe` execution path
- document the working-directory contract and add nested-directory regression coverage

## Root cause

Inline install hooks spawned their shell without setting a working directory, so they inherited the directory where `mise install` was invoked. Task-backed hooks already used `mise --cd <project-root> run`, which made equivalent hook forms behave differently.

## Impact

Relative paths in inline install hooks now resolve consistently from the project root. Hooks that need the invocation directory can continue to use `MISE_ORIGINAL_CWD`. Directory-change hooks retain their existing working-directory behavior.

## Validation

- `mise run test:e2e e2e/config/test_hooks`
- `mise run lint-fix`

Closes discussion #11854.

*AI-assisted — Tool: Codex; model: unavailable/unavailable; version: unavailable.*

---

> [!NOTE]
> **Low Risk**
> Narrow behavior fix for install hooks only; hooks that depended on invocation cwd must use `MISE_ORIGINAL_CWD`, which is documented.
> 
> **Overview**
> **Inline `preinstall` and `postinstall` hooks now run with the project root as their working directory**, including when `mise install` is started from a subdirectory. The invocation directory is still exposed via `MISE_ORIGINAL_CWD`.
> 
> In `hooks.rs`, install hooks set the child process cwd to the config `root` on both the normal duct path and the Windows `cmd.exe` verbatim path (replacing the previously commented `.dir(root)`). **Enter/leave/cd hooks are unchanged.**
> 
> Docs describe this contract; e2e coverage asserts `pwd` in install hooks matches the project root when installing from `foo/`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6bc0f8072cdce24f083b360c86d8b4fd42cdfff6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Improvements**
  * Preinstall and postinstall hooks now run from the project root, even when installation starts in a subdirectory.
  * The original invocation directory remains available through `MISE_ORIGINAL_CWD`.
* **Documentation**
  * Added documentation covering the updated hook behavior.
* **Tests**
  * Added coverage confirming hook execution from the project root.

PR #11865, opened 2026-08-11
test(task): isolate cargo writes in cache session e2e

jdx:
## Summary
- keep the cache-session fixture’s Cargo home and target directory inside its permitted `target/` sandbox path
- prevent fresh CI containers from failing when Cargo creates its global cache lock files

## Context
The E2E failure seen on #11857 was unrelated to the install-hook change. `cargo build` creates `.global-cache`, `.package-cache`, and `.package-cache-mutate` in `CARGO_HOME` even for the dependency-free fixture. On fresh CI containers, those files do not exist yet and the task’s `deny_write = true` sandbox rejects their creation outside `allow_write = ["target"]`.

## Testing
- `mise run lint-fix`
- `e2e/run_test tasks/test_task_action_cache_session`
- repeated the focused test with a fresh external `CARGO_HOME`; the fixture now succeeds because it scopes Cargo writes to `target/`

---

> [!NOTE]
> **Low Risk**
> E2E fixture-only change; no production task-cache or sandbox behavior is modified.
> 
> **Overview**
> Fixes the **cache-session** E2E `compile` task so `cargo build` no longer writes outside the sandbox.
> 
> The task uses `deny_write = true` with `allow_write = ["target"]`, but plain `cargo build` still creates lock/cache files under the default **`CARGO_HOME`**, which fails on fresh CI. The run script now sets **`CARGO_HOME=target/cargo-home`** and **`CARGO_TARGET_DIR=target`** so all Cargo writes stay under the permitted `target/` path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3fc941c18bd9e39b728424f6a26e70a764a9abf5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compile task caching by storing Cargo build data within the project’s target directory.
  * Preserved existing build verification behavior.

### DIFF
--- a/e2e/tasks/test_task_action_cache_session
+++ b/e2e/tasks/test_task_action_cache_session
@@ -33,7 +33,9 @@ rust_cache = true
 deny_write = true
 allow_write = ["target"]
 run = '''
-cargo build
+# Cargo creates lock files in CARGO_HOME even for a dependency-free crate. Keep
+# those writes inside the path permitted by this task's deny_write sandbox.
+CARGO_HOME=target/cargo-home CARGO_TARGET_DIR=target cargo build
 test -f target/debug/libcache_e2e.rlib
 '''
 EOF


JamBalaya56562:
The e2e run on this PR's head (`3fc941c1`) still fails the fixture it targets, on both retry attempts:

```
##[error][mise run compile 2>&1] command failed with status 1
##[error]tasks/test_task_action_cache_session
```

https://github.com/jdx/mise/actions/runs/31453472371/job/93663096515

The reason is that `cargo build` never starts. Landlock is unavailable on these runners, so `apply_landlock` bails inside the `pre_exec` hook, before `execve`:

```
mise sandbox: failed to apply landlock restrictions: RestrictionStatus { ruleset: NotEnforced,
no_new_privs: true, landlock: NotEnabled, log_same_exec: true, log_new_exec: false,
log_subdomains: true, all_threads: false }
```

https://github.com/jdx/mise/actions/runs/31449920970/job/93652822330 — an e2e job carrying an extra diagnostic print I had on #11853. `landlock: NotEnabled` is the runner's own report. `restrict_self()` comes back `NotEnforced`, the check at the end of `apply_landlock` bails, and the child dies before exec. Nothing writes anywhere, so where Cargo would have put its lock files does not come into it.

The failure mode described here is real, just conditional on Landlock working. `e2e/run_test:43` resolves `CARGO_HOME` from the outer `$HOME` before `HOME` is swapped for the isolated one, so the default lands outside `/tmp`, and the `deny_write`-only branch of `apply_landlock` grants only `/` (read), `/tmp`, `/dev` and `allow_write`. Where Landlock is enabled, scoping Cargo under `target/` does fix it. Where it is `NotEnabled`, there is nothing to fix at that layer.

#11861 takes the other route — drop `deny_write`/`allow_write` from the fixture. Its e2e is green on the same runners (`PASS: tasks/test_task_action_cache_session`, https://github.com/jdx/mise/actions/runs/31457042598/job/93673507583), and every check on the PR passes.

The coverage that gives up is smaller than it looks. The isolated dir lives under `/tmp` — `/tmp/test_task_action_cache_session.*` in the logs above — and the `deny_write` branch grants `/tmp` full access, so the fixture's own writes (`target/`, config, cache) were never constrained. The only write the sandbox actually denied was Cargo's `CARGO_HOME`, which is not what a task-action-caching test asserts.

Either way is fine by me. The two changes are not exclusive: yours makes the fixture pass wherever Landlock works, and if you would rather keep the sandbox here I can close #11861 — the runners would then need Landlock for this to go green.

On why this sat unnoticed: the e2e workflow runs two tranches in a single `command:` and keeps only the second exit code, so failures in tranches 0-3 are discarded. I have three runs where this fixture failed and the job still reported success. #11853 fixes the aggregation and is what surfaced it.

PR #11877, opened 2026-08-11
fix(task): mask archive modes in remote cache nodes

stevenpollack:
## Summary

Remote task-cache archive nodes now contain portable permission modes. This prevents mise-cache from rejecting a valid publication after the blob upload completes.

## User-visible failure

The remote cache server returns HTTP 422 `invalid directory entry` after mise successfully uploads the blobs, so the task result cannot be published.

## Root cause

Tar archive headers include Unix file-type bits in `mode`, such as `0o100664`. Those bits were being sent in the mise-cache protocol node mode, which expects permission bits while representing executable semantics separately.

## Fix

Mask archive modes with `& 0o7777` when constructing protocol nodes. This is limited to the protocol-node conversion and retains permission and executable semantics without leaking Unix file-type bits.

## Regression test

`archive_is_split_into_directory_and_file_cas_objects` now supplies directory and file headers with Unix file-type bits and asserts that emitted nodes contain `0o755`. Before the fix, the directory mode was emitted as `16877` (`0o40755`) instead of `493` (`0o755`).

## Interoperability qualification

Against pinned mise-cache commit `a684221baf43516732294b5723539d1593690b0d`, seven blob uploads and action-result publication succeeded. A subsequent client restored the outputs without executing the task. Missing and incorrect tokens returned 401, and teardown plus token scanning passed.

*AI-assisted — Tool: OpenCode; model: openai/gpt-5.6-luna; version: unavailable.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Normalized permissions for archived directories, files, and symbolic links.
  * Archives created with full file-type mode values now receive the expected permission settings.


stevenpollack:
The failed hosted checks appear unrelated to this change. The local protocol regression test, formatting, Clippy with `-D warnings`, and lint pass. The failed jobs hit GitHub API `403 rate limit exceeded`: [Linux E2E](https://github.com/jdx/mise/actions/runs/31508971907/job/93839513027) and [Windows E2E](https://github.com/jdx/mise/actions/runs/31508971907/job/93841589612); [test-ci](https://github.com/jdx/mise/actions/runs/31508971907/job/93848098138) failed because those jobs failed. The failed-job rerun was unavailable to this account because GitHub requires repository admin rights. *AI-assisted — Tool: OpenCode; model: openai/gpt-5.6-luna; version: unavailable.*

PR #11938, opened 2026-08-12
fix(rust): adjust rustup provider selection

AlternateRT:
System package manager installations of rustup also use `RUSTUP_HOME` and `CARGO_HOME` if they have been set. The current implementation brought by #11840 to address #8006 did not account for this, so it would still initialise another instance of rustup rather than reuse the existing installation.

Applying the [model suggested](https://github.com/jdx/mise/pull/11938#issuecomment-5280514983) by @jdx, the source of the `RUSTUP_`/`CARGO_` homes should no longer affect the provider selection of rustup. The provider should be selected as such:

1. If an usable provider recorded for an existing installation exists, that should be used.
2. If the above does not exist, a complete usable proxy set in the selected `CARGO_HOME/bin` should be used.
3. If the above does not exist either, then a complete usable proxy set from the original `PATH` should be used.
4. If none of the above exist, then download/run `rustup-init`.

## New Features
- Rust setup can reuse compatible `rustup`, `cargo`, and `rustc` installations from rustup.rs or package managers when properly configured.
- Previously configured Rust tool providers are prioritized when available.

## Bug Fixes
- Improved Rust runtime selection, fallback behavior, and validation for configured, managed, and external installations.

## Documentation
- Clarified Rust installation reuse, Rust/Cargo home-directory settings, configuration precedence, and management of toolchains, components, and targets.

### DIFF
--- a/docs/lang/rust.md
+++ b/docs/lang/rust.md
@@ -1,12 +1,14 @@
 # Rust
 
-Rust/cargo can be installed which uses rustup under the hood. mise will install rustup if it is not already
-installed and install the requested toolchain, components, and targets. By default, mise respects the `RUSTUP_HOME` and `CARGO_HOME` environment
-variables for the home directories and falls back to their standard location (`~/.rustup` and `~/.cargo`) if they are
-not set. You can change this by setting the `MISE_RUSTUP_HOME` and `MISE_CARGO_HOME` environment variables if you'd like
-to isolate mise's rustup/cargo from your other rustup/cargo installations.
+`mise` can be used to manage Rust/Cargo. Under the hood, it uses `rustup` to install the requested toolchains,
+components, and targets. `mise` will install `rustup` if it is not on your system, reuse an existing installation from
+[rustup.rs](https://rustup.rs), or reuse a system package manager installation (as provided by package managers such as Homebrew,
+APT, and pacman) if the original `PATH` contains a directory with the `rustup`, `cargo`, and `rustc` proxies. By
+default, `mise` respects the `RUSTUP_HOME` and `CARGO_HOME` environment variables for the home directories, and falls back
+to their standard location (`~/.rustup` and `~/.cargo`) if they are not set. You can change this by setting the
+`MISE_RUSTUP_HOME` and `MISE_CARGO_HOME` environment variables if you'd like to isolate `mise`'s Rust/Cargo from your other Rust/Cargo installations.
 
-These variables can also be set in mise configuration. They are applied to Rust operations in the same mise invocation:
+These variables can also be set in `mise` configuration. They are applied to Rust operations in the same `mise` invocation:
 
 ```toml
 [env]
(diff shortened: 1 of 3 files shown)


jdx:
I do not think the source of the home value should affect provider selection. Once mise has resolved the effective `RUSTUP_HOME` and `CARGO_HOME`, the same values should behave identically whether they came from the parent environment, `[env]`, `MISE_*`, or Rust settings. The source should determine precedence only.

`RUSTUP_HOME` and `CARGO_HOME` select Rustup state locations; they do not select the Rustup executable. A functional package-manager Rustup can operate on the resolved homes, including fresh custom homes. `MISE_*` can still isolate the state without requiring mise to download a separate Rustup executable.

I would model this as two independent steps:

1. Resolve the homes using the precedence established in #11798.
2. Select a provider: prefer the usable provider recorded for an existing installation, then a complete usable proxy set in the selected `CARGO_HOME/bin`, then a complete usable proxy set from the original `PATH`, and only download/run `rustup-init` when none exists.

Every provider should receive the same resolved `RUSTUP_HOME` and `CARGO_HOME`. Changing the homes should make mise verify or install the requested toolchain in the new Rustup state, but should not by itself force a new Rustup executable.

As written, rules 1/2 also conflict with rule 5: the first rules say to initialize regardless of `PATH`, while rule 5 says never to initialize when functional external proxies exist. I would remove the `explicit` home-source distinction and document one behavior for all sources.

The external-provider test should store its fake toolchain state under `RUSTUP_HOME`; currently it stores state under the provider prefix, so it cannot catch home/provider coupling bugs.

I would also retain the conceptual distinction between resolving the desired runtime and resolving the recorded runtime, even if the two functions happen to be identical today. Provider path plus `RUSTUP_HOME` and `CARGO_HOME` collectively identify the installation lifecycle, especially after `PATH` or home changes.

jdx:
The underlying bug is valid, but I am closing this because the current implementation has accumulated too many architectural issues to merge safely.

The provider model should be simpler: resolve `RUSTUP_HOME` and `CARGO_HOME` once, then select a provider independently of where those values originated. The source should affect precedence only.

The main issues with this implementation are:

- `rustup`, `cargo`, and `rustc` are synchronously probed from `list_bin_paths` and `which`, putting multiple subprocess launches on shim dispatch, activation, `mise env`, `mise exec`, and other hot paths. These probes can also block indefinitely.
- The desired-runtime and recorded-runtime resolvers are currently identical. Normal execution should trust the provider recorded by the install symlink; provider validation and fallback belong on install/reconciliation paths.
- The separate ambient `MISE_CARGO_HOME` and `MISE_RUSTUP_HOME` inputs appear redundant because those variables are already aliases for the Rust settings.
- Inherited values are read from the current process environment rather than the reconstructed pristine environment, which can blur values inherited from the user with values injected by a prior mise activation.
- The external-provider test still ties fake toolchain state to the provider prefix rather than `RUSTUP_HOME`, so it does not verify the key requirement that one provider executable can operate on newly resolved homes.
- The test also still expects explicit homes to replace the recorded provider, which is the opposite of the intended behavior.
- The documentation does not clearly describe the complete home-resolution and provider-selection sequence.

A replacement should:

1. Resolve homes from configured `[env]` values, the resolved Rust settings, pristine inherited direct variables, and defaults.
2. Use the install symlink as the recorded provider on normal execution, with no subprocess probes.
3. Validate the recorded provider, resolved `CARGO_HOME/bin`, and original `PATH` only during install or reconciliation, using bounded asynchronous probes.
4. Pass the same resolved homes to whichever provider is selected.
5. Update the external-provider e2e fixture so all state is stored under `RUSTUP_HOME`.

The provider-order concept is sound, but I think replacing this implementation will be safer and simpler than continuing to patch it.

PR #11968, opened 2026-08-13
fix(upgrade): allow typing to filter tools in interactive upgrade picker

jakedgy:
## Summary
- `mise upgrade`'s interactive multiselect (`get_interactive_tool_set` in `src/cli/upgrade.rs`) built its `demand::MultiSelect` with `.filterable(true)` but not `.filtering(true)`.
- Without `.filtering(true)`, the widget starts in browse mode: `j`/`k`/`h`/`l` are bound to movement and any other typed character is swallowed until `/` is pressed to enter filter mode. So typing e.g. `kube` to narrow the list does nothing visible — no filter box appears, the cursor doesn't move, and the letters vanish.
- The other three interactive pickers in the codebase (`mise use`, `mise search -i`, the task list picker) all already set `.filtering(true)` and work correctly — `mise upgrade` was the only outlier. Verified there are no other `Select`/`MultiSelect` construction sites in the repo, and `mise install` has no interactive picker at all.

## Fix
One-line addition of `.filtering(true)` to match the other three call sites.

## Test plan
- [x] Built a standalone repro against the exact `demand` v2.0.5 builder call used here and drove it over a real pty with `expect`, confirming: before the fix, typing `kube` did nothing; after adding `.filtering(true)`, it correctly narrows/highlights matches (`kubectl`, `kube-linter`, `kubent`).
- [ ] Manual check: `mise upgrade -i` (or `cargo run -- upgrade -i`) with several outdated tools, type to filter.

*AI-assisted — Tool: Claude Code; model: claude-sonnet-5; version: unavailable.*

## Summary by CodeRabbit

- **New Features**
  - Added filtering support to the interactive upgrade selection menu, making it easier to find and choose upgrades.

### DIFF
--- a/src/cli/upgrade.rs
+++ b/src/cli/upgrade.rs
@@ -639,6 +639,7 @@ impl Upgrade {
         let theme = crate::ui::theme::get_theme();
         let mut ms = demand::MultiSelect::new("mise upgrade")
             .description("Select tools to upgrade")
+            .filtering(true)
             .filterable(true)
             .theme(&theme);
         for out in outdated {


jakedgy:
Closing — on further review this doesn't look like a bug. \`.filterable(true)\` without \`.filtering(true)\` has been the design since the very first \`--interactive\` commit for this multiselect, unchanged since. Adding \`.filtering(true)\` starts the picker already in filter mode, which in \`demand\`'s \`MultiSelect\` also silently drops the \`a\` (toggle all) and \`x\` (alt toggle) shortcuts from both the key handler and the help footer — a real regression for what's probably the most common interaction here (select-all-to-upgrade). The original behavior is also discoverable as-is: the footer already shows \`/ filter\`.

*AI-assisted — Tool: Claude Code; model: claude-sonnet-5; version: unavailable.*

PR #11993, opened 2026-08-14
fix(test): add missing content field to the link_req test helper

stevenpollack:
`FileRequest` gained `content: Option<String>` in d96c3a6db (#11983), but the `link_req` helper in `src/system/files.rs`'s own `mod tests` was not updated, so every test target fails to compile on current `main`:

```
error[E0063]: missing field `content` in initializer of `files::FileRequest`
    --> src/system/files.rs:2110:9
```

Verified on `08cb51e70`, both directions:

| | `cargo check --all-features --all-targets` |
| -- | -- |
| `main` as-is | `error[E0063]` — exit 101 |
| with this one line | clean — exit 0 |

## Why it was not caught

`cargo build` is unaffected, so nothing user-facing broke. `test.yml` runs on `pull_request` into `main`, and on pushes to tags and the `mise`/`release` branches — but not on pushes to `main` itself, so no test target was compiled after the merge. The runs recorded on `main` at `08cb51e70` are `docs`, `release-plz` and `perf`.

The effect is on open pull requests rather than on `main`: `cargo test --all-features`, `cargo clippy --all-features --all-targets -- -D warnings` and `cargo check ... --all-targets` all run in `test.yml`, and all three compile test targets, so any PR opened against `main` currently goes red for a reason unrelated to its own diff.

`None` matches the helper's intent — it builds link and copy requests, which take their content from `source` rather than from an inline literal.

## Summary by CodeRabbit

* **Tests**
  * Updated test setup to correctly handle file requests without content.

### DIFF
--- a/src/system/files.rs
+++ b/src/system/files.rs
@@ -2111,6 +2111,7 @@ mod tests {
             target_raw: target.to_string_lossy().to_string(),
             target: target.to_path_buf(),
             source: source.to_path_buf(),
+            content: None,
             mode,
             exclude: vec![],
             base: source.parent().expect("source parent").to_path_buf(),


stevenpollack:
Closing as a duplicate of #11989, which was opened first and makes the same change (with a clearer comment). I should have checked open PRs before filing — apologies for the noise.

Leaving the verification here in case it is useful for prioritising #11989, since it blocks CI on every open PR rather than affecting `main` itself:

| | `cargo check --all-features --all-targets` on `08cb51e70` |
| -- | -- |
| as-is | `error[E0063]: missing field `content`` — exit 101 |
| with the fix | clean — exit 0 |

`test.yml` runs on `pull_request` into `main` and invokes `cargo test --all-features`, `cargo clippy --all-features --all-targets -- -D warnings` and `cargo check ... --all-targets`, all of which compile test targets. It does not run on pushes to `main`, which is why the break is invisible there — only `docs`, `release-plz` and `perf` ran at that commit.

PR #11997, opened 2026-08-14
feat(task): add task.cache.audit_report for the full audit report

stevenpollack:
> Discussion: https://github.com/jdx/mise/discussions/11996

Adds `task.cache.audit_report`, an optional path that receives the complete cache-audit
report. Default behavior is unchanged.

## Problem

`cache.audit = true` reports filesystem accesses a task performs that aren't covered by its
`sources`/`outputs`. The reporter caps console output at 20 paths per task and then prints
only a count of the remainder:

```
mise WARN  task build cache audit detected undeclared read: data/file1.txt
... 19 more ...
mise WARN  task build cache audit omitted 80 additional paths
```

There is no way to see the omitted paths — no flag, no setting, no file. That makes the
audit unusable for the tasks that need it most. A Jest run over `node_modules` reports on
the order of 40,000 undeclared paths; enumerating them 20 at a time would take ~2,000 runs
with a declare-and-rerun loop between each.

What the cap drops is not a random sample. `undeclared` is a
`BTreeSet<(AccessKind, PathBuf)>` and `AccessKind::Read` orders before `Write`, so every
undeclared read sorts ahead of every undeclared write and `take(MAX_REPORTED_PATHS)` yields
reads first. A task with 20 or more undeclared reads shows **no undeclared writes at all**.
On a 100-file repro the console prints 20 lines, none of them writes, and reports `omitted 81
additional paths` — the one undeclared write is inside that remainder. Writes are the more
consequential half of a cache contract: an undeclared write makes a cached result wrong for
the next consumer, whereas an undeclared read mostly makes the key incomplete.

This PR mitigates that rather than fixing it. The report file contains every entry, so the
writes are recoverable — but the console is untouched and still shows zero writes by default
for such a task. Reserving part of the console budget for writes, or reporting the two kinds
separately, is a separate change and deliberately not made here.

## Change

Set `task.cache.audit_report` (env `MISE_TASK_CACHE_AUDIT_REPORT`) to a file path and mise
writes every undeclared path there as JSON Lines:

```
{"task":"build","kind":"read","path":"data/file1.txt"}
{"task":"build","kind":"write","path":"dist/out.txt"}
```

`path` is the same rendered string the console warning uses, so the file needs no separate
path convention and stays in step with however the audit chooses to render paths. A
companion change adjusts that rendering for accesses above the task directory; the two
compose with no changes here.

```shell
MISE_TASK_CACHE_AUDIT_REPORT=audit.jsonl mise run --force build
```

Console output stays capped at 20. The only console change is that the omitted-count
warning names the report file:

```
mise WARN  task build cache audit omitted 80 additional paths; full report written to audit.jsonl
```

## Default behavior is unchanged

Not asserted — measured. A binary was built from `main` and confirmed to predate the change
by the fact that it rejects the new key outright:

```
main:        $ mise settings get task.cache.audit_report
             Error: Unknown setting: task.cache.audit_report
this branch: $ mise settings get task.cache.audit_report
             Error: Setting [task.cache.audit_report] is not set
```

Both binaries were then run against the same 100-file repro with the setting unset, and
their audit output diffed:

```
$ diff base.log new-default.log
$                       # empty — byte-identical, 20 warnings then "omitted 80 additional paths"
```

## Semantics

**One file per invocation.** The first audited task to execute truncates the file; later
tasks append. A run therefore replaces any earlier run's report rather than growing one
without bound. Cached tasks do not execute and write nothing, so a cache-hit run leaves the
previous report in place — consistent with the existing rule that cached tasks produce no
audit report at all.

**Every entry is tagged by task**, so one file covers all audited tasks in a run and stays
attributable. Concurrency is safe: each task serializes its whole block through one lock, so
blocks from tasks running in parallel cannot interleave.

**Writing is advisory.** A write failure warns and continues, and clears the report path so
the omitted-count warning cannot name a file that was not written. The audit must never fail
the task.

## Design notes

**A report file rather than a higher cap.** The obvious alternative is a
`task.cache.audit_max_paths` knob. It answers the title but not the problem: 40k lines
through `warn!` interleave with the task's own output and force the user to strip the
`mise WARN ` prefix back off before the list is usable. A file supersedes it — anyone who
wants more than 20 lines can read the file — so shipping both would be two knobs for one job.

**JSON Lines rather than a delimited format.** Paths can contain tabs, newlines and quotes;
this module has a test (`parses_escaped_quoted_paths`) for exactly the `a"b` case. Any
delimiter is lossy on the inputs the parser is built to handle. JSONL is the smallest
lossless option, and the file exists to be machine-consumed — the point is turning 40k
entries into `sources` declarations.

**A setting rather than a `mise run` flag.** A setting gets the env var for free, which is
what CI wants, and it still works when `mise run` is invoked indirectly.

**Non-UTF-8 paths are rendered lossily**, with invalid sequences replaced by U+FFFD, so such
a path will not round-trip back to a real file. This matches the existing console output,
which formats the same path with `Path::display()` and is lossy in exactly the same way — the
file is no less faithful than the warnings it replaces, and diverging here would mean two
renderings of the same path in the same run.

**Dotted `[task.cache.audit_report]`, with no flat `task.cache_audit_report` alias.** The
flat `task.cache_*` family is not a parallel form new settings are expected to appear in.
All six `task.cache_remote_*` entries are back-compat shims for settings that used to be
flat — each carries `deprecated = "Use task.cache.<x> instead."`, `hide = true`,
`deprecated_warn_at = "2027.2.0"` and `deprecated_remove_at = "2027.8.0"`, with migration
glue in `src/config/settings.rs` and a test named
`test_task_cache_hidden_aliases_map_to_nested_settings`. The flat entries that are canonical
(`task.cache_dir`, `task.cache_max_age`, `task.cache_max_size`) simply have no dotted twin
and were never migrated. Adding a flat alias for a brand-new setting would mean shipping a
pre-deprecated alias for a name that never existed.

## Testing

New e2e test `e2e/tasks/test_task_artifact_cache_audit_report`, following the existing
`test_task_artifact_cache_audit` pattern (fake `strace` on `PATH` emitting a synthetic
trace, so it needs no real strace). It emits 25 undeclared reads and asserts that the
default run prints exactly 20 lines plus `omitted 5 additional paths` and creates no file,
and that with the setting the console is still exactly 20 lines while the file holds all 25
entries.

Also verified by hand against real `strace` on Linux: 100 undeclared reads produce 20
console lines and a 100-entry report; two audited tasks in one invocation land in one file
correctly tagged; a re-run truncates rather than appends; four tasks at `--jobs 4` produce
800 entries with no interleaving; an unwritable report path warns and the task still exits
0; and relative paths resolve against mise's working directory, including under `--cd`.

`cargo clippy --workspace --all-features --all-targets -- -D warnings` passes with no
`#[allow]`/`#[expect]`/`-A` exclusions. `schema/mise.json` regenerated with
`mise run render:schema`.

The single undeclared write in that 100-file run turned out to be a genuine false positive in
the audit — a task's own declared output *directory*, reported on the run that creates it,
because the reporter skips directory reads but has no equivalent skip for directory writes.
It is reported separately and not addressed here; it is noted only because it had been sitting
inside the truncated remainder, invisible on the console, since the audit was introduced.

---

## Summary by CodeRabbit

* **New Features**
  * Added an optional `task.cache.audit_report` setting to write complete cache-audit results to a JSON Lines file.
  * Reports include the task, access type, and path for all undeclared accesses.
  * Supports configuration files or the `MISE_TASK_CACHE_AUDIT_REPORT` environment variable.
  * Reports accumulate results across audited tasks, while cached tasks leave the file unchanged.
  * Console warnings continue to be truncated when a full report is enabled.

* **Documentation**
  * Added configuration guidance and usage examples for audit reports.

### DIFF
--- a/docs/tasks/task-configuration.md
+++ b/docs/tasks/task-configuration.md
@@ -613,6 +613,17 @@ Audit mode requires `strace` on `PATH`. Mise warns and runs the task normally wh
 available; other platforms are not currently supported. Cached tasks are not executed and therefore
 produce no audit report, so use `mise run --force <task>` when checking an existing cache entry.
 
+Console warnings are limited to the first 20 paths per task, which is not enough to classify a task
+that reads thousands of undeclared files. Set
+[`task.cache.audit_report`](/configuration/settings.html#task-cache-audit-report) to also write every
+undeclared path as JSON Lines, one `{"task", "kind", "path"}` object per entry. The first audited
+task to execute truncates the file and later tasks append to it, so it covers every audited task
+that ran.
+
+```shell
+MISE_TASK_CACHE_AUDIT_REPORT=audit.jsonl mise run --force build
+```
+
 ```mise-toml
 [tasks.build]
 run = "npm run build"
--- a/docs/tasks/task-configuration.md
+++ b/docs/tasks/task-configuration.md
@@ -616,9 +616,10 @@ produce no audit report, so use `mise run --force <task>` when checking an exist
 Console warnings are limited to the first 20 paths per task, which is not enough to classify a task
 that reads thousands of undeclared files. Set
 [`task.cache.audit_report`](/configuration/settings.html#task-cache-audit-report) to also write every
-undeclared path as JSON Lines, one `{"task", "kind", "path"}` object per entry. The first audited
-task to execute truncates the file and later tasks append to it, so it covers every audited task
-that ran.
+undeclared path as JSON Lines, one `{"task", "kind", "path"}` object per entry. Truncation happens
+once per `mise` invocation: the first audited task in each invocation truncates the file and later
+audited tasks in that invocation append to it, so one file holds that run's report for every audited
+task and a later run replaces it rather than adding to it.
 
 ```shell
 MISE_TASK_CACHE_AUDIT_REPORT=audit.jsonl mise run --force build
--- a/e2e/tasks/test_task_artifact_cache_audit_report
+++ b/e2e/tasks/test_task_artifact_cache_audit_report
@@ -54,11 +54,11 @@ cache = { enabled = true, audit = true }
 EOF
 
 for task in alpha beta; do
-	mkdir -p "$task"
-	printf 'declared\n' >"$task/declared.txt"
-	for i in $(seq 1 25); do
-		printf 'undeclared\n' >"$task/undeclared$i.txt"
-	done
+  mkdir -p "$task"
+  printf 'declared\n' >"$task/declared.txt"
+  for i in $(seq 1 25); do
+    printf 'undeclared\n' >"$task/undeclared$i.txt"
+  done
 done
 
 export PATH="$PWD/bin:$PATH"
--- a/e2e/tasks/test_task_artifact_cache_audit_report
+++ b/e2e/tasks/test_task_artifact_cache_audit_report
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
 
 mkdir -p bin
+# One fake strace serves both tasks; it is spawned in each task's directory, so
+# every task gets its own trace. The write it emits is a plain file so these
+# assertions do not depend on how directory writes are reported.
 cat <<'EOF' >bin/strace
 #!/usr/bin/env bash
 set -euo pipefail
@@ -26,6 +29,7 @@ fi
 for i in $(seq 1 25); do
   echo "123 openat(AT_FDCWD<$PWD>, \"undeclared$i.txt\", O_RDONLY) = 3" >>"$trace"
 done
+echo "123 openat(AT_FDCWD<$PWD>, \"generated.txt\", O_WRONLY|O_CREAT|O_TRUNC, 0666) = 3" >>"$trace"
 exec "$@"
 EOF
 chmod +x bin/strace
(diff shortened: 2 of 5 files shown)


stevenpollack:
Both review findings addressed in `06843e884`.

## e2e coverage — agreed, extended

The test now runs two audited tasks concurrently against one report file, each with its own `dir` and its own trace, and asserts:

- 52 records total, 26 per task
- both tasks and both access kinds present
- write records for both tasks
- 52 again on a second run rather than 104, which pins per-invocation truncation

Parsing every line with `jq -s` doubles as the interleaving check, since a torn concurrent write fails to parse.

One deliberate choice: the write is a plain file rather than a directory. A directory write is currently reported as undeclared even when `outputs` covers it (reported separately in https://github.com/jdx/mise/discussions/11990), so asserting on a directory would couple this test to a defect not addressed here — and would start failing when that one is fixed.

A property fell out that had no test anywhere. Because reads sort ahead of writes in the `BTreeSet`, each task’s twenty console lines are all reads, so the same test now pins the behaviour this feature exists to work around: the console shows no write records while the report file contains them.

## Truncation scope in docs — agreed, fixed

Now stated as: the first audited task in each invocation truncates the file and later audited tasks in that invocation append, so one file holds that run’s report and a later run replaces it.

Applied the same correction to the `docs` block in `settings.toml`, which was not flagged but carries the identical wording into generated output. `mise run render:schema` produces no diff — the schema carries a setting’s `description`, not its `docs`.

Verified locally: both e2e tests pass individually via `./e2e/run_test`, `cargo clippy --workspace --all-features --all-targets -- -D warnings` clean with no `allow`/`expect`/`-A`, and `cargo test --all-features --bin mise settings` 77 passed.

CI will be red until #11989 merges — `main` does not currently compile its own test targets, unrelated to this change; the `--all-targets` gates above were run with that one-line fix applied locally.

stevenpollack:
Good catch — addressed in `dbf598690`.

You are right that `jq -s` does not cover this. It fails on a torn record, but alternating whole records still parse, so the assertion said nothing about the property the per-task lock exists to provide. The comment above it claimed otherwise and has been corrected too, since that was the thing most likely to mislead the next reader into thinking the case was covered.

Checked that contiguity is actually guaranteed before asserting it: `write_report` builds the whole block into a `String`, *then* takes the lock, then opens and writes, so the lock spans the entire `write_all` rather than a single syscall. No other task in the process can write between the first and last byte of a block. (It would not hold across two concurrent mise processes sharing one report path — the mutex is per-process, and the feature does not claim otherwise.)

The assertion:

```bash
assert "jq -r .task audit.jsonl | uniq | wc -l | tr -d ' '" "2"
```

`uniq` collapses consecutive runs, so a contiguous file yields one line per block — two blocks for two tasks, in whichever order they finished.

Rather than verify it discriminates once locally and ship an assertion nobody can see fail, the negative case is in the test:

```bash
paste -d '\n' <(jq -c 'select(.task == "alpha")' audit.jsonl) <(jq -c 'select(.task == "beta")' audit.jsonl) >interleaved.jsonl
assert "jq -s length interleaved.jsonl" "52"
assert "jq -r .task interleaved.jsonl | uniq | wc -l | tr -d ' '" "52"
```

The first of those demonstrates your point executably — `jq -s` passes on a deliberately interleaved file — and the second shows the contiguity check reporting 52 blocks instead of 2. So removing the lock fails the real assertion, and weakening the contiguity check fails the guard.

Separately verified under genuine concurrency, since the two tasks in the fixture are short enough that they might serialize and make contiguity trivially true. Four tasks of 300 reads each at `--jobs 4`, three runs:

```
run 1: records=1200 blocks=4 order=[a b c d]
run 2: records=1200 blocks=4 order=[d a c b]
run 3: records=1200 blocks=4 order=[b a c d]
```

Four blocks every time, with the block order varying between runs — which is what shows the tasks actually raced rather than quietly serializing.

PR #12006, opened 2026-08-14
fix(install): reinstall when alias backends change

Marukome0743:
## Summary
- reinstall a requested version when its recorded backend identity differs from the backend currently selected by an alias
- retain backend identity per installed version so aliases can migrate incrementally without deleting unrelated versions
- keep legacy and metadata-free installs compatible while filtering old-backend versions out of resolution

## Implementation
The install manifest now records a backend identity for each concrete version while retaining the existing tool-level identity as a fallback for older manifests. On the first successful install after a backend change, existing versions are assigned the previous identity and only the installed version moves to the current identity.

The common install-satisfaction path performs the identity comparison before backend-specific checks, including Cargo, GitHub, and Rust. Exact, fuzzy, latest, and npm range resolution no longer reuse versions known to belong to another backend. Inventory commands continue to show those versions until they are requested and migrated.

Dry runs leave files and metadata untouched, failed installs preserve prior identities, and installs without metadata retain the legacy path-exists behavior.

The cross-process manifest-lock regression test re-executes the test binary without rerunning its destructive shared test-directory initialization. This keeps the lock-contention assertion portable to Windows, where the parent process may still hold files in that directory.

## Verification
- `mise exec -- cargo test --all-features toolset::install_state::tests::manifest_file_locks_serialize_across_processes`
- `mise exec -- cargo test --all-features toolset::install_state::tests`
- `mise exec -- cargo fmt --all -- --check`
- `mise run test:e2e e2e/backend/test_alias_backend_reinstall e2e/backend/test_install_manifest e2e/backend/test_registry_backend_migration e2e/backend/test_github_alias_versions`
- `mise run test:e2e e2e/cli/test_install_system`
- `mise run test:e2e e2e/cli/test_shared_install_dirs`
- `mise exec -- cargo check --all-features`
- `mise exec -- shellcheck e2e/backend/test_alias_backend_reinstall e2e/cli/test_shared_install_dirs`
- `mise exec -- shfmt -d e2e/backend/test_alias_backend_reinstall e2e/cli/test_shared_install_dirs`

Addresses https://github.com/jdx/mise/discussions/6772

### DIFF
--- a/e2e/backend/test_alias_backend_reinstall
+++ b/e2e/backend/test_alias_backend_reinstall
@@ -1,0 +1,90 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+FIXTURES="$TMPDIR/alias-backend-reinstall-fixtures"
+PORT_FILE="$TMPDIR/alias-backend-reinstall-port"
+mkdir -p "$FIXTURES"
+
+for backend in a b; do
+  for version in 1.0.0 2.0.0 3.0.0; do
+    cat >"$FIXTURES/backend-$backend-$version" <<EOF
+#!/usr/bin/env bash
+echo "backend-$backend $version"
+EOF
+    chmod +x "$FIXTURES/backend-$backend-$version"
+  done
+done
+
+python3 -u - "$FIXTURES" "$PORT_FILE" <<'PY' >/dev/null 2>&1 &
+import functools
+import http.server
+import socketserver
+import sys
+from pathlib import Path
+
+root = sys.argv[1]
+port_file = Path(sys.argv[2])
+handler = functools.partial(http.server.SimpleHTTPRequestHandler, directory=root)
+with socketserver.TCPServer(("127.0.0.1", 0), handler) as httpd:
+    port_file.write_text(str(httpd.server_address[1]))
+    httpd.serve_forever()
+PY
+SERVER_PID=$!
+trap 'kill "$SERVER_PID" 2>/dev/null || true' EXIT
+wait_for_file "$PORT_FILE" "alias backend test server port file" 30 "$SERVER_PID"
+PORT=$(cat "$PORT_FILE")
+
+write_config() {
+  local backend="$1"
+  local version="$2"
+  local url_version="${3:-$version}"
+  cat >mise.toml <<EOF
+[tool_alias]
+bd = "http:backend-$backend"
+
+[tools.bd]
+version = "$version"
+url = "http://127.0.0.1:$PORT/backend-$backend-$url_version"
+bin = "bd"
+EOF
+}
+
+cat >mise.toml <<EOF
+[tool_alias]
+bd = "http:backend-a"
+
+[tools]
+bd = [
+  { version = "1.0.0", url = "http://127.0.0.1:$PORT/backend-a-1.0.0", bin = "bd" },
+  { version = "2.0.0", url = "http://127.0.0.1:$PORT/backend-a-2.0.0", bin = "bd" },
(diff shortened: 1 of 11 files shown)


jdx:
Thanks for the effort here, but I'm going to close this one. It touches the core install-state logic in subtle ways (per-version identity records, dual-file manifest merging, migration paths) and I lack the confidence to review this properly. The risk of regressions in this area outweighs the bug it fixes for me right now.

To make that concrete, an in-depth review turned up a number of issues, most of which trace back to two design decisions rather than implementation slips:

**1. Unknown versions inherit the tool-level identity.** `get_version_backend` falls back to the tool-level `full` for any version without an explicit record, so most identity data is inferred rather than recorded, and any exact-string mismatch flips installs to "not satisfied":

- Read-only shared install dirs (`MISE_SHARED_INSTALL_DIRS`) whose recorded `full` doesn't exactly match the client's resolved backend become permanently unsatisfiable — `write_backend_meta` never runs for shared-satisfied tools, so they can never heal, and each client silently reinstalls into its primary dir. The `test_shared_install_dirs` fixture edit (`asdf:tiny` → `vfox:tiny`) works around exactly this instead of preserving compatibility, and even the new fixture mismatches for a fresh client (plugin form `vfox:tiny` vs registry form `vfox:jdx/vfox-tiny`).
- For non-explicit shorts, `BackendArg::full()` intentionally follows the registry's current first backend, so a routine registry backend reorder would now invalidate every installed version of every migrated tool for every user without a lockfile — a fleet-wide reinstall storm, and hard breakage offline/behind proxies.
- An older mise binary rewriting `.mise-installs.toml` silently drops `version_backends` (plain serde, no unknown-field preservation); the next run backfills every version from the tool-level `full`, permanently misattributing old-backend artifacts to the new backend — silently defeating the PR's own invariant under mixed-version usage (project-pinned mise is common).

**2. A mismatch triggers a destructive reinstall.** `create_install_dirs` runs `remove_all` on the existing install before the new backend downloads anything:

- If the new backend's download fails (404, offline), the previously working install is destroyed. The PR's own e2e test ends in exactly this state after the 404 — it asserts the manifest contents but never that `installs/bd/3.0.0/bd` still exists (it doesn't).
- The identity check also fires for `system`/`path:` pseudo-versions (`tv_pathname()` is never recorded, so they inherit the stale tool-level identity). For `path:` pins, `install_path` is the user's absolute path, so a mismatch deletes a directory mise doesn't own, on every run.

Other confirmed issues: exact/prefix resolution no longer short-circuits to installed old-identity versions, so `mise x bd@2.0.0` hard-fails offline where it used to run locally; runtime `latest`/prefix symlinks are rebuilt unfiltered, so they point at foreign-identity artifacts the filtered resolver refuses to return; `merge_manifest_tool` reverses the old sidecar-wins precedence, so a stale consolidated entry (uninstall never prunes them) misattributes restored/copied installs; the `BracketedOptions` migration arm never converges for sidecar-only entries (bracketed identity never matches the stripped `full`, plus a schedule→lock→read→no-op cycle on every invocation); and the `MISE_TEST_SKIP_INIT` guard skips the entire unit-test sandbox, so a leaked env var runs `cargo test` destructively against the real `$HOME`.

If someone wants to pick this problem up again, the shape I'd be more comfortable with is: record identity eagerly on successful installs, but enforce it only when an explicit per-version record exists (no fallback to tool-level `full`, never for `system`/`path:`), and make a mismatch warn or require `--force` rather than wiping before the new download has succeeded. That likely removes the need for the dual-file merge protocol and migration machinery entirely.

PR #12009, opened 2026-08-14
e2e: install libgtk-3-0t64 so GTK tools can run their registry test

dmezhnov:
One package, `libgtk-3-0t64`, so that a registry tool shipping a GTK 3 binary can answer its `test` command inside `ghcr.io/jdx/mise:e2e`.

The image already installs the rest of the desktop runtime — `libgl1`, `libglib2.0-0`, `libsm6`, `libwayland-client0`, `libxkbcommon0`, `libxi6`, `libxfixes3`, `libasound2t64` — but not GTK. For a GTK binary that is fatal before `main` runs, so even a `--version` that deliberately never opens a window cannot be tested:

```console
$ docker run --rm -v /tmp/kc:/kc:ro ghcr.io/jdx/mise:e2e sh -lc '/kc/knitcalc --version'
/kc/knitcalc: error while loading shared libraries: libgtk-3.so.0: cannot open shared object file: No such file or directory
$ echo $?
127
```

With this line added (measured on `ghcr.io/jdx/mise:e2e@sha256:905b8812be0d63aa9858d8f2a343f0f70b5489c0499376f147979fd239201260`, Ubuntu 26.04, plus `apt-get install -y --no-install-recommends libgtk-3-0t64`):

```console
$ docker run --rm -v /tmp/kc:/kc:ro mise-e2e-gtk sh -lc '/kc/knitcalc --version'
knitcalc 1.8.79+102
$ echo $?
0
```

which is what `test = { cmd = "knitcalc --version", expected = "knitcalc {{version}}" }` needs.

I would like to follow this with a `registry/knitcalc.toml` entry (`github:dmezhnov/knitcalc` — a cross-platform desktop app whose native runners answer `--version` headless on Linux, macOS and Windows precisely so that a package manager can test the binary). Since the e2e image is rebuilt only on a mise release tag, that PR cannot be green until this one is merged and a new image is published, so I am sending them in order instead of together.

If you would rather not carry a GTK runtime in the e2e image, say so and I will look for a different arrangement for the entry.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runtime compatibility for end-to-end testing by including the required GTK library in the application environment.

### DIFF
--- a/packaging/e2e/Dockerfile
+++ b/packaging/e2e/Dockerfile
@@ -32,6 +32,7 @@ apt-get update && apt-get install -y --no-install-recommends \
     libffi-dev \
     libgl1 \
     libglib2.0-0 \
+    libgtk-3-0t64 \
     libicu-dev \
     libncurses-dev \
     libreadline-dev \


dmezhnov:
Thanks for merging this!

One follow-up: the published `ghcr.io/jdx/mise:e2e` has not been rebuilt since, so the GTK library is not in the image the registry tests actually pull.

- `ghcr.io/jdx/mise:e2e` is still `sha256:905b8812…`, created 2026-08-07T21:21Z.
- The `docker` workflow's most recent run is for tag `v2026.8.3` (2026-08-07); `v2026.8.4`, `v2026.8.5` and `v2026.8.6` produced no run, even though `release.yml` has the identical `on: push: tags: ["v[0-9]*"]` trigger and did run for all three. The workflow is not disabled, so I can't tell from the outside what stops it.

Could you kick off a `workflow_dispatch` of `docker` (or otherwise refresh the `e2e` tag)? `registry.yml` pulls that tag rather than building it, so until the image is refreshed any registry entry for a GTK-linked tool still fails on `libgtk-3.so.0` — which is what this PR was meant to fix. I'm holding back the follow-up PR that adds `registry/knitcalc.toml` until then, so it lands green.

dmezhnov:
Update: no action needed here — `docker` fired on its own for `v2026.8.7` (2026-08-17 03:48Z) and the published `ghcr.io/jdx/mise:e2e` now carries the GTK runtime, so no manual dispatch is required. The registry entry is up as #12087 and its `test-tool` job passes on that image.

For the record, the gap itself is still unexplained: `docker` has no runs for `v2026.8.4`, `.5` and `.6` although the workflow is unchanged and `release.yml` ran on the same tags — so it may recur. Happy to open a separate issue if that is useful.

PR #12010, opened 2026-08-14
fix(backend): preserve installed backend metadata

jdx:
## Summary

- preserve backend identity loaded from installed tool metadata when a same-named plugin is discovered
- retain plugin discovery as a fallback when no installed metadata exists
- add regression coverage for both precedence paths

## Root cause

Plugin discovery unconditionally replaced the backend recorded in `.mise.backend.toml`. When a GitHub-backed tool shared a shorthand with an installed asdf plugin, mise interpreted the existing install through the plugin and could add `<install>/bin` to `PATH` instead of the install root.

Related: https://github.com/jdx/mise/discussions/12005

## Validation

- `cargo test --bin mise toolset::install_state::tests::plugin_discovery`
- scoped `hk run fix --safe` (`cargo-fmt`, `cargo-check --all-features`)

---

> [!NOTE]
> **Medium Risk**
> Changes install-state backend resolution used for PATH and backend selection; behavior is narrowed to fix incorrect overwrites, with targeted tests.
> 
> **Overview**
> Fixes a bug where **plugin discovery overwrote** the backend identity already loaded from installed tool metadata (`.mise.backend.toml` / manifest). A GitHub-backed install that shared a shorthand with an asdf plugin could be treated as the plugin backend, which led to wrong **PATH** layout (e.g. `<install>/bin` vs install root).
> 
> Plugin merging is moved into **`merge_plugin_tools`**, which only sets `tool.full` when it is still missing (`get_or_insert`), so on-disk metadata wins and plugin IDs remain a fallback for tools without installed metadata. **Regression tests** cover both precedence paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f45dc944154b109c12b87342b4920d9f0c3cb778. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved plugin tool discovery and initialization for more reliable availability.
  - Preserved existing installed tool metadata when plugin information is also available.
  - Added missing metadata for plugin-provided tools when no existing information is present.
  - Ensured the combined tool list is consistently cached and returned after initialization, providing stable results across subsequent use.

### DIFF
--- a/src/toolset/install_state.rs
+++ b/src/toolset/install_state.rs
@@ -413,7 +413,17 @@ async fn init_tools() -> MutexResult<InstallStateTools> {
         }
     }
 
-    for (short, pt) in init_plugins().await?.iter() {
+    let plugins = init_plugins().await?;
+    merge_plugin_tools(&mut tools, plugins.as_ref());
+    let tools = Arc::new(tools);
+    *INSTALL_STATE_TOOLS
+        .lock()
+        .expect("INSTALL_STATE_TOOLS lock failed") = Some(tools.clone());
+    Ok(tools)
+}
+
+fn merge_plugin_tools(tools: &mut InstallStateTools, plugins: &InstallStatePlugins) {
+    for (short, pt) in plugins {
         let full = match pt {
             PluginType::Asdf => format!("asdf:{short}"),
             PluginType::Vfox => format!("vfox:{short}"),
@@ -430,13 +440,10 @@ async fn init_tools() -> MutexResult<InstallStateTools> {
                 opts: BTreeMap::new(),
                 installs_path: None,
             });
-        tool.full = Some(full);
+        // Installed metadata describes the versions already on disk. Plugin
+        // discovery should only supply an identity when no metadata exists.
+        tool.full.get_or_insert(full);
     }
-    let tools = Arc::new(tools);
-    *INSTALL_STATE_TOOLS
-        .lock()
-        .expect("INSTALL_STATE_TOOLS lock failed") = Some(tools.clone());
-    Ok(tools)
 }
 
 pub fn list_plugins() -> Arc<BTreeMap<String, PluginType>> {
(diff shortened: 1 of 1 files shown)


stevenpollack:
Heads-up: this appears to have broken `cli/test_where` on `main`, which is failing the `e2e` job on unrelated PRs.

**Bisected**, building and running `./e2e/run_test cli/test_where` at each point:

| commit | result |
| --- | --- |
| `f51149149` (this commit's parent) | passes |
| `1e6cdb4db` (this commit) | **fails** |
| `0742b33f8` (current main) | **fails** |

```
$ mise i
$ mise where dummy
Error:
   0: dummy not found in mise tool registry
Location: src/cli/args/backend_arg.rs:380
```

**Mechanism.** The install-dir scan populates `full` from the on-disk manifest before `merge_plugin_tools` runs (`install_state.rs:410-412`). In this test the manifest at `installs/dummy/.mise.backend.toml` reads:

```toml
short = "dummy"
full = "core:dummy"
explicit_backend = true
```

`core:dummy` does not resolve — `dummy` is an asdf plugin in the e2e fixture. Previously `tool.full = Some(full)` overwrote it with `asdf:dummy`, so plugin discovery was silently repairing the recorded identity. With `tool.full.get_or_insert(full)` the stale `core:dummy` is preserved and the lookup falls through to the registry.

So the two cases are harder to separate than "installed metadata wins": here the installed metadata is the wrong one, and the plugin identity was the correct one.

**Why the new tests don't catch it.** `plugin_discovery_preserves_installed_backend_metadata` uses `full = Some("github:babashka/babashka")` and `plugin_discovery_adds_missing_tool_metadata` uses `full = None`. Neither covers a recorded `full` that names a backend which doesn't resolve while an installed plugin of the same short name does — which is the shape that regressed. A case with `full = Some("core:dummy")` plus `PluginType::Asdf` would fail today.

I have no strong opinion on the right fix — whether to prefer the plugin identity when the recorded backend doesn't resolve, or to stop writing `core:` into the manifest for plugin-backed tools — so I haven't put up a patch. Happy to if you'd like one.

---

stevenpollack:
Resolved by #12029 — confirming so nobody re-chases this: `cli/test_where` passes on current `main` (verified 3/3 runs at `badd22b71`). The fixture now records the real backend rather than `core:dummy`, so plugin discovery no longer has anything to correct. Thanks.

PR #12015, opened 2026-08-14
feat(bootstrap): add launchd throttle and queue keys

CallumKerson:
Adds two additive launchd plist keys to
`[bootstrap.macos.launchd.agents]`:

- `throttle_interval` -> [`ThrottleInterval`](https://keith.github.io/xcode-man-pages/launchd.plist.5.html#ThrottleInterval)
- `queue_directories` -> [`QueueDirectories`](https://keith.github.io/xcode-man-pages/launchd.plist.5.html#QueueDirectories)

`queue_directories` entries expand bare `~`/`~/` like the other path-valued keys.

## New Features
- Added macOS launchd support for throttle intervals and queue-directory triggers.
- Supports configuring one or multiple queue directories.
- Expands queue-directory paths before generating launchd configuration.

## Bug Fixes
- Rejects empty, whitespace-only, relative, or invalid queue-directory entries.
- Validates throttle intervals to prevent negative values.

## Documentation
- Documented supported launchd settings, throttling behavior, path expansion, and queue-directory requirements.

### DIFF
--- a/docs/bootstrap/launchd.md
+++ b/docs/bootstrap/launchd.md
@@ -32,16 +32,19 @@ numbers, `.`, `_`, and `-`. mise owns only the plist files it creates with the
 | `run_at_load`             | `RunAtLoad`               |
 | `keep_alive`              | `KeepAlive`               |
 | `start_interval`          | `StartInterval`           |
+| `throttle_interval`       | `ThrottleInterval`        |
 | `start_calendar_interval` | `StartCalendarInterval`   |
+| `queue_directories`       | `QueueDirectories`        |
 | `environment`             | `EnvironmentVariables`    |
 | `working_directory`       | `WorkingDirectory`        |
 | `stdout_path`             | `StandardOutPath`         |
 | `stderr_path`             | `StandardErrorPath`       |
 | `kickstart`               | run `launchctl kickstart` |
 
-`program`, `working_directory`, `stdout_path`, and `stderr_path` expand bare
-`~` and `~/` to the current user's home directory before writing the plist.
-`args` are passed through exactly as written.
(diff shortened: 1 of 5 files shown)


CallumKerson:
Dropped `limit_load_to_session_type` from this PR (it was in the original title/description, now updated).

`apply()` always bootstraps into `gui/$UID` and writes to `~/Library/LaunchAgents`, and `is_loaded()` checks that same domain. That makes `Aqua` the only value that can work — and launchd already defaults to `Aqua` when the key is omitted, so every working configuration would have been a no-op.

The non-`Aqua` values are worse than useless. Either `launchctl bootstrap` fails, in which case `apply()` returns early and every agent later in the list is silently skipped, or the job loads nowhere and `is_loaded()` stays false — which leaves `status` reporting `unloaded` permanently, `status --missing` exiting 1 on every run, and `apply_launchd_with_report` (which filters on `state != Loaded`) rewriting and re-bootstrapping the agent on every `mise bootstrap`. The array form doesn't help either: `["Aqua", "Background"]` won't load into the background context, since that needs a separate bootstrap into `user/$UID` that mise never performs. Per Apple's docs, `Background` agents generally aren't managed from `~/Library/LaunchAgents` at all, and `System` is for `/Library/LaunchDaemons`.

Restricting the accepted set to `Aqua` was the alternative, but that just makes the no-op explicit. Since a shipped config key would need a 12-month deprecation window to remove, and adding it later is cheap, it seemed better to leave it out until mise can actually target a domain other than `gui/$UID`.

The remaining two keys are unaffected. `queue_directories` now also rejects empty and whitespace-only entries at config time, so launchd never receives a plist it would refuse to load.

PR #12038, opened 2026-08-15
fix(go): rebuild tools when go version changes

Marukome0743:
## Summary

- Record the concrete mise-managed Go compiler used to build each `go:` tool version.
- Rebuild an installed Go tool when the selected managed Go backend or resolved version changes.
- Preserve the existing path-only freshness behavior when Go comes from the system or ambient `PATH`.

## Root cause

The Go backend treated an existing install directory as sufficient, even though `go install` produces binaries tied to the compiler used for the build. As a result, changing the configured Go version left tools such as Staticcheck built with the previous compiler.

Each successful managed build now writes an atomic `.mise-go-install.toml` sidecar containing the compiler backend identity and opaque resolved version. Missing, malformed, unsupported, or mismatched state is considered stale. Existing installs therefore migrate with one rebuild, while failed installs never record the new identity.

External Go installations are intentionally not probed with `go version` during freshness checks, avoiding a new process on every install check.

## Verification

- `mise exec -- cargo test --all-features backend::go::tests --no-fail-fast -- --nocapture` — 21 passed
- `mise exec -- cargo test --all-features backend::go::tests::compiler -- --nocapture` — 3 passed after the final state-format update
- `mise run test:e2e e2e/backend/test_go_compiler_identity`
- `mise run test:e2e e2e/backend/test_go_shim_recursion`
- `mise exec -- cargo check --all-features`
- `mise run docs:build`
- `shellcheck e2e/backend/test_go_compiler_identity`
- `shfmt -d e2e/backend/test_go_compiler_identity`
- `bun x prettier --check docs/dev-tools/backends/go.md`

The workspace-wide Clippy command currently stops on the pre-existing `clippy::needless_return` warning in `src/system/packages/brew/cask.rs` on the latest main. The aggregate `mise run format` and `mise run lint` tasks also cannot initialize hk in this native Sapling checkout because hk requires Git repository discovery; their underlying checks listed above were run directly.

Addresses https://github.com/jdx/mise/discussions/8070

## Summary by CodeRabbit

* **New Features**
  * Go-managed tools now track the compiler version used to install them.
  * Tools are automatically rebuilt when the configured Go compiler changes.
  * Existing installations are updated to support compiler version tracking.

* **Bug Fixes**
  * Improved detection of outdated Go tools and clearer dry-run reporting.
  * Preserved compatibility for Go installations provided through the system PATH.
  * Failed installations no longer retain incomplete compiler state.

* **Documentation**
  * Added guidance on Go compiler version tracking and installation behavior.

### DIFF
--- a/docs/dev-tools/backends/go.md
+++ b/docs/dev-tools/backends/go.md
@@ -18,6 +18,12 @@ Any method of installing `go` is fine if you want to install go some other way.
 mise will use whatever `go` is on PATH.
 :::
 
+When Go is configured through mise, the backend records both the compiler backend identity and the
+concrete Go version used to build each command. Changing either the managed Go backend or version
+causes existing `go:` tools to be rebuilt the next time mise installs missing tools. Existing
+installs created before this metadata was added are rebuilt once to record their compiler. Go
+installations supplied only through `PATH` keep the legacy path-based install check.
+
 ## Usage
 
 The following installs the latest version of [hivemind](https://github.com/DarthSim/hivemind) and
--- a/e2e/backend/test_go_compiler_identity
+++ b/e2e/backend/test_go_compiler_identity
@@ -1,0 +1,96 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+export MISE_YES=1
+install_log="$HOME/go-install.log"
+
+make_go_stub() {
+  local version="$1"
+  local fail="${2:-false}"
+  local bin_dir="$MISE_DATA_DIR/installs/go/$version/bin"
+  mkdir -p "$bin_dir"
+  cat >"$bin_dir/go" <<EOF
+#!/usr/bin/env bash
+set -euo pipefail
+compiler_version="$version"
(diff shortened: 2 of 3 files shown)


jdx:
Thanks for working on this. I don't think we should make every `go:` tool installation depend on the currently selected Go compiler version.

mise installations are globally keyed by the tool version, not by the compiler version. Two projects using the same `go:` tool version with different Go versions would therefore repeatedly rebuild and overwrite the same installation as users move between projects. Most Go binaries are standalone and do not require rebuilding when Go changes; Staticcheck's analyzer compatibility constraint is a special case and shouldn't impose this behavior on the entire backend.

This would also trigger a one-time rebuild of every existing managed-Go tool, and a failed rebuild after a compiler change can remove the previously working installation.

For now, an explicit forced reinstall is preferable. If we automate this later, it should be opt-in or part of a generic dependency-aware install identity that models separate artifacts correctly.

Closing, but I appreciate the investigation and the thorough tests.

PR #12082, opened 2026-08-17
mise: armv7: use hf suffix for gnueabi

neheb:
Fixes usage on Debian 11/12 and armhf.

## Summary by CodeRabbit

* **Chores**
  * Updated Linux ARMv7 release builds to use the hard-float target.
  * Updated the ARMv7 cross-compilation environment to match, improving compatibility for supported ARMv7 devices.


neheb:
The Cross.toml change is telling. This was always supposed to be hf.

PR #12145, opened 2026-08-18
perf(vfox): cache plugin metadata on disk

jdx:
## Problem

Loading a filesystem vfox plugin's metadata boots a Lua VM, registers the Lua stdlib modules, and executes the plugin's **top-level metadata.lua** — which can do arbitrary work (`jdx/vfox-mysql` probes `apt-cache show libaio1t64` at load time, ~375ms). This runs in every mise process that queries a plugin's idiomatic filenames, tool dependencies, or system dependencies. The asdf backend already disk-caches the equivalent (`idiomatic_filenames.msgpack.z`); the vfox backend never got the same treatment.

Companion to #12143, which stops metadata from being queried at startup at all when idiomatic version files aren't enabled; this PR makes the remaining legitimate queries (enabled tools, dependency resolution, system deps) hit disk instead of Lua.

## Fix

Cache the metadata-derived values (`legacy_filenames`, `depends`, `system_dependencies`) per plugin in `<cache>/<tool>/metadata.msgpack.z`, invalidated by the plugin dir and `metadata.lua` mtimes — the same freshness scheme the asdf backend uses. `_idiomatic_filenames`, `get_dependencies`, and `system_dependencies` all read through the snapshot. Embedded plugins are unaffected: they have no plugin dir, their metadata is compiled in, and loading it is sub-millisecond.

`vfox::SystemDependency` gains serde derives so the snapshot can round-trip (plain-data struct; the `serde` dependency was already present, this just enables `derive`).

## Tests

New e2e test `e2e/backend/test_vfox_metadata_cache`: a plugin whose `metadata.lua` writes a marker file must run once, then be served from cache (no marker on re-run), then re-run after `touch metadata.lua`.

---

> [!NOTE]
> **Medium Risk**
> Stale metadata could misreport idiomatic files, tool deps, or system deps if the Lua fingerprint ever misses a change; behavior is localized to vfox filesystem plugins and is covered by invalidation tests.
> 
> **Overview**
> **Filesystem vfox plugins no longer boot Lua on every mise run** when resolving idiomatic filenames, `depends`, or `systemDependencies`. Those fields are stored in `metadata.msgpack.z` and loaded through a new `VfoxMetadataSnapshot` path in `VfoxBackend`.
> 
> **Cache invalidation keys off a hash of every `.lua` file under the plugin tree**, not `metadata.lua` mtime alone, so changes to `require`’d sibling modules still refresh the snapshot. Embedded plugins (no on-disk plugin dir) skip the cache and keep the existing cheap load path.
> 
> `vfox::SystemDependency` gains **serde** derives (and `serde`’s `derive` feature in `crates/vfox`) so the snapshot can round-trip. New e2e `test_vfox_metadata_cache` asserts one Lua execution, cache reuse, and refresh when `lib/meta.lua` or `metadata.lua` changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 974e1acee1311c495ed7aad7e82ad49ad26a7d2a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added caching for plugin metadata to speed up repeated environment and tool lookups.
  - Cached metadata includes supported filenames, tool dependencies, and system dependencies.
  - System dependencies can now provide an ordered list of candidate package names.
  - Metadata automatically refreshes when plugin source files or dependencies change.
  - Improved filename detection using cached plugin information.
  - Embedded plugins continue loading metadata without disk caching.

- **Tests**
  - Added end-to-end coverage for metadata caching, reuse, and invalidation.

### DIFF
--- a/crates/vfox/Cargo.toml
+++ b/crates/vfox/Cargo.toml
@@ -44,7 +44,7 @@ once_cell = "1"
 reqwest = { version = "0.13", default-features = false, features = [
   "json",
 ] } # TODO: replace with xx
-serde = "1"
+serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 toml = "1"
 thiserror = "2"
--- a/crates/vfox/src/metadata.rs
+++ b/crates/vfox/src/metadata.rs
@@ -29,7 +29,7 @@ pub struct Metadata {
 /// apt's `libaio1` became `libaio1t64` in the 64-bit time_t transition). The
 /// consumer picks the first candidate that exists, so plugins never have to
 /// probe the host while their metadata loads.
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone, Default, serde::Serialize, serde::Deserialize)]
 pub struct SystemDependency {
     pub bin: Option<String>,
     pub pkgconfig: Option<String>,
--- a/e2e/backend/test_vfox_metadata_cache
+++ b/e2e/backend/test_vfox_metadata_cache
@@ -2,16 +2,21 @@
 # vfox plugin metadata is cached on disk per plugin. Loading it boots a Lua VM
 # and executes the plugin's top-level metadata.lua, which can do arbitrary work
 # (e.g. probe system packages), so it must not re-run on every mise invocation —
-# only when the plugin changes.
+# only when the plugin's Lua sources change.
 
 marker="$PWD/metadata-loaded"
+PLUGIN_DIR="$MISE_DATA_DIR/plugins/sideeffect"
+mkdir -p "$PLUGIN_DIR/lib"
 
-mkdir -p "$MISE_DATA_DIR/plugins/sideeffect"
-cat <<EOF >"$MISE_DATA_DIR/plugins/sideeffect/metadata.lua"
+# metadata.lua derives a field from a sibling module, so the cache has to track
+# that module too.
+write_metadata() { # $1 = version
+  cat <<EOF >"$PLUGIN_DIR/metadata.lua"
+local meta = require("meta")
 PLUGIN = {
     name = "sideeffect",
-    version = "0.1.0",
-    description = "records when its metadata is loaded",
+    version = "$1",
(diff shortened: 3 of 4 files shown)


jdx:
Thanks both — addressed in `8a1e2f1`/latest push.

**@greptile — sibling module changes stay stale: valid, fixed.** You're right that the two tracked mtimes miss it: writing a file in place changes neither its parent directory's mtime nor `metadata.lua`'s. The cache is now keyed on a hash of *every* `.lua` source under the plugin, computed lazily so plugins whose metadata is never requested still read nothing.

I verified the new e2e coverage actually catches this rather than just passing: with the cache reverted to the old mtime keying, the sibling-module phase fails; with the fingerprint, all five phases pass.

**@coderabbitai — embedded plugin dependency metadata: this is not a regression.** The pre-existing `load_metadata_deps` on `main` is:

```rust
let plugin_path = dirs::PLUGINS.join(&self.pathname);
if !plugin_path.exists() {
    return Ok(vec![]);   // embedded plugins have no plugin dir
}
```

`load_system_deps` is identical. Embedded plugins already returned no dependencies before this PR, so the snapshot path preserves behavior exactly. `_idiomatic_filenames` differs because its pre-existing implementation went through `vfox.metadata()`, which *does* work for embedded plugins — which is why that one keeps a fallback and these two do not.

That said, you've surfaced a real **pre-existing** bug: four embedded plugins (`vfox-lua`, `vfox-redis`, `vfox-chicken`, `vfox-chezscheme`) declare `systemDependencies` that mise silently ignores, so e.g. `mise install lua` compiles from source and fails partway instead of reporting missing build tools up front. I've deliberately kept it out of this PR to keep the caching change behavior-neutral — `system_deps` defaults to `prompt`, so honoring those declarations is a user-visible change that deserves its own PR and its own release note.

PR #12149, opened 2026-08-18
feat(system): let plugins declare candidate package names

jdx:
## Problem

A plugin's `systemDependencies` hint maps a package manager to exactly **one** package name. But the same capability is packaged under different names across distro releases — apt renamed `libaio1` to `libaio1t64` in the 64-bit `time_t` transition (Ubuntu ≥ 24.04, Debian ≥ 13), and `libncurses5`/`6` and `libssl1.1`/`3` have the same shape. There is no axis in the API to express that.

So the only workaround is probing the host from `metadata.lua` — which is what [jdx/vfox-mysql](https://github.com/jdx/vfox-mysql) does (`apt-cache show libaio1t64` at the top level). That is bad twice over:

1. **It runs constantly.** `metadata.lua`'s top level executes whenever *any* metadata field is read — including `legacyFilenames` and `depends` on ordinary `mise x`/`mise run`. Measured ~350ms per invocation. (#12143 and #12145 stop those reads; this PR removes the reason to probe at all.)
2. **It poisons caching.** Once plugin metadata is cached (#12145), a probe result is frozen on disk keyed on plugin mtime — so a user who upgrades Ubuntu 22.04 → 24.04 keeps a stale hint and gets offered a package that no longer exists. Caching *data* is correct; caching a *machine probe* is not, and metadata cannot tell the two apart.

## Fix

Accept a list of candidates per manager, resolved to the first the manager actually has:

```lua
{ sharedlib = "libaio.so.1",
  packages = { apt = { "libaio1t64", "libaio1" }, dnf = "libaio" } },
```

**Resolution is lazy by construction.** Package hints are consulted only for a dependency that already failed its check ([helpers.rs:125](src/toolset/helpers.rs#L125), reached only after `missing_required` is non-empty) — i.e. when mise is about to run a package install behind a prompt anyway. A single-candidate hint queries nothing at all, so the common case is unchanged.

Adds `SystemPackageManager::available()`, which is *availability*, not installed state — `installed()` cannot answer it, since apt's asks dpkg, which only knows what is already on the box. The default implementation reports every name as available, so managers that do not implement it pick the first candidate exactly as before. apt implements it with `apt-cache policy`, which correctly excludes a name apt knows only as a transitional package with `Candidate: (none)`.

Fully backward compatible: a bare string is still valid and means a single candidate.

## Tests

- vfox metadata: candidate lists and bare strings both parse.
- `resolve_package`: picks the first *available* candidate (not merely the first listed), keeps the first when none exist so the user still sees a name, issues **no query** for a single candidate, and falls back when the manager errors.
- `parse_apt_cache_policy`: fixture covering an installable package, a `Candidate: (none)` transitional package, and a name with no stanza.
- `test_available_against_real_apt_cache`: runs the real `apt-cache` where present (Linux CI) so the parser cannot drift from apt's output format.
- New e2e `test_plugin_system_deps_package_candidates`: a plugin lists a nonexistent name **first**, and mise's hint resolves past it. Skips where apt-cache is absent.

Verified end-to-end on Ubuntu noble against real apt: a plugin declaring `{ "libaio1", "libaio1t64" }` (stale name first) produced `install with: sudo apt-get install -y libaio1t64`.

## Follow-up

jdx/vfox-mysql#2 currently reads `/etc/os-release` to dodge the subprocess; once this lands, that heuristic can be deleted in favor of a candidate list. The docs example in this PR is updated for the same reason — it previously showed `apt = "libaio1"`, which is wrong on current Debian/Ubuntu.

---

> [!NOTE]
> **Medium Risk**
> Changes system dependency remediation and apt package queries on the install path; behavior is backward compatible for single-name hints but new apt-cache calls run when multiple candidates are declared.
> 
> **Overview**
> Plugin **`systemDependencies`** package hints can now be a **single name or an ordered list of candidates** per manager (e.g. `apt = { "libaio1t64", "libaio1" }`), so the same capability can be declared across distro renames without probing the host from `metadata.lua`. Bare strings still mean one candidate and stay backward compatible.
> 
> **Remediation** resolves hints lazily when a dependency is missing: one candidate skips any query; multiple candidates call a new **`SystemPackageManager::available`** hook and pick the first installable name (others default to the first listed). Hints are trimmed, empty lists are treated as unremediable, and **apt** implements availability via **`apt-cache policy`** (including arch-qualified names).
> 
> Docs and an e2e cover candidate syntax and apt resolution behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 664e594b18205ebdac80f1b062e94deb7affe71d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - System dependencies can specify ordered package-name candidates while retaining single-name syntax.
  - The first available candidate is selected automatically, with graceful fallback when checks are unavailable.
  - APT environments detect installable candidates, including architecture-qualified names.

- **Bug Fixes**
  - Prevents stale or unavailable candidates from being reported when a valid alternative exists.
  - Ignores blank or unusable candidate entries.

- **Documentation**
  - Updated plugin guidance with candidate-list syntax, resolution behavior, and examples.


jdx:
All three addressed in the latest push.

**@cursor — apt availability ignores locale: valid, fixed.** apt translates the stanza labels the parser reads, so on a non-English host every candidate would look unavailable and resolution would silently fall back to the first name — defeating the feature exactly where it matters. The query now pins `LC_ALL=C` and `LANGUAGE=C`. (I could not reproduce locally: this container ships no apt translation catalogs, so there was nothing to mistranslate. The fix is correct regardless, and pinning the locale is the right default for parsing tool output.)

**@greptile / @coderabbitai — empty candidate lists become package names: valid, fixed.** `packages = { apt = {} }` selected the manager and then produced an empty operand. `pick_manager` now skips entries whose candidate list is empty, and `build_requests` reports the dependency as unremediable when nothing resolves, instead of defaulting to `""`. Added `test_empty_candidate_list_is_unremediable` covering exactly `packages = { apt = {} }`.

**@coderabbitai — document the default-availability fallback: valid, fixed.** The docs now say that only managers which can be queried for availability (currently apt) do the selection, that everything else uses the first candidate, and that candidates should be ordered newest-name-first.

jdx:
@cursor good catch — removed in `b3a504a`, and the PR is now down to the six source files it should touch.

**How it happened:** mise's unit-test harness redirects `HOME` to `<repo>/test` ([src/test.rs](https://github.com/jdx/mise/blob/main/src/test.rs)) but does not redirect `XDG_CACHE_HOME`, so aube writes its cache to `$HOME/.cache/aube` → `test/.cache/aube/`. Running the test suite creates it. I then staged with `git add -A` instead of naming paths, which swept it into the commit. My mistake — the artifacts were never intended as fixtures.

**Worth knowing:** `test/.gitignore` already ignores `cache/`, `data/`, `state/`, and `cwd/`, but not `.cache/` — so this directory is untracked-but-not-ignored for anyone who runs `cargo test`. That is the trap that made the accident possible, and it will catch others. Happy to send a one-line `chore:` PR adding `.cache/` to `test/.gitignore`; I have kept it out of this PR since it is unrelated to the feature.

I re-audited the other two PRs in this series (#12143, #12145) — both are clean, no stray files.

jdx:
All three addressed in `2242e3f`.

**@cursor — arch fallback poisons availability: valid, and it reproduces exactly as described.** I confirmed against real apt before fixing:

```
$ apt-cache policy bash:amd64      # header is 'bash:', NOT 'bash:amd64:'
$ apt-cache policy bash:arm64      # no output at all
$ apt-cache policy bash:arm64 bash # one 'bash:' stanza — indistinguishable
```

apt heads every stanza with the *bare* name, so a batched result genuinely cannot tell `bash:arm64` from `bash`, and my `dpkg_name` fallback made the unavailable one look installable. Note the reverse is equally true, so simply dropping the fallback would have been wrong too: exact-matching would report an installable `bash:amd64` as unavailable.

Fixed by splitting the query — bare names in one batch, each arch-qualified name on its own, where apt disambiguates by emitting a stanza or nothing (exit 0 either way, including for a bogus arch). The common case still costs one process. The real-apt test now covers the poisoning case directly: `available(["bash:mise-not-an-arch", "bash"])` must be `[false, true]`.

**@cursor — vacuous empty-candidate unit test: correct, it asserted nothing.** The dep was keyed on the fictional manager `stub`, and `pick_manager` only iterates `all_managers()`, so it returned `None` for want of a real manager rather than for the empty list — the assertions held with a non-empty list too. Replaced with two tests that do exercise the paths: one driving `resolve_package` directly through the stub, and one that picks a manager actually available on the host (skipping where none is). I mutation-tested the latter — removing the `!c.is_empty()` check in `pick_manager` makes it fail, which the old test did not.

**@greptile — blank candidates remain remediable: valid, fixed.** `{ "" }` passed the non-empty-vector check and produced an empty operand. Blank names are now dropped when converting plugin metadata, and a hint left with no usable name is removed entirely so the dep reports as unremediable. Covered by `test_blank_candidate_names_are_dropped`.

Also rebased onto `6015a21`; I closed #12150 since it duplicated that `.cache/` line.

jdx:
@greptile — valid, fixed in `664e594`. The filter tested `c.trim().is_empty()` but collected the *original* string, so `"  libaio  "` passed validation and would have reached both the install hint and the apt operand list padded. Now trimmed before the emptiness check, as suggested.

Extended `test_blank_candidate_names_are_dropped` to cover it, and confirmed the test actually pins the behavior rather than passing by luck — restoring the untrimmed collect fails it:

```
assertion `left == right` failed
  left: Some(["  libaio  "])
 right: Some(["libaio"])
```

Apologies for the delay on this one — it landed at 21:35, just inside the window I had filtered from on my previous sweep, so I picked it up a round late.

PR #12194, opened 2026-08-20
fix(config): skip disabled idiomatic tracked configs

xqm32:
## Summary

Fixes a regression introduced by #12143. If an idiomatic file such as `package.json` was tracked while enabled and is later disabled, read-only tracked-config consumers can report that mise "cannot update" the file.

This PR distinguishes a confirmed-disabled idiomatic file from an unknown file or a discovery failure. Tracked scans skip only the confirmed-disabled case and keep the tracking entry for possible re-enablement.

## Reproduction

Originally reproduced with mise v2026.8.9:

```bash
(
  set -eu
  tmp="$(mktemp -d)"
  trap 'rm -rf "$tmp"' EXIT

  export MISE_CACHE_DIR="$tmp/cache"
  export MISE_CONFIG_DIR="$tmp/config"
  export MISE_DATA_DIR="$tmp/data"
  export MISE_STATE_DIR="$tmp/state"
  export MISE_SYSTEM_DIR="$tmp/system"
  export MISE_TRUSTED_CONFIG_PATHS="$tmp"

  mkdir -p "$tmp/project" "$tmp/elsewhere"

  cat >"$tmp/project/package.json" <<'EOF'
{"devEngines":{"runtime":{"name":"node","version":"22.0.0"}}}
EOF
  cat >"$tmp/project/mise.toml" <<'EOF'
[settings]
idiomatic_version_file_enable_tools = ["node"]
EOF

  # Parse and track package.json while its idiomatic parser is enabled.
  mise -C "$tmp/project" current node >/dev/null

  # Keep the tracking entry but disable the parser.
  cat >"$tmp/project/mise.toml" <<'EOF'
[settings]
idiomatic_version_file_enable_tools = []
EOF

  mise -C "$tmp/elsewhere" ls --all-sources
)
```

### Actual behavior

```text
mise WARN  error loading tracked config file …/package.json: cannot update idiomatic version file …/package.json; use mise.toml, .tool-versions, or --path to choose a writable config file
```

No write was requested; `mise ls --all-sources` was only reading tracked configs.

### Expected behavior

The disabled `package.json` is omitted without a warning. Its tracking entry remains, so re-enabling node idiomatic files makes it active again.

## Implementation

- represent config detection as explicit `recognized`, `disabled`, `unknown`, and `discovery failed` outcomes
- share one root-aware Settings source pipeline, parameterized by source and trust policy, between normal Settings initialization and rooted tracked/monorepo resolution
- expose only the two-field idiomatic Settings projection to rooted callers and cache that projection per config root
- classify fixed config filenames before rooted Settings resolution, avoiding unnecessary hierarchy scans and duplicate diagnostics for ordinary tracked configs
- filter untrusted owner settings before plugin discovery, while preserving safe mode's trust exemption even when paranoid mode is also enabled
- retain disabled matches returned by dynamic plugin discovery instead of degrading them to `unknown`
- reuse one detection result for trust checks and parsing while preserving malformed-config and backend-discovery diagnostics
- continue rejecting explicit writes to idiomatic version files

The rooted path uses the canonical Settings source ordering, parsing, and TOML discovery, so environment precedence, `--no-config`, safe mode, ignored paths, global/system files, and ancestor precedence do not have a separate implementation that can drift. The full canonical Settings schema is intentionally parsed before projecting the two idiomatic fields; a reduced parser would create a second validation path.

Registry metadata provides the disabled fallback when a backend is not enabled and therefore cannot be queried. Dynamically supplied filenames are also classified as disabled when metadata discovery succeeds. Metadata failures remain explicit discovery failures rather than being guessed away.

## Validation

- `RUSTUP_TOOLCHAIN=1.95.0 cargo check --all-features --all-targets`
- `RUSTUP_TOOLCHAIN=1.95.0 cargo check -p mise --no-default-features --features rustls-native-roots,vfox/vendored-lua,self_update --all-targets`
- `RUSTUP_TOOLCHAIN=1.95.0 cargo test --all-features --bin mise config::` (288 passed)
- focused E2E coverage for registry and dynamically discovered disabled files, discovery failures, monorepo idiomatic settings, safe+paranoid mode, global precedence, tracked trust/prune, and explicit write consistency
- `RUSTUP_TOOLCHAIN=1.95.0 mise run lint-fix`
- `RUSTUP_TOOLCHAIN=1.95.0 mise run lint`
- `RUSTUP_TOOLCHAIN=1.94.1 cargo clippy --workspace --all-features --all-targets --ignore-rust-version -- -D warnings`

Rust 1.95 Clippy additionally reports two new baseline `collapsible_match` lints in unchanged upstream files (`src/backend/cargo.rs` and `src/task/task_source_checker.rs`); this PR does not suppress or modify them.

No documentation change is needed because this restores read-only tracked-config behavior without changing CLI or configuration syntax.

*AI-assisted — Tool: Shelley; model: openai/gpt-5.6-sol; version: 0.971.911043651.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved loading of tracked configuration files across nested project hierarchies.
  * Correctly applies configuration precedence, environment settings, safe mode, and `--no-config`.
  * Disabled or unrecognized configuration files are handled more clearly without masking parse warnings.
  * Improved error messages when configuration discovery fails.
  * Prevented untrusted settings from triggering plugin metadata discovery.
* **Tests**
  * Added end-to-end coverage for configuration precedence, disabled files, discovery failures, and paranoid mode.


jdx:
I think the `ConfigFileDetection` refactor is solid, but the settings-resolution portion should be refactored before merge.

`tracked_idiomatic_default_settings()` and `idiomatic_version_file_settings_for_root()` effectively create a second settings loader. They independently reproduce environment locking, config filename discovery, ignore and trust filtering, safe mode, and global/system/ancestor precedence. That appears correct in this patch, but it is easy for this path to drift from the canonical `Settings` loader as those rules evolve.

Could we expose a reusable settings-loading API that accepts a starting root and a source/trust policy, then use it both for normal settings loading and tracked-root idiomatic settings? `get_tracked_config_files()` should ideally only request the two idiomatic fields for a root, cache that result, and perform classification/parsing.

The explicit recognized/disabled/unknown/discovery-failed states are worth keeping; my concern is specifically the parallel precedence implementation.

xqm32:
> I think the `ConfigFileDetection` refactor is solid, but the settings-resolution portion should be refactored before merge.
> 
> `tracked_idiomatic_default_settings()` and `idiomatic_version_file_settings_for_root()` effectively create a second settings loader. They independently reproduce environment locking, config filename discovery, ignore and trust filtering, safe mode, and global/system/ancestor precedence. That appears correct in this patch, but it is easy for this path to drift from the canonical `Settings` loader as those rules evolve.
> 
> Could we expose a reusable settings-loading API that accepts a starting root and a source/trust policy, then use it both for normal settings loading and tracked-root idiomatic settings? `get_tracked_config_files()` should ideally only request the two idiomatic fields for a root, cache that result, and perform classification/parsing.
> 
> The explicit recognized/disabled/unknown/discovery-failed states are worth keeping; my concern is specifically the parallel precedence implementation.
> 
> _AI-assisted — Tool: Codex; model: unavailable/unavailable; version: unavailable._

Will work on this, thanks for advice!

xqm32:
Hi @jdx, would you mind reviewing this PR again? Thank you!

PR #12199, opened 2026-08-20
fix(backend): normalize shorthand archive stems

risu729:
## Summary

- fix the `.tbz` / `.tbz2` preferred-name bug still present on `main`
- replace the manually ordered archive-stem suffix list with `ExtractionFormat`-validated parsing
- reuse the same suffix logic for stem-only checksum companions, including shorthand and mixed-case suffixes
- preserve the `.txz` / `.tzst` behavior already fixed by #12156 without boosting unsupported archive types

### Already fixed by #12156

Preferred-name handling for `.txz` and `.tzst` is working on current `main`. Those complete suffixes appear before the shorter `.xz` and `.zst` entries, so they no longer leave a trailing `t`.

The Elide-style selection between `elide.linux-amd64.tgz` and `elide.linux-amd64.txz` also works on `main`: both names receive the preferred-name bonus, and the xz format score makes `.txz` win.

Therefore, the claim that current `main` still strips `.txz` as `.xz` is incorrect. That describes the historical failure before #12156, not the current code.

### Still broken on `main`

The equivalent bzip shorthand cases were missed:

- `.tbz2` is not in the stem suffix list, so the later `.bz2` entry matches and leaves a trailing `t`.
- `.tbz` is not in the list at all, so the suffix remains attached.

For example:

```text
elide.linux-amd64.tbz2
                     └─ main removes only .bz2

result: elide.linux-amd64.t
```

That result no longer matches the preferred repository name, so the asset loses the +20 preferred-name bonus and can lose to a different archive.

Checksum companion discovery has a separate incomplete suffix list. On `main`, assets such as `tool.txz`, `tool.tzst`, `tool.tbz2`, `tool.tbz`, and `tool.tar.zst` cannot find a stem-only checksum named `tool.sha256`. Mixed-case suffixes such as `tool.TXZ` have the same problem.

## What was incomplete in #12156

The Elide failure that motivated #12156 was a `preferred_name` problem, not a case of `.tgz` and `.txz` being treated as the same asset. `asset_name_stem` is only used when applying the preferred-name bonus, and the GitHub backend sets that preferred name to the repository name.

Historically, stripping `.txz` as the shorter `.xz` suffix left a trailing `t`. `.tgz` then received the +20 bonus while `.txz` did not. Before the format scores were separated, the shortest-name and alphabetical tiebreak also favored `.tgz`.

#12156 fixed that specific case by landing both the complete `.txz` / `.tzst` stem entries and the xz/zstd score ordering. Its weakness was the diagnosis and coverage: it added manual special cases without testing the actual Elide preferred-name path, did not generalize the fix to `.tbz` / `.tbz2`, and left checksum suffix handling out of sync.

This PR fixes those residual stem and checksum gaps. It deliberately excludes unsupported or semantically distinct `ExtractionFormat` variants such as `.rar`, `.tbr`, and `.vsix` from the preferred-name bonus.

The separate draft #12200 fixes the direct scoring regression introduced by #12156: `TarZst` tying the Windows ZIP score instead of remaining strictly below it.

## Tests

- `cargo test --bin mise backend::asset_matcher`
- `mise run lint-fix`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved archive matching by prioritizing TarZst and TarXz formats while retaining reliable ZIP matching.
  - Improved preferred-name matching across case variations and supported executable and archive suffixes.
  - Added support for shorthand archive extensions, including `.tbz2`, when matching names and checksums.
  - Prevented ZIP-based extension packages such as `.vsix` from being incorrectly treated as standard ZIP archives.
  - Improved checksum matching and excluded unsupported formats from name processing.

### DIFF
--- a/src/backend/asset_matcher.rs
+++ b/src/backend/asset_matcher.rs
@@ -593,6 +593,8 @@ impl AssetPicker {
 
         if format == ExtractionFormat::Zip {
             if self.target_os == "windows" {
+                // Native Windows ZIP extraction is faster than tar. Keep this
+                // strictly above every tar format score below.
                 return 15;
             } else {
                 return 5;
@@ -601,12 +603,11 @@ impl AssetPicker {
 
         if format.is_archive() {
             return match format {
-                // Since we are downloading, prefer small archives: zstd and xz
-                // tend to be smaller archives.
-                ExtractionFormat::TarZst => 15,
-                // xz comes in after zst because it can cost more CPU to decompress.
-                ExtractionFormat::TarXz => 13,
-                // All other archive formats roughly created equal.
+                // Prefer zstd over xz over other tarballs: zstd decompresses
+                // faster, which usually beats xz's slightly smaller download.
+                // Stay below the Windows ZIP score so zip still wins there.
+                ExtractionFormat::TarZst => 12,
+                ExtractionFormat::TarXz => 11,
                 _ => 10,
             };
         }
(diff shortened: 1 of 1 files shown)


sgammon:
oops, didn't catch this drift from main. thank you!

PR #12201, opened 2026-08-20
fix(sync): aggregate incomplete marker cleanup errors

risu729:
## Summary

- collect incomplete-marker cleanup errors while continuing to reconcile later external versions
- rebuild shims and runtime links for successfully reconciled versions
- return one non-zero error after reconciliation and rebuild finish

## Bug

External sync processes multiple versions, but marker cleanup currently returns immediately on the first failure. Versions later in the provider listing are never reconciled, and runtime links for versions that did succeed are not rebuilt.

For example, suppose nvm provides `node@18`, `node@19`, and `node@22`, while the marker paths for `node@18` and `node@19` cannot be removed because they are unexpectedly directories:

```text
$MISE_CACHE_DIR/node/18.0.0/incomplete/  # existing link; removal fails
$MISE_CACHE_DIR/node/19.0.0/incomplete/  # new link; removal fails
$MISE_CACHE_DIR/node/22.0.0/incomplete   # normal stale marker
```

Before this fix, `mise sync node --nvm` stops at `node@18`; `node@19` and `node@22` may remain unsynced or incomplete, and runtime aliases are not rebuilt. After this fix, sync creates the `node@19` and `node@22` links, clears the normal `node@22` marker, rebuilds runtime aliases, then exits non-zero with one error identifying both `node@18` and `node@19`.

## Stack

#11682 has merged as `b7be4a1da`. This PR is rebased directly onto current `main` at `aa22e7c45` and now contains only the marker-error aggregation change and its focused regression coverage.

## Testing

- scoped `hk run check --safe --format json`
- `cargo test --all-features --bin mise cli::sync::reconcile::tests`
- `mise run test:e2e e2e/sync/test_sync_nvm e2e/sync/test_sync_python_uv`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved runtime synchronization across nvm, nodenv, Homebrew, pyenv, and uv.
  - Preserved managed installations, runtime aliases, and links owned by other providers.
  - Removed only stale links owned by the relevant provider.
  - Applied consistent provider precedence when multiple sources provide the same runtime.
  - Improved handling of incomplete or invalid installations and clearly reported marker-cleanup failures.
  - Continued rebuilding shims and runtime links reliably after synchronization.


risu729:
Closing this because the failure mode requires an abnormal filesystem state, such as an `incomplete` marker becoming a directory or otherwise being unremovable. The existing fail-fast behavior is reasonable, and rerunning sync after correcting the filesystem completes reconciliation. Without a real-world report, the partial-success outcome type and Node/Python/Ruby caller complexity do not appear justified.

PR #12203, opened 2026-08-20
fix(task): preserve user-global task ownership

risu729:
## Stack

Built on #12229, which makes task includes override within each user/system global scope and prevents system task metadata from decorating user-global tasks. That PR has merged, and this branch is now rebased onto `main`, so the diff contains only this PR's own changes.

The corresponding system-only root behavior remains in #12204 and will be rebased after this PR merges.

## Summary

- keep the effective user-global include decision authoritative when the local config walk reaches HOME or a custom global root
- skip incidental default file sources before include expansion, parsing, template rendering, and inline-task merging
- retain genuine HOME-local inline definitions and file tasks with explicit local task context
- use effective cascade state when deciding whether HOME-local context intentionally owns file tasks
- preserve the existing global-wins same-source behavior for ordinary project and monorepo directories

## Bug examples

- A user-global config can replace the default task directories with `includes = ["~/dotfiles/tasks"]`. The later HOME-local walk previously rediscovered `~/.config/mise/tasks`, leaving an omitted default task runnable and allowing its script to replace a same-name HOME-local inline task.
- Even when an omitted default task was removed from the final task list, its `#MISE` template metadata was previously rendered first. An effectful template could therefore run—and a malformed one could fail task loading—for a task the user-global include decision had excluded.
- A HOME-local `.mise/conf.d` fragment that explicitly changes the task directory, global inputs, or Rust cache intentionally owns its local file tasks. Those tasks remain local, including when the user-global scope loaded the same physical source.
- With a custom global root, an ancestor can supply cascading task context such as `dir` while a root-local config sets `cascade = false`. The ownership check previously still used that ancestor context and resurrected default tasks omitted by the user-global includes. It now uses the effective context, so those omitted tasks remain unavailable.

## Testing

- `mise run lint-fix`
- `mise run test:e2e e2e/tasks/test_task_ls_global e2e/tasks/test_task_global_config_dedup e2e/tasks/test_task_monorepo_config_dedup`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task discovery across local, global, and system configurations.
  * Prevented duplicate tasks when configurations overlap.
  * Respected disabled cascading and global task precedence.
  * Preserved discovery for relative includes and inline task definitions.
  * Improved handling of task directories, working directories, shell settings, and global inputs.
* **Tests**
  * Expanded coverage for configuration inheritance, collisions, monorepos, and global task execution.

### DIFF
--- a/e2e/tasks/test_task_global_config_dedup
+++ b/e2e/tasks/test_task_global_config_dedup
@@ -107,3 +107,18 @@ EOF
 assert \
   "MISE_GLOBAL_CONFIG_FILE='$HOME/global-config/include-order.toml' MISE_SYSTEM_CONFIG_FILE='$HOME/global-config/missing.toml' mise run include-winner" \
   "$HOME/second-global-tasks/include-winner"
+
+local_include_marker="$MISE_TMP_DIR/local-include-render-count"
+: >"$local_include_marker"
+mkdir -p "$HOME/.config/mise/tasks"
+cat >"$HOME/mise.toml" <<EOF
+[task_config]
+includes = ["{{ exec(command='echo .config/mise/tasks; echo rendered >> $local_include_marker') }}"]
+EOF
+
+# Detecting explicit local task context must inspect the raw include option.
+# Rendering it again would repeat effectful template functions during one load.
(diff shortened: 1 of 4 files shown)


risu729:
@coderabbitai resume

PR #12228, opened 2026-08-20
fix(lockfile): stop monorepo upgrades from resolving to the old version

pikeas:
## Problem

With unified monorepo lockfiles enabled, `mise upgrade` could append the upgraded concrete version without removing the prior pin. A loose request such as `jq = "1.8"` would then continue resolving through the older first matching lockfile entry, even after that installation had been pruned.

For example, upgrading jq 1.8.0 to 1.8.2 produced both `[[tools.jq]]` entries. `mise ls jq` showed 1.8.0 as active but missing, and `mise which jq` failed instead of selecting installed 1.8.2.

## Root cause

Lockfile updates from ordinary monorepo commands are intentionally conservative because the active config hierarchy does not include every sibling project. `mise upgrade` used that same partial view after installing the replacement, so the preservation pass restored the superseded entry. Prefix resolution subsequently returned the first matching entry.

## Fix

- Prune only what the upgrade superseded. For each tool whose replacement installed successfully, the pre-upgrade resolved entry (`(lockfile, short, version, options)`, built by the same function that writes entries) is removed from the root lockfile unless the monorepo union still pins that exact version. Every other absent entry is preserved, as before. Earlier revisions of this PR pruned every absent version of every active tool that the union did not mention; that was both wider than the bug and wrong for the common layout below.
- Sibling requests satisfied by the new version pull forward. The union is resolved with the lockfile, so a sibling's loose `jq = "1.8"` (or `"1"`) resolves to the stale entry; those requests are evicted from the keep-set when the newly installed version satisfies them, using the same prefix-match predicate the lockfile reader uses. A sibling's exact `jq = "1.8.0"` is not satisfied by `1.8.2` and keeps its entry. Without this, the upgrade was silently undone whenever two projects shared a loose pin.
- Reuse the existing monorepo-union loader for that check, only when a superseded entry routes to a monorepo root lockfile and lockfiles are enabled. The upgrade-only variant also includes the monorepo root directory's own effective toolset: `monorepo_union` merges the root config underneath each declared config root, so a sibling's request for the same tool shadows the root's and the root's exact pin was otherwise absent from the keep-set and pruned. Other callers of `monorepo_union` are unchanged. If the union cannot be loaded or resolution fails, nothing is pruned. Sibling tools that are request-level `os`-gated or fail to resolve disable pruning for their short only.
- Attribute each newly installed version to its own request's config source when grouping tools for lockfile writes. Previously all new versions of one backend were attributed to the first version's source, so a single command producing new versions of the same tool from two configs could write a project-scoped version into the global lockfile. This path runs on every lockfile write, so the fix applies outside monorepos too.
- Keep the active project toolset authoritative for lockfile contents, shim/runtime rebuilding, platform snapshots, and auto-lock candidates.

Builder-filtered tools (unknown backends, Windows asdf, registry OS gates, settings filters) need no special handling: the active toolset and the union are built by the same `ToolRequestSetBuilder`, so such tools never reach the preservation loop (documented at the call site).

## Reproduction

1. Configure a unified monorepo lockfile with `[settings] lockfile = true`, `[monorepo] lockfile = true`, and a valid config root.
2. Set `jq = "1.8.0"` and run `mise install`.
3. Change the request to `jq = "1.8"`.
4. Run `mise upgrade jq`.

Before this change, the lockfile retained both 1.8.0 and 1.8.2 and selected stale 1.8.0. After this change, the superseded 1.8.0 entry is pruned while versions still present in the current monorepo union remain.

## Tests

- Unit: upgrade, lockfile, tool-request-set, install-context, and config-fixture tests (`112 passed` under `--bin mise`), covering superseded-key pruning, retention when an exact sibling pin remains, preservation of unrelated absent versions, satisfied-by eviction across differently spelled specs, union-failure and unresolved/OS-gated fallbacks.
- e2e `test_lockfile_monorepo`: distinct sibling versions (original case); identical loose pins in root and sibling (the upgrade previously left both entries and both projects resolved the old one), including a subsequent sibling `mise install` to confirm the result survives the next lockfile write; differently spelled overlapping loose pins; an exact sibling pin retained alongside the new version; a root exact pin retained when the sibling's loose request is upgraded. All 48 non-slow `e2e/lockfile` files pass against the debug binary in isolated state.
- `cargo fmt --all -- --check` and the CI clippy invocation (`--all-features --all-targets -D warnings`) pass.

## Limitations

- A flat unified lockfile records which concrete versions exist, not which request resolved to which. When a loose request matches more than one retained entry, `get_locked_version` returns the first in file order. So whenever any config keeps an exact pin of an older version, every loose request in the same prefix family — in any config, in either direction — keeps resolving that older entry, and an upgrade of the loose request reports success without effect. This is identical to `main`. No pruning can fix it without dropping an entry another project needs; it needs the lockfile to record request bindings (see Future work).
- Only the directly superseded entry is pruned. Duplicates accumulated by earlier mise versions, hand edits, or merge conflicts converge over successive upgrades; `mise lock` remains the authoritative one-shot cleanup.
- Other partial monorepo mutation paths (`mise use`, `mise install`, auto-lock) retain conservative sibling preservation and can still leave superseded entries. This change is intentionally scoped to `mise upgrade`.
- Configs nested below a declared `[monorepo].config_roots` glob are invisible to the union but route to the root lockfile; sibling configs are template-rendered with the invoking directory's environment. With superseded-only pruning, either can only cause loss of an entry whose version is exactly the one being replaced, not unrelated history.

## Future work

- **Request-keyed lockfile bindings.** Recording `(tool, request spec, options) -> concrete version` alongside the existing entries would make lookup an exact match on the spec, with no ordering of opaque version strings and no ownership tracking, and would resolve the first limitation above; pruning would then become hygiene rather than a correctness mechanism. Proposed in #12298.
- **Sibling configs render templates with the active root's environment** (`monorepo_union`; affects `mise install --monorepo` and `mise ls` today). Needs root-scoped env/vars evaluation.
- **Configs nested below a config root are invisible to the union.** Either recursive discovery or per-entry attribution.

*AI-assisted — Tools: Codex CLI 0.146.0 (OpenAI/gpt-5.6-sol); Claude Code (anthropic/claude-fable-5, anthropic/claude-opus-5).*


pikeas:
@coderabbitai resume

jdx:
## Verification

Checked out the branch and ran everything:

- `cargo build --all-features` — clean
- `cargo clippy --all-features --all-targets` — zero warnings
- Unit tests: `lockfile::` / `cli::upgrade::` / `tool_request_set::` — 99 + 19 passed
- `mise run test:e2e e2e/lockfile/test_lockfile_monorepo` — passed, and the new section demonstrably fixes the bug (installs 1.1.0, root reports `1.1.0`, `packages/a` reports `2.0.0`)

I also traced every non-upgrade caller. All 13 other `rebuild_shims_and_runtime_symlinks` call sites route to `LockfileUpdateScope::Active`, which produces `AbsentEntryPolicy::PreserveAll` for a monorepo root lockfile and `Drop` otherwise — identical to the old `if is_monorepo_root_lockfile { preserve_absent_tool_entries(...) }`. The refactor is behavior-preserving outside `upgrade`. `Config::reset()` at `src/cli/upgrade.rs:468` lands before the union load, so `--bump` rewrites are visible to it, and `ResolveOptions::default()` has `use_locked_version: true`, so siblings normally resolve back to their own lock entries. **No correctness bug found in the fix itself.**

## Findings

### 1. OS-gated or unresolvable requests disable pruning monorepo-wide instead of per-short — `src/cli/upgrade.rs:912`

`monorepo_toolset_is_complete` returns false if *any* tool anywhere in the monorepo has an OS-gated request or failed to resolve, dropping the entire upgrade back to `PreserveAll` for every short. But this PR already built the right instrument for that — per-short `PreserveAll` via `omitted_shorts` — and only wires it to `filtered_tools`/`unknown_tools`.

The practical effect: one `os = ["macos"]` request in one sibling, or one sibling tool whose plugin isn't installed yet (`ToolVersionList::resolve` fails → `warn_once` → short version list), and the fix silently never engages for that repo. That's a likely configuration in exactly the large monorepos this targets. Marking those shorts `PreserveAll` rather than bailing globally is strictly safer-or-equal and far more likely to actually fire.

### 2. The fallback warning misreports the OS-gated case — `src/cli/upgrade.rs:602`

The `Ok(())` arm warns "could not completely resolve the monorepo toolset for lockfile update". For a monorepo that legitimately platform-gates a tool, nothing failed — yet every `mise upgrade` on the non-matching OS prints this. Given finding 1, this is the common path, not the edge case.

### 3. Reachability of the `filtered_tools` mechanism is narrower than the PR implies

`Toolset::is_disabled` (`src/toolset/mod.rs:661`) applies the same `!ba.is_os_supported()` and `tool_enabled(..)` filters as `ToolRequestSetBuilder::is_disabled`, so a short excluded from the union by *either* of those is equally excluded from the active toolset — it never enters the `for (short, entries) in tools_by_short` loop, and its entries are untouched regardless of policy. The only real divergence between the two filters is `BackendType::Unknown` (which can't resolve, so also never reaches the loop) and `cfg!(windows) && BackendType::Asdf`. So as far as I can tell the new per-short preservation can only change output on Windows + asdf, and two of the three new `filter_disabled_tools` tests document behavior that cannot affect a lockfile. Worth saying so in a comment rather than leaving readers to infer broader coverage.

### 4. The union is loaded and resolved even when the lockfile write is a no-op — `src/cli/upgrade.rs:585`

The gate is `config.monorepo_lockfile_root().is_some()`, which only checks `[monorepo].lockfile` — not `Settings::lockfile_enabled()`. `update_lockfiles` bails immediately on `!lockfile_enabled()`, so `[monorepo] lockfile = true` + `MISE_LOCKFILE=0` pays a full sibling-config load plus a resolve of every tool in the monorepo for nothing. One-line guard.

Related, and worth stating in the PR body: a resolve of the *whole* union is now on the `mise upgrade` critical path for every monorepo-lockfile repo. The body says "adds no per-tool network work" for omitted tools, which is true, but a sibling tool that is neither locked nor installed will still trigger remote version listing.

### 5. `UpgradeLockfileUpdate` + `upgrade_lockfile_update` are an identity mapping over `Option` — `src/cli/upgrade.rs:875-895`

The enum is isomorphic to the `Option<(&Toolset, &HashSet<String>)>` the function takes, the function just re-wraps it, and the call site immediately matches back. `test_upgrade_uses_complete_monorepo_union_for_lockfile_update` asserts that `Some` maps to the `Some` variant. That's ~30 lines of code and ~40 of test for zero behavior; matching the `Option` directly at the call site is equivalent.

### 6. `empty_versions` is a parameter that exists only to satisfy a borrow — `src/lockfile.rs:1722`, `src/lockfile.rs:3066`

`BTreeSet::new()` is `const`, so a file-level `static EMPTY_VERSIONS: BTreeSet<String> = BTreeSet::new();` removes the sixth parameter of `absent_entry_policy_for_update`, the local, and the threading through both test call sites.

### 7. Dead arm — `src/lockfile.rs:3131`

`merge_tool_entries_for_update` already skips `preserve_absent_tool_entries` for `Drop`, so the `AbsentEntryPolicy::Drop => false` arm inside is unreachable from that path. Keep the inner check, drop the outer `if !matches!(...)`.

### 8. `merge` still drops both metadata vectors — `src/toolset/tool_request_set.rs:337`

The PR body documents this as future work with a correct safety argument (`filter_disabled_tools` runs after every merge in `build()`). But unioning them is two lines here, versus leaving a landmine for whoever populates these vectors earlier. `FromIterator` and `filter_by_tool` drop them too.

### 9. Registry-data-dependent test — `test_filter_disabled_tools_records_registry_os_gate`

Hardcodes `tart`/`systemctl-tui`. Registry entries churn constantly in this repo, so an unrelated `os` gate change breaks the test. The leading `assert!(!ba.is_os_supported())` at least fails legibly, and since `filter_disabled_tools` reads OS support off the registry a synthetic short can't reproduce the gate — so this is low priority, just worth a comment noting the coupling.

## Assessment

The core fix is sound, well-scoped, honestly documented (the four limitations in the body are all real and correctly characterized, including the sibling-checksum-loss one), and behavior-preserving for every non-upgrade path. Findings 5–7 are cleanup. Finding 1 is the one I'd want addressed before merge — the machinery for it already exists in the PR, and without it the fix is likely inert in the multi-platform monorepos it's aimed at, with finding 2's warning as the only visible symptom.

pikeas:
Thanks for the thorough review — all nine findings addressed: a01bcb4dd (1–7, 9), 644878c70 (8), bc2cd62c1 (follow-ups).

  1. OS-gated shorts are detected from the request set *before* resolution — request-level `os = [...]` resolves fine on a
  non-matching OS, so no post-resolution check can catch it. Resolution failures are attributed afterward via
  `versions.len() != requests.len()`, an exact predicate (resolve is fail-fast, one push per request). Only unattributable
  failures — union load errors, propagated resolver errors — still fall back to global preservation. Regression tests
  cover both new per-short cases.
  2. Gone; the once-only warning names the sorted shorts as unavailable, disabled, platform-gated, or unresolved.
  3. Confirmed and documented, with one amendment: after finding 1, reachability is wider than Windows+asdf —
  `Toolset::is_disabled` checks only the registry gate, so a request-gated or unresolved sibling reaches the active update
  while its short is preserved. The two builder-filter tests now note they can't change a lockfile.
  4. Done — `Settings::lockfile_enabled()` gates the union load.
  5. Done — enum, helper, and the isomorphism test deleted; the decision tests were rewritten against the per-short
  functions.
  6. Done — `static EMPTY_VERSIONS`.
  7. Done, with a twist: `preserve_absent_tool_entries` returns early on `Drop`, so ordinary lockfile writes skip the
  key-set build and re-sort; the inner match stays as the policy definition.
  8. Done — `merge`, `filter_by_tool`, and the install-dependency path retain both vectors; `FromIterator` is documented
  as inherently lossy (its item type carries no metadata). Inert today, so no behavior change.
  9. Documented.

  PR body updated to match, including two disclosures this surfaced: the union resolve is on the upgrade critical path for
  lockfile-enabled monorepos (an unlocked, uninstalled sibling can trigger remote version listing), and pruning
  correctness depends on `ResolveOptions::default()` keeping `use_locked_version: true`.

  *AI-assisted — Tools: Codex CLI 0.146.0 (OpenAI/gpt-5.6-sol); Claude Code (anthropic/claude-fable-5,
  anthropic/claude-opus-5).*

jdx (commented):
## Re-review

All nine findings addressed, and I verified each fix is actually in the tree rather than just described:

- Finding 1 — `monorepo_toolset_is_complete` is gone; `monorepo_request_set_omitted_shorts` + `monorepo_toolset_unresolved_shorts` now attribute per short. This is the important one and it's done right.
- Finding 4 — `Settings::get().lockfile_enabled()` guard is on the gate.
- Finding 5 — enum gone, the call site matches the `Option` directly.
- Finding 6 — better than what I suggested: `PreserveVersions(Option<&HashSet<String>>)` removes the static entirely.
- Finding 7 — outer `Drop` early-return kept for the allocation-free path with the inner arm documented as authoritative. Fine, and the comment explains why.
- Finding 8 — `merge` unions both vectors; `filter_by_tool` and `InstallDependencyDeclarations::matching_requests` re-set them after the `FromIterator` collect.
- Findings 2, 3, 9 — comments and warning text corrected.

I also checked the predicate from finding 1 rather than taking my own earlier claim on faith. `ToolVersionList::resolve` clears, pushes exactly one version per request, and returns `Err` on the first failure, while `Toolset::resolve` warns and retains the partial list — so `versions.len() != requests.len()` is exact and cannot produce a false "complete". And `outdated.is_empty()` returns early at `src/cli/upgrade.rs:215`, so the union resolve is only on the path of an upgrade that actually does something.

## One change I'd want before merge

`AbsentEntryPolicy::PreserveVersions(None)` prunes — `src/lockfile.rs:3155`:

```rust
AbsentEntryPolicy::PreserveVersions(versions) => {
    versions.is_some_and(|versions| versions.contains(&existing_tool.version))
}
```

`None` means the union has no version for this short at this lockfile path, i.e. precisely the *unattributable* case. Everywhere else this PR is careful to fall back to `PreserveAll` when attribution fails — union load error, propagated resolver error, unresolved short, OS-gated short. This is the one arm that lands on the aggressive side of that principle, and it's the mechanism behind limitation #3 rather than an abstract concern.

It's reachable. `lockfile_path_for_config` routes *any* config under the monorepo root to the root lockfile (`config_path.starts_with(root)`, `src/lockfile.rs:1412`), while the union only loads directories matched exactly by `config_roots`. So with `config_roots = ["packages/*"]`, a `packages/a/sub/mise.toml` pinning `dummy = "3.0.0"` writes into the root `mise.lock`, is invisible to the union, and its entry plus checksums are silently deleted by the next `mise upgrade dummy` from anywhere in the repo. Silent checksum loss is the failure mode I care most about in this file: the entry comes back on the next `mise install` from that directory, but re-derived from whatever is downloaded then rather than the value that was pinned.

Mapping `None` to `PreserveAll` closes it and does not weaken the fix. `base_config_files` is merged into every root's config set (`src/config/mod.rs:812`), so the monorepo root config's own tools are always in the union — the e2e scenario still gets `PreserveVersions(Some({"1.1.0", "2.0.0"}))` and still prunes `1.0.0`. `None` degrades to current behavior only for entries nothing can attribute, which is the right default for a lockfile. It also lets limitation #3 drop from "may silently prune" to "may retain a stale entry", which is the correct direction to fail in.

Worth a regression test pinning that a nested undeclared config's entry survives an upgrade of the same short elsewhere.

## Non-blocking

- `dbfa8595a` post-dates my review and is the only commit that changes behavior for non-monorepo users. I read it: attributing each new version to its own `request.source()` instead of `tvs[0]`'s is strictly more correct, and the old `chunk_by` grouping was order-dependent so nothing could have depended on it. The `retain` dedup narrowing from cross-source to per-source is also right — identical concrete versions still collapse in `merge_tool_entries` via the `(version, platform)` key, and genuinely distinct ones should both be written. No objection, just flagging that it's the widest-reaching and least-reviewed part of the PR.
- The new `warn_once` will fire on every upgrade for any monorepo with a single `os = [...]`-gated tool. The message is accurate now, but this is the common path in multi-platform monorepos, not an edge case. Fine to ship and watch.

Otherwise this is good — well-scoped, behavior-preserving for all 13 non-upgrade `rebuild_shims_and_runtime_symlinks` callers, opt-in until the `2027.6.0` default flip, and the four limitations in the body are all real and correctly characterized.

jdx (commented):
## Upgrading the warning note from "non-blocking" to "fix this too"

I dug further into the `warn_once!` at `src/cli/upgrade.rs:596` and it's a worse trade than I described. The problem isn't volume — `warn_once` dedups on the formatted string, so it's once per process, and `--dry-run` returns at `src/cli/upgrade.rs:358` before reaching it. The problem is that **the warning's trigger set is a strict superset of the mechanism's effective set, and the difference is dominated by cases that structurally cannot touch a lockfile.**

The PR already says so in its own code comment at `src/lockfile.rs:3083`:

> Registry OS gates and settings filters exclude the same short from the active update too, and unknown backends produce no active resolved entry; Windows asdf is the filtered-tool exception.

That is exactly right, and I confirmed the two predicates:

```rust
// Toolset::is_disabled
!ba.is_os_supported() || !tool_enabled(enable_tools.as_ref(), &disable_tools, &ba.short.to_string())

// ToolRequestSetBuilder::is_disabled
backend_type == BackendType::Unknown
    || (cfg!(windows) && backend_type == BackendType::Asdf)
    || !ba.is_os_supported()
    || !tool_enabled(self.enable_tools.as_ref(), &self.disable_tools, ba)
```

The `is_os_supported` and `tool_enabled` terms are shared. So a short excluded from the union by either one is equally excluded from the active toolset, never enters the `for (short, entries) in tools_by_short` loop, and has its entries untouched regardless of policy. Sorting the seven contributors to `omitted_shorts`:

| Contributor | Can it change a lockfile write? |
| --- | --- |
| `filtered_tools` via registry OS gate | No — same predicate in `Toolset::is_disabled` |
| `filtered_tools` via `enable_tools`/`disable_tools` | No — same predicate |
| `filtered_tools` via `BackendType::Unknown` | No — cannot resolve, never in the loop |
| `unknown_tools` | No — same |
| `filtered_tools` via `cfg!(windows) && Asdf` | Yes, Windows only |
| request-level `os = [...]` | Yes |
| unresolved short | Yes |

On non-Windows only the last two can matter — and the five that can't are the *common* ones in a monorepo. One registry-OS-gated tool in one sibling, or one `disable_tools` entry, or one sibling tool whose backend isn't resolvable locally, and every `mise upgrade` in the repo prints a warning naming tools whose entries were never at risk.

It also can't help but overclaim, because it's emitted before `update_lockfiles` is called and therefore has no idea whether any named short even has an existing entry. A sibling's macOS-only tool that has never been installed on this machine and has no lockfile entry still produces `preserving existing entries for: tart`. Nothing was preserved; there was nothing to preserve.

A routine, unactionable warning about correct behavior is worse than silence — it's the kind that teaches people to skim past mise's warnings, which costs us the ones that matter.

### Suggested fix

Keep `omitted_shorts` exactly as it is — it's harmless and defensively correct for the keep-set. Just narrow what gets *reported* to the shorts that can actually diverge from the active update: `BackendType::Unknown`, `cfg!(windows) && Asdf`, request-level `os = [...]`, and unresolved. Concretely, have `monorepo_request_set_omitted_shorts` return the full set plus a reportable subset (or a small struct with `all` / `reportable`), warn only on the latter, and `debug!` the remainder.

If you'd rather not split the set, demoting the whole thing to `debug!` is strictly better than the status quo: the message is diagnostic, not actionable, and the conservative fallbacks that *are* worth surfacing already have their own `warn!` arms at `src/cli/upgrade.rs:603` and `:611`.

Combined with the `PreserveVersions(None)` change from my previous review, that's two small diffs, both local. I'd hold for both rather than ship and follow up.

pikeas:
Both changes done, then re-scoped after finding the branch did not fix the common layout. Head fd9497cb3; body updated.

**Bug found.** Root and sibling both `dummy = "1"`, locked 1.0.0. `mise upgrade dummy` left both entries; both projects still resolved 1.0.0. The union resolved the sibling's `"1"` through the stale lock, so the keep-set protected it. Now an e2e case.

**Changes.**
- Prune only the entry the upgrade replaced, and only if the union does not still pin that exact version (148b6027f).
- Drop union requests that the new version satisfies, using the reader's own match predicate (e3ecf655d). Exact pins keep their entry.
- `PreserveVersions(None)` is gone; policy is `Drop` / `PreserveAll` / `PruneSuperseded`.
- `filtered_tools` and the warning are gone. The lockfile write uses `config.get_toolset()`, built by the same `ToolRequestSetBuilder` as the union, so builder-filtered shorts — Windows asdf included — never reach the preservation loop.
- The upgrade keep-set now includes the monorepo root's own toolset (fd9497cb3). `monorepo_union` merges the root config under each declared root, so a sibling's request shadows the root's; the root's exact pin was being pruned.
- e2e: shared loose pin (+ `mise install` from the sibling afterwards), overlapping pins `"1.0"`/`"1"`, exact sibling pin kept, exact root pin kept, original case.

**Not fixed.** An exact pin of an older version blocks every loose request in the same prefix, in either direction; same as `main`. Listed as the first limitation. #12298 proposes recording the resolved version per spec.

*AI-assisted — Codex CLI 0.146.0 (OpenAI/gpt-5.6-sol); Claude Code (anthropic/claude-fable-5, anthropic/claude-opus-5).*

pikeas:
@jdx Should this hold until #12299 lands?

pikeas:
Closing this: #12299 landed lockfile v1 request bindings, which solves the underlying stale-selection problem this PR was working around, so the approach here is obsolete.

While validating v1 against the scenarios from this thread I found three remaining bugs — two fixed in much smaller PRs, one to be filed as an issue:

- #12382 — monorepo root lockfile maintenance: `mise lock` (both modes) drops the root's own entry when a sibling requests the same tool.
- #12381 — `update_lockfiles` attributes all new versions of a backend to the first request's source, writing entries into the wrong lockfile.
- Issue to follow: superseded rows accumulate in monorepo root lockfiles after `mise upgrade` (inert under v1 bindings, but never cleaned; pruning them safely needs a complete liveness model).

Thanks for the reviews here — they shaped what those PRs check for.

jdx:
apologies, I forgot about this PR

PR #12229, opened 2026-08-20
fix(task): enforce global task scope precedence

risu729:
This intentionally changes two legacy additive behaviors in global task loading: independently loaded `task_config.includes` per fragment, and system task metadata decorating a user-global task. If either compatibility break is not acceptable without deprecation, I will revise the approach.

## Summary

- compose user-global task configuration files as one precedence scope
- compose system task configuration files as a separate precedence scope
- use the highest-precedence `task_config.includes` within each scope instead of aggregating every include set
- make user-global tasks cleanly replace same-named system tasks without inheriting system metadata
- keep differently named system tasks available

### Global config fragments

Global config files currently load their task include sets independently. A lower-precedence `conf.d` fragment can therefore keep contributing tasks even when a higher-precedence `config.toml` defines a replacement `task_config.includes` list.

For example, if `config.toml` includes `~/dotfiles/tasks` and `conf.d/machine.toml` includes `~/machine/tasks`, the old loader discovers tasks from both directories. This PR uses only the higher-precedence include list. Users who want both directories should list both in the higher-precedence config.

Because the files now compose like project-local fragments, a lower-precedence same-name `[tasks.*]` block can still decorate a file task selected by the effective include list. Its metadata such as `env` applies to that selected script; an alternate `file` value in the inline block does not replace the selected file-task base or get evaluated merely for source comparison.

### User and system scopes

The old loader can also apply lower-precedence system inline metadata to a user-global task when both scopes discover the same physical script. For example, a system `[tasks.build] env = { TARGET = "system" }` block can modify the `build` script selected by the user-global config.

This PR treats the scope boundary as replacement instead: the user-global `build` task wins unchanged. System tasks with other names remain available.

This PR is limited to global task-scope composition. It does not change HOME-local ancestor traversal or duplicate-source ownership.

## Maintainer question

@jdx, is it acceptable to make these behavior changes immediately without a deprecation period?

I am sorry that I previously retained the independently loaded global behavior without reconsidering the overall configuration semantics carefully, and then documented it in #8905. Your earlier suggestion in #8740 to make global tasks use the same override semantics as local tasks was the cleaner direction. I also found that system-to-user task decoration was an incidental consequence of the generic aggregation introduced in #11103 rather than a separately designed policy.

If an immediate change is not acceptable, I will revise the approach based on your preference.

## Testing

- `mise run lint-fix`
- `mise run test:e2e e2e/tasks/test_task_global_config_dedup e2e/tasks/test_task_ls_global`
- `mise run test:e2e e2e/tasks/test_task_source_comparison_no_exec`

## Summary by CodeRabbit

- **New Features**
  - Global task configurations maintain separate user and system scopes.
  - User-defined tasks with matching names replace system tasks without inheriting their metadata.
  - Differently named system tasks remain available.
  - Relative task includes resolve within their configured scope.
  - Same-name task definitions can compose across global fragments, with higher-priority files determining execution and lower-priority definitions contributing supported metadata.

- **Bug Fixes**
  - Prevented unintended metadata inheritance between task scopes.
  - Preserved predictable ordering when multiple includes appear in the same configuration.

### DIFF
--- a/docs/tasks/task-configuration.md
+++ b/docs/tasks/task-configuration.md
@@ -1503,7 +1503,10 @@ For local and monorepo task discovery, mise uses the nearest config file that de
 `task_config.includes`. When the parent has `task_config.cascade = true`, its includes are inherited
 until a child defines its own. A child config's `includes` replaces both the defaults and any
 inherited `includes` for that directory.
-Global config files are loaded independently, so each global config file uses its own `task_config.includes` or the default directories if `includes` is unset.
+User-global config files form one config scope, as do system config files. Within each scope, the
+highest-precedence config that defines `task_config.includes` replaces lower-precedence includes and
+the default directories. User-global and system scopes remain independent, so user includes do not
+remove tasks provided by the system config.
 
 Entries are evaluated in order, and when more than one include defines a task with the same name the **last** entry in the list wins.
 This applies uniformly to directory, toml-file, and `git::` includes, so to override a task coming from a `git::` include with a local one, list the local directory after the `git::` entry:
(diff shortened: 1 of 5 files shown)


risu729:
@coderabbitai resume

PR #12236, opened 2026-08-21
perf: load install state per tool instead of scanning every install

jdx:
Follow-up to the nested-invocation work (#12143/#12145): those removed the pathological startup cost; this attacks the largest remaining item in the fixed floor.

## Problem

Every invocation scans the **entire installs dir** at startup, before argument parsing: a `.mise.backend.toml` read + a readdir per installed tool, plus an lstat (runtime-symlink check), a stat (`incomplete` marker), and a `Versioning` sort per version. With ~60 tools installed that measured **5.8ms of a 14ms `mise x` floor** (`MISE_TIMINGS`), and every nested `mise run`/`mise x` in a script tree pays it again — for tools the invocation never touches.

## Change

**Per-tool, on demand.** Resolution asks install state for the tools it actually uses: `get_tool(short)` probes the tool's own dir (mise creates install dirs as the kebab-cased short, and one memoized read of the consolidated root manifest covers dirs recorded under a different name), merges shared-dir versions and plugin identity, and memoizes — including known-absent results. The full scan still exists, memoized, for callers that genuinely enumerate — `ls`, `doctor`, `prune`, shims, runtime symlinks, `ls-remote` — reached via `list_tools()` / `backend::list()`.

**Nothing new is persisted.** We tried a consolidated state file once (`.mise-installs.toml` as the identity source) and moved away from it in #9941 because a file detached from the install dirs desyncs when installs are copied or merged. This is not that: the directories remain the only source of truth, the same reads happen — just only for the tools in play — so there is no invalidation logic to get wrong. Legacy-format migration still runs on full scans; the per-tool path is strictly read-only.

Two supporting changes:
- `backend::load_tools` no longer builds a backend for every installed tool at startup; `backend::list()` fills them in (memoized) when an enumerator asks. `backend::get()` already fell back to `arg_to_backend` per name.
- Config loading enumerated every installed tool on every invocation to collect version aliases, though only core tools and asdf/vfox plugins override `get_aliases` — everything else returns the default empty map. `backend::alias_backends()` now asks only those.

## Measurements

Against main, both release builds, same machine, ~60 tools installed, project with 2 active tools + templated `[env]`:

| command | main | this PR |
|---|---|---|
| `mise x -- true` | 17.5ms | **11.7ms** (−33%) |
| `mise env` | 16.9ms | **11.1ms** (−34%) |
| 3-level nested `mise run` | 70.9ms | 65.0ms |
| `mise ls` | 22.6ms | 24.0ms (within noise; scan moved from a parallel init task to on-demand) |

`mise ls` output is **byte-identical** to main on both the test project and the 60-tool global dir. The nested-`run` chain improves less than 3× the `x` saving because the old scan ran on a parallel tokio task, so part of its span overlapped other startup work; the remaining `run` overhead is task machinery, which is future work (a nested-invocation fast path), not this PR.

## Tests

- New e2e `test_install_state_lazy`: `mise x`/`mise env` must not emit `full_scan_tools` in `MISE_TIMINGS` (verified to fail when `get_tool` is reverted to full-scanning); a tool installed by another process resolves through the lazy path; `mise ls` still enumerates everything; uninstall is reflected.
- `install_state` writes from the same process (`add_tool_version`, `add_plugin`) update/invalidate both memo layers, so same-run dependency resolution sees fresh installs — the existing behavior #9941 fixed.
- Full unit suites for the touched areas pass (toolset 91, backend 594, config 276, shims 11), plus the e2e sweep over `ls`/`uninstall`/`prune`/`use`/runtime-symlinks.

## Known edge

A dir whose only identity is a sidecar under a name that doesn't kebab-match its short (and isn't in the root manifest) is invisible to the per-tool path — mise itself never produces that layout (dirs are created as kebab(short) and sidecars record the same short), and full-scan callers still see such dirs. Called out in the `get_tool` docs.

---

> [!NOTE]
> **Medium Risk**
> Touches core install-state and backend loading, including memo races and shim listing. A scan failure now errors instead of looking empty, which is safer for shim rebuilds but changes failure behavior.
> 
> **Overview**
> Stops scanning every installed tool at startup. `mise x` / `mise env` now load install identity and versions **per short** via `get_tool`, so a large installs dir is no longer on the hot path.
> 
> The full scan still runs once for enumerators (`ls`, prune, shims, `backend::list`). `load_tools` seeds only core/registry backends; installed tools are merged later without overwriting command-scoped `get` entries. Alias loading uses `alias_backends()` so config no longer constructs a backend for every install.
> 
> Also memoizes lockfile path discovery (it ran on every backend-identity lookup) and treats an unreadable installs dir as an error rather than an empty list, so shim rebuild cannot delete everything on a failed scan. Same-run `add_tool_version` / `add_plugin` update both memo layers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3042f731173ebf2c3ecc79cc8cf7ff0084c5eb3e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Performance**
  * Tool and installation lookups now load information only when needed, improving command responsiveness.
  * Full installation scans are deferred until commands requiring complete lists are used.
  * Repeated lockfile and manifest lookups are faster through caching.

* **Bug Fixes**
  * Commands detect tools installed by another process without requiring a prior full scan.
  * Installation listings update correctly after tools are uninstalled.
  * Alternate tool directory names are resolved correctly.
  * Unreadable installation directories report an error, while missing directories are handled as empty.


jdx:
All four addressed in `2a194054d`; the fifth is a test suggestion I handled differently.

**@cursor / @coderabbitai — installed backends lose to core: valid, fixed.** `load_tools` collected core, then idiomatic, then installed into one map, so a later installed entry overwrote a core/registry one with the same short and kept its recorded identity. Adding installed tools afterwards with `or_insert` silently reversed that. Installed entries now displace *seeded* shorts (tracked from `load_tools`) while leaving anything `backend::get` resolved later alone — those can carry command-scoped options, which is the case @coderabbitai flagged about not blindly replacing configured entries.

Worth recording what I could *not* reproduce: `e2e/plugins/test_core_overloading` shadows core `node` with an asdf plugin and passes **either way**, because tool resolution goes through install_state-derived `BackendArg`s rather than the `TOOLS` map. So this is a semantics-preservation fix for `backend::list()` enumeration; I found no user-visible reproducer. I pinned the intended precedence with a unit test on the extracted `installed_replaces` predicate rather than asserting on a copy of the loop.

**@cursor / @coderabbitai — memo can clobber a concurrent install: valid, fixed.** `load_tool` runs without the memo lock, so an `add_tool_version` completing in that window could be overwritten by the older read, and the fresh version would read as *not installed* for the rest of the process (no full scan runs on the lazy path to correct it). The insert is now vacant-only, so the writer that saw the install wins.

**@coderabbitai — `add_plugin` misses the full-scan cache: valid, fixed.** `get_tool` serves a completed full scan ahead of the memo, so invalidating only the memo left a newly installed plugin invisible when a scan had already run in that process. `add_plugin` now applies the plugin identity to `INSTALL_STATE_TOOLS` too.

**@coderabbitai — a failed scan must not read as empty: valid, and the most important one.** Your conclusion is right; the mechanism is one step over from the example. `prune`&#39;s version deletion is safe-fail (`to_delete` *starts* from the installed list, so empty deletes nothing), but the shim rebuild in the same command is not: `get_desired_shims` derives from installed versions and `extra = actual − desired`, so an empty scan marks **every existing shim** stale and removes it. `try_list_tools` now propagates the error and `Toolset::list_installed_versions` fails loudly instead of returning an empty list. A *missing* installs dir is still not an error — `file::dir_subdirs` returns empty for it, which is what a fresh install looks like — so only real IO failures abort.

**@coderabbitai — same-process backend cache coverage:** I did not add this one. `backend::list()` caching installed tools for the process lifetime is unchanged from before this PR (`TOOLS` was built once at startup, so a tool installed later did not appear there either), and install flows that need fresh state call `backend::reset()`. Adding a test asserting the current caching would pin behavior this PR did not introduce; happy to add it if you want that pinned regardless.

Re-verified after the fixes: perf holds (`mise x` 17.9ms → 12.1ms vs main, both release), `mise ls` output still byte-identical on a 60-tool dir, unit suites green (toolset 91, backend 595, config 276, shims 11), and the e2e sweep over `test_install_state_lazy`, `test_core_overloading`, `ls`, `prune`, `uninstall` passes.

jdx:
Both new findings fixed in `f15cd06e7`, plus the perf gate investigated properly — details below, and the gate is not closed yet.

**@coderabbitai — per-tool scan failures swallowed: valid, fixed.** You were right that the root-scan guard only covered one level. `scan_versions` turned an unreadable *individual* install dir into an empty version list, `scan_tool_dir` then dropped the tool, and the full scan still returned `Ok` — so the incomplete state passed the guard and reached shim cleanup. Primary install dirs now propagate the error; read-only shared dirs stay best-effort via an explicit `scan_versions_best_effort`.

**@cursor — memo race drops installed versions: valid, and my previous fix caused it.** Switching to `or_insert` traded last-writer-wins for first-writer-wins, and you are right that this loses the complete scan when `add_tool_version` got there first with only the new version. Neither side is authoritative — one scanned every version on disk, the other knows an install that may have finished after that scan — so picking either is wrong. They now merge: the recorded identity stands and every version either side saw is kept.

## The `ls` perf gate

I could not close it, and I want to be straight about what I found rather than hand-wave. I reproduced the gate locally with `tak` and then with callgrind.

**My earlier "within noise" claim was wrong.** I judged `mise ls` by wall clock; the gate counts instructions, and the regression is real.

What I ruled out by measurement, not reasoning: backend count (154 both), `backend::list()` calls (2 new vs 3 main), `load_tool` calls (8), installed-version-listing calls (1 both), lockfile read calls (32 both), scan ordering (moving the scan before config load changed nothing), the seeded-shorts copy, and whole-struct clones in the field accessors. Each of those cost under 1M instructions or was identical.

What I did fix: the installs manifest was parsed twice (once per-tool, once by the full scan) — now shared; alias collection constructed ~15 backends per config load — now reuses what `load_tools` already built; a missing install dir did a sidecar read plus two legacy probes before concluding "absent" — now one stat; manifest lookups by short walked every entry — now indexed. Together: 142.3M → 140.4M.

**The reframing that matters:** this change trades syscalls for CPU. Measured locally against main (both release, 60 installed tools):

| | main | branch |
|---|---|---|
| `ls` instructions | 110.0M | 140.4M |
| `env` instructions | 96.0M | 125.3M |
| `env` wall | 16.9ms | 11.1ms |

`env` burns ~30M more instructions while running ~6ms faster — the eager scan's cost was readdir/stat latency, which dominates wall clock but not instruction count. On a machine with many installed tools that trade is clearly good; on CI's fixture there are few syscalls to save, so only the CPU side shows, which is why `ls` gates at +3.33%.

`BackendArg::full()` is where the CPU goes: 23.7M inclusive on main vs 49.3M here, ~783 calls. It is *not* the install-state lookups inside it — those are only 11 calls — so the remaining work is to find why the same function costs 2× per call, and I did not get to the bottom of that.

Correctness is verified regardless: `mise alias ls`, fixture `mise ls` and global `mise ls` are byte-identical to main, and unit suites (toolset 91, backend 595, config 276, shims 11) plus the e2e sweep pass.

I would rather not paper over the gate. Happy to keep digging into `full()`, or to narrow this PR to the parts that do not move instruction counts, whichever you prefer.

jdx:
Kept digging — found the root cause, and it was not what I assumed. `b0dee62b7`.

## Root cause

`BackendArg::full()` consults the lockfile for a locked backend identity, and `read_all_lockfiles` **rebuilt its cache key on every call**: resolve the monorepo lockfile root, map every config path through `lockfile_path_for_config`, collect legacy paths, dedup through a `HashSet<PathBuf>`, then hash that vector to do the lookup. That is the `Path`/`PathBuf` hashing and the `Components::parse_next_component_back` the profile kept pointing at.

Why this PR exposed it: `full()` short-circuits before the lockfile branch when `install_state` already knows the tool. Building installed-tool backends used to happen in `load_tools` **before config load**, where `config::is_loaded()` is false and the whole lockfile branch is skipped. Deferring that construction moved those calls *after* config load, so they now enter the branch. Measured: `read_all_lockfiles` calls went 210 → 342 on `mise env`.

So the regression was not the lazy scan, and not the per-tool lookups (only 11 `get_tool_full` calls on that path). It was ~130 extra calls into a function that was quietly expensive per call.

## Fix

Derive the key once per config and hold it there, and move the work only the miss path needs below the cache lookup.

| | main | before | now |
|---|---|---|---|
| `ls` instructions | 109,988,312 | 140,402,304 | **122,429,416** |
| `env` instructions | 95,955,029 | 125,339,686 | **101,144,978** |

`env` is +5.4% over main and `ls` +11.3%, down from +31% and +28%. Note this fix helps main's paths too — those 210 calls were already rebuilding the key.

## What is left

`Lockfile::read` accounts for most of the residual: 9 parses on main vs 11 here, +9.1M inclusive. Same `read_lockfile_at` call count (32 both), so two extra distinct cache keys are being produced somewhere; I have not tracked down which yet. The remaining `full()` delta on `ls` also reflects 688 → 783 calls.

Verified unchanged along the way: `mise ls` (fixture and global), `mise env`, and `mise alias ls` are byte-identical to main; lockfile unit tests (72) and the full `test_lock*` e2e suite pass, plus config/toolset/backend units and clippy.

jdx:
All addressed in `e5327b2bd` except one test suggestion — details below.

**@cursor / @coderabbitai / @greptile-apps — `BLIST` debug print: valid, removed.** Leftover from the callgrind session in the previous round; entirely my mess. It's also what failed every e2e tranche: quiet/silent output assertions saw the extra stderr line, and `cli/test_broken_stderr` panicked outright because `eprintln!` aborts when the stderr write fails (`2>/dev/full`). The previously-failing tests pass again locally.

**@coderabbitai — `load_tools` not single-flight: valid, fixed.** Two tasks racing past the memo check both built seed maps, and the loser's commit replaced a map `list()` may already have extended with installed tools — with `TOOLS_INCLUDE_INSTALLED` left set, so the installed entries were never restored. `load_tools` now rechecks `TOOLS` while committing and adopts an already-published map.

**@coderabbitai — full scan can hide a concurrent install: valid, fixed structurally.** Rather than merging the per-tool memo into the scanned map at publish (which would also fold in plain memoized *reads*, resurrecting state that's stale relative to the scan), `full_scan_tools` now holds the full-map lock from before the first directory read until publish. `add_tool_version` takes the same lock, so it either completes before the scan starts reading — its version is on disk by then — or blocks until the publish and updates the published map. The bad interleaving (scan reads early, installer records, publish shadows it) can no longer be scheduled. This also makes the scan single-flight.

**@coderabbitai — shared manifest alternate-dir mapping invisible to `get_tool`: valid, fixed.** The per-tool path probed only the kebab-cased dir in each shared install dir; the full scan finds renamed dirs by enumeration. `load_tool` now reads the shared manifest mapping and probes the recorded dir as well. Covered in `e2e/cli/test_install_state_lazy` (`mise latest --installed dummy@2` resolves a version that only exists in a shared dir under a non-kebab name, still without `full_scan_tools` in the timings) — verified to fail with the probe reverted. Worth noting the fix is about identity/version visibility: execution-path resolution (`find_in_shared_installs`) probes by kebab name and never handled renamed shared dirs, on main or here.

**@coderabbitai — memoize shared manifests: valid, fixed.** Per-dir memo next to the root manifest memo, cleared in `reset`, and the full scan reads through it too.

**@coderabbitai — lockfile discovery repeated on the cache-miss path: valid, fixed.** The once-per-config derivation now returns a `LockfileDiscovery` carrying the cache key plus the monorepo root and legacy paths it had to compute anyway, and the miss path reuses them instead of re-running the discovery.

**@coderabbitai — same-process `backend::list()` install/uninstall e2e coverage: not taken as an e2e.** Each e2e assertion is a separate process, so it can't observe the cached map across a mutation; the same-process semantics are pinned by the `installed_replaces` unit test from round 1 plus `add_tool_version`/`add_plugin` updating both memo layers (and now the published map, per the lock change above).

Validation: `cargo clippy --workspace --all-features --all-targets -- -D warnings` clean, 973 unit tests across backend/config/shims/toolset pass, lockfile+install_state+outdated_info suites pass, and an e2e sweep over the CI-failed tests (`test_use`, `test_exit_status`, `test_task_help`, `test_task_silent_streams`, `test_task_double_dash_behavior`, `test_task_edit_nested_names`, `test_broken_stderr`, `test_ls`, `test_uninstall`) plus `test_install_state_lazy` and `test_prune_shared_installs` is green.

PR #12261, opened 2026-08-21
fix(install-script): default the pinned binary under the data dir, not the cache dir

Guria:
Reported in https://github.com/jdx/mise/discussions/12245.

## The problem

The bash wrapper generated by `mise generate install-script` self-installs the pinned mise binary to `$XDG_CACHE_HOME/mise/mise-$VERSION` — inside mise's own cache directory. The runtime that every subsequent invocation execs gets deleted by:

- `mise cache clean` / `cache prune`
- auto-prune (atime-based, fires on ~1% of invocations)
- any "rm -rf ~/.cache/mise" troubleshooting step

On macOS it is worse: the wrapper hardcodes `~/.cache`, but mise resolves its cache dir to `~/Library/Caches/mise` there — so the binary survives cache wipes, but leaves a stray `~/.cache/mise` that nothing owns or cleans.

## The change

Defaults the bash wrapper's install path to `${MISE_DATA_DIR:-${XDG_DATA_HOME:-$HOME/.local/share}/mise}/bootstrap/mise-$VERSION` — matching how the Windows launcher already installs into `%LOCALAPPDATA%\mise`, which is `MISE_DATA_DIR` on Windows — and moves the Windows side under a matching `bootstrap\` subdirectory so both halves of the command agree.

Migration: wrappers keep running a binary left at the old location instead of re-downloading it. On Unix that is the old cache path; on Windows it is directly under `%install_dir%`. The fallback only applies when `MISE_INSTALL_PATH` was **not set** — an explicit value still wins even while its file does not exist yet (#11401), including a value that happens to equal the new default. The bash probe uses `-x`, so a non-executable leftover reinstalls rather than dying at `exec`.

The localized variant is untouched on both platforms: its directory is project-local, not a cache, and its binary was always a sibling of its own data dir. Same disease class as #7267.

## Local review findings addressed

The revision was reviewed locally before opening this PR; these findings are already fixed in it:

1. **[blocker]** The first cut moved the *localized* Windows launcher under `bootstrap\` too (the cmd template is shared), breaking the non-recursive `.mise\mise-*.exe` glob in `e2e-win/generate_bootstrap.Tests.ps1` and diverging from the untouched localized bash wrapper. `windows_script()` now takes a `localized` flag; the localized launcher keeps `%install_dir%\mise-%resolved_version%.exe` and no legacy fallback.
2. **[major]** The fallback originally reconstructed "was the path defaulted?" by string-equality against the re-derived default, which silently hijacked an explicit `MISE_INSTALL_PATH` equal to the default. Bash now gates on `[ -z "${MISE_INSTALL_PATH:-}" ]`; cmd nests the fallback inside the existing `if not defined MISE_INSTALL_PATH (...)` block, which also avoids reading `%MISE_INSTALL_PATH%` inside a parenthesized block (delayed expansion is intentionally disabled).
3. **[minor]** The bash legacy probe used `[ -f` — a leftover without the exec bit would pass the probe and die at `exec` instead of reinstalling. Now `[ -x`.
4. **[minor]** The unix e2e only proved the cache fallback. It now also stubs the data-dir `bootstrap/` path (fresh installs land there), asserts the data-dir binary wins when both generations exist, and covers a run-time `XDG_CACHE_HOME` override for the legacy probe.
5. **[nit]** Docs now scope the new default to the non-localized wrapper and describe the old Windows location (`%LOCALAPPDATA%\mise` root, not a cache dir) correctly.

Known pre-existing limitations, deliberately left out of scope: the primary skip-install check is `test -f` (bash) / `if exist` (cmd), which accept a non-executable file or a directory at the install path — both predate this change.

## Review-bot feedback addressed (round 2)

Rebased onto `main`, which renamed the command to `generate install-script` (#12247); the change now applies on top of the rename. In addition:

6. **[minor] (CodeRabbit, Copilot)** Docs said the localized wrapper keeps its binary "beside the script" — inaccurate: with `bin/mise`, the binary lives in the localized data directory. Docs now say exactly that.
7. **[major] (Copilot)** `MISE_DATA_DIR: ~/...` (the exact shape mise's own CI YAML supplies) was not tilde-expanded by the wrapper, because parameter expansion does not do tilde expansion while mise itself does (`env::var_path` → `file::replace_path`). The wrapper now expands a leading `~/` (and a bare `~`) against `$HOME`; covered by a unit test, an e2e case with a literal `'~/…'` value, and both bash 3.2 and modern bash.
8. **[out of scope, noted] (Copilot)** The Windows launcher's `install_dir` remains the hard-coded `%LOCALAPPDATA%\mise` and does not honor `MISE_DATA_DIR`/`XDG_DATA_HOME`. This is pre-existing behavior — the old default ignored those variables too — and changing it means tilde expansion and variable precedence logic inside a cmd script that is deliberately written around `DisableDelayedExpansion`. Prefer a follow-up if wanted.

## Testing

- New unit tests for the bash default/fallback/override gating and the windows localized vs non-localized defaults; existing template tests extended (`cargo test --bin mise generate::bootstrap`, 17 passed).
- Behavioral matrix against the real generated wrapper in a sandboxed `$HOME`: fresh install → data dir; legacy-only → cache reuse; both → data dir wins; explicit override equal to the default → not redirected; non-executable legacy → ignored; `XDG_CACHE_HOME`/`MISE_DATA_DIR` run-time overrides honored (8/8).
- `e2e/generate/test_generate_bootstrap` extended with the stub cases above.

*AI-assisted — Tool: pi; model: zai/glm-5.3;*

## Summary by CodeRabbit

* **New Features**
  * Install scripts now use version-specific locations in the data directory by default.
  * Existing binaries from older installation locations remain usable through automatic fallback.
  * Custom cache and data-directory paths are honored, including paths beginning with `~`.
  * Windows localized launchers continue storing binaries beside the launcher.

* **Documentation**
  * Updated continuous-integration documentation with the latest installation-path and compatibility guidance.

* **Tests**
  * Expanded coverage for migration, custom paths, tilde expansion, and legacy binary reuse.


Guria:
Bot review feedback processed in f504c59:

- **Docs wording (CodeRabbit + Copilot):** "beside the script" replaced with "in its localized data directory (`.mise/` by default)".
- **Tilde expansion (Copilot):** the wrapper now expands a leading `~/` / bare `~` in `MISE_DATA_DIR`/`XDG_DATA_HOME` against `$HOME`, matching mise's own `env::var_path` → `file::replace_path`. Unit test, an e2e case passing a literal `'~/…'` value, and a bash-3.2 + modern-bash behavioral check cover it.
- **Windows `MISE_DATA_DIR`/`XDG_DATA_HOME` overrides (Copilot):** left as a pre-existing limitation — the old default ignored them too, and honoring them needs tilde/precedence logic in a cmd script deliberately built around `DisableDelayedExpansion`. Called out in the PR description; happy to do a follow-up if wanted.
- Rebased onto `main` to follow the `generate bootstrap` → `generate install-script` rename (#12247); commit rescoped to `fix(install-script)`.

PR #12265, opened 2026-08-22
fix(ls-remote): distinguish unknown from stable in JSON prerelease output

risu729:
## Summary

Make `VersionInfo.prerelease` an `Option<bool>` so `ls-remote --json` output distinguishes three states instead of two:

| output | meaning |
| --- | --- |
| `"prerelease": true` | pre-release, per the source |
| `"prerelease": false` | confirmed stable, per the source |
| key absent | the source has no pre-release signal |

## Problem

Today the field is a plain `bool` serialized with `skip_serializing_if = "is_false"`, so the JSON output only ever contains `"prerelease": true` or nothing:

```console
$ mise ls-remote --prerelease --json pinact
{"version":"4.1.0-0", …, "prerelease":true}
{"version":"4.1.0",   …}
{"version":"4.1.1-0", …, "prerelease":true}
{"version":"4.1.1",   …}
```

An absent key is ambiguous: for `pinact` (aqua → GitHub releases) mise *knows* `4.1.1` is stable — the GitHub API returns `prerelease: false` — but the output is indistinguishable from `node`, where the core plugin has no pre-release signal at all.

Consumers that cache or aggregate this output cannot tell "confirmed stable" from "unknown", so they cannot safely overwrite previously recorded flags. Concrete fallout: mise-versions recorded `pinact 4.1.1` as a pre-release while it briefly was one, and since mise never emits `false`, nothing could ever clear the flag once the release was promoted — <https://mise-versions.jdx.dev/pinact> has served `4.1.0` as the newest version for three weeks (jdx/mise-versions#284 works around it on their side).

The `--all --json` doc comment even says the flag is "Always emitted so JSON consumers can rely on its presence", and the `after_long_help` example shows `"prerelease":false` — the single-tool path just never produced it.

## Changes

- `VersionInfo.prerelease: bool` → `Option<bool>` with `skip_serializing_if = "Option::is_none"`. `Default` gives `None`, so backends without a signal need no change and old msgpack caches deserialize as `None`.
- Sources that know populate `Some(..)`:
  - github / forgejo release listings: `Some(r.prerelease)` (GitHub's API always returns the boolean)
  - aqua: the release path carries `Some(r.prerelease)`; the `github_tag` path now yields `None` instead of a false "stable" (tags have no pre-release flag)
  - npm: `Some(is_semver_prerelease(..))` — semver grammar is total for npm versions
  - versions-host entries: pass through `Option<bool>`; absent host data maps to `None` instead of `false`
- Pattern-based detection (`mark_prerelease`, pipx's PEP 440 rule, java's version regex) only produces `Some(true)` on a match and never overrides an authoritative `Some(false)`; a regex miss proves nothing, so it leaves the value as-is (`None` for signal-less sources).
- Filtering behaviour is unchanged: `filter_cached_prereleases` drops `Some(true)` and keeps `Some(false)`/`None`, exactly as it previously dropped `true` and kept `false`.
- `--all --json` keeps its always-present contract: `true`/`false` when known, `null` when not.
- New unit test pins the three-state serialization; existing prerelease tests updated to assert `Some(..)`/`None` explicitly.

With this, single-tool output for a GitHub-backed tool matches what the long-help example already promised:

```console
$ mise ls-remote --prerelease --json pinact
{"version":"4.1.0","created_at":"2026-06-09T00:41:41Z","release_url":"…","prerelease":false}
{"version":"4.1.1-0","created_at":"2026-07-24T00:34:52Z","release_url":"…","prerelease":true}
{"version":"4.1.1","created_at":"2026-07-30T00:23:08Z","release_url":"…","prerelease":false}
```

while `node` (no signal) still omits the key.

## Compatibility

- Cache: `#[serde(default)]` reads old entries (absent → `None`, previously `false`); filters treat both as not-prerelease, so behaviour is identical without any cache invalidation.
- JSON consumers doing `.prerelease == true` are unaffected. Consumers that treated absence as "stable" now additionally receive explicit `false` where mise actually knows — which is the point.
- `--all --json`: `false` becomes `null` for tools with no signal. That is a behaviour change for consumers reading the raw boolean; the field itself remains present as documented.

## Transition analysis for caching consumers

For a consumer that stores flags across runs (mise-versions), the merge rule is: explicit boolean wins, absent keeps the stored value. What changes per transition, before vs after this PR:

| transition (same version, between runs) | before (two-state output) | after (three-state output) |
| --- | --- | --- |
| upstream promotes prerelease → release | key just disappears — indistinguishable from "no signal", stored `true` **stuck forever** (the pinact case) | emits `false` → cleared ✅ |
| upstream demotes release → prerelease | `true` appears → flagged ✅ | same ✅ |
| backend gains a signal (e.g. aqua `github_tag` → `releases`) | only new `true`s arrive; a wrong pattern-stamped `true` on a stable version stays stuck | `false` arrives → corrected ✅ |
| backend loses its signal (source switch, API error → fallback) | indistinguishable from every other run | key absent = "unknown" → stored value kept, self-heals when the signal returns |
| mise's pattern newly matches | `true` → flagged ✅ | same ✅ |
| mise's pattern stops matching | key disappears → stored `true` stuck | key absent → stored `true` still kept ⚠️ (see below) |

The one residual case is a retracted pattern stamp: a regex non-match proves nothing, so mise cannot emit `false` for it (the same principle that keeps java/pipx pattern detection from claiming "confirmed stable"), and the consumer cannot distinguish a stamped `true` from an upstream `true` — the output carries no provenance. Closing that would need a provenance field (upstream vs pattern), which is out of scope here. Every transition backed by an actual upstream signal now round-trips correctly; before this PR, no transition could ever clear a stored flag.

## Testing

- `cargo test --bin mise` (unit suite)
- `cargo clippy --all-features --tests` clean
- manually verified against live data: `pinact` emits explicit `true`/`false` per the GitHub flag; `node` omits the key

*AI-assisted — Tool: Claude Code; model: anthropic/claude-fable-5; version: 2.1.237.*


risu729:
Addressed the CodeRabbit finding in 757c52fb3: `mark_prerelease` and the pipx PEP 440 rule now stamp `Some(true)` only over `None`, so an authoritative `Some(false)` from a release flag survives pattern detection. Added a regression test for the `Some(false)` + regex-match case (the pipx closure isn't unit-testable in isolation, but it applies the identical rule).

*AI-assisted — Tool: Claude Code; model: anthropic/claude-fable-5; version: 2.1.237.*

risu729:
Both inline findings addressed:
- None-only stamping (mod.rs / pipx): fixed in 757c52fb3 with a `Some(false)`-survives-regex test.
- pipx dropping GitHub's flag: fixed in b8936464a — `versions_from_github_releases` now copies `Some(r.prerelease)`, with a regression test for a PEP 440-looking tag published as a full release.

*AI-assisted — Tool: Claude Code; model: anthropic/claude-fable-5; version: 2.1.237.*

risu729:
@coderabbitai resume

risu729:
Latest round addressed in 6199f0ce9:
- **pipx test duplicating the stamping guard** (valid): extracted `stamp_pep440_prerelease` and both the production path and the regression test now go through it.
- **"--all --json should emit null"** (not applicable): that contract lives on `VersionOutputAll`, whose `prerelease: Option<bool>` has no `skip_serializing_if` — unknown already serializes as `null`. The `skip_serializing_if` on `VersionInfo` (mod.rs:315) governs the single-tool output, where the documented contract is key-absent-when-unknown. Added a serde test pinning the `--all` null behavior so the two structs' contracts are both under test.

*AI-assisted — Tool: Claude Code; model: anthropic/claude-fable-5; version: 2.1.237.*

PR #12268, opened 2026-08-22
fix(config): don't prompt for trust when stdin is not a tty

Marukome0743:
Fixes the remaining half of #8940.

## Problem

`src/ui/prompt.rs` decided whether it could prompt from `console::user_attended_stderr()` alone — stdin was never checked. `demand` defines interactive as stdin **and** stderr (`demand::tty::is_tty`), and `Confirm` honors that, but `Dialog` has no non-tty branch: it drives `Term::stderr().read_key()` unconditionally.

So with a terminal on stderr but not on stdin — `docker run -t` without `-i`, a `script`-wrapped CI step, `mise env < /dev/null` — mise painted the trust dialog even though nothing could answer it. `console` then fell back to `/dev/tty` and either blocked forever waiting for a keypress nobody was there to type, or failed outright:

```
Error:
   0: error parsing config file: /tmp/proj/mise.toml
   1: Device not configured (os error 6)

Location:
   src/ui/prompt.rs:49
```

That errno **replaced** the actionable `Config files in … are not trusted. Trust them with 'mise trust'` message. It also emitted `ESC[?25l` without ever pairing it, leaving the terminal with an invisible cursor. With both streams piped the same command already failed correctly, so only this one fd shape was broken.

Investigating turned up two more defects with the same root cause:

**A completion keystroke could silently disable a config.** `trust_check` gated `add_ignored()` on `console::user_attended_stderr()`, but `confirm_with_all` also returns `false` when it *could not ask* — including when `__USAGE` is set. With stderr a terminal, `__USAGE=1 MISE_YES=0 mise env` in an untrusted directory exited 0 and persisted `state/ignored-configs/<root> -> <root>`. That marker is sticky and silent, so the config simply stopped applying with no message. Reproduced; the piped control run creates no marker.

**`prompt::confirm()` failed open on EOF.** `demand::Confirm`'s non-tty branch discards the 0-byte return of `read_line`, so EOF is indistinguishable from a bare newline and applies the default. `confirm()` defaults to `true`, so `mise implode < /dev/null` in a terminal auto-confirmed every deletion (verified — it removed the first path it asked about).

## Fix

- `confirm_with_all` requires a terminal on both ends, via a new `can_prompt_dialog()`. This restores `demand`'s own contract rather than inventing a new rule.
- `trust_check` gates `add_ignored` on the same predicate, so only a real "no" is remembered.
- `confirm_with_default` reads the line itself when stdin is not a terminal: EOF → `false`, empty line → the default, any prefix of `yes`/`no` answers, and anything else is an error rather than a silent "no". `echo y | mise …` keeps working.
- Prompt errors restore the cursor before propagating.

The other `confirm_with_all` callers now get `Ok(false)` instead of an errno, which matches what they already do when fully piped: `asdf_plugin`/`vfox_plugin` return `PluginNotInstalled`, `prune` skips the entry rather than deleting it.

## Behavior change worth calling out

Dialog prompts no longer reach the controlling terminal through `/dev/tty` when stdin is redirected. This is broader than one command: `foo | mise install`, `mise x -- cmd < file`, or `mise prune < /dev/null` from a live terminal all used to prompt and now fail closed with the actionable error instead.

Reading `/dev/tty` under a redirect is normal for terminal programs — `sudo` and `ssh` do it — so this is a genuine capability being given up, not just the removal of a quirk. The trade is deliberate: discussions#8940 is precisely the case where that dialog is unanswerable, and `Dialog` cannot consume a piped answer anyway, so the alternatives are an actionable "run `mise trust`" error or a hang. `confirm_with_default` is unaffected — it still answers from piped stdin.

## Tests

- Unit tests in `src/ui/prompt.rs` for the predicate and the answer parsing. `trust_check` returns early under `cfg!(test)`, so this is the only unit-level seam.
- `e2e/config/test_trust_prompt_non_interactive` runs mise under a pty with `/dev/null` on stdin and asserts the actionable error, no hang, and no ignore marker — for both the plain and `__USAGE` shapes. It skips where `script(1)` cannot allocate a pty (macOS needs a terminal on its own stdin, which the harness lacks).

Verified locally with a `pty.openpty()` harness: the errno is gone, the cursor is no longer left hidden, no ignore marker is written in either shape, and a real terminal still shows the Yes/No/All dialog. `cargo clippy --workspace --all-features --all-targets -- -D warnings` is clean (no lint exclusions added).

## Possible follow-up

If you would prefer it, `confirm_with_all` could return a tri-state so "couldn't ask" is structurally distinct from "the user declined", instead of two call sites consulting `can_prompt_dialog()` independently. That would make the `add_ignored` bug unrepresentable rather than merely fixed, at the cost of touching all four callers. Happy to do that instead.

*AI-assisted — Tool: Claude Code; model: anthropic/claude-opus-5; version: 2.1.231.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Trust prompts now behave correctly in non-interactive environments.
  * Failed or unavailable prompts no longer create persistent ignore entries.
  * Usage and completion generation no longer attempt interactive dialogs.
  * Prompts handle EOF, invalid input, input errors, and cursor restoration more reliably.
  * Recognized responses and default choices are handled consistently.

* **Tests**
  * Added coverage for non-interactive trust prompts and confirmation behavior, including EOF, default responses, explicit answers, and invalid input.

### DIFF
--- a/e2e/config/test_trust_prompt_non_interactive
+++ b/e2e/config/test_trust_prompt_non_interactive
@@ -1,0 +1,79 @@
+#!/usr/bin/env bash
+
+# The trust prompt is a `demand::Dialog`, which has no non-tty branch: it drives
+# `Term::stderr().read_key()` unconditionally. With a terminal on stderr but not on
+# stdin (`docker run -t` without `-i`, a `script`-wrapped CI step) `console` falls
+# back to /dev/tty and either blocks forever waiting for a keypress nobody is there
+# to type, or fails with an errno that replaces the actionable "not trusted" error.
+#
+# mise must fail the same way it does with plain pipes, and it must not record an
+# ignore marker -- that marker is sticky and silent, so the config would just stop
+# applying without the user ever having declined it.
+
+require_cmd script
+
+# A terminal on stderr is the whole point of this test, so it needs a pty. macOS's
+# script(1) additionally requires a terminal on its own stdin, which the harness
+# does not have, so skip where a pty cannot be allocated rather than fail.
+if ! script -qec true /dev/null >/dev/null 2>&1; then
+  echo "skipping: script(1) cannot allocate a pty here" >&2
+  exit 0
+fi
+
+export MISE_TRUSTED_CONFIG_PATHS=""
+
+cat <<'EOF' >mise.toml
+[env]
+TRUST_PROMPT_TEST = "1"
+EOF
+
+ignored_configs="$MISE_STATE_DIR/ignored-configs"
+
+# Gives mise a pty on stderr and /dev/null on stdin. Kept under a timeout because
+# the regression this guards against is a hang, not a failure. `-e` makes script(1)
+# exit with the child's status. 10s is well clear of the ~1s a passing run takes,
+# and keeps a regression under the harness's 20s "should be marked slow" warning.
+run_in_pty() {
+  timeout 10 script -qec "$1 </dev/null" /dev/null
+}
+
+marker_count() {
+  if [[ -d $ignored_configs ]]; then
+    find "$ignored_configs" -mindepth 1 | wc -l | tr -d ' '
+  else
+    echo 0
+  fi
+}
+
+# MISE_PARANOID=1 keeps the trust check deterministic by skipping the
+# implicitly-trust-active-config fast path.
+status=0
+output="$(run_in_pty 'MISE_YES=0 MISE_PARANOID=1 mise env' 2>&1)" || status=$?
+
+if [[ $status == 124 ]]; then
+  fail "mise blocked on a trust prompt that nothing could answer"
+fi
+if [[ $status == 0 ]]; then
+  fail "expected mise to fail on untrusted config, got: $output"
+fi
+assert_contains_text "$output" "not trusted"
+assert_contains_text "$output" "mise trust"
+
+if [[ $(marker_count) != 0 ]]; then
+  fail "an ignore marker was recorded without the user declining"
+fi
+ok "[trust] no ignore marker recorded when the prompt cannot be answered"
+
+# Same terminal shape while generating usage/completions. Before the fix this exited
+# 0 -- because it had just recorded the ignore marker -- so the marker is what
+# distinguishes the bug, not the exit status.
+status=0
+run_in_pty '__USAGE=1 MISE_YES=0 MISE_PARANOID=1 mise env' >/dev/null 2>&1 || status=$?
+
+if [[ $status == 124 ]]; then
+  fail "mise blocked while generating usage in an untrusted directory"
+fi
+if [[ $(marker_count) != 0 ]]; then
+  fail "generating usage recorded an ignore marker"
+fi
+ok "[trust] no ignore marker recorded while generating usage"
--- a/src/ui/prompt.rs
+++ b/src/ui/prompt.rs
@@ -1,3 +1,4 @@
+use std::io::{BufRead, IsTerminal};
 use std::sync::Mutex;
 
 use demand::{Confirm, Dialog, DialogButton};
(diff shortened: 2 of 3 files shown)


e2e/config/test_trust_prompt_non_interactive
```diff
@@ -0,0 +1,78 @@
...
+TRUST_PROMPT_TEST = "1"
+EOF
+
+ignored_configs="$MISE_STATE_DIR/ignored-configs"
+
+# Gives mise a pty on stderr and /dev/null on stdin. Kept under a timeout because
+# the regression this guards against is a hang, not a failure. `-e` makes script(1)
+# exit with the child's status.
+run_in_pty() {
+  timeout 30 script -qec "$1 </dev/null" /dev/null
```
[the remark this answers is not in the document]
Marukome0743: Skipping this one — `timeout` here isn't an external binary, so the 127 path can't happen, and adding the guard would break the test on stock macOS.

`timeout` is a shell function defined in `e2e/assert.sh:75`. When neither `timeout` nor `gtimeout` is on PATH it falls back to `run_with_timeout`, a pure-shell implementation that returns 124 on expiry. `run_test` sources `assert.sh` before the test body, so the function is always in scope and shadows the missing binary — the shell never gets far enough to return 127.

The proposed guard would also be checking the wrong thing: `require_cmd` uses `type -p`, which only matches an executable on disk and never a shell function. On a machine without coreutils (stock macOS has no `timeout` and no `gtimeout`) `require_cmd timeout` would abort the test with exit 2 even though the fallback works fine there.

`require_cmd script` stays, since that one really is an external binary with no fallback.

*AI-assisted — Tool: Claude Code; model: anthropic/claude-opus-5; version: 2.1.231.*

src/ui/prompt.rs
```diff
@@ -25,20 +61,54 @@ pub(crate) fn confirm_with_default<S: Into<String>>(
...
         return Ok(false);
     }
+    let message = message.into();
+    // Deliberately not `can_prompt_dialog()`: `Confirm` does have a non-tty
+    // branch, so `echo y | mise ...` is a supported way to answer. What it cannot
+    // do is tell EOF from a bare newline -- it discards the 0-byte return of
+    // `read_line` and applies the default -- so `mise implode < /dev/null` would
+    // answer "yes" to every deletion. Read the line here instead, so that no
+    // answer means no.
+    if !std::io::stdin().is_terminal() {
```
jdx, line 75: This early return happens before `_progress_pause` is acquired on line 74, so the non-tty path skips the progress suspension that #11421 (`fix: suspend progress while displaying prompts`) added deliberately. With a live progress bar and piped stdin — `install_into` is the likely caller, since it prompts mid-install — the prompt gets overwritten by progress redraws and mise then blocks on a read the user can't see.

Hoisting the `_progress_pause` binding above this branch fixes it; the guard is a no-op when no report is active.

jdx (commented):
Went through this against the `demand` and `console` sources rather than taking the writeup at face value, and the diagnosis holds up on every point:

- `demand::tty::is_tty()` is `stdin.is_terminal() && stderr.is_terminal()` (`demand-2.0.5/src/tty.rs:4`), but `Dialog::run` (`dialog.rs:141`) goes straight to `hide_cursor()` + `term.read_key()` with no non-tty branch.
- `console`'s `Input::unbuffered` (`unix_term.rs:87-100`) opens `/dev/tty` when stdin isn't a terminal, which is exactly the hang / `ENXIO` path.
- `Confirm::run`'s non-tty branch (`confirm.rs:148`) calls `tty::read_line()`, which discards `read_line`'s byte count — so EOF really is indistinguishable from an empty line and falls through to `self.selected`. With `confirm()` defaulting to `true`, `mise implode < /dev/null` auto-confirming every deletion is real. That one alone is worth the fix.

Also confirmed the four `confirm_with_all` callers all degrade safely on `Ok(false)`: `prune` skips the entry, asdf/vfox return `PluginNotInstalled`, `trust_check` raises the actionable "not trusted" error.

I built the branch and checked the e2e test isn't passing vacuously — reverting both predicates back to `console::user_attended_stderr()` and re-running gives:

```
ERROR: E2E assertion failed: mise blocked on a trust prompt that nothing could answer
config/test_trust_prompt_non_interactive: 32s
```

so it catches the hang through the `timeout`/124 path. Nice test.

One thing to fix before this goes in — see the inline note on the progress pause.

Smaller notes, none blocking:

- **The answer parser is stricter than what it replaces.** `demand`'s non-tty branch did prefix matching against the affirmative/negative labels and returned `Err(InvalidInput)` on anything unrecognized. `parse_confirm_answer` accepts only `y`/`yes`/`n`/`no` and silently maps everything else to `false`, so `echo ye | mise install-into ...` used to proceed and now quietly declines. Tiny blast radius, but silently reinterpreting something the user typed as "no" is arguably worse than erroring. Happy either way.
- **`/dev/tty` prompting going away is broader than the `mise prune < /dev/null` framing.** `foo | mise install` or `mise x -- cmd < file` from a live terminal used to prompt and now errors. Reading the controlling terminal under a redirect is normal for terminal programs (`sudo`, `ssh`), so I wouldn't call it "silently taking the terminal over" — but given discussions#8940 is precisely that dialog being unanswerable, failing closed on an actionable "run `mise trust`" is the right call. Fine as-is, just don't undersell it in the changelog.
- **Take the tri-state follow-up.** Two call sites independently deriving "could we have asked?" is the trap, and `trust_check` consulting `can_prompt_dialog()` a second time to decide whether to remember a "no" is the shape of the bug you just fixed. Make it unrepresentable rather than fixed — separate PR is fine.
- The 30s timeout means a regression takes ~32s and trips the harness's ">20s but not marked as slow" warning. Harmless (1s when passing), but ~10s would keep failures snappy.

src/ui/prompt.rs
Marukome0743: Good catch, and it was worse than an ordering nit — the non-tty branch was the one path that could *silently* deadlock. With a progress bar live, the prompt line gets scribbled over and then mise blocks on a read the user has no idea is waiting.

Fixed by hoisting the `_progress_pause` binding above the branch. It stays alive across the early return, since the return expression is evaluated before locals are dropped, so the pause covers `read_confirm_from_stdin` too.

Also took two of the smaller notes:

**Answer parser.** You're right that silently rereading a typed answer as "no" is worse than erroring, so `parse_confirm_answer` now matches what `demand` accepted: any prefix of `yes`/`no` (`ye`, `n`, ...), empty line for the default, and an error for anything unrecognized. EOF still returns `false` — that is the one case `demand` got wrong and the reason the branch exists at all. Unit test renamed to `unrecognized_answers_are_an_error`.

**e2e timeout.** 30s -> 10s. A passing run is ~1s, so a regression now trips at ~12s and stays under the harness's 20s "should be marked slow" warning.

**Changelog framing.** Rewrote that section of the description. It now says dialog prompts no longer reach the controlling terminal through `/dev/tty` at all when stdin is redirected — `foo | mise install` and `mise x -- cmd < file` as well as `mise prune < /dev/null` — and drops the "silently taking the terminal over" characterization, since reading `/dev/tty` under a redirect is normal behavior for terminal programs. It is a real capability being given up; the trade is that the alternative here is a hang.

**Tri-state.** Agreed, and agreed on why: `trust_check` asking `can_prompt_dialog()` a second time to decide whether to remember a "no" is the same shape as the bug. I'll do it as a separate PR rather than widening this one.

*AI-assisted — Tool: Claude Code; model: anthropic/claude-opus-5; version: 2.1.231.*

PR #12270, opened 2026-08-22
registry: add nushell (aqua:nushell/nushell)

i-api:
In the aqua standard registry, so `aqua` is the preferred backend, with `github` as a tier 1 fallback. The binary is `nu` rather than `nushell`, so `bins` and a `nu` alias are set; `mise search nu` currently returns no nushell result. All 100 release tags are strict `MAJOR.MINOR.PATCH`, hence `version_order = "semver"`.

## Popularity

- GitHub: 40,315 stars, 2,214 forks; latest release 0.115.0 published 2026-08-15
- Active maintenance: latest repository push 2026-08-21
- crates.io `nu`: 448,911 downloads
- Also packaged in Homebrew, nixpkgs, and winget

## Validation

- `mise test-tool nushell` -> `nushell: OK`
- `mise x aqua:nushell/nushell@0.115.0 -- nu --version` -> `0.115.0`
- `mise x github:nushell/nushell@0.115.0 -- nu --version` -> `0.115.0`

## Summary by CodeRabbit

* **New Features**
  * Added Nushell registry support, including version detection and semantic version ordering.
  * Configured metadata and download sources for the Nushell binary.

### DIFF
--- a/registry/nushell.toml
+++ b/registry/nushell.toml
@@ -1,0 +1,6 @@
+aliases = ["nu"]
+backends = ["aqua:nushell/nushell", "github:nushell/nushell"]
+bins = ["nu"]
+description = "A new type of shell"
+test = { cmd = "nu --version", expected = "{{version}}" }
+version_order = "semver"


jdx:
you don't want to install shells with mise

i-api:
@jdx if the concern is mise owning our login shell, I agree, that's not what we'd
want (although the registry does already include 'powershell' https://github.com/jdx/mise/pull/5449/changes). 

But Nu isn't only a shell - it's also a programming language. (statically parsed, type checked https://www.nushell.sh/book/thinking_in_nu.html)
People use it also like python and node, rather than just as a shell. 

Please reconsider adding it -  Nu is genuinely useful programming language for bootstrapping remote hosts, and for general scripting where you'd rather not reach for bash or python - which is why I'd
like mise to have it.

i-api:
Screenshot 2026-08-22 at 22 30 18
Screenshot 2026-08-22 at 22 19 18

tldr: nushell is also super useful as a statically parsed, type checked programming language too
https://github.com/nushell/nushell

jdx:
you can still use it, I don't want it in the registry

i-api:
ok, no problem

PR #12322, opened 2026-08-23
fix(tool-stub): keep env_with_path instead of rebuilding PATH

tmkx:
Non-cached tool-stub exec rebuilt PATH from pristine env, which dropped project _.path dirs from __MISE_DIFF and broke nested LookPath of sibling stubs. Rebuilding from the caller PATH could also let an outer task's install dirs shadow the stub-selected version.

Keep toolset.env_with_path (already includes _.path and filters outer install dirs) and only prepend backend dependency bins (#7777).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed tool execution so caller-provided `PATH` entries are preserved, including paths configured through environment settings.
  * Improved nested and sibling tool discovery when using non-cached tool stubs.
  * Ensured the tool version selected by a stub takes precedence over versions provided by an outer task.
* **Tests**
  * Added end-to-end coverage verifying sibling commands remain discoverable and the selected tool version is used during tool-stub execution.

### DIFF
--- a/e2e/cli/test_tool_stub_preserves_caller_path_slow
+++ b/e2e/cli/test_tool_stub_preserves_caller_path_slow
@@ -1,0 +1,82 @@
+#!/usr/bin/env bash
+# shellcheck disable=SC2103
+#
+# tool-stub non-cached exec must:
+# 1. Keep project _.path dirs so sibling stubs remain LookPath-able
+# 2. Prefer the stub-selected tool version over an outer task's tool bins
+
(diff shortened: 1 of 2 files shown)


tmkx:
https://github.com/jdx/mise/blob/23f86076b177f71be04b77394d391bbc06270fb7/e2e-win/exec_inactive_tool.Tests.ps1#L48-L51
the output of `usage --version` is changed in recent releases.

PR #12362, opened 2026-08-24
fix(elvish): separate the prepended entry from the existing PATH

NgoQuocViet2001:
## The bug

`Elvish::prepend_env` emitted the new value immediately followed by the existing one, with nothing between them:

```rust
format!("set-env {k} {v}(get-env {k})\n")
```

```
set-env PATH '/new/dir'(get-env PATH)
```

So prepending `/new/dir` to `/usr/bin:/bin` yields:

```
/new/dir/usr/bin:/bin
```

One bogus entry where two were intended, and the first real entry loses its leading component. Since this is the call mise uses to put a tool's `bin` on `PATH`, every tool it activates is unreachable under elvish.

## Every other shell separates

| shell | emitted |
|---|---|
| bash / zsh | `export {k}="{v}:${k}"` |
| pwsh | `'{v}'+[IO.Path]::PathSeparator+${env:{k}}` |
| nushell | `$env.{k} \| prepend r#'{v}'#` (a list, so separation is implicit) |
| **elvish** | `set-env {k} {v}(get-env {k})` — **nothing** |

## The fix

Emit the separator:

```rust
format!("set-env {k} {v}':'(get-env {k})\n")
```

The quoting keeps it a literal `:` rather than a barewords parse, so it composes with the already-escaped `{v}` and the `(get-env …)` capture.

I hardcoded `:` to match what `bash` and `zsh` already do in this same trait. If you would rather it be platform-aware like the pwsh arm, `$path:list-separator` is the elvish equivalent — I left it out because it needs the `path:` module in scope in the activate script, which is a bigger change than this bug warrants. Happy to switch if you prefer.

## Verified

```
$ cargo test --bin mise shell::elvish
test result: ok. 5 passed; 0 failed
```

The snapshot had recorded the missing separator, so it is updated in the same commit — and it does guard the behaviour. Reverting `elvish.rs` while keeping the new snapshot fails:

```
    1       │-set-env PATH '/some/dir:/2/dir'':'(get-env PATH)
          1 │+set-env PATH '/some/dir:/2/dir'(get-env PATH)
test result: FAILED. 0 passed; 1 failed
```

Built with cmake installed for the `libz-ng-sys` build script, on Windows.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed environment variable updates in the Elvish shell so prepended entries are correctly separated from existing values.
  * Prevented newly added entries from being concatenated with the first existing entry.

### DIFF
--- a/src/shell/elvish.rs
+++ b/src/shell/elvish.rs
@@ -83,7 +83,11 @@ impl Shell for Elvish {
     fn prepend_env(&self, k: &str, v: &str) -> String {
         let k = shell_escape::unix::escape(k.into());
         let v = shell_escape::unix::escape(v.into());
-        format!("set-env {k} {v}(get-env {k})\n")
+        // The list separator is required: without it the new entry is glued
+        // onto the first existing one, so prepending /new/dir to
+        // /usr/bin:/bin produced /new/dir/usr/bin:/bin. bash and zsh hardcode
+        // ":" here for the same reason.
+        format!("set-env {k} {v}':'(get-env {k})\n")
     }
 
     fn unset_env(&self, k: &str) -> String {
--- a/src/shell/snapshots/mise__shell__elvish__tests__prepend_env.snap
+++ b/src/shell/snapshots/mise__shell__elvish__tests__prepend_env.snap
@@ -3,4 +3,4 @@ source: src/shell/elvish.rs
 expression: "replace_path(&sh.prepend_env(\"PATH\", \"/some/dir:/2/dir\"))"
 snapshot_kind: text
 ---
-set-env PATH '/some/dir:/2/dir'(get-env PATH)
+set-env PATH '/some/dir:/2/dir'':'(get-env PATH)


NgoQuocViet2001:
Rebased onto current `main` and force-pushed to re-run CI.

**On the failing check.** It is `unit-macos` (and the `test-ci` gate that aggregates it). `lint`, `e2e`, `windows-e2e`, `windows-unit` and all `build-*` are green. Two things suggest it is not this diff:

- `windows-unit` runs the same `shell::elvish` snapshot tests this PR touches, and passes.
- The identical `unit-macos` failure appears on #12361, which branches from the same base and changes a different file (`src/shell/nushell.rs`, two lines).

A one-line change in `elvish.rs` and a two-line change in `nushell.rs` producing the same macOS-only failure points at something outside both. I cannot read the job log without an authenticated token, so I am not claiming certainty — if it reproduces after this rebase, please paste the failing test name and I will chase it properly.

**On the `cfg!(windows)` separator** that @greptile-apps and @coderabbitai both raised — I agree with the direction, and the PR description offered it, but I did not push it here because it would break the test it is meant to be covered by:

`prepend_env` is asserted through an `insta` snapshot (`mise__shell__elvish__tests__prepend_env.snap`), and that snapshot holds the literal emitted string. Resolving the separator at Rust compile time makes that string differ between `unit-macos` and `windows-unit`, so whichever platform the snapshot is not written for would fail. Adding a second failing job to a PR that already has one seemed like the wrong trade.

Two ways forward, happy with either:

1. **Leave as-is.** `bash` and `zsh` in this same trait hardcode `:` too, so elvish is at least consistent with its neighbours, and this PR stays a one-line fix for the missing separator.
2. **Resolve at shell runtime**, the way the pwsh arm does with `[IO.Path]::PathSeparator`. The elvish equivalent is `$path:list-separator`, which keeps the emitted string constant and therefore snapshot-safe — but it needs the `path:` module in scope in the activate script, which is a wider change than this bug.

Say which you prefer and I will push it. I would rather not guess on a platform-behaviour question in someone else's shell integration.

PR #12377, opened 2026-08-24
fix(github): skip provenance API calls when lockfile has checksum but no provenance

effati:
When the lockfile records a checksum but no provenance for a tool (because it
doesn't publish SLSA attestations), install-time should not probe the GitHub
releases API to detect provenance. The lockfile checksum already guarantees
artifact integrity.

These redundant API calls fail under rate limiting since
`provenance_api_failures_fatal` defaults to `true`, causing hard install failures
in CI environments with many concurrent jobs.

The skip applies whenever the lockfile has a checksum and no provenance,
regardless of whether `--locked` is passed. `force_provenance_verify` still
forces detection.

Also fixes a pre-existing issue where `asset.digest` / `api_digest` from release
metadata could overwrite a lockfile checksum. The lockfile checksum now takes
priority when present.

Both fixes are applied to the `github` and `aqua` backends.

## Security

Only tools where the lockfile records no provenance skip detection. Tools with
recorded provenance still verify it. The `force_provenance_verify` setting
overrides this and forces API verification regardless. The lockfile checksum
remains the primary integrity guarantee.

*AI-assisted — Tool: Claude Code; model: anthropic/claude-opus-4-6; version: unavailable.*

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved locked-mode installation behavior when checksum information lacks provenance.
  - Installations now correctly evaluate existing lockfile checksums before verification.
  - Preserved validation for lockfiles that already include provenance details.
  - Improved checksum validation consistency for primary and additional downloaded assets.
  - Prevented verification settings from incorrectly affecting checksum provenance decisions.

### DIFF
--- a/src/backend/aqua.rs
+++ b/src/backend/aqua.rs
@@ -636,11 +636,18 @@ impl Backend for AquaBackend {
             }
         }
 
+        // Snapshot before verify_checksum can generate a new checksum.
+        let lockfile_had_checksum = tv
+            .lock_platforms
+            .get(&self.get_platform_key())
+            .is_some_and(|p| p.checksum.is_some());
+
         // Advance to checksum operation if applicable
         if pkg.checksum.as_ref().is_some_and(|c| c.enabled()) || api_digest.is_some() {
             ctx.pr.next_operation();
         }
-        self.verify(ctx, &mut tv, &pkg, &v, &filename).await?;
+        self.verify(ctx, &mut tv, &pkg, &v, &filename, lockfile_had_checksum)
+            .await?;
 
         // Advance to extraction operation if applicable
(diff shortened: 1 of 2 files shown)


jdx:
Not sure --locked should be part of the heuristic. I think I agree that if it's locked it should be considered safe but probably regardless of --locked as long as the same version is the one being installed

effati:
@jdx Makes sense. Dropped the locked guard. Also fixed a bug where api_digest from release metadata could overwrite lockfile checksums, and extended both to aqua. Last couple commits fix a snapshot ordering issue that was failing the cosign/attestations e2e tests on fresh installs.

PR #12476, opened 2026-08-26
refactor(brew-cask): share parent directory creation

seuros:
This is the season finale of refactor(brew-cask) series.

I might have missed some, but i will catch them later.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when preparing directories during package installation.
  * Preserved required elevated permissions for protected installation locations.
  * Improved handling of paths without parent directories and invalid target paths.
  * Improved reliability when processing application bundles, fonts, binaries, commands, and completion files.

* **Refactor**
  * Standardized file, receipt, argument, and artifact handling for more consistent cask operations.
  * Improved consistency when resolving symbolic links and validating installation paths.

### DIFF
--- a/src/system/packages/brew/cask.rs
+++ b/src/system/packages/brew/cask.rs
@@ -1264,9 +1264,7 @@ async fn cask_ruby_bin() -> Result<PathBuf> {
 }
 
 fn ensure_cask_shim(path: &Path) -> Result<()> {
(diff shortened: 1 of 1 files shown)


jdx:
this one doesn't seem worth it

seuros:
Ok ! 

I did it because both block have exact same logic, if the logic changes it will be 1 place.

PR #12530, opened 2026-08-27
fix(task): accept a single string for sources

risu729:
Task `sources` is documented and schematized as `string | string[]`, but runtime deserialization only accepted an array. This updates tasks and task templates to use the existing `deserialize_arr` helper, and updates tracked file-task headers to accept the same scalar form.

Coverage includes scalar TOML tasks, scalar task templates, and scalar `#MISE sources=...` file-task headers.

Validation after rebasing onto current main:

- `mbx test --bin mise sources_single_string` (3 passed)
- scoped `hk fix` (cargo fmt and all-features cargo check)
- `git diff --check upstream/main...HEAD`

## Gap provenance

- [#8285](https://github.com/jdx/mise/pull/8285) (merged 2026-02-21) added a scalar-string branch for task `sources` to both JSON schemas, while the TOML task parser continued to require an array. This PR removes that schema-only form.

## Summary by CodeRabbit

- **Bug Fixes**
  - Task and task template source settings now accept either a single file path or a list of file paths.
  - Existing configurations using a single `sources` value are now parsed correctly.

- **Tests**
  - Added coverage for single-value source configurations in tasks and task templates.

### DIFF
--- a/src/config/config_file/mise_toml.rs
+++ b/src/config/config_file/mise_toml.rs
@@ -4112,24 +4112,6 @@ run = "cargo build"
     }
 
     #[test]
-    fn test_task_sources_array() {
-        let body = r#"
-[tasks.build]
-sources = ["Cargo.toml", "src/**/*.rs"]
-run = "cargo build"
-"#;
-
-        let path = std::path::Path::new("/tmp/mise.toml");
-        let rf = MiseToml::from_str(body, path).unwrap();
-        let task = rf.tasks.0.get("build").expect("build task should exist");
-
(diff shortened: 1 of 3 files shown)


risu729:
Addressed the review finding: file-task `#MISE sources="..."` headers went through `parse_array`, which silently produced an empty list for a scalar. The header path now uses the same `deserialize_arr` as `depends` (`src/task/mod.rs`), with a regression test (`test_from_path_sources_single_string`). All three unit tests pass.

PR #12531, opened 2026-08-27
fix(schema): enforce env required value combinations

risu729:
## Summary

- allow `required = false` alongside a concrete environment `value`
- reject `required = true` or help text alongside a concrete value
- require standalone `required` entries to be `true` or a help string
- correct the shared option description to match those combinations

This PR is now the focused required-value portion of the former combined env/vars schema change.

## Split work

The scopes are independent and can merge in any order:

- this PR: environment required/value combinations
- #12793: environment directive aliases and options
- #12794: vars values, directives, and ignored-tools restrictions

## Validation

- `mise run render:schema`
- `mise run test:e2e e2e/config/test_schema_tombi`
- `mise run lint-fix`

Schema-only change; runtime behavior is unchanged.

## Gap provenance

- #10697 added required/redacted variable option behavior without fully equivalent schema coverage.

## Summary by CodeRabbit

- **Bug Fixes**
  - Tightened environment variable validation to prevent entries with a concrete value from also being marked as required.
  - Preserved support for required-only variables using `true` or a help message.
  - Added validation coverage for invalid combinations and updated schema fixtures.

### DIFF
--- a/e2e/config/test_schema_tombi
+++ b/e2e/config/test_schema_tombi
@@ -143,8 +143,9 @@ TOML
 # placeholder values while running tombi through `mise x`.
 cat >"$HOME/workdir/mise-age.toml" <<'TOML'
 [env]
-VALUE_OBJECT = { value = 123, tools = true, redact = true, required = "set VALUE_OBJECT" }
+VALUE_OBJECT = { value = 123, tools = true, redact = true, required = false }
 VALUE_DEFAULT = { default = 123, tools = true, redact = true }
+REQUIRED_HELP = { required = "set REQUIRED_HELP", tools = true }
 SECRET_SIMPLE = { age = "AGE-SECRET", tools = true }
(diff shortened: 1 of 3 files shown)


risu729:
Addressed the CodeRabbit comments: added a general `env_expand` def for the per-variable `expand` (noting it only takes effect for `_.file`-loaded files) while `env_file_expand` stays on the file directive, and replaced the empty `enum: []` on `vars.mise` with `not: {}` — the empty enum would have failed the repo's ajv strict lint. Both schemas regenerated.

Declining the Greptile comment about `{ age = "...", value = "x" }`: verified that the untagged parser matches the age variant and the flattened options **silently discard** `value` (no error, no warning) — the schema rejecting that combination is the safer contract, mirroring the existing `age`+`default` runtime check. Adding the missing `age`+`value` runtime rejection would be a separate behavior change.

risu729:
@coderabbitai resume

PR #12573, opened 2026-08-29
fix(file): stop create_dir_all reporting success when it created nothing

JamBalaya56562:
## The measurement

`std::fs::create_dir_all` can answer `Ok` and leave nothing behind. From a standalone rustc probe,
run on the exact path shape mise builds:

```
installs/jq/zzz   exists()=false  create_dir_all=Ok  after: is_dir()=true
installs/jq/aux   exists()=false  create_dir_all=Ok  after: is_dir()=true
installs/jq/nul   exists()=false  create_dir_all=Ok  after: is_dir()=false     ← nothing there
```

Everything downstream then works from a directory that is not there, and the failure surfaces
somewhere else entirely.

## How it reached the user

`create_install_dirs` makes three of those calls and then writes a marker file underneath —
`install_state::incomplete_file_path` is `<CACHE>/<short>/<version>/incomplete`, whose parent is the
`cache_path` from the call just above:

```rust
file::create_dir_all(tv.install_path())?;   // installs/jq/nul  -> Ok, nothing created
file::create_dir_all(tv.download_path())?;  // ditto
file::create_dir_all(tv.cache_path())?;     // <CACHE>/jq/nul   -> Ok, nothing created
File::create(self.incomplete_file_path(tv))?;   // parent missing -> os error 3, unwrapped
```

```console
$ mise install jq@nul
mise ERROR Failed to install aqua:jqlang/jq@nul: The system cannot find the path specified. (os error 3)
```

True, and useless: it named neither the path nor the operation, and the cause was three calls
earlier. The controls — `aux`, `prn`, `nulx`, `auxx`, `zzz` — all gave the ordinary
`no asset released`, so the shape of the error offered no hint either. A trace confirms `nul` and
`zzz` are identical up through aqua's package resolution and diverge only after it.

## The fix

`create_dir_all` now checks, after an attempt, that a directory is actually there — and
distinguishes two states, because saying the wrong one sends the reader somewhere useless:

- **something is in the way** — `symlink_metadata` succeeds: a file, or a link whose target is gone,
  which the `exists()` guard above answers `false` for
- **nothing was created** — the case above

Measured, so the branch is not a guess: for the `nul` path both `metadata` and `symlink_metadata`
fail with error 1, so it takes the second branch; a plain file takes the first.

**The check runs only when creation was attempted.** A path that already existed as a file still
returns `Ok`, exactly as before. This narrows the change to the case that was measured rather than
tightening a helper the whole codebase leans on.

When the final component is one Windows reserves for a device, the message says so. That is a check
rather than a guess — the component either is one of those names or it is not — but deliberately
**not** a prediction: **`aux` created a directory perfectly well** on the machine this was written
on, so the note is only ever added to a failure that already happened, and is never used to refuse a
name in advance.

`create_install_dirs` also switches its marker file from `File::create` to `file::create`, which
already creates the parent and wraps the failure with the path. That is the half that turned a
specific problem into an anonymous one.

### The empty path has to stay a no-op

`Path::new("plainstub").parent()` is `Some("")`, and `file::create` passes that straight through. std
answers `Ok` for an empty path and creates nothing — the exact shape the new check looks for. So the
guard is not decoration: without it every tool stub written to the working directory failed, which
is what the first CI run reported — **six Windows and eight Linux e2e cases**, all with
`failed create_dir_all: ` and no path at all. An explicit early return keeps it the no-op it has
always been, pinned by its own test.

## `nul` is the reproduction, not the subject

Nobody needs `nul` as a version. What this fixes is a helper that reported success for work it had
not done — reachable by any path component the filesystem will not take, with the failure landing
wherever the first user of that directory happens to be.

## Tests

The unit tests state the fix as **an invariant rather than a verdict per name**: whatever
`create_dir_all` answers has to match what is on disk afterwards. "Reserved names fail" would be a
claim about Windows that the `aux` measurement contradicts, so it is not asserted.

- Windows: `create_dir_all(<tmp>/nul)` must now be `Err`, and name both the path and the component
- Windows: `zzz`, `aux`, `con`, `prn`, `lpt1` — `is_ok()` must equal `is_dir()`, whichever way each
  one goes on the runner
- unix: an ordinary nested directory still gets created, since the new check runs everywhere
- both: the empty path is still a no-op — `create_dir_all("")` and the `Path::new("").join("")` that
  `parent()` produces

The Windows e2e drives `mise install jq@nul` and asserts the message names the path and the
component, with `jq@zzz` as the control — it creates its directories fine and is refused by the aqua
registry instead, so a change that rewrote every install failure would fail there.

## Not verified locally

The message assertions were matched against a string built from the format string, not against
output from a build containing this change; mise here is `2026.8.14`, which predates it. The real
output is what CI checks.

`create_dir_all` is used throughout, so the broadest regression surface is CI itself rather than any
one test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved installation errors when paths contain Windows-reserved names, such as `nul`.
  * Error messages now identify the affected path and clarify whether it conflicts with a file or reserved device name.
  * Preserved standard “no asset released” errors for ordinary missing versions.
  * Improved directory creation reliability across Windows and Unix systems, including clearer handling of invalid paths.

### DIFF
--- a/e2e-win/install_reserved_device_name.Tests.ps1
+++ b/e2e-win/install_reserved_device_name.Tests.ps1
@@ -1,0 +1,79 @@
+Describe 'an install directory Windows will not create' {
+    # `std::fs::create_dir_all` answers Ok for a path ending in `nul` and creates nothing --
+    # measured with a standalone rustc probe:
+    #
+    #     installs/jq/zzz   create_dir_all=Ok   after: is_dir()=true
+    #     installs/jq/aux   create_dir_all=Ok   after: is_dir()=true
+    #     installs/jq/nul   create_dir_all=Ok   after: is_dir()=false
+    #
+    # `create_install_dirs` made three such calls and then wrote a marker file underneath, so the
+    # failure surfaced from `File::create` -- three calls away from the cause, with no path and no
+    # operation, as a bare `(os error 3)`.
+    #
+    # This does not make `nul` a usable version. It makes the refusal say which path and why.
+
+    BeforeAll {
+        $script:OriginalDir = Get-Location
+        # All four, every time. A probe that isolated only some of these once wrote junctions into
+        # the real installs directory.
+        $script:Saved = @{}
+        foreach ($v in 'MISE_DATA_DIR', 'MISE_CONFIG_DIR', 'MISE_CACHE_DIR', 'MISE_TRUSTED_CONFIG_PATHS') {
+            $script:Saved[$v] = [Environment]::GetEnvironmentVariable($v, 'Process')
+        }
+
+        # Not named for a device itself: an earlier attempt at this used `nul` for the probe root
+        # and Windows refused to create it, which broke the isolation rather than the subject.
+        $script:Root = Join-Path $TestDrive 'devname'
+        $cfg = Join-Path $script:Root 'cfg'
+        $cache = Join-Path $script:Root 'cache'
+        $data = Join-Path $script:Root 'data'
+        $proj = Join-Path $script:Root 'proj'
+        New-Item -ItemType Directory -Path $cfg, $cache, $data, $proj -Force | Out-Null
+
+        $env:MISE_DATA_DIR = $data
+        $env:MISE_CONFIG_DIR = $cfg
+        $env:MISE_CACHE_DIR = $cache
+        $env:MISE_TRUSTED_CONFIG_PATHS = $script:Root
+        Set-Location $proj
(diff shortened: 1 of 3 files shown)


jdx:
Thanks for investigating this. I don't think we should add Windows device-name handling and additional filesystem checks to this widely used helper for this case. The extra metadata calls also add overhead while holding the global directory-creation lock. A narrowly scoped improvement to the marker-file error context would be preferable, but this change is too broad and complex as written.

PR #12582, opened 2026-08-29
fix(fish): split PATH on the host's separator, not always ':'

JamBalaya56562:
## What happens

Measured on **mise 2026.8.14 windows-x64**. The only config is `_.path = ["C:/aaa/bin", "C:/bbb/bin"]`.

```
$ mise env -s fish
set -gx PATH C '/aaa/bin;C' '/bbb/bin;C' '\Users\Jam\bin;C' '\Program Files\Git\mingw64\bin;C' ...
```

Every drive letter is severed from its path, and the rest of each path is glued to the next one. `mise hook-env -s fish` — what actually runs after `mise activate fish`, via `build_env_commands` — prints the same thing.

The other six shells are fine on the same config:

| shell | `mise env -s <shell>` |
|---|---|
| bash / zsh | `export PATH='C:/aaa/bin;C:/bbb/bin;C:\Users\...'` ✅ |
| pwsh | `${Env:PATH}='C:/aaa/bin;C:/bbb/bin;...'` ✅ |
| nu / xonsh / elvish | `;`-separated value intact ✅ |
| **fish** | as above ❌ |

## Why

`Shell::set_env` for fish:

```rust
// Fish uses space-separated list for PATH, not colon-separated string
if key == "PATH" {
    let paths = v.split(':').map(|p| escape(p.into())).join(" ");
```

Two things are wrong on Windows, where the separator is `;` and PATH is spelled however the environment spells it:

1. `v.split(':')` — the measurement above.
2. `key == "PATH"` — `env::PATH_KEY` is the spelling the process actually has, and on Windows that is often `Path`. Then the branch is skipped and PATH comes out as a scalar, which is not what fish keeps it as.

`prepend_env` and `move_prepend_env`, eight lines below in the same file, already do this correctly with `env::PATH_KEY` and `env::split_paths`. Only `set_env` was left behind.

## The change

`set_env` now uses `env::is_path_key` and `env::split_paths`, the two helpers that already exist for exactly this. **On unix both are the identity on what the code did before** — `split_paths` splits on `:`, and `is_path_key` reduces to `key == "PATH"` — so unix output does not move, with one deliberate exception below.

Empty entries are now dropped. That is a behaviour change on unix for a value like `/a::/b`. It is intentional and pinned by a test:

- an empty element of a fish list is the current directory, so this was putting the cwd on PATH;
- `prepend_env` right below already drops them, as does `env::split_colon_list`;
- a Windows PATH usually ends in `;`, and `std::env::split_paths` yields an empty entry for it. I measured that rather than assuming it:

```
"C:\\a\\bin;C:\\b\\bin"  -> ["C:\\a\\bin", "C:\\b\\bin"]  (n=2)
"C:\\a\\bin;"            -> ["C:\\a\\bin", ""]            (n=2)
"C:\\a\\bin;;C:\\b\\bin" -> ["C:\\a\\bin", "", "C:\\b\\bin"] (n=3)
```

`prepend_env` and `move_prepend_env` are untouched — they were already right.

## Tests

There were none for this branch: `test_set_env` covered `set_env("FOO", "1")` and nothing else, and fish's whole `mod tests` is `#[cfg(all(test, not(windows)))]`, so a Windows test could not go in it. Hence a sibling `mod windows_tests`.

- **windows** — a `;` list becomes one fish element per directory; the `Path` spelling takes the same branch; a trailing `;` does not add an empty element; and, as the control, a `;` in any other variable survives whole.
- **unix** — the colon list still produces the same output as before, `Path` is still *not* PATH there, and the empty-segment case is pinned.
- **`e2e-win/env_fish_path.Tests.ps1`** — the unit tests are on a pure function, so this runs the path the measurement above came from. It needs no fish installed. Controls: pwsh still joins PATH into one `;`-separated value, and a non-PATH variable holding a `;` is left alone.

## Limits

- **The `Path` spelling is not measured.** `PATH_KEY` resolved to `PATH` on the machine I tested, so I only saw the separator half break. The unit test pins the `is_path_key` branch; it is not a reproduction.
- fish does not run natively on Windows. This affects MSYS2/Cygwin fish being handed a `;`-separated Windows PATH by a native `mise.exe`, and anyone reading `mise env -s fish`. The internal inconsistency — one of three functions in the file using a different separator — stands on its own regardless.
- bash and zsh hardcode `":"` in their `prepend_env` too, which is visible in `mise activate --shims` on Windows. Different function, different controls; not in this PR. elvish had the same problem and is fixed separately in #12584.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `mise env` PATH handling for Fish on Windows, including correctly recognizing `Path` and semicolon-separated values.
  * Prevented empty PATH segments from unintentionally adding the current directory.
  * Preserved non-PATH variables containing semicolons.
  * Maintained expected colon-separated PATH behavior on Unix systems.

* **Tests**
  * Added coverage for Fish and PowerShell PATH output across Windows and Unix scenarios.

### DIFF
--- a/e2e-win/env_fish_path.Tests.ps1
+++ b/e2e-win/env_fish_path.Tests.ps1
@@ -1,0 +1,66 @@
+Describe 'fish PATH on Windows' {
+    BeforeAll {
+        $script:OriginalDir = Get-Location
+        # Saved here, restored in AfterAll: Pester runs every suite in one process, so removing an
+        # inherited value would leave the next suite without it.
+        $script:OriginalTrusted = [Environment]::GetEnvironmentVariable('MISE_TRUSTED_CONFIG_PATHS', 'Process')
+        $script:TestRoot = Join-Path $TestDrive ([System.Guid]::NewGuid().ToString())
+        New-Item -ItemType Directory -Path $script:TestRoot | Out-Null
+        $env:MISE_TRUSTED_CONFIG_PATHS = $script:TestRoot
+
+        # `_.path` is what puts a drive-lettered directory in front of PATH. `FOO` is the control:
+        # it holds a `;` too, but it is not PATH, so nothing may split it.
+        @(
+            '[env]'
+            '_.path = ["C:/aaa/bin"]'
+            'FOO = "a;b"'
+        ) | Out-File -Encoding ascii (Join-Path $script:TestRoot "mise.toml")
+
+        Set-Location $script:TestRoot
+
+        function Get-FishLine {
+            param([string]$Pattern)
+            (mise env -s fish | Select-String -Pattern $Pattern | Select-Object -First 1).Line
+        }
+    }
+
+    AfterAll {
+        Set-Location $script:OriginalDir
+        if ($null -eq $script:OriginalTrusted) {
+            Remove-Item Env:MISE_TRUSTED_CONFIG_PATHS -ErrorAction Ignore
+        } else {
+            [Environment]::SetEnvironmentVariable('MISE_TRUSTED_CONFIG_PATHS', $script:OriginalTrusted, 'Process')
+        }
+    }
+
+    It 'gives fish one list element per directory' {
+        # The reproduction. This printed `set -gx PATH C '/aaa/bin;C' '\Users\...;C' ...`: fish
+        # keeps PATH as a list, and the value was split on `:`, so every drive letter became an
+        # element of its own and the rest of that path was glued to the next one. A `;` surviving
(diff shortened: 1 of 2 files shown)


lhs-12:
@JamBalaya56562 Thank you for your PR. I use the Fish Shell via MSYS2 on Windows, and previously I had to use the workaround documented in https://github.com/jdx/mise/discussions/3961#discussioncomment-17034051. Because of this separator limitation, the workaround for Fish was quite complex. Based on this change, it seems like we might be able to finally put this issue to rest.

PR #12622, opened 2026-08-31
feat(core): Add nim core plugin

c22:
Add a `nim` core plugin that installs the Nim toolchain and its package manager (`nimble`) as a mise-managed core tool.

- **Windows (x86_64 / x86)**: downloads the official prebuilt zip from nim-lang.org (`nim-<v>_<arch>.zip`) and verifies it against the per-file `.sha256` sidecar (stored in the lockfile on first install).
- **Linux (x86_64 / i686)**: downloads the official prebuilt `nim-<v>-linux_<arch>.tar.xz` with the same checksum verification.
- **All other platforms** (macOS, Linux arm64, ...): Nim ships no prebuilt artifact, so the toolchain is built from source. The plugin clones the repo recursively and runs `build_all.sh`, the koch bootstrap that compiles the compiler from C sources and then builds `bin/nim`, `bin/nimble`, and the other tools. Requires a C compiler (`gcc`/`clang`) and `git` on PATH.

Adds `nim.compile` (env: `MISE_NIM_COMPILE`), mirroring the existing `node.compile` / `python.compile` / `ruby.compile` / `erlang.compile` pattern and honoring the global `all_compile`:

- `true` - always build from source
- `false` - require a prebuilt binary; errors on platforms without one
- unset - use the prebuilt binary when available, build from source otherwise (the default)

Source installs are recorded with `install = "source"` in the lockfile so locked installs reproduce the build method.

`_list_remote_versions` uses a fast `git ls-remote --tags` path (stripping the `v` prefix, filtering to semver releases) and falls back to GitHub tags with dates under `MISE_LIST_ALL_VERSIONS`.

- Unit tests for the prebuilt URL mapping (Linux/Windows filenames and arch coverage) and version validation.
- `e2e/backend/test_nim_install_slow` installs `nim@2.2.0` and asserts both `nim` and `nimble` run; it self-skips on platforms that have neither a prebuilt binary nor a C compiler.
- Verified end-to-end in a Linux x86_64 container: prebuilt install (checksum verified, `nim 2.2.0` + `nimble 0.16.1` on PATH), forced source build via `MISE_NIM_COMPILE=true` (recursive clone, koch bootstrap, working `nim`/`nimble`), and the e2e harness all pass.
- Verified on Windows: the `nim-2.2.0_x64.zip` layout matches the extraction/rename logic (top-level `nim-<v>/` with `bin/nim.exe` and `bin/nimble.exe` plus bundled DLLs), the sidecar checksum matches, and mise executes both binaries from the install path (`nim 2.2.0 [Windows: amd64]`, `nimble 0.16.1`).
- Docs: `docs/dev-tools/backends/nim.md` (incl. the `nim.compile` setting), backends index entry, and regenerated `schema/mise.json`.

## Summary by CodeRabbit

* **New Features**
  * Added support for installing and using Nim through mise.
  * Uses prebuilt Nim binaries where available, with source compilation fallback.
  * Added the `nim.compile` setting and `MISE_NIM_COMPILE` environment variable to control installation behavior.
  * Nim installations include Nimble and verify both tools after setup.

* **Documentation**
  * Added Nim backend documentation covering installation, configuration, platform support, and usage.

* **Tests**
  * Added end-to-end coverage for installing Nim 2.2.0 and verifying Nim and Nimble versions.

### DIFF
--- a/docs/dev-tools/backends/index.md
+++ b/docs/dev-tools/backends/index.md
@@ -20,6 +20,7 @@ Below is a list of the available backends in mise:
 - [go](/dev-tools/backends/go)
 - [http](/dev-tools/backends/http)
 - [npm](/dev-tools/backends/npm)
+- [nim](/dev-tools/backends/nim)
 - [pipx](/dev-tools/backends/pipx)
 - [pkgx](/dev-tools/backends/pkgx) <Badge type="warning" text="experimental" />
 - [s3](/dev-tools/backends/s3)
--- a/docs/dev-tools/backends/nim.md
+++ b/docs/dev-tools/backends/nim.md
@@ -1,0 +1,55 @@
+# Nim Backend
+
+You may install the [Nim](https://nim-lang.org) toolchain and its package manager
+(`nimble`) as a mise-managed core tool — no external plugin is required.
+
+The code for this is inside of the mise repository at
+[`./src/plugins/core/nim.rs`](https://github.com/jdx/mise/blob/main/src/plugins/core/nim.rs).
+
+## Dependencies
+
+- **Windows (x86_64 / x86):** the official prebuilt zip is downloaded from
+  nim-lang.org, so no extra tooling is required.
+- **Linux (x86_64 / i686):** the official prebuilt tarball is downloaded from
+  nim-lang.org, so no extra tooling is required.
+- **Other platforms (macOS, Linux arm64, Windows arm64, …):** Nim is built
+  from source, which requires `git` and a C compiler on PATH:
+  `gcc`/`clang` on Unix (`build_all.sh`, the koch bootstrap) or mingw `gcc`
+  on Windows (`build_all.bat`; Nim's csources bootstrap does not support
+  MSVC).
+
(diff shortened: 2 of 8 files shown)


c22:
@jdx when you have time it would be good to know why this was closed. Would it be better for me to start a discussion?

I probably should have mentioned that I'm aware of the asdf plugin for nim but this would make nim part of core instead and I have a separate branch adding support for the nimble package manage (similar to how cargo works).

I'm happy to address all the AI feedback if it helps.

risu729:
@c22 Please read https://mise.jdx.dev/contributing.html#adding-backends

c22:
Thanks @risu729. There are quite a few gaps with the asdf forked [mise-nim](https://github.com/mise-plugins/mise-nim) plugin compared to the work I've done here

I think I can port most of the functionality I have here except maybe the build from source functionality and lockfile support over to a plugin based architecture.

Ideally we would have access to more traits in the plugin system and I could write a more robust plugin.

eg.

BackendResolveLockInfo
BackendSecurityInfo
BackendOptionalDependencies

PR #12625, opened 2026-08-31
fix(github): stop picking a release's source.tar.gz as its binary

JamBalaya56562:
#12578 was closed with:

> seems like it could have false positives

Fair, so the rule is smaller. It matched three stems — `source`, `*-source`, `*_source`. It now matches **one exact stem**, `source`, which is the only one the measurement ever needed.

| asset name | #12578 | this PR |
| --- | --- | --- |
| `source.tar.gz`, `source.zip` | excluded | excluded |
| `v1.2.3-source.tar.gz` | excluded | **kept** |
| `tool_source.zip` | excluded | **kept** |
| `sourcery-2.2.7-macos-arm64.zip` | kept | kept |
| `opensource-tool-linux-x64.tar.gz` | kept | kept |
| `tool-source-map-linux.tar.gz` | kept | kept |

An asset whose name merely *ends in* "source" can plausibly be a real build. One called plainly `source` cannot — that is a release's own source archive, and it is what cargo-dist emits.

## Measured

```console
$ mise install libsql-server@latest        # Windows, isolated MISE_* dirs

.dockerignore  .env  .gitignore  .gitmodules  Cargo.lock  Cargo.toml
CODE_OF_CONDUCT.md  CONTRIBUTING.md         # a source checkout. nothing runnable
```

The release builds for linux and darwin only:

```
libsql-server-{aarch64,x86_64}-apple-darwin.tar.xz
libsql-server-{aarch64,x86_64}-unknown-linux-gnu.tar.xz
libsql-server-installer.sh   sqld.rb   dist-manifest.json
source.tar.gz
```

There is no Windows asset, and **"no asset for this platform" is a sentence mise already knows how to say** — `pick_best_asset` returning `None` reaches a choke point that reports it and lists what was available.

## Why it never got there

`pick_best_asset` keeps only assets scoring above zero. On a platform the release skipped:

| asset | os | format | total | |
| --- | --- | --- | --- | --- |
| `libsql-server-x86_64-unknown-linux-gnu.tar.xz` | **−100** (names another OS) | +11 | negative | dropped |
| `sqld.rb` | 0 | 0 (not an archive) | **0** | dropped — needs `> 0` |
| `libsql-server-installer.sh` | 0 | 0 | **0** | dropped |
| `source.tar.gz` | **0** (names no OS, so nothing subtracts) | **+10** (is an archive) | **10** | **picked** |

**Nothing subtracts for a source tarball.** Where real assets exist it loses to them by 100; where they don't, it is the last thing standing and **wins by elimination**.

`is_package_or_installer_asset` cannot catch this — it excludes by extension, and a source tarball shares `.tar.gz` with the real ones. The name is the only thing that separates them.

## The change

```rust
fn is_source_archive_asset(asset: &str) -> bool {
    asset_name_stem(asset) == "source"
}
```

added to the existing filter in `pick_best_asset`.

**Behaviour changes only where nothing else scores above zero.** Linux and macOS keep resolving `libsql-server` to exactly the asset they resolved before — asserted as a control, because the regression surface here is every github/forgejo/gitlab asset selection, not one tool.

## What this does not do

It does not make `libsql-server` usable on Windows; upstream ships nothing for it. It stops mise installing something that was never the tool and reporting it as a success.

`sqld.rb` and `libsql-server-installer.sh` scoring 0 is existing behaviour and is left alone.

The `libsql-server` registry entry is deliberately not touched: the mis-selection is general, and patching one entry would leave every release with the same asset layout picking a source tree.

## Tests

The release's asset list is a literal in the test, not fetched. A test that reads a third party's release list fails when that party changes it — which is what moved `test_tool_stub_basic` off `stylelint` the same week this was written.

- windows/x86_64 over that list → `None` (today: `Some("source.tar.gz")`)
- **control**: linux/x86_64 and macos/aarch64 over the same list still pick their real assets
- **control**: the table above, as an assertion — including the two names #12578 dropped and this one keeps

## Limits

A source archive named some other way — `<tool>-source.tar.gz`, say — is no longer excluded. That is the trade for a rule that cannot plausibly fire on a real asset.

## Summary by CodeRabbit

* **Bug Fixes**
  * Release asset selection now excludes source archives from platform-specific downloads.
  * Windows releases correctly show no downloadable asset when only a source archive is available.
  * Linux and macOS continue selecting their appropriate release archives.
  * Assets are excluded only when their name is exactly `source`.

### DIFF
--- a/src/backend/asset_matcher.rs
+++ b/src/backend/asset_matcher.rs
@@ -348,6 +348,7 @@ impl AssetPicker {
                 *score > 0
                     && !self.has_arch_mismatch(asset)
                     && !is_package_or_installer_asset(asset)
+                    && !is_source_archive_asset(asset)
             })
             .min_by(|(score_a, name_a), (score_b, name_b)| {
                 score_b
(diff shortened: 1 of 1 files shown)


jdx:
this should not be part of the backend

PR #12675, opened 2026-09-01
fix(shim): support shared executable directories

jdx:
## Summary
- identify mise shims directly instead of treating every executable in `shims_dir` as managed
- preserve unmanaged files and symlinks when publishing or pruning a shared shim directory
- mark generated plugin scripts while recognizing the legacy script format
- document `~/.local/bin` and `/usr/local/bin` as supported shim directories

## Tests
- `cargo test --bin mise mise_shim_detection`
- `cargo test --bin mise staged_shim_publication`
- `mise run lint`

---

> [!NOTE]
> **Medium Risk**
> Reshim and shim pruning logic changes how files in `shims_dir` are classified and deleted, which can affect users who pointed `MISE_SHIMS_DIR` at a shared PATH directory or relied on old `--force` wiping the whole farm.
> 
> **Overview**
> **`mise reshim`** can target shared bins like `~/.local/bin` without wiping unrelated executables. It now **identifies mise-owned shims** (generated script headers, native shim marker, symlink targets, legacy plugin layout) and only **replaces or prunes** those entries; unmanaged same-name files are left in place, with warnings when a collision would overwrite different content.
> 
> **Dedicated** default shim farms keep **full cleanup** on force rebuild; **shared** directories use conservative pruning and treat intentional name collisions as occupied so `mise doctor` and post-install reshims do not report false “missing shims.” **`--force`** is documented as rebuilding mise-owned shims rather than deleting everything in the directory first.
> 
> Generated shims add stable **“mise generated shim”** markers; **`mise-shim`** embeds a **native shim marker** byte sequence for Windows exe-shim fingerprinting. Docs and settings clarify that activation/hook-env still expect a **dedicated** `shims_dir`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 58ba6493845f8a694d83ab7ff2f1e11298666c57. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - `mise reshim` supports shared executable directories.
  - Native shims now include an ownership marker.

- **Bug Fixes**
  - Preserves unmanaged or externally modified executables.
  - Removes or replaces only recognized mise-owned shims.
  - `--force` now rebuilds mise-owned shims instead of removing all entries.
  - Improved handling of shims in symlinked directories.

- **Documentation**
  - Updated CLI help, man pages, and configuration guidance to describe safer shim handling.

### DIFF
--- a/docs/dev-tools/shims.md
+++ b/docs/dev-tools/shims.md
@@ -174,10 +174,12 @@ Note that `mise` already runs a reshim anytime a tool is installed/updated/remov
 `mise reshim` only creates/removes the shims. Some users sometimes use it as a
 "fix it" button, but it is only necessary if `~/.local/share/mise/shims` doesn't contain something it should.
 
-The configured shim directory can safely be a shared executable directory such as `~/.local/bin`
-or `/usr/local/bin`. `mise reshim` only replaces or removes entries it recognizes as mise shims.
-When an unmanaged file has the same name as a desired shim, mise keeps the existing file and prints
-a warning.
+For `mise reshim`, the configured shim directory may be a shared executable directory such as
+`~/.local/bin` or `/usr/local/bin`: reshim only replaces or removes entries it recognizes as mise
+shims, and leaves a same-named unmanaged file in place. Other mise features still identify shim
+directories as whole `PATH` entries, however, so a shared directory is not yet supported with
+`mise activate`, hook-env, or internal dependency lookups. Use a dedicated `shims_dir` if you use
+those features.
 
 ## Command wrappers
 
--- a/docs/dev-tools/shims.md
+++ b/docs/dev-tools/shims.md
@@ -174,7 +174,10 @@ Note that `mise` already runs a reshim anytime a tool is installed/updated/remov
 `mise reshim` only creates/removes the shims. Some users sometimes use it as a
 "fix it" button, but it is only necessary if `~/.local/share/mise/shims` doesn't contain something it should.
 
-Do not add additional executable in the `mise` directory, `mise` will delete them with the next reshim.
+The configured shim directory can safely be a shared executable directory such as `~/.local/bin`
+or `/usr/local/bin`. `mise reshim` only replaces or removes entries it recognizes as mise shims.
+When an unmanaged file has the same name as a desired shim, mise keeps the existing file and prints
+a warning.
 
 ## Command wrappers
 
--- a/src/cli/reshim.rs
+++ b/src/cli/reshim.rs
@@ -31,7 +31,7 @@ pub(crate) struct Reshim {
     #[usage(hide = true)]
     pub version: Option<String>,
 
-    /// Removes all shims before reshimming
+    /// Rebuilds all mise-owned shims
     #[usage(long, short)]
     pub force: bool,
 
--- a/src/shims.rs
+++ b/src/shims.rs
@@ -518,9 +518,26 @@ fn is_mise_shim(path: &Path, mise_bin: &Path) -> Result<bool> {
 
     #[cfg(windows)]
     {
-        // Native copy and hardlink shims do not carry an ownership marker.
-        // Retain the existing Windows shim-farm behavior.
-        return Ok(path.is_file());
+        if !path.is_file() {
+            return Ok(false);
+        }
+        let parent = path.parent().unwrap_or_else(|| Path::new("."));
+        let is_default_farm = file::storage_paths_eq(parent, &dirs::DATA.join("shims"))
+            || file::storage_paths_eq(parent, &env::MISE_SYSTEM_DATA_DIR.join("shims"));
+        if is_default_farm {
+            // Preserve the existing upgrade behavior in mise's dedicated
+            // farms. Native copies from an older mise cannot be identified by
+            // their contents after mise-shim.exe changes.
+            return Ok(true);
+        }
+
+        if is_generated_windows_file_shim(path) {
+            return Ok(true);
+        }
+        let matches_mise = files_identical(path, mise_bin).unwrap_or(false);
+        let matches_launcher = find_mise_shim_bin(mise_bin)
+            .is_some_and(|launcher| files_identical(path, &launcher).unwrap_or(false));
+        return Ok(matches_mise || matches_launcher);
     }
 
     #[cfg(not(windows))]
@@ -532,6 +549,29 @@ fn is_mise_shim(path: &Path, mise_bin: &Path) -> Result<bool> {
         Ok(contents.starts_with("#!/bin/sh\n# mise generated shim\n")
             || (contents.starts_with("#!/bin/sh\nexport ASDF_DATA_DIR=")
                 && contents.contains("\nmise x -- ")))
+    }
+}
+
+#[cfg(any(windows, test))]
+fn is_generated_windows_file_shim(path: &Path) -> bool {
+    let Some(name) = path.file_stem().and_then(|name| name.to_str()) else {
+        return false;
+    };
+    let Ok(contents) = fs::read_to_string(path) else {
+        return false;
+    };
+    if path
+        .extension()
+        .is_some_and(|ext| ext.eq_ignore_ascii_case("cmd"))
+    {
+        contents == windows_file_shim_body(name)
+    } else if path.extension().is_none() {
+        #[cfg(windows)]
+        return contents == bash_shim_script(name);
+        #[cfg(not(windows))]
+        return false;
+    } else {
+        false
     }
 }
 
@@ -1613,6 +1653,17 @@ mod tests {
         let body = windows_file_shim_body("gh");
         assert!(body.ends_with("\r\n"));
         assert_eq!(body.matches('\n').count(), body.matches("\r\n").count());
+    }
+
+    #[test]
+    fn windows_file_shim_detection_requires_the_exact_generated_body() {
+        let dir = tempfile::tempdir().unwrap();
+        let shim = dir.path().join("gh.cmd");
+        fs::write(&shim, windows_file_shim_body("gh")).unwrap();
+        assert!(is_generated_windows_file_shim(&shim));
+
+        fs::write(&shim, "@echo off\r\necho user script\r\n").unwrap();
+        assert!(!is_generated_windows_file_shim(&shim));
     }
 
     #[cfg(windows)]
--- a/src/shims.rs
+++ b/src/shims.rs
@@ -17,13 +17,15 @@ use crate::toolset::{ResolveOptions, ToolVersion, Toolset, ToolsetBuilder};
 use crate::{backend, dirs, env, fake_asdf, file};
 use color_eyre::eyre::{Result, bail, eyre};
 use eyre::WrapErr;
+#[cfg(windows)]
 use indoc::formatdoc;
 use itertools::Itertools;
 use path_absolutize::Absolutize;
 use tokio::task::JoinSet;
 
 #[cfg(any(windows, test))]
-const NATIVE_SHIM_MARKER: &[u8] = b"mise generated native shim v1";
+const NATIVE_SHIM_MARKER: &[u8] = include_bytes!("../crates/mise-shim/native-shim-marker");
+const GENERATED_SHELL_SHIM_HEADER: &str = "#!/bin/sh\n# mise generated shim\n";
 
 // executes as if it was a shim if the command is not "mise", e.g.: "node"
 pub(crate) async fn handle_shim() -> Result<()> {
@@ -344,15 +346,19 @@ pub(crate) async fn reshim_for(
     let full_rebuild = force || shim_mode_changed || shim_version_changed;
     file::create_dir_all(&shims_dir)?;
 
-    // Always materialize the complete farm first. Besides keeping publication
-    // atomic per shim, this gives us the exact content used to decide whether a
-    // same-named entry in a shared directory is ours or a collision.
-    let desired = get_desired_shims(config, &mise_bin, ts, scope).await?;
+    let (desired, shims_to_stage) = if full_rebuild {
+        let desired = get_desired_shims(config, &mise_bin, ts, scope).await?;
+        let shims_to_stage = desired.iter().cloned().collect();
+        (desired, shims_to_stage)
+    } else {
+        let diffs = get_shim_diffs(config, &mise_bin, ts, &shims_dir, scope).await?;
+        (diffs.desired, diffs.missing)
+    };
     let staging = stage_shim_farm(
         &shims_dir,
         &mise_bin,
         scope,
-        &desired.into_iter().collect::<BTreeSet<_>>(),
+        &shims_to_stage,
         &shim_mode,
         shim_version,
     )
@@ -361,6 +367,7 @@ pub(crate) async fn reshim_for(
         &shims_dir,
         &mise_bin,
         staging,
+        desired,
         is_dedicated_shims_dir(&shims_dir),
     )?;
 
@@ -437,6 +444,7 @@ fn publish_staged_shim_farm(
     shims_dir: &Path,
     mise_bin: &Path,
     staging: tempfile::TempDir,
+    mut desired: HashSet<String>,
     prune_unmanaged: bool,
 ) -> Result<()> {
     let mut entries = staging
@@ -453,16 +461,17 @@ fn publish_staged_shim_farm(
     // the next reshim retry instead of treating a partially published farm as
     // current.
     entries.sort_by_key(|entry| is_hidden_shim_name(&entry.file_name()));
-    let desired = entries
-        .iter()
-        .filter_map(|entry| entry.file_name().into_string().ok())
-        .collect::<HashSet<_>>();
+    desired.extend(
+        entries
+            .iter()
+            .filter_map(|entry| entry.file_name().into_string().ok()),
+    );
     for entry in entries {
         let source = entry.path();
         let destination = shims_dir.join(entry.file_name());
         if (destination.exists() || destination.is_symlink())
             && !is_mise_shim(&destination, mise_bin)?
-            && !files_identical(&source, &destination)?
+            && !files_identical(&source, &destination).unwrap_or(false)
         {
             warn!(
                 "not replacing unmanaged file in shims directory: {}",
@@ -506,8 +515,13 @@ fn publish_staged_shim_farm(
 }
 
 fn is_dedicated_shims_dir(path: &Path) -> bool {
-    file::storage_paths_eq(path, &dirs::DATA.join("shims"))
-        || file::storage_paths_eq(path, &env::MISE_SYSTEM_DATA_DIR.join("shims"))
+    matches_unredirected_dedicated_dir(path, &dirs::DATA.join("shims"))
+        || matches_unredirected_dedicated_dir(path, &env::MISE_SYSTEM_DATA_DIR.join("shims"))
+}
+
+fn matches_unredirected_dedicated_dir(path: &Path, dedicated: &Path) -> bool {
+    file::paths_eq(path, dedicated)
+        && fs::canonicalize(path).is_ok_and(|resolved| file::paths_eq(&resolved, path))
 }
 
 fn files_identical(a: &Path, b: &Path) -> Result<bool> {
@@ -585,6 +599,7 @@ fn is_mise_shim(path: &Path, mise_bin: &Path) -> Result<bool> {
     }
 }
 
+#[cfg(any(windows, test))]
 fn bytes_contain(haystack: &[u8], needle: &[u8]) -> bool {
     !needle.is_empty()
         && haystack
@@ -593,9 +608,25 @@ fn bytes_contain(haystack: &[u8], needle: &[u8]) -> bool {
 }
 
 fn is_generated_shell_shim(contents: &[u8]) -> bool {
-    contents.starts_with(b"#!/bin/sh\n# mise generated shim\n")
-        || (contents.starts_with(b"#!/bin/sh\nexport ASDF_DATA_DIR=")
-            && bytes_contain(contents, b"\nmise x -- "))
+    contents.starts_with(GENERATED_SHELL_SHIM_HEADER.as_bytes()) || is_legacy_plugin_shim(contents)
+}
+
+fn is_legacy_plugin_shim(contents: &[u8]) -> bool {
+    let Ok(contents) = std::str::from_utf8(contents) else {
+        return false;
+    };
+    let mut lines = contents.lines();
+    lines.next() == Some("#!/bin/sh")
+        && lines
+            .next()
+            .is_some_and(|line| line.starts_with("export ASDF_DATA_DIR=") && line.len() > 21)
+        && lines
+            .next()
+            .is_some_and(|line| line.starts_with("export PATH=\"") && line.ends_with(":$PATH\""))
+        && lines
+            .next()
+            .is_some_and(|line| line.starts_with("mise x -- ") && line.ends_with(" \"$@\""))
+        && lines.next().is_none()
 }
 
 #[cfg(any(windows, test))]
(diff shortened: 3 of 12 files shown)


src/shims.rs
```diff
@@ -496,6 +492,49 @@ fn publish_staged_shim_farm(shims_dir: &Path, staging: tempfile::TempDir) -> Res
...
+        let target = fs::read_link(path)?;
+        let target = if target.is_absolute() {
+            target
+        } else {
+            path.parent().unwrap_or_else(|| Path::new(".")).join(target)
+        };
+        return Ok(file::paths_eq(
+            &file::canonicalize_or_self(&target),
+            &file::canonicalize_or_self(mise_bin),
+        ));
```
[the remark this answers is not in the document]
jdx: Fixed in 0d7e7ac57: filenames recognized as mise binaries are explicitly excluded from shim ownership, so a real `mise` dispatcher symlink in a shared directory is never pruned.

src/shims.rs
```diff
@@ -464,9 +444,23 @@ fn publish_staged_shim_farm(shims_dir: &Path, staging: tempfile::TempDir) -> Res
...
         let destination = shims_dir.join(entry.file_name());
+        if (destination.exists() || destination.is_symlink())
+            && !is_mise_shim(&destination, mise_bin)?
+            && !files_identical(&source, &destination)?
+        {
+            warn!(
+                "not replacing unmanaged file in shims directory: {}",
+                display_path(&destination)
+            );
+            continue;
```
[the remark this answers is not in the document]
jdx: Fixed in 0d7e7ac57: symlinks whose target basename identifies a mise binary are recognized even when the old target is dangling, while the dispatcher-name exclusion still protects the real `mise` link.

src/shims.rs
```diff
@@ -464,9 +444,23 @@ fn publish_staged_shim_farm(shims_dir: &Path, staging: tempfile::TempDir) -> Res
...
+        }
```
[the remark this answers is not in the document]
jdx: Fixed in 0d7e7ac57: Windows publication again calls `remove_shim_with_rename_fallback` after the ownership/collision check and before publishing the staged replacement.

src/shims.rs
```diff
@@ -496,6 +492,89 @@ fn publish_staged_shim_farm(shims_dir: &Path, staging: tempfile::TempDir) -> Res
...
+            return Ok(true);
+        }
+
+        if is_generated_windows_file_shim(path) {
+            return Ok(true);
+        }
+        let matches_mise = files_identical(path, mise_bin).unwrap_or(false);
+        let matches_launcher = find_mise_shim_bin(mise_bin)
+            .is_some_and(|launcher| files_identical(path, &launcher).unwrap_or(false));
+        return Ok(matches_mise || matches_launcher);
```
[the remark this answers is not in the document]
jdx: Fixed in 9530530cf without a manifest: new native shims contain a stable embedded marker, and conservative legacy fingerprints recognize pre-marker `mise.exe` and `mise-shim.exe` copies so the first upgrade can replace them safely.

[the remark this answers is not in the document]
jdx: Fixed in 9530530cf: the generated shell-shim marker and legacy plugin-script shape are now recognized on Windows as well as Unix, so plugin shims remain replaceable and prunable.

src/shims.rs
```diff
@@ -341,43 +344,20 @@ pub(crate) async fn reshim_for(
...
+    let staging = stage_shim_farm(
+        &shims_dir,
+        &mise_bin,
+        scope,
+        &desired.into_iter().collect::<BTreeSet<_>>(),
+        &shim_mode,
+        shim_version,
+    )
+    .await?;
+    publish_staged_shim_farm(&shims_dir, &mise_bin, staging)?;
```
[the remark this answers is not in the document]
jdx: Fixed in 227dcf255: non-forced reshims now compute shim diffs and stage only missing or stale tool shims, while the publisher receives the full desired set so unstaged current shims are retained. Forced, mode-change, and version-change rebuilds still stage the complete farm.\n\n*AI-assisted — Tool: Codex; model: openai/gpt-5; version: unavailable.*

src/shims.rs
```diff
@@ -464,9 +447,26 @@ fn publish_staged_shim_farm(shims_dir: &Path, staging: tempfile::TempDir) -> Res
...
+        if (destination.exists() || destination.is_symlink())
+            && !is_mise_shim(&destination, mise_bin)?
+            && !files_identical(&source, &destination)?
+        {
+            warn!(
+                "not replacing unmanaged file in shims directory: {}",
+                display_path(&destination)
+            );
+            continue;
+        }
```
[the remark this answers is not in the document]
jdx: Fixed in 227dcf255: an identity-read failure is treated as non-identical, so an unreadable unrecognized collision follows the existing warning-and-skip path instead of aborting publication.\n\n*AI-assisted — Tool: Codex; model: openai/gpt-5; version: unavailable.*

src/shims.rs
```diff
@@ -496,6 +505,137 @@ fn publish_staged_shim_farm(shims_dir: &Path, staging: tempfile::TempDir) -> Res
     Ok(())
 }
 
+fn is_dedicated_shims_dir(path: &Path) -> bool {
+    file::storage_paths_eq(path, &dirs::DATA.join("shims"))
+        || file::storage_paths_eq(path, &env::MISE_SYSTEM_DATA_DIR.join("shims"))
+}
```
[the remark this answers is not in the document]
jdx: Fixed in 227dcf255: broad pruning now requires both the default path spelling and an unredirected canonical path. A default farm symlinked into a shared directory is treated conservatively, with a Unix regression test covering the case.\n\n*AI-assisted — Tool: Codex; model: openai/gpt-5; version: unavailable.*

src/shims.rs
```diff
@@ -496,6 +514,159 @@ fn publish_staged_shim_farm(shims_dir: &Path, staging: tempfile::TempDir) -> Res
...
 }
 
+fn is_dedicated_shims_dir(path: &Path) -> bool {
+    matches_unredirected_dedicated_dir(path, &dirs::DATA.join("shims"))
+        || matches_unredirected_dedicated_dir(path, &env::MISE_SYSTEM_DATA_DIR.join("shims"))
+}
+
+fn matches_unredirected_dedicated_dir(path: &Path, dedicated: &Path) -> bool {
+    file::paths_eq(path, dedicated)
+        && fs::canonicalize(path).is_ok_and(|resolved| file::paths_eq(&resolved, path))
```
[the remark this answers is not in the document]
jdx: Fixed in daaf6881e: the redirect guard now uses `dunce::canonicalize`, which strips Windows verbatim-path prefixes before the platform-aware comparison while still resolving actual symlink/junction redirects away from the dedicated path. The regression test and full lint check pass.\n\n*AI-assisted — Tool: Codex; model: openai/gpt-5; version: unavailable.*

src/shims.rs
```diff
@@ -496,6 +514,169 @@ fn publish_staged_shim_farm(shims_dir: &Path, staging: tempfile::TempDir) -> Res
...
+            return Ok(false);
+        }
+        let parent = path.parent().unwrap_or_else(|| Path::new("."));
+        let is_default_farm = file::storage_paths_eq(parent, &dirs::DATA.join("shims"))
+            || file::storage_paths_eq(parent, &env::MISE_SYSTEM_DATA_DIR.join("shims"));
+        if is_default_farm {
+            // Preserve the existing upgrade behavior in mise's dedicated
+            // farms. Native copies from an older mise cannot be identified by
+            // their contents after mise-shim.exe changes.
+            return Ok(true);
```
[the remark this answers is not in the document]
jdx: Fixed in c095d2134. Windows ownership now uses the same `is_dedicated_shims_dir` predicate as pruning, so a default path redirected through a junction/symlink is treated as shared in both places and unrelated regular files must pass direct shim fingerprint/identity checks. The shim unit suite and `mise run lint-fix` pass locally.

docs/dev-tools/shims.md
```diff
@@ -174,7 +174,10 @@ Note that `mise` already runs a reshim anytime a tool is installed/updated/remov
 `mise reshim` only creates/removes the shims. Some users sometimes use it as a
 "fix it" button, but it is only necessary if `~/.local/share/mise/shims` doesn't contain something it should.
 
-Do not add additional executable in the `mise` directory, `mise` will delete them with the next reshim.
+The configured shim directory can safely be a shared executable directory such as `~/.local/bin`
```
jdx: **Docs claim is ahead of the implementation.** `file::is_mise_shims_dir` matches the *whole* configured directory, and it is used everywhere mise wants to avoid its own shims:

- `mise activate` calls `remove_shims_from_path()` when `not_found_auto_install` is false (`src/cli/activate.rs:309`), and hook-env skips the dir on every prompt (`src/cli/hook_env.rs:338`). With `shims_dir = ~/.local/bin`, every non-mise executable there (pipx/uv tools, the `~/.local/bin/mise` launcher itself) disappears from activated shells.
- `which_no_shims` / `path_env_without_shims` / `strip_shims_from_path` (`src/file.rs:1812,1860`) and `remove_dependency_shims` (`src/backend/mod.rs:4045`) hide the dir from `uv` for `uv_venv_auto`, from the `gh auth token` credential command, and from dependency tools during installs.
- `mise_bin_for_shims` uses `which_no_shims("mise")` (`src/shims.rs:759`), so with `shims_dir = /usr/local/bin` it skips the stable `/usr/local/bin/mise` launcher and points every shim at `current_exe`; the next package upgrade dangles them.

Only reshim was fixed. I'd either hold this paragraph (and the matching text in `docs/directories.md` and the `shims_dir` setting docs) until dispatch-dir handling filters per entry rather than per directory, or say explicitly that shared dirs are supported for reshim only.

src/shims.rs
```diff
@@ -464,9 +461,27 @@ fn publish_staged_shim_farm(shims_dir: &Path, staging: tempfile::TempDir) -> Res
...
     entries.sort_by_key(|entry| is_hidden_shim_name(&entry.file_name()));
+    desired.extend(
+        entries
+            .iter()
+            .filter_map(|entry| entry.file_name().into_string().ok()),
+    );
     for entry in entries {
         let source = entry.path();
         let destination = shims_dir.join(entry.file_name());
+        if (destination.exists() || destination.is_symlink())
```
jdx: **A same-named unmanaged file blocks its shim forever, even with `--force`.** This guard ignores `prune_unmanaged` and runs identically on the forced and incremental paths.

(a) Dedicated farm on unix: a stray regular file `~/.local/share/mise/shims/node` (interrupted write, user copy) fails `is_mise_shim` (no header), `files_identical(symlink, file)` is false, so we warn and `continue`. Pre-PR `mise reshim --force` renamed over it unconditionally; now nothing can, while the prune loop below still deletes *other* unmanaged files in that same farm. On Windows the `is_dedicated_shims_dir` early return in `is_mise_shim` makes the same file replaceable, so the platforms diverge.

(b) Shared dir: a pipx-installed `~/.local/bin/black` alongside `pipx:black` in mise fails `is_current_mise_shim`, so it lands in `diffs.missing` on every reshim. Every `mise install` re-stages it and emits `warn!` (not `warn_once!`), and `mise doctor` raises the hard error "shims are missing, run mise reshim" (`src/cli/doctor/mod.rs:829`) that reshim can never clear.

Suggestion: when `prune_unmanaged` (or `force`) is set, replace the collision as before; otherwise exclude skipped names from `missing` so doctor and the warning don't repeat forever.

src/shims.rs
```diff
@@ -464,9 +461,27 @@ fn publish_staged_shim_farm(shims_dir: &Path, staging: tempfile::TempDir) -> Res
...
+            && !is_mise_shim(&destination, mise_bin)?
+            && !files_identical(&source, &destination).unwrap_or(false)
+        {
+            warn!(
+                "not replacing unmanaged file in shims directory: {}",
```
jdx, line 506: **Windows, shared dir: `.mode`/`.version` can never be updated.** The metadata files staged by `write_shim_metadata` go through this same guard. In a non-dedicated `MISE_SHIMS_DIR` on Windows, `is_mise_shim(<dir>\.version)` is false (not a symlink, parent not dedicated, the extensionless branch compares against `bash_shim_script(".version")`, no fingerprint, not identical to `mise.exe`), and once the version string differs `files_identical` is false too, so the new marker is skipped with the unmanaged-file warning.

The first publish works because the destination doesn't exist yet. After the first mise upgrade, the live `.version` is stuck at the old value, `shim_version_stale` is true on every subsequent reshim, and every `mise install` does a full rebuild plus a spurious warning. Same for `.mode` after a `windows_shim_mode` change. `write_shim_metadata` only ever writes into staging, so nothing else refreshes these.

Fix: treat hidden metadata names (`is_hidden_shim_name`) as always mise-owned in this guard.

src/shims.rs
```diff
@@ -496,6 +514,169 @@ fn publish_staged_shim_farm(shims_dir: &Path, staging: tempfile::TempDir) -> Res
...
+        return Ok(false);
+    }
+    if path.is_symlink() {
+        let target = fs::read_link(path)?;
+        let target = if target.is_absolute() {
+            target
+        } else {
+            path.parent().unwrap_or_else(|| Path::new(".")).join(target)
+        };
+        let target_names_mise = target
```
jdx: **Over-broad ownership: any symlink whose *target basename* is `mise`, `mise.*` or `mise-*` is claimed and pruned.** `env::is_mise_binary` is the argv[0] dispatch heuristic, not an ownership proof. In a shared `~/.local/bin`, a user's `mw -> ~/scripts/mise-wrapper.sh` or `oldmise -> /opt/mise-2025.1/mise` is not in `desired`, passes `list_shims_in` (extensionless symlink), returns true here, and gets `file::remove_all`'d by the prune loop on the next `mise install`. That contradicts the new docs promise that only recognized mise shims are removed.

The dangling-shim repair case the test covers only needs the canonical-target-equals-`mise_bin` half plus a target basename of exactly `mise`; I'd drop the prefix matching here.

src/shims.rs
```diff
@@ -920,14 +1101,29 @@ async fn get_actual_shims(mise_bin: impl AsRef<Path>, shims_dir: &Path) -> Resul
 
     Ok(list_shims_in(shims_dir)?
         .into_iter()
-        .filter(|bin| {
-            let path = shims_dir.join(bin);
-
-            !path.is_symlink() || path.read_link().is_ok_and(|p| p == mise_bin)
-        })
+        .filter(|bin| is_current_mise_shim(&shims_dir.join(bin), mise_bin).unwrap_or(false))
```
jdx: **Doctor's unused-shims list now diverges from what prune deletes, and incremental pruning widened.** Pre-PR `get_actual_shims` kept `!is_symlink || read_link == mise_bin`, so a regular non-shim file in the default farm showed up in `diffs.extra` (doctor: "Unused shims") and only that set was removed on an incremental reshim; foreign symlinks like `python3 -> /usr/bin/python3` survived until `--force`.

Now `is_current_mise_shim` drops unmanaged regular files, stale `bar -> /old/path/mise` links, and foreign symlinks from `extra`, so `mise doctor` (`src/cli/doctor/mod.rs:836`) is silent about all of them, yet the prune loop at line 502 deletes all three in the default farm on the next install-triggered reshim. The comment "dedicated farms retain their historical full cleanup" describes the old `--force` path, not the old incremental path. Either doctor should report what prune will remove, or incremental pruning should stay scoped to `diffs.extra`.

src/shims.rs
```diff
@@ -920,14 +1101,29 @@ async fn get_actual_shims(mise_bin: impl AsRef<Path>, shims_dir: &Path) -> Resul
...
+    if !path.is_symlink() {
+        return Ok(true);
+    }
+    let target = fs::read_link(path)?;
+    let target = if target.is_absolute() {
+        target
+    } else {
+        path.parent().unwrap_or_else(|| Path::new(".")).join(target)
+    };
+    Ok(file::paths_eq(
```
jdx: **Canonicalized comparison stops migrating shims to the stable launcher.** `mise_bin_for_shims` deliberately targets the PATH-visible launcher (#1222: "homebrew puts the version in the path"). Pre-PR, a shim pointing at a versioned real path (written when `which_no_shims("mise")` failed during an auto-reshim, which happens on Linuxbrew/Nix where `current_exe` resolves symlinks) mismatched `read_link == mise_bin`, landed in `missing`, and was rewritten to the launcher on the next reshim. Now it canonicalizes to the same file, counts as current, and stays versioned until `brew upgrade` removes the old Cellar and every shim dangles until the next reshim.

The window is bounded (the dangling link is later repaired via the target-name check), but the proactive migration is lost. Consider keeping the raw `read_link == mise_bin` fast path as the "current" test and using canonical equality only for ownership.

src/shims.rs
```diff
@@ -496,6 +514,169 @@ fn publish_staged_shim_farm(shims_dir: &Path, staging: tempfile::TempDir) -> Res
...
+#[cfg(any(windows, test))]
+fn is_generated_windows_file_shim_contents(path: &Path, contents: &[u8]) -> bool {
+    let Some(name) = path.file_stem().and_then(|name| name.to_str()) else {
+        return false;
+    };
+    if path
+        .extension()
+        .is_some_and(|ext| ext.eq_ignore_ascii_case("cmd"))
+    {
+        contents == windows_file_shim_body(name).as_bytes()
```
jdx: **Windows file-mode shims have no stable ownership marker** (plausible rather than confirmed). `.cmd` and bash shims are recognized only by byte-exact equality with the *current* template. Today they also happen to match the "transition" arm of `has_mise_native_shim_fingerprint` (`__MISE_SHIM_PATH` + `recursive shim invocation detected` + `mise x --`), but that arm is labelled temporary, uses hardcoded literals rather than `env::MISE_SHIM_PATH_ENV`, and no test asserts `has_mise_native_shim_fingerprint(windows_file_shim_body(..))`. Pre-#10982 bodies (`@echo off / setlocal / mise x -- foo %*`) already match nothing, so in a shared dir they are re-staged, warned about, and never replaced or pruned.

`is_generated_launcher` / `WINDOWS_LAUNCHER_MARKER` in `src/cli/generate/mod.rs` already documents why a stable marker line beats exact-body equality. A `rem mise generated shim` line in the `.cmd` body and `# mise generated shim` in the bash body, mirroring `GENERATED_SHELL_SHIM_HEADER`, would make this robust to the next template change.

src/shims.rs
```diff
@@ -496,6 +514,169 @@ fn publish_staged_shim_farm(shims_dir: &Path, staging: tempfile::TempDir) -> Res
...
+            .is_some_and(|launcher| files_identical(path, &launcher).unwrap_or(false));
+        return Ok(matches_mise || matches_launcher);
+    }
+
+    #[cfg(not(windows))]
+    {
+        if !path.is_file() {
+            return Ok(false);
+        }
+        Ok(is_generated_shell_shim(&fs::read(path).unwrap_or_default()))
```
jdx: **Efficiency: every file in a shared shims dir is read in full, twice per reshim.** `fs::read(path)` here tests a 32-byte header. It runs once per entry in `get_actual_shims` and again per non-desired entry in the prune loop (`prune_unmanaged` is false for shared dirs), so a `~/.local/bin` with a dozen 50–150 MB static binaries costs multiple GB of reads on every `mise install`. On Windows each unmanaged `.exe` is read ~6 times (up to 7 naive `windows()` scans, then `files_identical` against `mise.exe` and `mise-shim.exe` with no length pre-check), plus a `find_mise_shim_bin` PATH walk and two `dunce::canonicalize` calls per file.

Cheap wins: bounded read (first 4 KiB, or reject by `metadata().len()`); compare lengths before contents in `files_identical`; resolve `find_mise_shim_bin` / `is_dedicated_shims_dir` once per reshim; reuse the `get_actual_shims` classification instead of re-scanning; `memchr::memmem` or `aho-corasick` (already in the dep graph) instead of `windows()`.

src/shims.rs
```diff
@@ -920,14 +1101,29 @@ async fn get_actual_shims(mise_bin: impl AsRef<Path>, shims_dir: &Path) -> Resul
...
-        .filter(|bin| {
-            let path = shims_dir.join(bin);
-
-            !path.is_symlink() || path.read_link().is_ok_and(|p| p == mise_bin)
-        })
+        .filter(|bin| is_current_mise_shim(&shims_dir.join(bin), mise_bin).unwrap_or(false))
         .collect::<HashSet<_>>())
 }
 
+fn is_current_mise_shim(path: &Path, mise_bin: &Path) -> Result<bool> {
```
jdx: Nit: this symlink-target resolution (read_link → join to parent → `canonicalize_or_self` → `paths_eq`) duplicates the block in `is_mise_shim` (lines 557–573), and re-does the `read_link` that `is_mise_shim` just performed. `file::same_file` / `file::is_symlink_to` already cover the canonical comparison. One shared helper would keep `get_actual_shims` and `publish_staged_shim_farm` from disagreeing about ownership if either copy is later fixed.

src/cli/reshim.rs
```diff
@@ -31,7 +31,7 @@ pub(crate) struct Reshim {
     #[usage(hide = true)]
     pub version: Option<String>,
 
-    /// Removes all shims before reshimming
+    /// Rebuilds all mise-owned shims
```
jdx, line 34: Nit (conventions): the PR title and all ten commits use the `shims` scope; AGENTS.md lists the subsystem scope as `shim`. Easy to fix at squash-merge with `fix(shim): ...`.

jdx (commented):
Review of the shared-shim-directory change. Ten findings below as inline comments, most severe first. Seven are confirmed against the code, one is plausible, two are cleanup/conventions.

Summary: the reshim-side ownership logic is careful, but the headline docs claim does not hold yet — the rest of mise still treats `dirs::shims()` as a dispatch-only directory and strips it from PATH. There are also two regressions in the dedicated farm (`--force` can no longer displace a stray same-named file; doctor's unused-shims list no longer matches what prune deletes), a Windows bug where `.mode`/`.version` can never be updated in a shared dir, and an over-broad symlink ownership rule.

Unit tests in `shims::tests` pass locally (26 passed).

src/shims.rs
jdx: Current symlink detection now uses raw resolved-target path equality, while canonical equality remains ownership evidence. This restores proactive migration from a versioned executable path to the stable package-manager launcher.

src/shims.rs
jdx: Added stable generated markers to both Windows `.cmd` and extensionless bash file-mode shims, retained exact-current detection, and added recognition for the historical minimal cmd shim body. Tests cover marker-based future bodies and the legacy body.

src/shims.rs
jdx: Doctor and incremental pruning now use the same computed `extra` set. Dedicated farms retain the historical classification of regular files and raw-current symlinks; foreign and stale symlinks stay outside incremental pruning, while full rebuilds retain broad dedicated-farm cleanup.

src/cli/reshim.rs
jdx: Updated the PR title and follow-up commit to the repository convention: `fix(shim): ...`.

docs/dev-tools/shims.md
jdx: Scoped the documentation to `mise reshim` coexistence and explicitly documented that activation, hook-env, and internal dependency lookups still require a dedicated shim directory because they filter whole PATH entries.

src/shims.rs
jdx: Restored full-rebuild replacement in unredirected dedicated farms. In shared directories, an unmanaged same-named entry now counts as occupied, so automatic reshims do not restage or warn repeatedly and doctor no longer reports an unrepairable missing shim; even `--force` remains conservative there.

src/shims.rs
jdx: Ownership classification is now reused between diffing, publication, and pruning instead of rescanning entries. Script inspection is bounded to 16 KiB, and byte identity checks compare file lengths before reading contents. Native fingerprint scans remain available for upgrade ownership but no longer repeat during an incremental reshim.

src/shims.rs
jdx: Hidden metadata entries are now always treated as mise-owned during publication, so `.mode` and `.version` update in shared Windows directories.

src/shims.rs
jdx: Factored symlink target resolution into `resolved_symlink_target`; ownership and current-target checks now share it and avoid duplicate read-link logic.

src/shims.rs
jdx: Narrowed general symlink ownership to canonical identity with the current mise binary. A target basename of exactly `mise` is now used only while publishing a staged desired shim, allowing dangling desired shims to heal without pruning unrelated `oldmise` or wrapper symlinks.

src/shims.rs
```diff
@@ -888,6 +1161,51 @@ pub(crate) struct ShimDiffs {
...
+    } else {
+        (
+            desired
+                .iter()
+                .filter(|name| {
+                    !actual.current.contains(*name)
+                        && (!actual.occupied.contains(*name) || actual.repairable.contains(*name))
+                })
+                .cloned()
+                .collect(),
```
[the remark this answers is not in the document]
jdx: Fixed in 58ba64938. Shared-directory diffing now stages desired shims that are positively mise-owned but do not point at the current launcher, while unmanaged same-name collisions remain untouched. The launcher migration test now runs through get_actual_shims and calculate_shim_diffs to cover the on-disk versioned-target case. Focused shim tests (32) and mise run lint-fix pass.\n\n*AI-assisted — Tool: Codex; model: openai/gpt-5; version: unavailable.*

PR #12686, opened 2026-09-01
fix(brew-cask): keep the extracted payload beside a stage-sourced binary

azohra:
## Problem

A cask can declare one binary artifact and still ship a package layout around it. `codex` is the case I hit, reported in #12684:

```ruby
binary "bin/codex"
```

```
bin/codex
bin/codex-code-mode-host
codex-package.json
codex-path/rg
codex-resources/zsh/bin/zsh
```

`stage_binary` copies `bin/codex` into the caskroom on its own, and `file::remove_all(stage)` takes the rest. The launcher arrives without the helper it execs or the manifest naming its resource directories, so it falls back to a `codex-code-mode-host` from an unrelated install — an older one, in my case — and every tool call fails on a protocol field the newer CLI requires:

```text
failed to decode code-mode IPC frame: missing field `code_mode_host_duration_ns`
```

Nothing reports a problem. The cask installs, `codex --version` is correct, and the package is unusable.

## Fix

The rule this needs is already here. `links_rather_than_copies_a_binary_behind_a_flight_symlink` asserts "must link into the SDK tree, not copy the launcher out of it", and `durable_binary_link_target` explains why: a CLI that derives its own root from the resolved path of `$0` has to keep its tree. Stage content was the one source that got flattened instead, because its tree was ephemeral. So make the tree durable and apply the same rule.

`durabilize_stage_payload` copies the extracted payload into the temporary caskroom before binaries are staged, which is the layout Homebrew leaves behind. `stage_binary` then links into that copy, or leaves it alone where the payload already puts the binary at the target path:

```
Caskroom/codex/0.152.0/
├── bin/codex                    the binary target, no longer a lone copy
├── bin/codex-code-mode-host
├── codex-package.json
├── codex-path/rg
└── codex-resources/zsh/bin/zsh
```

Scope is deliberately narrow:

- only a cask with a binary artifact whose source resolves inside the stage durabilizes anything, so nothing else changes shape
- app bundles are skipped, so #12072 stays fixed and a self-updating cask still keeps no caskroom copy of its app
- entries resolving outside the stage are skipped, so the gcloud-cli path from #12293 still links to the SDK its preflight installed rather than gaining a copy
- entries an artifact phase already placed are left alone

## Tests

Two, both `#[cfg(unix)]` beside the existing binary staging tests:

- `keeps_the_payload_beside_a_stage_sourced_binary` — codex-shaped, asserting the helper, manifest, and resources survive the stage teardown and the linked binary still resolves
- `links_a_stage_sourced_binary_into_its_payload_when_the_target_moves_it` — payload nesting the binary deeper than its target path, asserting the caskroom entry links back into the tree instead of standing outside it

`cargo test --bin mise -- system::packages::brew::cask` goes from 242 to 244 passing here. Eleven app and appdir tests fail identically before and after on this machine, which I take to be local privileges rather than anything in this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cask installation reliability for binaries that reference their surrounding extracted files.
  * Preserved required payload contents after installation so these binaries continue working after temporary installation files are removed.

### DIFF
--- a/src/system/packages/brew/cask.rs
+++ b/src/system/packages/brew/cask.rs
@@ -4747,6 +4747,11 @@ fn stage_binary(
     // needs nothing further, and must not be removed to be re-copied onto
     // itself.
     if let Some(payload) = payload_binary_path(stage, caskroom, binary) {
+        // A cask can declare a binary the payload does not ship executable, so
+        // the bit is set on the payload copy itself: it is what the target
+        // resolves to, whether the link points at it or the layout already put
+        // it at the target path.
+        file::make_executable(&payload)?;
         if payload == caskroom_binary {
             return Ok(());
         }
@@ -13779,6 +13784,10 @@ end
         let target = binary.target_path(Path::new("/Applications"))?;
         assert_eq!(std::fs::read_link(&target)?, caskroom.join("bin/tool"));
         assert_eq!(crate::file::read_to_string(&target)?, "launcher");
+        assert!(
+            crate::file::is_executable(&caskroom.join("bin/tool")),
+            "a payload the cask ships without the bit still has to run"
+        );
         for (path, contents) in [
             ("bin/tool-helper", "helper"),
             ("package.json", "{}"),
@@ -13824,6 +13833,7 @@ end
             caskroom.join("pkg/bin/tool"),
             "must link into the payload, not copy the launcher out of it"
         );
+        assert!(crate::file::is_executable(&caskroom.join("pkg/bin/tool")));
         file::remove_all(&stage)?;
         assert_eq!(crate::file::read_to_string(&staged)?, "launcher");
         assert!(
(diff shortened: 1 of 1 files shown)


azohra:
Thanks — the executable bit was a real regression. The copy path guaranteed it with `make_executable`, and linking into the payload dropped that guarantee for a cask whose payload does not ship the binary executable. `fs::copy` carries the mode across, so it survives in the common case, which is exactly why it would have gone unnoticed. Fixed in ab08a57 by setting the bit on the payload copy itself, before both the early return and the link, since that copy is what the target resolves to either way. Both tests now assert it.

On the merge-risk note about absolute payload links breaking when the temporary caskroom is removed: I looked, and that one does not hold. `symlinks_under(installed_caskroom, 1)` takes a *minimum* depth, not a maximum, so `retarget_transient_symlinks` already walks the whole tree and rewrites a link nested under `bin/` onto the final caskroom after activation renames it. I have added `retargets_a_payload_binary_link_when_the_caskroom_is_renamed` to pin that, since it was worth proving rather than asserting.

PR #12736, opened 2026-09-03
fix(activate): translate paths for the calling windows posix shell

jdx:
Native Windows mise previously emitted Windows PATH values into MSYS/Cygwin shells. This breaks command lookup after activation. Translate shell PATH assignments and generated executable references using the runtime of the process that actually invoked mise, while keeping internal and child-process PATH values native.

Runtime detection uses Windows process APIs and the calling shell's runtime DLL, so raw Git Bash/MSYS2 launches work without `MSYSTEM` or `SHELL`, and inherited hints do not misidentify native callers. Mapping reads runtime defaults and persistent fstab mounts, including Git for Windows' `usertemp` mount, custom drive prefixes, and nested mounts. Saved original-PATH values remain opaque native Windows strings to avoid MSYS double conversion. There are no path-conversion subprocesses in activation or hooks.

Tests exercise generated Bash/Zsh/Fish hooks, command lookup after directory changes, native grandchildren, deactivation, and shims. Hook duplicate preservation and `mise env`'s existing deduplication behavior are asserted separately. The documentation describes the supported boundary and excludes session-only mounts and generating scripts in one runtime for later use in another.

## Validation

- Rebased onto main `0d625d4a42a49e731138187d31d7b8c339864ccb`.
- Local Windows 11 ARM64 with x64 Git Bash, MSYS2, and Cygwin runtimes: all 9 POSIX activation/hook E2E cases pass, including executable paths containing spaces and an apostrophe.
- Full mise unit run: 2,901 passed. The final mapper harness additionally passes all 15 cases, including the subsequently added usertemp regression.
- Broader Windows PATH E2E selection: 63 passed; one unrelated Python 3.12.0 test cannot install that version because no Windows ARM64 package exists.
- Six real persistent-mount comparisons against Cygwin pass.
- `mise run lint-fix`, Rust formatting, TOML formatting, and documentation lint checks completed. Unrelated formatter changes were excluded.
- Workspace/all-features/all-targets Clippy with `-D warnings` reports four existing errors, reproduced on stable Rust 1.95 and beta 1.96, in unchanged `src/backend/cargo.rs`, `src/shims.rs`, `src/system/remote.rs`, and `src/task/task_source_checker.rs`. No lint exclusions were added.

This was tested on Windows ARM64 under x64 runtime emulation, not a separate native x64 Windows machine. Runtime discovery and mapper initialization measured roughly 6 ms per fresh process on this host (debug harness, 30 samples); this includes process enumeration and mount reads and excludes process startup.

Context: discussion #3961; previous fixes and reversions #4048, #5581, #6633, #7002, #9547, #10147, #10190, #12696, and #12582. In particular, this preserves #12696's native child-PATH inheritance rather than restoring pre-spawn conversion.

---

> [!NOTE]
> **Medium Risk**
> Changes core activation and PATH restoration on Windows with complex mount and caller detection; mistakes could mis-order PATH or mis-identify POSIX vs native callers, though scope is limited to emitted shell scripts and E2E coverage is broad.
> 
> **Overview**
> Fixes native Windows mise emitting semicolon-delimited `PATH` and Windows paths into Git Bash, MSYS2, and Cygwin, which broke lookup after `activate`, `hook-env`, and directory changes.
> 
> Adds **`windows_posix`**: mise keeps Windows PATH internally (including opaque `__MISE_ORIG_PATH`), detects the **parent shell** via process APIs and runtime DLLs (not `SHELL`/`MSYSTEM` alone), and maps only **shell-facing** PATH assignments and `mise.exe` references using runtime defaults plus persistent `fstab` / `fstab.d` (including `usertemp` and custom `cygdrive`). Bash, Zsh, and Fish generators and Fish `fish_add_path` paths go through the mapper; hook-env restores original PATH through `orig_path_for_windows`. Native callers with inherited POSIX env vars stay on Windows paths.
> 
> **Windows E2E** runs real executables through activation, hooks, `mise env`, deactivation, shims, and a PowerShell grandchild PATH probe. Docs and **AGENTS.md** codify the boundary rules (no `cygpath` subprocesses, no pre-converting child PATH).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bbcf71efb13fa00081200b83556f9ddb548c8283. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Improved Windows compatibility for Git Bash, MSYS2, and Cygwin with Bash, Zsh, and Fish.
  - PATH and executable paths now translate correctly between native Windows and POSIX shell formats.
  - Activation, deactivation, hooks, tasks, and shims preserve PATH ordering while avoiding duplicate or empty entries.
  - Added support for persistent custom mount configurations when mapping paths.

- **Bug Fixes**
  - Fixed PATH restoration and environment-variable handling across Windows and POSIX-style shells.

- **Documentation**
  - Added guidance for embedded tool version updates and Windows PATH behavior.

### DIFF
--- a/e2e-win/activate_posix_path.Tests.ps1
+++ b/e2e-win/activate_posix_path.Tests.ps1
@@ -1,5 +1,11 @@
 Describe 'POSIX activation PATH on Windows' {
-    It 'keeps Git Bash usable while native mise updates PATH' {
+    It 'keeps <Runtime> usable while native mise updates PATH' -TestCases @(
+        @{ Runtime = 'Git Bash'; Candidate = 'C:\Program Files\Git\bin\bash.exe' },
+        @{ Runtime = 'raw Git Bash'; Candidate = 'C:\Program Files\Git\usr\bin\bash.exe' },
+        @{ Runtime = 'MSYS2'; Candidate = "$(if ($env:MISE_TEST_MSYS_ROOT) { $env:MISE_TEST_MSYS_ROOT } else { 'C:\msys64' })\usr\bin\bash.exe" },
+        @{ Runtime = 'Cygwin'; Candidate = "$(if ($env:MISE_TEST_CYGWIN_ROOT) { $env:MISE_TEST_CYGWIN_ROOT } else { 'C:\cygwin64' })\bin\bash.exe" }
+    ) {
+        param($Runtime, $Candidate)
         function Test-PosixRuntimeBash([string] $Path) {
             if (-not $Path -or -not (Test-Path $Path)) {
                 return $false
@@ -15,19 +21,9 @@ Describe 'POSIX activation PATH on Windows' {
             return $false
         }
 
-        $bashCandidates = @(
-            'C:\Program Files\Git\bin\bash.exe',
-            'C:\Program Files (x86)\Git\bin\bash.exe',
-            "$env:LOCALAPPDATA\Programs\Git\bin\bash.exe",
-            'C:\msys64\usr\bin\bash.exe',
-            'C:\msys32\usr\bin\bash.exe'
-        ) + @(Get-Command bash.exe -All -ErrorAction SilentlyContinue | ForEach-Object { $_.Source })
-        $wslLauncher = Join-Path $env:SystemRoot 'System32\bash.exe'
-        $bash = $bashCandidates |
-            Where-Object { $_ -and ($_ -ne $wslLauncher) -and (Test-PosixRuntimeBash $_) } |
-            Select-Object -First 1
-        if (-not $bash) {
-            Set-ItResult -Skipped -Because 'Git Bash/MSYS2 is not installed'
+        $bash = $Candidate
+        if (-not (Test-PosixRuntimeBash $bash)) {
+            Set-ItResult -Skipped -Because "$Runtime is not installed at $Candidate"
             return
         }
 
@@ -65,7 +61,7 @@ Write-Output $env:PATH
         $saved = @{}
         foreach ($name in 'MISE_E2E_EXE', 'MISE_E2E_ROOT', 'MISE_E2E_NESTED', 'MISE_E2E_ORIGINAL_A',
             'MISE_E2E_ORIGINAL_B', 'MISE_E2E_PREPENDED', 'MISE_E2E_INTERLEAVED', 'MISE_E2E_APPENDED',
-            'MISE_E2E_NATIVE_PROBE', 'MISE_TRUSTED_CONFIG_PATHS') {
+            'MISE_E2E_NATIVE_PROBE', 'MISE_TRUSTED_CONFIG_PATHS', 'PATH', 'SHELL', 'MSYSTEM') {
             $saved[$name] = [Environment]::GetEnvironmentVariable($name, 'Process')
         }
         $env:MISE_E2E_EXE = (Get-Command mise.exe).Source
@@ -78,9 +74,17 @@ Write-Output $env:PATH
         $env:MISE_E2E_APPENDED = $appended
         $env:MISE_E2E_NATIVE_PROBE = Join-Path $TestDrive 'native-path.ps1'
         $env:MISE_TRUSTED_CONFIG_PATHS = $TestDrive
+        $env:PATH = "$(Split-Path $bash);$env:PATH"
+        Remove-Item Env:SHELL, Env:MSYSTEM -ErrorAction Ignore
 
         $probe = @'
-set -e
+set -eu
+mise_exe=$(cygpath -u "$MISE_E2E_EXE")
+apply_mise() {
+  local code
+  code=$("$mise_exe" "$@") || return $?
+  eval "$code"
+}
 original_a=$(cygpath -u "$MISE_E2E_ORIGINAL_A")
 original_b=$(cygpath -u "$MISE_E2E_ORIGINAL_B")
 prepended=$(cygpath -u "$MISE_E2E_PREPENDED")
@@ -90,7 +94,8 @@ root=$(cygpath -u "$MISE_E2E_ROOT")
 nested=$(cygpath -u "$MISE_E2E_NESTED")
 export PATH="$original_a:$original_b:$PATH"
 
-eval "$("$MISE_E2E_EXE" activate bash)"
+apply_mise activate bash
+mise --version >/dev/null
 original-marker.exe cmd.exe >/dev/null
 rest=
 IFS=: read -ra path_entries <<< "$PATH"
@@ -103,30 +108,34 @@ done
 export PATH="$prepended:$original_b:$interleaved:$original_a:$rest:$original_a:$appended"
 
 cd "$root"
-eval "$("$MISE_E2E_EXE" hook-env -s bash)"
+# Exercise the generated cd hook before a direct invocation could conceal failure.
 root-marker.exe cmd.exe >/dev/null
-eval "$("$MISE_E2E_EXE" env -s bash)"
+apply_mise hook-env -s bash
 root-marker.exe cmd.exe >/dev/null
 printf 'ROOT_PATH=%s\n' "$PATH"
+apply_mise env -s bash
+root-marker.exe cmd.exe >/dev/null
+printf 'ENV_PATH=%s\n' "$PATH"
 
 cd "$nested"
-eval "$("$MISE_E2E_EXE" hook-env -s bash)"
 nested-marker.exe cmd.exe >/dev/null
+apply_mise hook-env -s bash
 printf 'NESTED_PATH=%s\n' "$PATH"
 
 cd "$root"
-eval "$("$MISE_E2E_EXE" hook-env -s bash)"
 root-marker.exe cmd.exe >/dev/null
+apply_mise hook-env -s bash
 printf 'RETURN_PATH=%s\n' "$PATH"
 
+powershell.exe -NoProfile -Command '(Get-Command root-marker.exe).Source' | sed 's/^/NATIVE_LOOKUP=/'
 powershell.exe -NoProfile -File "$MISE_E2E_NATIVE_PROBE" | sed 's/^/NATIVE_PATH=/'
-eval "$("$MISE_E2E_EXE" deactivate)"
+apply_mise deactivate
 command -v rm >/dev/null
 command -v cygpath >/dev/null
 original-marker.exe cmd.exe >/dev/null
 printf 'DEACTIVATED_PATH=%s\n' "$PATH"
 
-eval "$("$MISE_E2E_EXE" activate bash --shims)"
+apply_mise activate bash --shims
 printf 'SHIMS_PATH=%s\n' "$PATH"
 '@
 
@@ -143,7 +152,17 @@ printf 'SHIMS_PATH=%s\n' "$PATH"
             $rootPath.IndexOf((Split-Path $originalB -Leaf)) | Should -BeLessThan $rootPath.IndexOf((Split-Path $originalA -Leaf))
             $rootPath | Should -Match ([regex]::Escape((Split-Path $interleaved -Leaf)))
             $rootPath.IndexOf((Split-Path $interleaved -Leaf)) | Should -BeLessThan $rootPath.IndexOf((Split-Path $appended -Leaf))
-            ([regex]::Matches($rootPath, [regex]::Escape((Split-Path $originalA -Leaf))).Count) | Should -Be 2
+            ([regex]::Matches($rootPath, [regex]::Escape((Split-Path $originalA -Leaf))).Count) | Should -Be 2 -Because $rootPath
+            # `mise env` computes a deduplicated environment; hook-env preserves the
+            # live shell's user-owned duplicates. Pin these contracts separately.
+            $envPath = ($output -split "`r?`n" | Where-Object { $_ -like 'ENV_PATH=*' } | Select-Object -Last 1)
+            ([regex]::Matches($envPath, [regex]::Escape((Split-Path $originalA -Leaf))).Count) | Should -Be 1
+            foreach ($line in ($output -split "`r?`n" | Where-Object { $_ -match '^(ROOT|ENV|NESTED|RETURN|DEACTIVATED|SHIMS)_PATH=' })) {
+                $line | Should -Not -Match '(?:^|=|:)[A-Za-z]:[\\/]'
+                $line | Should -Not -Match '::'
+            }
+            $output | Should -Match ('NATIVE_LOOKUP=' + [regex]::Escape((Join-Path $rootBin 'root-marker.exe')))
+            $output | Should -Match 'SHIMS_PATH=[^\r\n]*/shims:'
 
             $nativePath = ($output -split "`r?`n" | Where-Object { $_ -like 'NATIVE_PATH=*' } | Select-Object -Last 1)
             $nativePath | Should -Not -BeNullOrEmpty
--- a/e2e-win/activate_posix_path.Tests.ps1
+++ b/e2e-win/activate_posix_path.Tests.ps1
@@ -1,0 +1,164 @@
+Describe 'POSIX activation PATH on Windows' {
+    It 'keeps Git Bash usable while native mise updates PATH' {
+        function Test-PosixRuntimeBash([string] $Path) {
+            if (-not $Path -or -not (Test-Path $Path)) {
+                return $false
+            }
+            $ancestor = (Get-Item $Path).Directory
+            for ($depth = 0; $depth -lt 5 -and $ancestor; $depth++) {
+                if ((Test-Path (Join-Path $ancestor.FullName 'usr/bin/msys-2.0.dll')) -or
+                    (Test-Path (Join-Path $ancestor.FullName 'bin/cygwin1.dll'))) {
+                    return $true
+                }
+                $ancestor = $ancestor.Parent
+            }
+            return $false
+        }
+
+        $bashCandidates = @(
+            'C:\Program Files\Git\bin\bash.exe',
+            'C:\Program Files (x86)\Git\bin\bash.exe',
+            "$env:LOCALAPPDATA\Programs\Git\bin\bash.exe",
+            'C:\msys64\usr\bin\bash.exe',
+            'C:\msys32\usr\bin\bash.exe'
+        ) + @(Get-Command bash.exe -All -ErrorAction SilentlyContinue | ForEach-Object { $_.Source })
+        $wslLauncher = Join-Path $env:SystemRoot 'System32\bash.exe'
+        $bash = $bashCandidates |
+            Where-Object { $_ -and ($_ -ne $wslLauncher) -and (Test-PosixRuntimeBash $_) } |
+            Select-Object -First 1
+        if (-not $bash) {
+            Set-ItResult -Skipped -Because 'Git Bash/MSYS2 is not installed'
+            return
+        }
+
+        $root = Join-Path $TestDrive 'project root'
+        $nested = Join-Path $root 'nested'
+        $originalA = Join-Path $TestDrive 'original a bin'
+        $originalB = Join-Path $TestDrive 'original b bin'
+        $prepended = Join-Path $TestDrive 'prepended bin'
+        $interleaved = Join-Path $TestDrive 'interleaved bin'
+        $appended = Join-Path $TestDrive 'appended bin'
+        $rootBin = Join-Path $root 'bin'
+        $nestedBin = Join-Path $nested 'bin'
+        foreach ($dir in $root, $nested, $originalA, $originalB, $prepended, $interleaved, $appended,
+            $rootBin, $nestedBin) {
+            New-Item -ItemType Directory -Path $dir -Force | Out-Null
+        }
+
+        # Real native executables make these functional command-resolution checks, not output-shape checks.
+        $where = Join-Path $env:SystemRoot 'System32\where.exe'
+        Copy-Item $where (Join-Path $originalA 'original-marker.exe')
+        Copy-Item $where (Join-Path $rootBin 'root-marker.exe')
+        Copy-Item $where (Join-Path $nestedBin 'nested-marker.exe')
+        @'
+[env]
+_.path = ["bin"]
+'@ | Out-File (Join-Path $root 'mise.toml') -Encoding utf8NoBOM
+        @'
+[env]
+_.path = ["bin"]
+'@ | Out-File (Join-Path $nested 'mise.toml') -Encoding utf8NoBOM
+        @'
+Write-Output $env:PATH
+'@ | Out-File (Join-Path $TestDrive 'native-path.ps1') -Encoding utf8NoBOM
+
+        $saved = @{}
+        foreach ($name in 'MISE_E2E_EXE', 'MISE_E2E_ROOT', 'MISE_E2E_NESTED', 'MISE_E2E_ORIGINAL_A',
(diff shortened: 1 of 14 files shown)


src/windows_posix.rs
```diff
@@ -0,0 +1,761 @@
...
+
+fn ensure_leading_slash(tail: &str) -> Cow<'_, str> {
+    if tail.starts_with('/') {
+        Cow::Borrowed(tail)
+    } else {
+        Cow::Owned(format!("/{tail}"))
+    }
+}
+
+fn split_posix_path_list(value: &str) -> Vec<&str> {
```
jdx, line 412: **Corrupts native `;` lists.** The drive-letter guard only fires when the current entry is exactly one character, so the second and later `X:` boundaries split. Running this function standalone:

```
"C:\a;D:\b" -> ["C:\a;D", "\b"]
"C:\Users\x\bin;C:\Program Files\Git\usr\bin;C:\Windows"
  -> ["C:\Users\x\bin;C", "\Program Files\Git\usr\bin;C", "\Windows"]
```

`posix_to_windows` leaves those untouched and `map_posix_path_list` rejoins with `;`, giving `C:\a;D;\b`.

This is the common case, not an edge: MSYS2 runs every non-PATH env value through its argument heuristic on native spawn (msys2.org/docs/filesystem-paths, "paths in environment variables get converted too", `/foo:/bar` -> `C:\msys64\foo;C:\msys64\bar`), so the `export __MISE_ORIG_PATH="$PATH"` from `activate.sh` reaches `mise.exe` already in `;` form unless `MSYS2_ENV_CONV_EXCL`/`MSYS_NO_PATHCONV` is set. The `PRISTINE_ENV` fallback in `compute_deactivated_path` (src/hook_env.rs:673) is *always* native. Effect: `orig_set` in `cli/hook_env.rs:321` matches nothing so every pristine entry is treated as a user "pre" path, and `mise deactivate` restores fragments.

The splitter needs to decide POSIX-vs-native per list (any `;`, or `X:\`/`X:/` at an entry start => already native, return as-is), not per colon. A unit test feeding `map_posix_path_list` a `;` list would pin it.

src/windows_posix.rs
```diff
@@ -0,0 +1,761 @@
...
+
+    let Some((kind, root)) = executable.as_deref().and_then(runtime_root) else {
+        return if intent == RuntimeIntent::RequiredMsys {
+            Err("MSYSTEM identifies an active MSYS shell, but mise could not locate its runtime root from PATH".into())
+        } else {
+            // This includes BusyBox Bash and ordinary native Windows callers.
+            Ok(None)
+        };
+    };
+    if intent == RuntimeIntent::ProbeCygwin && kind != RuntimeKind::Cygwin {
```
jdx: **Positive MSYS evidence is discarded when `MSYSTEM` is unset.** `runtime_root` has just proven this is MSYS (`usr/bin/msys-2.0.dll` beside the shell), but `ProbeCygwin && kind != Cygwin` returns `Ok(None)`, so `Bash::set_env` emits `export PATH='C:\Program Files\Git\usr\bin;...'` into a real MSYS shell and every following command is not found.

`MSYSTEM` is only set by the `bin\bash.exe` git-wrapper, `msys2_shell.cmd`, and `/etc/profile`. It is unset for `usr\bin\bash.exe` launched directly, Windows OpenSSH `DefaultShell`, IDE terminal profiles pointing at `usr\bin\bash.exe`, and any `--noprofile --norc` shell. The new E2E lists `C:\msys64\usr\bin\bash.exe` as a candidate and runs it with `--noprofile --norc`, so on an MSYS2-only runner it would fail its own `Should -Not -Match ';'` assertions.

`PathMapper::new` already takes `subsystem: Option<&str>`, so accepting the Msys result here (without the `/<msystem>` mount) costs nothing. That also matches the AGENTS.md rule this PR adds: the DLL is the runtime evidence, not `MSYSTEM`.

src/windows_posix.rs
```diff
@@ -0,0 +1,761 @@
...
+                name.to_ascii_lowercase().as_str(),
+                "sh" | "sh.exe" | "bash" | "bash.exe" | "zsh" | "zsh.exe" | "fish" | "fish.exe"
+            )
+        })
+        .unwrap_or("bash.exe");
+    let executable = find_runtime_executable(candidate);
+
+    let Some((kind, root)) = executable.as_deref().and_then(runtime_root) else {
+        return if intent == RuntimeIntent::RequiredMsys {
+            Err("MSYSTEM identifies an active MSYS shell, but mise could not locate its runtime root from PATH".into())
```
jdx: **Hard error through `require_shell` for a leaked `MSYSTEM`.** `find_runtime_executable` stops at the first `bash.exe` on PATH and never tries later entries. With `MSYSTEM` non-empty (inherited by pwsh/cmd/VS Code spawned from an MSYS shell, or exported job-wide by `msys2/setup-msys2`), `SHELL` unset, and `C:\Windows\System32` ahead of `Git\usr\bin` (the stock order), the WSL launcher `System32\bash.exe` is picked, `runtime_root` returns None, and this `Err` propagates via `ensure_available` from `require_shell` (`shell/mod.rs:192`) and `Env::print` (`cli/env.rs:186`).

That makes `mise hook-env -s bash` fail on every prompt, and `eval "$(mise deactivate)"` fails too, so the user cannot back out. The E2E already knows to skip the WSL launcher (`$wslLauncher`); the runtime lookup should do the same, keep walking PATH past non-runtime matches, and degrade to "no translation" with a warning rather than erroring.

jdx (commented):
Reviewed the mapper and the shell/hook-env wiring, with each finding checked by running the module's functions standalone or against the MSYS2 runtime docs. Eight items are confirmed, two are lower-confidence hardening notes. Inline comments carry the details; the short version:

**Blocking**
1. `split_posix_path_list` corrupts a native `;`-delimited list. MSYS2 converts every path-list-shaped env var (including `__MISE_ORIG_PATH`) to `;` form before spawning `mise.exe`, and the `PRISTINE_ENV` fallback in `compute_deactivated_path` is always native, so `orig_path_for_windows` normally receives `C:\a;D:\b` and returns `C:\a;D;\b`. That garbles `orig_set` in `hook-env` and the PATH restored by `mise deactivate`. The E2E doesn't catch it because its ordering assertions hold whichever way the original entries are classified.
2. An MSYS shell without `MSYSTEM` is rejected even though `msys-2.0.dll` was found next to the shell (`usr\bin\bash.exe` launched directly, OpenSSH DefaultShell, `--noprofile --norc`). The E2E's own `C:\msys64\usr\bin\bash.exe` candidate would hit this.
3. A leaked `MSYSTEM` with a non-MSYS first `bash.exe` on PATH (System32's WSL launcher is the stock order) turns `hook-env`, `activate`, `env -s`, and `deactivate` into hard errors via `require_shell`.

**Mount table**
4. `add_mount` deletes nested child mounts when a parent mountpoint is added; a Cygwin fstab with an explicit `/` line wipes `/usr/bin` and `/usr/lib`.
5. `/d` maps to `D:` (drive-relative) and `D:\` maps back to `/d/`, so drive roots never round-trip.
6. MSYS2's one real runtime mount, `/bin` -> `usr\bin`, is missing.

**Other**
7. The new `#[cfg(windows)]` bash test and two ungated fish tests read the real environment through `ACTIVE_MAPPER`, so `cargo test` from Git Bash fails them.
8. `env_diff.rs` still uses `path::unix_path_to_windows`, so `_.source` PATH readback and activation now disagree on `/usr/bin`, `/mingw64/bin`, and custom cygdrive prefixes.
9. Bare `split(';')` drops the quoted-entry handling `std::env::split_paths` has.
10. Empty entries are emitted as `:` (cwd on PATH) for bash/zsh while fish drops them.

src/windows_posix.rs
```diff
@@ -0,0 +1,761 @@
...
+
+    #[cfg(any(windows, test))]
+    fn add_mount(&mut self, windows: &str, posix: &str) {
+        let windows = normalize_windows(windows);
+        let posix = normalize_posix(posix);
+        if windows.is_empty() || !posix.starts_with('/') {
+            return;
+        }
+        // A later fstab entry replaces a default/system entry at the same mountpoint.
+        self.mounts.retain(|mount| {
```
jdx: **Adding a parent mount deletes its nested children.** The second `retain` term removes every existing mount whose posix path is a descendant of the new one. Cygwin/MSYS mounts nest (`mount_info::add_item` only replaces an identical mountpoint), and `best_*_mount` already picks the longest prefix, so this term is both unnecessary and wrong. Standalone run of the module code:

- fstab `D:/tools /usr/local ntfs` then `E:/usr /usr ntfs`: `/usr/local` is gone, `posix_to_windows("/usr/local/bin")` = `E:\usr\local\bin` (should be `D:\tools\bin`).
- Cygwin with the User's Guide's own example (`C:/cygwin/bin /usr/bin`, `C:/cygwin/lib /usr/lib`, `C:/cygwin / ntfs override,binary 0 0`): `prefix_matches(_, "/")` is `starts_with('/')`, so the root line wipes every mount; `/usr/bin` -> `C:\cygwin\usr\bin` (nonexistent), `C:\cygwin\bin` -> `/bin` instead of `/usr/bin`.

Keeping only the `eq_ascii` term fixes both. Note `fstab_overrides_defaults_and_supports_bind_and_octal_escapes` currently asserts the descendant-removal behaviour (`E:/usr /usr` making `/usr/bin` -> `E:\usr\bin`), so it would need to change.

src/windows_posix.rs
```diff
@@ -0,0 +1,761 @@
...
+        let normalized = normalize_posix(input);
+
+        if let Some(mount) = self.best_posix_mount(&normalized)
+            && mount.posix != "/"
+        {
+            return append_windows_tail(&mount.windows, &normalized[mount.posix.len()..]);
+        }
+        // Drive mounts take precedence over the implicit `/` runtime mount.
+        if let Some((drive, tail)) = self.split_cygdrive(&normalized) {
+            return append_windows_tail(&format!("{}:", drive.to_ascii_uppercase()), tail);
```
jdx: **Drive roots don't round-trip.** With an empty tail this yields `D:`, which in a Windows PATH means "current directory on drive D", not `D:\`. Standalone: `posix_to_windows("/d")` = `D:`, `posix_to_windows("/cygdrive/d")` = `D:`, `windows_to_posix("D:\\")` = `/d/` (MSYS says `/d`), and `windows_to_posix("D:")` passes through untouched since it isn't `is_windows_absolute`, so `/d -> D: -> D:` sticks.

`PathBuf::from("D:") != PathBuf::from("D:\\")`, so in `cli/hook_env.rs` a drive root in `orig_set` never matches the `D:\` MSYS puts in `current_paths` and gets reclassified as a user entry each hook. The existing `crate::path::unix_path_to_windows("/c")` returns `C:\` (src/path.rs:1079), so this also disagrees with the converter already in tree. Emit `D:\` for an empty tail and strip the lone `/` tail in `windows_to_posix`.

src/windows_posix.rs
```diff
@@ -0,0 +1,761 @@
...
+            #[cfg(any(windows, test))]
+            next_precedence: 0,
+        };
+        mapper.add_mount(root.to_string_lossy().as_ref(), "/");
+        let usr_dir = if kind == RuntimeKind::Cygwin {
+            root.to_path_buf()
+        } else {
+            root.join("usr")
+        };
+        mapper.add_mount(usr_dir.join("bin").to_string_lossy().as_ref(), "/usr/bin");
```
jdx, line 63: **MSYS2's real runtime mount is `/bin`, not `/usr/bin`.** `mount_info::init` in msys2-runtime adds exactly one automatic mount besides `/`: `root\usr\bin` -> `/bin`. The `/usr/bin` and `/usr/lib` entries here are identities under `/` (harmless), but `/bin` is missing, so `posix_to_windows("/bin")` = `root\bin`, a directory that doesn't exist. `/bin` is in every default MSYS2/Git Bash PATH (`MSYS2_PATH="/usr/local/bin:/usr/bin:/bin"`).

Reachable whenever `__MISE_ORIG_PATH` arrives in POSIX form, i.e. with `MSYS_NO_PATHCONV=1` (which docs/troubleshooting.md tells Git Bash users to set) or `MSYS2_ENV_CONV_EXCL`. One-line fix: add `root\usr\bin` -> `/bin` for `RuntimeKind::Msys`.

src/shell/bash.rs
```diff
@@ -149,6 +173,23 @@ mod tests {
...
+    fn path_preludes_quote_spaces_and_single_quotes() {
+        assert_eq!(
+            Bash::default().prepend_env("PATH", "/a b/say'hi"),
+            "export PATH=\"/a b/say'hi:$PATH\"\n"
+        );
+    }
+
+    #[cfg(windows)]
+    #[test]
+    fn an_unrecognized_windows_context_keeps_native_path_entries() {
```
jdx, line 183: **Environment-dependent test.** This goes through the process-global `ACTIVE_MAPPER`, which reads `MSYSTEM`/`SHELL`/`PATH` from the real environment with no test seam. `cargo test` from Git Bash on Windows (`MSYSTEM=MINGW64`, Git root on PATH) activates the mapper and `set_env("Path", r"C:\a bin;D:\b")` becomes `export PATH='/c/a bin:/d/b'`, failing both asserts. The ungated fish tests `a_windows_path_becomes_one_fish_element_per_directory` and `the_windows_spelling_of_path_takes_the_list_branch_too` fail the same way. CI is green only because `windows-unit` runs under pwsh. `remove_var` in the test can't help since the `LazyLock` may already be initialised by another test; an injectable/`#[cfg(test)]`-overridable mapper is needed.

src/windows_posix.rs
```diff
@@ -0,0 +1,761 @@
...
+            };
+            return append_tail(&base, tail);
+        }
+        if let Some(tail) = normalized.strip_prefix("//") {
+            return format!("//{tail}");
+        }
+        input.to_string()
+    }
+
+    fn posix_to_windows(&self, input: &str) -> String {
```
jdx, line 164: **Second POSIX->Windows converter.** `crate::path::unix_path_to_windows` (src/path.rs:428) is still used by `src/env_diff.rs:336` to read back `_.source`-prepended PATH entries, with a fixed `/c/` + `/cygdrive/c/` heuristic that deliberately skips `/usr/bin`, `/mingw64/bin`, and custom cygdrive prefixes. After this PR activation maps those correctly while sourced scripts drop them. Routing env_diff through the mapper (a per-entry `posix_path_to_windows` here) and deleting `path::unix_path_to_windows` keeps one set of semantics.

src/windows_posix.rs
```diff
@@ -0,0 +1,761 @@
...
+    match ACTIVE_MAPPER.as_ref() {
+        Ok(Some(mapper)) => Cow::Owned(map_windows_path_list(mapper, value)),
+        _ => Cow::Borrowed(value),
+    }
+}
+
+pub(crate) fn windows_path_entries_for_shell(value: &str) -> Vec<String> {
+    match ACTIVE_MAPPER.as_ref() {
+        Ok(Some(mapper)) => value
+            .split(';')
```
jdx: Lower confidence: the active branch uses a bare `split(';')` while the inactive branch (and the pre-PR fish code) used `std::env::split_paths`, which on Windows honours double-quoted entries. `std::env::join_paths` (used by `hook_env.rs:753` / `cli/hook_env.rs:485` to build the emitted PATH) quotes any entry containing `;`, so such an entry comes out as `"C:\dir`, `x"`, `C:\b` with literal quotes (`"C:\dir` isn't `is_windows_absolute`). Rare input, but using `split_paths` in both branches removes the divergence.

src/windows_posix.rs
```diff
@@ -0,0 +1,761 @@
...
+        entries.push(&value[start..idx]);
+        start = idx + 1;
+    }
+    entries.push(&value[start..]);
+    entries
+}
+
+fn map_windows_path_list(mapper: &PathMapper, value: &str) -> String {
+    value
+        .split(';')
```
jdx: Lower confidence: empties are preserved here (`;;` -> `::` is asserted in the tests) and bash/zsh emit them, while fish drops them precisely because an empty POSIX PATH element is the cwd. MSYS's own PATH conversion strips empties so the default Git Bash flow is safe, but `mise env -s bash` fed a registry-shaped trailing-`;` PATH would emit `export PATH='/c/a:/c/b:'`. Dropping empties here would also make bash/zsh consistent with fish.

src/shell/fish.rs
```diff
@@ -311,6 +325,14 @@ mod tests {
         );
     }
 
+    #[test]
+    fn original_path_is_emitted_as_a_fish_path_list() {
+        assert_eq!(
+            Fish::default().set_env("__MISE_ORIG_PATH", "/a path:/say'hi"),
+            concat!(r"set -gx __MISE_ORIG_PATH '/a path' '/say'\''hi'", "\n")
+        );
+    }
```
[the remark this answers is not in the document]
jdx: This test is already inside the module gated by #[cfg(all(test, not(windows)))] at src/shell/fish.rs:260. It is excluded from Windows builds as a whole; the separate windows_tests module covers Windows behavior. The current Windows unit CI also passed, so no additional per-test gate is needed.

src/windows_posix.rs
```diff
@@ -0,0 +1,849 @@
...
+        None
+    }
+}
+
+fn prefix_matches(path: &str, prefix: &str) -> bool {
+    if prefix == "/" {
+        return path.starts_with('/');
+    }
+    strip_prefix_ascii(path, prefix).is_some_and(|tail| tail.is_empty() || tail.starts_with('/'))
+}
```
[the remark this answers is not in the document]
jdx: Fixed in bbcf71efb: a prefix ending in a separator already establishes the directory boundary. Added coverage for descendants, the mount root, more-specific child mounts, other drives, and drive-relative inputs. All 16 mapper tests pass. A production mapper probe against real Cygwin fstab drive-root and child mounts agrees with cygpath (normalizing the optional trailing slash at the root).

e2e-win/activate_posix_shells.Tests.ps1
```diff
@@ -0,0 +1,115 @@
...
+            $output | Should -Match ([regex]::Escape((Join-Path $fixture 'project\bin\root-marker.exe')))
+            $output | Should -Match 'SHIMS_PATH=[^\r\n]*/shims:'
+            foreach ($line in ($output -split "`r?`n" | Where-Object { $_ -match '^(PROJECT|SHIMS)_PATH=' })) {
+                $line | Should -Not -Match '(?:^|=|:)[A-Za-z]:[\\/]|;|::'
+            }
+        }
+        finally {
+            foreach ($name in $saved.Keys) {
+                [Environment]::SetEnvironmentVariable($name, $saved[$name], 'Process')
+            }
```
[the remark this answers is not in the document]
jdx: Fixed both cleanup blocks in bbcf71efb: originally absent process variables are explicitly removed, and existing values are restored. All 9 Windows POSIX activation tests passed using Git Bash, MSYS2 and Cygwin, including Zsh and Fish. The repository lint-fix task and formatting checks completed successfully.

e2e-win/activate_posix_path.Tests.ps1
```diff
@@ -0,0 +1,164 @@
...
+            $rootPath.IndexOf((Split-Path $originalB -Leaf)) | Should -BeLessThan $rootPath.IndexOf((Split-Path $originalA -Leaf))
+            $rootPath | Should -Match ([regex]::Escape((Split-Path $interleaved -Leaf)))
+            $rootPath.IndexOf((Split-Path $interleaved -Leaf)) | Should -BeLessThan $rootPath.IndexOf((Split-Path $appended -Leaf))
+            ([regex]::Matches($rootPath, [regex]::Escape((Split-Path $originalA -Leaf))).Count) | Should -Be 2
+
+            $nativePath = ($output -split "`r?`n" | Where-Object { $_ -like 'NATIVE_PATH=*' } | Select-Object -Last 1)
+            $nativePath | Should -Not -BeNullOrEmpty
+            $nativePath | Should -Match ';'
+            $nativePath | Should -Match ([regex]::Escape($originalA))
+            $nativePath | Should -Match ([regex]::Escape($originalB))
```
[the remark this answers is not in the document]
jdx: Addressed by the functional NATIVE_LOOKUP assertion: a native PowerShell grandchild resolves root-marker.exe through PATH and the test checks the exact project-bin executable path (activate_posix_path.Tests.ps1:130,164). This proves the active project path reaches the native child. The latest Windows E2E CI passed.

src/windows_posix.rs
jdx: Fixed in a0cb07143: native semicolon-delimited original PATH snapshots remain unchanged. native_original_paths_are_not_converted_twice covers multi-drive and single-entry native values; new activation snapshots remain native and opaque.

src/windows_posix.rs
jdx: Fixed in a0cb07143: detection uses the immediate parent executable plus its runtime DLL, without requiring MSYSTEM. Local functional tests passed for raw Git Bash, MSYS2, Cygwin, Zsh and Fish with shell hints removed.

src/windows_posix.rs
jdx: Fixed in a0cb07143: PATH scanning and the require_shell hard error were removed. Native callers remain native despite leaked SHELL/MSYSTEM/WSL hints; the negative Windows E2E case verifies this.

src/windows_posix.rs
jdx: Fixed in a0cb07143: only an identical POSIX mountpoint is replaced. parent_mounts_do_not_remove_child_mounts and real Cygwin parent/child mount probes pass.

src/windows_posix.rs
jdx: Fixed in a0cb07143: drive roots map to absolute Windows roots including the trailing backslash. The regression covers ordinary and /proc/cygdrive roots. bbcf71efb additionally fixes descendant matching for persisted drive-root mounts.

src/windows_posix.rs
jdx: Fixed in a0cb07143: the MSYS /bin alias maps to the runtime usr/bin directory. drive_roots_and_msys_aliases_are_absolute covers this mapping.

src/shell/bash.rs
jdx: Fixed in a0cb07143: unit-test builds use the explicit mapper seam rather than ambient process detection. Mapper unit tests construct their runtime explicitly; production detection is exercised by real Windows shell E2E tests. Latest Windows unit and E2E CI passed.

src/windows_posix.rs
jdx: This remains a separate limitation in sourced-script readback. Reusing ACTIVE_MAPPER here would be incorrect: it identifies the shell calling mise, while _.source runs a child shell that can use a different runtime (or be launched from a native caller with no active mapper). Correctly sharing conversion requires identifying the actual source executor and passing its mapper explicitly. This PR keeps the requested activation/emission boundary and native child inheritance intact; it does not claim to fix _.source mount-aware readback.

src/windows_posix.rs
jdx: Fixed in a0cb07143: windows_path_entries preserves Windows double-quoted semicolons and removes the quoting before mapping. windows_path_quotes_protect_semicolons_without_adding_empty_entries covers this case.

src/windows_posix.rs
jdx: Fixed in a0cb07143: empty Windows PATH entries are filtered before emitting POSIX PATH values. The mapper tests cover repeated and trailing semicolons while retaining nonempty relative and Unicode entries.

PR #12760, opened 2026-09-04
fix(schema): accept package state in bootstrap packages

risu729:
## Summary

- add the missing `state` property to the `[bootstrap.packages]` table form in `schema/mise.json`
- constrain it to `"present"` / `"absent"` with a `"present"` default, matching the parser
- add tombi regression coverage for the accepted and rejected values

## Gap provenance

`state` was introduced by #12716, merged 2026-09-02, without adding it to the JSON schema. The parser accepts it today — `PackageOptionsTomlConfig` in `src/system/mod.rs` deserializes `state` into `PackageDesiredStateTomlConfig` (`present` / `absent`, defaulting to `present`) — but the schema's table form is `additionalProperties: false` with only `version`, `os`, and `adopt`, so a documented config such as:

```toml
[bootstrap.packages]
"pacman:libreoffice-fresh" = { state = "absent" }
```

is rejected by editors and by any schema-based validation, even though mise itself applies it and the [bootstrap packages docs](https://mise.jdx.dev/bootstrap/packages/) document it.

The manager-specific part of the constraint is intentionally not encoded: the manager lives in the property key (`"pacman:libreoffice-fresh"`), which is not portably expressible in JSON Schema. Removal support is enforced at apply time instead — `mise bootstrap packages apply` bails with `<manager> does not support declarative package removal` when a manager without `supports_remove()` has an installed package declared absent — so the description points at that behavior.

## Validation

- `mise run render:schema` (only `schema/mise.json` changed; `schema/mise-task.json` unaffected)
- `mise run test:e2e e2e/config/test_schema_tombi`
- `mise run lint-fix`

## Summary by CodeRabbit

* **New Features**
  * Bootstrap package definitions can specify a desired state of `present` or `absent`.
  * The `absent` state is supported for Pacman packages and removes the package when applied.
  * Packages default to the `present` state when no state is specified.

* **Bug Fixes**
  * Improved validation to reject unsupported package states and invalid package configuration values.

### DIFF
--- a/e2e/config/test_schema_tombi
+++ b/e2e/config/test_schema_tombi
@@ -76,10 +76,7 @@ e2e_args = { default = "--headless", redact = true }
 "11.0" = "temurin-11"
 
 [bootstrap.packages]
-"brew:coreutils" = "latest"
-"brew-cask:1password" = { os = "macos", adopt = true }
 "pacman:libreoffice-fresh" = { state = "absent" }
-"pacman:vim" = { version = "latest", os = "linux", state = "present" }
 
 [bootstrap.macos.launchd.agents.daily]
 program = "/bin/echo"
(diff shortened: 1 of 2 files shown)


risu729:
Superseded by #12785, which landed the same package-state schema support and focused regression on `main`.\n\n*AI-assisted — Tool: Codex; model: openai/gpt-5; version: unavailable.*

PR #12816, opened 2026-09-05
fix(rust): reconcile required profile components

risu729:
Existing rustup toolchains can lose required components such as cargo, rustc, or the host standard library. Reapplying a profile alone does not restore those components. This change explicitly reconciles minimal/default profile requirements, including rust-mingw on GNU Windows, and prevents cross-target components from satisfying host requirements.

Depends on #12814. This PR targets main, so its GitHub diff includes that prerequisite. [Review this layer only](https://github.com/risu729/mise/compare/fix/rust-profile-aliases...fix/rust-required-profile-components).

Validation: independent subagent review; focused Rust unit and reconciliation/custom-home/external-provider E2E coverage passed on the combined stack. The profile requirements and selected-host behavior were checked against rustup source.

*AI-assisted — Tool: Codex; model: openai/gpt-6; version: unavailable.*

## Summary by CodeRabbit

- **Bug Fixes**
  - Rust toolchain installation now restores missing components from an existing toolchain.
  - Component detection now accounts for the active host platform and supported RISC-V targets.
  - Rust profiles recognize `minimal`, `default`, and `complete` aliases consistently.
  - Minimal and default profiles install the appropriate standard components, while Windows GNU targets receive required components.

- **Documentation**
  - Clarified when `mise install` restores components implied by the selected Rust profile.

### DIFF
--- a/docs/lang/rust.md
+++ b/docs/lang/rust.md
@@ -140,6 +140,9 @@ If not set, it defaults to the profile configured in `rustup`. You can check you
 "rust" = { version = "1.83.0", profile = "minimal" }
 ```
 
+If the Rust toolchain is already installed, `mise install` restores missing components implied by
+the `minimal` or `default` profile.
+
 ### `targets`
 
 The `targets` option specifies platforms to install for cross-compilation. Multiple targets can
--- a/e2e/core/test_rust_components_reconcile
+++ b/e2e/core/test_rust_components_reconcile
@@ -65,7 +65,7 @@ case "$1 $2" in
 					;;
 			esac
 			done
-		if ! $toolchain_exists && [[ -z "$profile" || "$profile" == "default" ]]; then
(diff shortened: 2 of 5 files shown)


risu729:
Independent subagent review found that rustup's RISC-V host names were missing from the architecture list used by host-component reconciliation. Fixed in 377cd587e by recognizing riscv64gc and riscv64a23, matching [rustup's supported architecture list](https://github.com/rust-lang/rustup/blob/60e495a99dc409c13c52dff1889217f5f9a1ae61/src/dist/target_tuple/known.rs#L58-L60).

The regression covers dated toolchain names and confirms installed cargo, rustc, and host rust-std satisfy the minimal profile on both architectures. The subagent re-reviewed the patch with no further findings. Lint and all 33 focused Rust-plugin unit tests passed. The fix has been propagated to both dependent PRs.

*AI-assisted — Tool: Codex; model: openai/gpt-6; version: unavailable.*

PR #12857, opened 2026-09-06
fix(brew-cask): upgrade brew-cask packages whose `auto_updates` is enabled

himkt:
`mise bootstrap packages upgrade` currently skips all casks with `auto_updates true`, even when an upgrade is available. The upgrade command should follow Homebrew's ordinary bulk-upgrade defaults and upgrade these apps when their live installed version is older than the current cask version.

The [Homebrew FAQ](https://docs.brew.sh/FAQ) describes this behavior:

> Casks for self-updating apps declare auto_updates true. For a versioned cask that installs a single readable app bundle, Homebrew compares the bundle’s version metadata with the cask’s current version. The default brew upgrade includes the cask when the installed app appears older and skips it when the installed app appears to be the same version or newer.

---

`mise bootstrap packages upgrade` now considers versioned, self-updating casks using Homebrew's default upgrade decision. Matching receipts and `latest` casks skip; otherwise a single owned app upgrades when its bundle's short/build version indicates it is older. Current, newer, unreadable, and incomparable versions remain unchanged.

For casks whose upstream definition declares `version :latest`, this PR also aligns with ordinary `brew upgrade`: these casks are skipped because there is no concrete upstream version to compare. On current mise main, a non-self-updating cask can instead proceed with an upgrade when its recorded version differs from upstream, including when upstream becomes `latest`. This PR makes the `latest` skip explicit for all cask upgrades. This refers to upstream cask metadata, separately from requesting `latest` in `mise.toml`.

This alignment is with Homebrew's ordinary bulk-upgrade defaults. Homebrew can include `version :latest` casks when named explicitly or with `--greedy-latest`, downloading the artifact and comparing its checksum when available. This PR applies the default skip even to explicitly named casks. A follow-up could add `--greedy-latest` with artifact-change detection; a broader `--greedy` option would also need to define overrides for self-updating apps. See the [Homebrew FAQ](https://docs.brew.sh/FAQ#how-does-brew-upgrade-handle-apps-that-update-themselves).

The change reuses existing installation and replacement paths, checks again after acquiring the install lock, and preserves install/apply and dependency behavior. Homebrew's version comparison stays local to casks. The E2E test is included in macOS CI, addressing the review feedback.

Validated on macOS: 269 cask unit tests, the focused upgrade E2E, strict workspace Clippy, and scoped formatting/lint checks passed.

*AI-assisted — Tool: Codex; model: OpenAI/GPT-6; version: unavailable.*

## Summary by CodeRabbit

* **New Features**
  * Homebrew cask upgrades now handle self-updating applications by checking installed app versions before replacement.
  * Outdated applications can be upgraded, while current, newer, unreadable, or incomparable versions are preserved.
  * Dry runs report whether an application would be upgraded without replacing it.
  * Upgrade decisions are rechecked after downloading to account for external app updates.

* **Bug Fixes**
  * `latest` casks are skipped during upgrade operations.
  * Version comparisons support Homebrew-compatible short, build, combined, and prerelease versions.


himkt:
Addressed the [macOS CI coverage finding](https://github.com/jdx/mise/pull/12857#discussion_r3942794109) in 3d2cfdc35: the existing macOS workflow now runs `e2e/cli/test_system_upgrade_brew_cask_auto_updates`. The focused E2E passed locally, as did 269 cask unit tests, strict workspace Clippy, and the workflow checks.

The updated implementation also follows Homebrew's default live bundle version assessment for self-updating casks. The PR description reflects the final behavior.

*AI-assisted — Tool: Codex; model: OpenAI/GPT-6; version: unavailable.*

himkt:
Sorry @jdx that I created the PR without any discussion on the community. I want to have this kind of change since it's impossible to upgrade brew-cask packages installed by mise on demand and some packages declare `auto_update` are enabled but that is not triggered correctly (sadly)...

PR #12902, opened 2026-09-06
feat(install): concise fractional install progress for non-TTY output

jdx:
Install output in CI logs and captured stderr currently comes from `VerboseReport`, which prints a line for every status change. A 22-tool `mise install` turns into hundreds of rows of archive names, checksum lines and rapidly-changing npm counters, and none of them answer the two questions you actually have when reading the log: *what finished*, and *what is taking so long*.

This adds a dedicated append-only reporter for non-TTY stderr (and for known CI, which often allocates a PTY and then keeps every redraw frame in the log). Interactive terminals, `--verbose`, `--raw`, `--quiet` and `--dry-run` are unchanged.

## Output

```text
mise installing 22 tools

  ✓ cargo-insta@1.48.0               0.8s
  ✓ bun@1.4.0                        1.2s
  ✓ hk@1.56.1                        1.5s

  ██████░░░░░░░░░░  8/22 · 3.0s
    aube@2.2.9                      extracting            1.2s
    lua-language-server@3.19.1       downloading           0.8s
    10 queued

  ✓ stylua@2.5.2                     0.8s
  ✓ aube@2.2.9                       1.8s

  ████████████████ 22/22 · installed 22 tools in 10.8s
```

- One permanent line per finished tool, with the time that tool took.
- One periodic snapshot (every 3s, one timer for the whole install regardless of concurrency) showing the bar, what each active tool is doing, its elapsed time, and how many are still queued.
- Per-tool timers start when the worker begins, so queue time is not charged to a tool. The bar's timer covers the whole install.
- The summary distinguishes installed, already-installed and failed, so a full bar can't imply everything succeeded.

There is deliberately no overall ETA or percentage-of-bytes: a 2MB binary and a Rust toolchain are not comparable units, and a completed-tool count is a number the log can actually justify.

## Progress that means something

The bar fills fractionally: a tool halfway through its own install occupies half the width a finished one would. Each tool's fraction is its completed operations plus byte progress through the current one, weighted by a new `Backend::install_operation_weights` (default: the fetch is 0.7, verification and extraction split the rest — the shape nearly every backend here has; a backend that compiles from source can override it). The count beside the bar stays whole tools, since that is the number you can check against the completion lines above it.

The fraction is monotonic — a retried download that reports byte zero again cannot walk the bar backwards — and a tool caps at 0.99 until its worker returns, because postinstall hooks run after the last declared operation. This makes the bar smooth, not time-proportional: three tools are still three equal shares, whatever their size.

Active rows show bytes and rate (`42.1/78.3 MB · 10.5 MB/s`), averaged over the transfer so a 3-second snapshot doesn't report whichever instant it caught. A resumed download's existing offset is excluded from the rate, every announced length opens a fresh transfer so a retry cannot inherit the failed attempt's bytes and time, and a server that sends no `content-length` still gets a running count (no fraction — the bar must not pretend to know).

Two more things the log is asked after the fact. The artifact the backend chose is kept, dimmed, on the row and the completion line (`✓ python@3.12.3  7.1s  cpython-3.12.3+….tar.gz`). And a fetch skipped because the file was already in the download cache says `· cached` — aqua and java returned silently in that case and now report `cached <name>`, the wording node used under a different phrasing. The http backend's `using cached tarball` is deliberately *not* that: it fires after a real download when the unpacked tree is cached, so it is now the phase `extracting from cache`.

Failures take their share of the filled bar in red: `████████████░░░░` over `installed 0 tools · 1 failed` would otherwise read as success. When a tool's custom postinstall hook starts, every declared operation is treated as complete — plugin backends (asdf, vfox) never call `next_operation`, so their bar sat at zero until the worker returned. An empty install creates no reporter at all.

## What stays visible

Only the phase messages mise generates itself are folded into the snapshot. A child process's own stdout — cargo, npm, build scripts — is the diagnostic value of an install log, so it keeps printing as it arrives.

That needed a split: `CmdLineRunner` routed child stdout through `pr.set_message`, the same call backends use for their own phase text. `set_process_output` now carries child stdout and defaults to `set_message`, so the progress and quiet reporters are unchanged; the verbose and text reporters show it live and declare `shows_process_output`, which suppresses the failure replay that would otherwise print it twice.

## Fixes found along the way

- A tool with a custom `postinstall` hook was reported as finished *before* the hook ran (`finish_with_message` on the reporter, then `run_postinstall_hook`). A failing hook still fails the command, but the log had already shown a completed install. The reporter now sets a status message and the worker's own result decides the outcome.
- Backends that short-circuit on an already-installed version now mark the report skipped, so "already installed" is distinguishable from a fresh install rather than silently counting as one.

## Tests

- Unit tests for snapshot layout, queue-time vs work-time, duplicate terminal results, and heartbeat shutdown.
- `e2e/cli/test_install_text_progress` covers captured stderr, a slow postinstall hook (timer keeps ticking, completion only after the hook returns), a failing postinstall hook (no success checkmark), the unchanged `--quiet`/`--verbose`/`--raw`/`--dry-run` modes, and a CI-with-PTY run that must still be append-only.

---

> [!NOTE]
> **Medium Risk**
> Broad install UX and progress-reporting changes across backends and hook timing; behavior is well-tested but affects every default non-TTY install path.
> 
> **Overview**
> **Adds append-only install progress** for captured stderr and CI logs (including PTY-backed CI), while interactive TTY, `--quiet`, `--verbose`, `--raw`, and `--dry-run` stay on the existing reporters.
> 
> The new **`TextInstallProgress`** path prints one completion line per tool, periodic ~3s snapshots (bar, active phases, byte/rate hints, queue count), and a final summary that separates installed, already-installed, and failed. Progress bars use **weighted install operations** (`start_operations_weighted` / `Backend::install_operation_weights`, default ~70% download) plus byte transfer state, with monotonic fractions and red bar segments for failures.
> 
> **Reporting plumbing changes:** child stdout goes through **`set_process_output`** (immediate for text/verbose reporters; **`shows_process_output`** skips duplicate failure replay). Backends report **`cached <file>`** when skipping download fetches; HTTP cache hits say **`extracting from cache`** instead of implying a skipped download. **Postinstall** no longer finishes the progress report before the hook runs—the worker result (and hook failure) drives success/failure lines. Already-installed short-circuits emit a **skipped** icon.
> 
> **Tests:** new `e2e/cli/test_install_text_progress`; attestation/precompile e2e installs use **`--verbose`** where assertions depend on per-phase log lines.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 616faf4817d931d20f41b179c826bfc03deec7d5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added concise text-based install progress with fractional progress bars, download details, transfer rates, artifact names, and postinstall status.
  - Progress now reflects weighted installation steps for smoother, more accurate pacing.
  - Failed tools display their specific error messages and failed progress portions.

- **Improvements**
  - Cached downloads and artifacts now show clear status messages, including cached filenames and extraction progress.
  - Progress output avoids duplicated process logs and remains clean in terminal and CI environments.

### DIFF
--- a/e2e/backend/test_aqua_cosign
+++ b/e2e/backend/test_aqua_cosign
@@ -15,7 +15,7 @@ echo "=== Testing Native Cosign Verification (checksum-level) ==="
 echo "Installing fork-cleaner with native Cosign verification..."
 
 # Capture the installation output to verify the native verification is being used
-output=$(mise install aqua:caarlos0/fork-cleaner@2.4.0 2>&1)
+output=$(mise install --verbose aqua:caarlos0/fork-cleaner@2.4.0 2>&1)
 echo "$output"
 
 # Verify the native Cosign verification was used
@@ -35,7 +35,7 @@ echo "✓ fork-cleaner installed and working correctly"
 echo "=== Testing Native Cosign Verification (top-level binary config) ==="
 echo "Installing envsense with top-level binary Cosign verification..."
 
-output=$(mise install aqua:technicalpickles/envsense@0.3.4 2>&1)
+output=$(mise install --verbose aqua:technicalpickles/envsense@0.3.4 2>&1)
 echo "$output"
 
 if echo "$output" | grep -q "Cosign verified" && ! echo "$output" | grep -q "verify checksums with cosign"; then
@@ -53,7 +53,7 @@ echo "✓ envsense installed and working correctly"
 echo "=== Testing Native Cosign Verification (public key bundle config) ==="
 echo "Installing kube-linter with native public-key bundle verification..."
 
-output=$(mise install aqua:stackrox/kube-linter@0.8.3 2>&1)
+output=$(mise install --verbose aqua:stackrox/kube-linter@0.8.3 2>&1)
 echo "$output"
 
 if echo "$output" | grep -q "Cosign verified"; then
--- a/e2e/backend/test_aqua_github_attestations
+++ b/e2e/backend/test_aqua_github_attestations
@@ -14,7 +14,7 @@ echo "=== Testing Native GitHub Artifact Attestations Verification ==="
 echo "Installing goreleaser with native GitHub artifact attestations verification..."
 
 # Capture the installation output to verify the native verification is being used
-output=$(mise install aqua:goreleaser/goreleaser@2.14.1 2>&1)
+output=$(mise install --verbose aqua:goreleaser/goreleaser@2.14.1 2>&1)
 echo "$output"
 
 # Verify the native GitHub artifact attestations verification was used
--- a/e2e/backend/test_aqua_slsa
+++ b/e2e/backend/test_aqua_slsa
@@ -13,7 +13,7 @@ echo "=== Testing Native SLSA Verification ==="
 echo "Installing sops with native SLSA verification..."
 
 # Capture the installation output to verify the native verification is being used
-output=$(mise install aqua:getsops/sops@3.9.0 2>&1)
+output=$(mise install --verbose aqua:getsops/sops@3.9.0 2>&1)
 echo "$output"
 
 # Verify the native SLSA verification was used
--- a/e2e/cli/test_install_text_progress
+++ b/e2e/cli/test_install_text_progress
@@ -49,6 +49,11 @@ assert "grep -c 'hook-failed' output.log" "1"
 if [[ $output == *'✓ dummy@1.0.2'* ]]; then
   fail "postinstall failure was reported as a successful install: $output"
 fi
+# The reason on the failure line is the worker's error, not whatever phase the
+# tool happened to be in when it died.
+if [[ $output == *'failed: running postinstall hook'* ]]; then
+  fail "failure line reported a phase as the reason: $output"
+fi
 
 # Explicit output modes keep their existing behavior.
 echo '[tools]' >mise.toml
--- a/e2e/cli/test_install_text_progress
+++ b/e2e/cli/test_install_text_progress
@@ -1,0 +1,93 @@
+#!/usr/bin/env bash
+
+# Captured stderr gets concise progress even outside CI.
+export CI=0 MISE_COLOR=0 MISE_DEBUG=0 MISE_TRACE=0 MISE_LOG_LEVEL=info
+export MISE_FORCE_PROGRESS=0
+
+output=$(mise install dummy@1.0.0 2>&1)
+printf '%s\n' "$output" >output.log
+assert_contains 'cat output.log' 'installing 1 tool'
+assert_contains 'cat output.log' '✓ dummy@1.0.0'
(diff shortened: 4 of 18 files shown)


jdx:
Review findings, addressed in 4fd69616a:

**Failure line used the phase instead of the error** (Greptile, Bugbot) — valid. `complete()` finalized a failure with whatever `tool.message` held, and `finish()` then skipped it because it already had an outcome, so the line could read `failed: running postinstall hook` or even `failed: ✓ Cosign verified`. The worker's own error is now passed into `complete()`; `finish()` only handles tools that never reached a worker (blocked dependencies, backend and plugin errors). Covered by a unit test and an e2e assertion.

**Child stdout skipped redaction** (Bugbot) — valid, and the one I'd have most regretted. `safe_eprintln!` bypasses the logger, and redaction happens in the logger (`Config::redact`, `src/logger.rs:81`). Stdout previously reached it via `set_message`, so a token written by a hook, npm or a build script would have started appearing unredacted in CI logs. Child output now goes through `info!` like every other line mise prints.

**`VerboseReport` dropped repeated child output** (Greptile) — valid. It inherited `set_process_output` from `set_message`, whose dedup is meant for repeated phase text; combined with the suppressed failure replay, a child's repeated stdout lines would have been lost. It now prints them directly.

**Delaying the summary until after the global postinstall hook** (CodeRabbit) — not taking this one. It matches existing behavior: `mpr.footer_finish()` is already called before the hooks, with the comment "Finish the global footer before running hooks so output isn't masked". The hook is user-configured work that runs after tools are installed, so folding it into "installed N tools in X" would misreport the install duration. There is also a concrete reason to stop the heartbeat early: tools blocked by a failed dependency have no outcome until `finish()` runs, so a later stop would keep printing them as queued through the config reload.

jdx:
Second review round, addressed in 616faf481:

**Retry rate includes earlier bytes** (Greptile) — valid. `set_length` with an equal total kept the old transfer, so a retried attempt's rate spanned the failed one. Every announced length now opens a fresh transfer; the downloader announces it once per attempt, and a resume then reports its offset through `set_position`, which is excluded from the rate. Unit-tested.

**Unknown-length transfers invisible** (Bugbot) — valid, and contradicted my own doc comment. `inc`/`set_position` now open a transfer with an unknown total, so a chunked download gets a running byte count and no fraction. Unit-tested.

**Extraction cache marked as reuse** (Bugbot) — valid, and I had just made it worse by renaming that message to `cached <file>`. It fires after a real download when the unpacked tree is cached, so it is now the phase `extracting from cache`; only a genuinely skipped fetch (aqua, java, node) reports `cached <name>` and earns `· cached` on the completion line.

Also in this commit: failures take their share of the bar in red, a postinstall hook marks every declared operation complete (plugin backends never call `next_operation`), and an empty install creates no reporter.

PR #12906, opened 2026-09-06
feat(install): live install region for interactive terminals

jdx:
#12902 added the append-only reporter for non-TTY output. This PR brings the interactive terminal up to the same information, from the same model.

## Before

```
mise 2026.9.1 by @jdx                                              [12/22]
node@24.20.0     download node-v24.20.0-linux-x64.tar.xz  42 MB 3s ████████░░░░ ◜
python@3.12.3    extract cpython-3.12.3+20240415-x86_64-unknown-linux-gnu-install_only.tar.gz ◝
hk@1.56.1        installed
gh@2.98.0        installed
... (18 more finished rows)
```

The header count was the only aggregate — no bar behind it (its `progress_total` only drove OSC). Rows carried raw phase text with the filename, so a long artifact name pushed the byte bar off a narrow terminal. Finished rows piled up (clx's default `Keep`), so on a 22-tool install the live work scrolled out of view. Rows never said how long a tool had been going, and the per-tool bar restarted for every operation — download to 100%, then checksum 0→100% again.

## After

```
  ✓ hk@1.56.1            1.5s  hk-v1.56.1-linux-x64.tar.xz
  ✓ gh@2.98.0            2.1s  gh_2.98.0_linux_amd64.tar.gz
  ✓ bun@1.4.0            0.3s · cached  bun-linux-x64.zip
mise 2026.9.1  ████████████░░░░░░░░░░░░  12/22 · 18.4 MB/s · 3 queued · 6.2s
node@24.20.0    downloading  42.1/78.3 MB · 12.4 MB/s       ██████░░░░  1.8s  node-v24.20.0-linux-x64.tar.xz ◜
python@3.12.3   extracting                                  █████████░  3.1s  cpython-3.12.3+….tar.gz ◝
npm:prettier    fetching                                    ████░░░░░░  1.0s ◞
  ↳ 32/48 pkgs · 4.2 MiB
markdownlint    waiting for node@24.20.0                                       ⏸
```

**One model, two renderers.** The `State`/`Tool` model from #12902 — weighted fractions, transfer rate, artifact, reuse — is now the only source of truth. Handler logic moved from the text reporter onto `Tool` (`apply_message`, `set_length`, …); the text and TTY renderers are thin. `InstallProgress`/`ToolProgress` traits let the scheduler not care which one it has. The two modes cannot drift apart because they read the same numbers.

**Finished tools become permanent lines, not live rows.** `job.println` writes above the live region and the row leaves it. The terminal shows the *same* `✓ hk@1.56.1  1.5s` lines CI does; the live region holds only the header and what's active. Scrollback is the record.

**Header carries the install-wide bar** — fractional, failures in red, whole-tool count, aggregate transfer rate, queued count, elapsed — and still drives the terminal's OSC progress indicator. It opens `mise VERSION by @jdx`, as the header always did; the append-only reporter's first line reads `mise by @jdx – installing 7 tools`.

**Rows: phase, detail, one bar, elapsed, artifact.** The row bar is the tool's weighted fraction, filled once across download → verify → extract instead of restarting per operation. Elapsed is per tool. The artifact is dimmed at the end so its length can't disturb the columns.

**"waiting for node@24.20.0."** The scheduler already knew a tool was held behind a dependency; it just had no row. `DepsGraph::dependencies_of` exposes what a queued tool is behind, the row is paused (⏸) rather than spinning, and the list drains as dependencies finish. Tools merely waiting for a job slot are a count in the header. The append-only reporter gets the same rows.

**Embedded npm installs drive the bar through aube's own phases.** The npm backend declares aube's bar weights (`0.15 / 0.80 / 0.05`) through `install_operation_weights`, so the plan — including the removal a forced reinstall puts in front — is set once by the install wrapper. `AubeProgress` walks through it: it reads the phase off each snapshot (resolving → fetching → linking → complete), advances an operation per change, and drops a snapshot from a phase already left rather than repainting the row with its counts. The package tally goes through a new `SingleReport::set_items(done, total)`, so it fills the bar as whole items with no size or rate attached. Before aube reports any phase it vets the package (typosquat, download-count and age gates) and resolves the requested version against the registry; that stretch is now labelled `checking package` instead of sitting on the wrapper's `installing`. Unit-tested with a recording reporter.

**Sub-progress gets a child row.** New `SingleReport::set_detail` carries secondary text that isn't a byte transfer — the npm path reports its phase as the message and `32/48 pkgs · 4.2 MiB` as detail. The TTY shows it as an indented `↳` row under the tool, and the row also carries a bare item tally when a backend drives the bar with counts but sends no detail; the parent row shows byte transfers only, so nothing is printed twice. The classic `ProgressReport` folds phase and detail back into its single status cell, and `VerboseReport` prints one line per change, so single-line reporters lose nothing.

**`prune`, `uninstall` and `upgrade`'s old versions use the same session.** They kept one `remove <path>` row per version in the live region — a prune of ~500 versions is what produced the megabytes of repeated rows in an agent terminal, and even a well-behaved terminal scrolled the active work away. Each finished removal is now one permanent `✓ tool@version  0.3s` line, the header reads `removing N tools`, the summary `removed N tools in …`, and `uninstall`/`remove <dir>` phases are normalized to `removing` (the path was what wrapped). `prune` asks its confirmations first, then removes, so a prompt never lands inside the live region. A removal row reads `removing` from the start (there is nothing to resolve), and a session dropped on an error path still closes with its summary, so the header never sits under the error. Dry runs keep their per-line report. The verbose operation counter is capped, so aube's `Complete` step no longer prints `[4/3]`. e2e-covered.

**The live region fits the terminal it is in.** A layout is decided from the two widths that hold still during an install — the terminal and the prefix column — so a row never gains and loses cells as its message changes. Cells go in order of how little they say: the artifact name first, then the transfer figures, then the row bar; the header gives up its version, then bar cells down to ten, then (below 62 columns) the byline and (below 54) the aggregate rate; the bar never grows back as the terminal shrinks. Completion lines drop the artifact when the full line would wrap (a wrapped record reads as two). At 52 columns a seven-tool install renders without a single wrapped line; at 120 everything is shown. The append-only reporter has no width to read (`COLUMNS` is not exported to pipes or CI), so it is narrow by construction instead.

**Append-only rows are narrow by construction.** A pipe or CI log has no width to detect (`COLUMNS` is a shell variable, almost never exported), so the snapshot rows are shaped to fit instead: fixed cells first — prefix, phase, elapsed — then the one variable cell, the transfer detail, unpadded and last. The artifact name is no longer repeated on every 3-second row (it wrapped mid-name in the GitHub log viewer); the completion line still records it once, in full, where it can be searched. `verify`/`verify size` phases are normalized like `download`/`extract`.

**Agent terminals get the append-only output.** The Claude Code terminal (and other AI-agent terminals) is a PTY that *records* frames instead of redrawing them: `isatty` says yes, but cursor-up and erase-line are logged, so an animated display becomes one copy of the whole live region per tick — a `mise prune` of ~500 versions produced megabytes of repeated rows. `CLAUDECODE=1` / `AI_AGENT` now select the same append-only reporter CI gets. e2e-covered.

**An install waiting on another mise's lock says so.** `lock_tool_version` already had a contention callback; the row now reads `waiting for install lock` only when actually contended, instead of sitting in "resolving". Unit-tested.

**A closing line.** `✓ installed 22 tools in 10.8s — 22 tools` (or `✗ 21 installed · 1 failed`), then the live region is gone.

`--verbose`, `--raw`, `--quiet` and `--dry-run` are unchanged; the classic per-tool `ProgressReport` remains for callers outside an install session.

## Tests

- Model tests in `text_install_progress.rs` cover the shared behavior (including the new waiting rows and non-transfer detail).
- `e2e/cli/test_install_tty_progress` drives `mise install` through a PTY with `CI` unset and a real window size: cursor controls prove a redraw, header bar and count present, phase and hook output visible, completion line and summary in the record, old `[cur/total]` header gone, `already installed` on re-run, a 50-column run with no line over 50 cells and a bar no wider than ten, an agent terminal (`CLAUDECODE=1`) getting the append-only output, `--verbose` unchanged.
- `e2e/cli/test_install_text_progress` covers the append-only reporter, including a removal session.
- Layout thresholds are unit-tested in `tty_install_progress.rs`.

---

> [!NOTE]
> **Medium Risk**
> Large, user-visible changes to install/uninstall/progress across many code paths, though behavior for `--verbose`/`--raw`/`--quiet` is preserved and coverage is extensive.
> 
> **Overview**
> Interactive installs now use a **shared progress model** with two renderers: a **live TTY region** (header bar, active rows, permanent completion lines in scrollback) and the existing **append-only** reporter for CI, captured stderr, and agent PTYs (`CLAUDECODE` / `AI_AGENT`).
> 
> The scheduler goes through **`InstallProgress` / `ToolProgress`**, surfaces **dependency waits** and **install-lock contention**, and reuses the same session for **`uninstall` / `prune` / upgrade removals** (one line per tool, no path spam). **NPM embedded installs** map aube phases to weighted operations via new **`set_detail` / `set_items`** on `SingleReport`.
> 
> E2E adds PTY install coverage and extends text-progress tests for uninstall output; layout and aube mapping have unit tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fb0312ee2d28d4f0fb16ece79458d676b2379850. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added clearer interactive installation progress with live updates, dependency waiting states, transfer details, and improved narrow-terminal layouts.
  - Added consistent progress reporting for uninstalling, pruning, and removing older tool versions.
  - Added append-only progress output for CI and AI-agent terminals.

- **Bug Fixes**
  - Improved progress accuracy across installation phases, including package checks, item counts, completion status, and concurrent installations.
  - Prevented progress displays from overflowing narrow terminals or showing stale updates.

- **Tests**
  - Expanded end-to-end coverage for interactive, verbose, narrow-terminal, CI, uninstall, and removal progress scenarios.

### DIFF
--- a/e2e/cli/test_install_text_progress
+++ b/e2e/cli/test_install_text_progress
@@ -7,6 +7,7 @@ export MISE_FORCE_PROGRESS=0
 output=$(mise install dummy@1.0.0 2>&1)
 printf '%s\n' "$output" >output.log
 assert_contains 'cat output.log' 'installing 1 tool'
+assert_contains 'cat output.log' 'by @jdx – installing 1 tool'
 assert_contains 'cat output.log' '✓ dummy@1.0.0'
 assert_contains 'cat output.log' '████████████████ 1/1 · installed 1 tool in'
 if [[ $output == *'[1/'* || $output == *$'\033['* ]]; then
--- a/e2e/cli/test_install_text_progress
+++ b/e2e/cli/test_install_text_progress
@@ -55,6 +55,18 @@ if [[ $output == *'failed: running postinstall hook'* ]]; then
   fail "failure line reported a phase as the reason: $output"
 fi
 
+# Removals get the same session: one permanent line per tool and a summary in
+# removal verbs, instead of a kept "remove <path>" row per version.
+output=$(mise uninstall dummy@1.0.0 dummy@1.0.1 2>&1)
+printf '%s\n' "$output" >output.log
+assert_contains 'cat output.log' 'removing 2 tools'
+assert_contains 'cat output.log' '✓ dummy@1.0.0'
+assert_contains 'cat output.log' '✓ dummy@1.0.1'
+assert_contains 'cat output.log' '2/2 · removed 2 tools in'
+if [[ $output == *'remove ~'* || $output == *'remove /'* ]]; then
+  fail "removal rows should not print the install path: $output"
+fi
+
 # Explicit output modes keep their existing behavior.
 echo '[tools]' >mise.toml
 for mode in '--quiet' '--verbose' '--raw' '--dry-run'; do
--- a/e2e/cli/test_install_tty_progress
+++ b/e2e/cli/test_install_tty_progress
@@ -57,6 +57,7 @@ plain = re.sub(r"\x1b\[[0-9;?]*[a-zA-Z]", "", raw)
 assert re.search(r"\x1b\[[0-9;]*[ABCDEFGHJKST]|\x1b\[\?25[lh]", raw), raw
 assert "░" in plain or "█" in plain, plain
 assert re.search(r"[01]/1", plain), plain
+assert "by @jdx" in plain, plain
 
 # The row named the phase. (A hook's stdout is a transient status line in the
 # interactive display, as it always was; the append-only reporter is what
@@ -95,6 +96,8 @@ assert re.search(r"✓ dummy@1\.0\.0\s+\d+(\.\d+)?(ms|s)", plain), plain
 for line in plain.replace("\r", "\n").splitlines():
     assert len(line.rstrip()) <= 50, f"{len(line)} cells at 50 columns: {line!r}\n{plain}"
 assert "█" * 11 not in plain, plain
+# The byline is the last thing the header gives up, and at 50 columns it goes.
+assert "by @jdx" not in plain, plain
 
 # An AI agent's terminal is a PTY that records frames instead of redrawing
 # them. It must get the append-only output CI gets, or every tick reprints
--- a/e2e/cli/test_install_tty_progress
+++ b/e2e/cli/test_install_tty_progress
@@ -1,0 +1,86 @@
+#!/usr/bin/env bash
+
+# An interactive terminal gets the live install region: a header with the
+# install-wide bar, one row per active tool, and the same permanent completion
+# lines the append-only reporter prints. This drives mise through a PTY with CI
+# unset so the animated display is actually chosen.
+export CI=0 MISE_DEBUG=0 MISE_TRACE=0 MISE_LOG_LEVEL=info
+export MISE_FORCE_PROGRESS=0
+
+cat >mise.toml <<'TOML'
+[tools]
+dummy = { version = "1.0.0", postinstall = "sleep 1; echo hook-finished" }
+TOML
+
+python3 - <<'PY'
+import errno
+import os
+import pty
+import re
+import subprocess
+
+def run(args, env):
+    master, slave = pty.openpty()
+    proc = subprocess.Popen(args, stdout=subprocess.DEVNULL, stderr=slave, env=env)
+    os.close(slave)
+    output = bytearray()
+    try:
+        while True:
+            try:
+                chunk = os.read(master, 8192)
+            except OSError as error:
+                if error.errno == errno.EIO:
+                    break
+                raise
+            if not chunk:
+                break
+            output.extend(chunk)
+    finally:
+        os.close(master)
+    return proc.wait(), output.decode(errors="replace")
+
+env = dict(os.environ, COLUMNS="120", LINES="40")
+env.pop("CI", None)
+env["CI"] = "0"
+code, raw = run(["mise", "install"], env)
+assert code == 0, raw
+plain = re.sub(r"\x1b\[[0-9;?]*[a-zA-Z]", "", raw)
+
+# The live region was drawn: cursor controls appear, and the header carries
+# the install-wide bar with its whole-tool count.
+assert "\x1b[" in raw, raw
+assert "░" in plain or "█" in plain, plain
+assert re.search(r"[01]/1", plain), plain
+
+# The row named the phase. (A hook's stdout is a transient status line in the
+# interactive display, as it always was; the append-only reporter is what
+# prints it per line.)
+assert "running postinstall hook" in plain, plain
+
+# The bar never reads full while the tool is still working: with one tool the
+# header shows at most 23 of 24 cells until the completion line lands.
+for line in plain.splitlines():
+    if "0/1" in line and "█" * 24 in line:
+        raise AssertionError("bar was full before the install finished:\n" + plain)
+
+# The permanent record matches the append-only reporter: one completion line
+# per tool with its duration, then a summary. The old per-tool "installed"
+# row and the old "[cur/total]" header are gone.
+assert re.search(r"✓ dummy@1\.0\.0\s+\d+(\.\d+)?(ms|s)", plain), plain
+assert "installed 1 tool in" in plain, plain
+assert "[0/1]" not in plain and "[1/1]" not in plain, plain
+
+# A version that is already installed is reported as such, not as a fresh
+# install. (A bare `mise install` with nothing to do says "all tools are
+# installed" before any reporter exists, so ask for the version explicitly.)
+code, raw = run(["mise", "install", "dummy@1.0.0"], env)
+assert code == 0, raw
+plain = re.sub(r"\x1b\[[0-9;?]*[a-zA-Z]", "", raw)
+assert "already installed" in plain, plain
+
+# --verbose keeps the classic per-line reporter.
+code, raw = run(["mise", "install", "--force", "--verbose"], env)
+assert code == 0, raw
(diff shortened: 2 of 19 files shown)


jdx:
First review round, addressed:

**npm detail appeared twice** (Greptile, Bugbot) — valid. The message was the folded `fetching 32/48 pkgs` *and* the tally was sent as detail, and `apply_message` didn't recognize the phase words, so both renderers showed the tally twice. Now the phase is the message and the tally is detail, and the classic reporters fold them back themselves: `ProgressReport` into its single status cell, `VerboseReport` as one line per change. Beyond that, npm progress now drives the bar properly: `AubeProgress` maps aube's phase snapshots onto weighted operations (`0.15 / 0.80 / 0.05`, aube's own bar weights) and reports the tally through a new `set_items(done, total)` so it fills the bar as whole items with no size or rate (98e6208ce, unit-tested with a recording reporter).

**TTY repeated non-transfer detail** (Bugbot) — valid. The parent row's `transfer_detail` fell back to the same detail the `↳` child row shows. New `Tool::transfer_bytes` returns only genuine byte transfers (item counts and detail are excluded), and the parent row uses that. Unit-tested.

**Waiting rows went stale** (Bugbot) — valid. Once `waiting_on` drained, the row kept the old `waiting for …` text. A queued tool with nothing left to wait on now reads `queued`.

**Header queued count inflated** (Greptile, Bugbot) — valid. `State::queued_count` now excludes tools that have a `waiting for …` row, and the text snapshot uses the same method, so the two renderers agree. Unit-tested.

Also since the last round: agent terminals (`CLAUDECODE=1` / `AI_AGENT`) get the append-only reporter — they are PTYs that log frames instead of redrawing them, which is how a `mise prune` produced megabytes of repeated rows (8adeae164, e2e-covered); and an install blocked on another mise's lock says `waiting for install lock`, only when actually contended (c54114540, unit-tested).

Plus a layout fix from reading the CI log at narrow width: snapshot rows no longer repeat the artifact name (it wrapped mid-name every three seconds) and the transfer detail is last and unpadded, so rows without one carry no gap. The completion line keeps the full artifact once.

jdx:
Second round, addressed:

**Package tallies disappeared from the TTY child row** (Greptile, Bugbot, CodeRabbit) — valid. `sync_child` kept the `↳` row only while the tool had no transfer, and `set_items` stores the tally as a transfer, so every aube snapshot created the row and removed it again. New `Tool::child_detail` decides what the child shows — an explicit detail or the item tally, never the byte figure the parent row already carries — and is unit-tested.

**`AubeProgress` reset the operation plan** (Bugbot) — valid. The install wrapper already declares the plan through `install_operation_weights` (with the removal a forced reinstall puts in front), and the first snapshot re-declared it, dropping that leading operation. `AubeProgress` now only walks through the declared plan.

**Stale snapshots repainted the row** (CodeRabbit) — valid. The operation never rewound but the label, detail and counts of a late snapshot still landed. A snapshot from a phase already left is now dropped entirely; the unit test covers it.

**Fixed 200ms sleep in the lock-notice test** (CodeRabbit) — valid; it waits on a channel now.

**`\x1b[` proves nothing on its own** (CodeRabbit) — valid; the TTY e2e asserts a cursor-movement or erase sequence.

Also in this push, from running the live region in a 52-column pane: the header (version + 24-cell bar + status), the rows (bar + elapsed + spinner) and the completion lines (artifact) all wrapped. A `Layout` is now decided from the terminal width and the prefix column, giving up cells in order of how little they say (artifact → transfer figures → row bar; the header drops its version, then bar cells down to ten), and completion lines drop the artifact when the full line would wrap. Unit-tested thresholds; the e2e PTY now has a real window size and runs once at 50 columns asserting no line exceeds it. And the stretch before aube reports its first phase — the supply-chain gates and the version lookup against the registry — is labelled `checking package` instead of sitting on the wrapper's `installing` for as long as the registry takes.

jdx:
Third round, addressed:

**Removal sessions skipped finalization on error** (Greptile, Bugbot) — valid. `result?` (and the later `?`s) returned before `progress.finish`, leaving the header in the live region under the error. Both renderers now finish themselves on drop when nobody called `finish`: the live region closes and the summary line says how far the session got before the error that follows. The explicit `finish` calls stay for the normal path.

**Removal rows read `resolving`** (Bugbot) — valid. `State::start_tool` now sets the opening phase from the session's action — `removing` for a removal — and the unit test covers it.

Also in this push: the header opens `mise VERSION by @jdx` again, as it always did, and the append-only reporter's first line reads `mise by @jdx – installing 7 tools`. In the narrow layout the byline is the last thing the header gives up (below 62 columns).

jdx:
**Ten-cell bar could still wrap below 54 columns** (CodeRabbit, Greptile) — valid. The status budget assumed the aggregate rate is always shown. Below 54 columns (where `mise`, the gaps, a ten-cell bar and the full status no longer fit) the rate now leaves the status, so the header is at most 40 cells. Cells freed by a dropped byline or rate are not handed back to the bar — a shrinking terminal never sees the bar grow — and the layout test sweeps 140 → 30 columns asserting the bar is monotonic.

PR #12997, opened 2026-09-09
fix(install): install a lazy tool's missing dependencies on first use

balintant:
### Issue

A lazy tool's `depends` target is commonly lazy itself, and in that case nothing ever installs it. `mise install` skips lazy tools by design, and first-use dispatch installs only the provider, so the dependency preflight fails. The hint it prints suggests `mise install`, which skips lazy tools, so there is no way out except to stop marking the dependency lazy.

> We hit this with [hk](https://hk.jdx.dev). Both `python` and `ruff` were marked lazy, and hk invokes `ruff` itself during a check run, so it failed mid-workflow while `mise install` kept reporting that everything was already installed.

### Fix

With this change, upon the first-use dispatch through a shim, `mise x`, or a task, mise now installs the provider together with its still-missing configured dependencies.

### Reproduction

```toml
# mise.toml
[tools.python]
lazy = true
lazy_bins = ["python", "python3"]
version = "3"

[tools.ruff]
depends = ["python"]
lazy = true
version = "latest"
```

```sh
$ mise install
#> mise all tools are installed          # lazy tools are skipped by design

$ ruff --version
#> mise ERROR failed to install lazy tool ruff@0.16.6
#> mise ERROR Failed to install aqua:astral-sh/ruff@latest: tool 'ruff@latest' requires
#>   configured install dependency 'python@3', but its selected version is not installed
#> hint: Run `mise install python@3` before installing 'ruff@latest'. Remove the
#> dependency from configuration to allow the install hook to rely on system PATH instead.
```

*AI-assisted — Tool: Claude Code; model: anthropic/claude-opus-5[1m]; version: 2.1.220.*

## Summary by CodeRabbit

- **New Features**
  - Lazy tool execution now installs configured tools along with missing direct and transitive dependencies.
  - Dependency installation works consistently during direct execution and through `mise x`.
  - Already-installed dependencies are reused without duplicate installation.
  - Each dependency is installed in the scope associated with its configuration, including support for multiple configured versions.

- **Documentation**
  - Updated developer documentation to reflect lazy tool dependency installation.

- **Tests**
  - Added end-to-end coverage for dependency installation, reuse, multiple versions, and configuration scopes.

### DIFF
--- a/docs/dev-tools/shims.md
+++ b/docs/dev-tools/shims.md
@@ -100,9 +100,11 @@ declare their command names with `lazy_bins`:
 
 Run `mise reshim` after editing a lazy declaration directly. Commands such as
 `mise use` that update tool configuration rebuild the shim farm automatically.
-Invoking a lazy shim installs only its configured provider and then executes
-it. This is independent of `not_found_auto_install`; an explicit project tool
-selection is never bypassed by a lower-precedence lazy declaration.
+Invoking a lazy shim installs its configured provider, plus any configured tools
+that provider [depends](/dev-tools/#tool-dependencies) on that are not installed
+yet, and then executes it. Nothing else in the toolset is installed. This is
+independent of `not_found_auto_install`; an explicit project tool selection is
+never bypassed by a lower-precedence lazy declaration.
 
 A bare `mise install` skips missing lazy tools. Pass `--include-lazy` to install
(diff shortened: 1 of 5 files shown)


balintant:
Regarding @coderabbitai's **Docstring Coverage** requirement:
- Both new functions are documented.
- The pre-existing helpers were already missing docstrings.

PR #13020, opened 2026-09-09
feat(install): add per-tool force option

thespags:
Allows `node = { version = "24", force = true }` to always reinstall that tool on `mise install`, without requiring the global `--force` flag. Composes with `lazy` for tools that should skip bare installs but always reinstall fresh once triggered.

This is a bit of an edge case, so you may not want to pollute mise with it. Specifically, if I'm trying to support users who want to point their tool to `some_go_lib@stable`. To ensure they have the "newest" version of `stable` this allows an option to `force` for that individual tool any time `mise install` is run, or if the user has `mise install` on `cd,` to ensure they are up to date.

Thank you.

## Summary by CodeRabbit

* **New Features**
  * Added a per-tool `force` option that reinstalls a tool even when its requested version is already installed.
  * The option can be combined with lazy installation behavior and applies to both automatic and explicit installs.

* **Documentation**
  * Added guidance for configuring and using the per-tool force reinstall option.

* **Tests**
  * Added coverage confirming forced tools reinstall while other satisfied tools remain unchanged.

### DIFF
--- a/docs/dev-tools/index.md
+++ b/docs/dev-tools/index.md
@@ -164,6 +164,23 @@ precedence. Registry entries may set this option for tools known to follow
 semantic versioning; users can set `version_order = "source"` to restore the
 backend's default ordering.
 
+### Force reinstall
+
+Set `force = true` on a tool to have it always reinstall, even when a bare
+`mise install` would otherwise consider it already satisfied. This is the
+per-tool equivalent of [`mise install --force`](/cli/install.html), scoped to
+just that tool:
+
+```toml
+[tools]
+node = { version = "24", force = true }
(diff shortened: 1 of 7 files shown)


thespags:
HI @jdx, is this something you don't wish to implement or am I duplicating something? My apologies; I didn't see anything in the issues sections or open MRs.

jdx:
i don't want this

thespags:
Bummer, thanks for the update though.

thespags:
@jdx 
Reading a little more, with fuzzy matching, the closest equivalent would be running `mise upgrade` ?

PR #13042, opened 2026-09-10
fix(config): scope runtime lockfile lookup to effective tool config

nettlesh:
Runtime requests can lose the configuration file that owns a tool. Lockfile lookup then merges independent project/global pins and reports a false ambiguity, or selects a pin from a definition the project has overridden. This change preserves the effective config source for lock-aware runtime arguments and uses that source’s corresponding lockfile.

EDIT: pushed the separation—requests now keep where they came from and carry lockfile ownership separately. The original precedence fix is still covered, local install/use/upgrade regressions pass, and I’m keeping this draft until CI finishes.

## Reproducing case

The original failure involved `hk = "latest"` in both the project and global configuration, with version-1 lockfiles using `aqua:jdx/hk` and the same options:

| Source | Binding |
| --- | --- |
| Project `mise.toml` / `mise.lock` | `latest` → `1.58.1` |
| Global `config.toml` / `mise.lock` | `latest` → `1.57.0` |

From the project, `mise which hk --tool hk@latest` reported `lockfile contains multiple resolutions for hk@latest with the same options`, although ordinary config-scoped lookup selected 1.58.1. It now selects the project pin.

If the project still defines hk but has no matching lock entry, the global pin is ignored. Normal unlocked resolution proceeds, or locked mode reports a missing lock entry—even for read-only lookup of an already-installed tool. If the project does not define hk, the inherited definition and its lockfile remain available.

## Implementation and scope

We attempted to implement the agreed behavior with the smallest blast radius:

- Preserve the winning config source when the runtime builder applies arguments. Keep the requested selector and argument-level error reporting, and reuse existing source-to-lockfile mapping and locked-scope policy, including idiomatic version files.
- Interpret request bindings using the recorded backend, so a shorthand does not reject a valid project pin because backend discovery initially found another scope’s backend. Unavailable or disabled recorded backends are non-matches, so a stale candidate cannot block a valid pin or fall back to a cached shorthand backend.
- Reuse the missing-lock-entry check for prefix and ref requests, before they can return through an unlocked shortcut.
- Enable the targeted warning only for `mise which --tool`: when the effective source has no matching pin, inspect lower-precedence configs that actually define the tool and check their corresponding lockfiles. Matching effective pins and unrelated stale lockfile entries do not warn. Run this diagnostic only after successful toolset resolution so its errors cannot replace a returned resolution error.
- Retain intentional lockfile bypasses, including explicit `mise exec hk@latest`. Ambiguous bindings within the selected lockfile remain errors; ambiguity in an ignored source can produce the warning without failing the effective lookup.

The change reuses the existing config order and lockfile cache. It does not change lockfile formats, writers, migration, or shared-monorepo merge rules. Runtime behavior is documented in the lockfile guide.

## Validation

- 203 focused unit tests passed: 98 lockfile tests and 105 toolset tests
- Six focused e2e suites passed: request precedence, idiomatic version files, lockfile versions, backends, environment resolution, and explicit exec-latest bypass
- `cargo clippy --workspace --all-features --all-targets -- -D warnings` passed
- Scoped safe hk checks passed, including formatting, Cargo check, shell checks, and documentation checks
- Verified the original `mise which hk --tool hk@latest` command in the affected dotfiles project: it selects 1.58.1

## Summary by CodeRabbit

- **New Features**
  - Lockfile resolution now follows the configuration that defines each tool.
  - Runtime requests honor matching lockfile pins, including dependency tables.
  - Version, prefix, and reference requests respect locked-mode requirements.
  - Explicit tool versions continue to override lockfile selections.
  - Warnings identify when an overriding configuration lacks a matching pin.

- **Bug Fixes**
  - Improved handling of inherited, environment-specific, aliased, and non-TOML configurations.
  - Clearer errors appear for ambiguous or missing lockfile entries.
  - Invalid or unavailable candidates no longer hide valid lockfile matches.

- **Documentation**
  - Added guidance on runtime resolution, lockfile overrides, warnings, and locked-mode behavior.

### DIFF
--- a/e2e/lockfile/test_lockfile_missing_plugin
+++ b/e2e/lockfile/test_lockfile_missing_plugin
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+
+export MISE_LOCKFILE=1
+export MISE_LOCKED=1
+
+# Temporarily move the harness's fixture symlink so resolution sees a missing plugin.
+mv "$MISE_DATA_DIR/plugins/dummy" "$HOME/dummy-plugin"
+
+for selector in 1 prefix:1 ref:main; do
+  # Keep the backend known even when the plugin cannot supply its identity.
+  printf '[tools]\ndummy = "%s"\n\n[tool_alias]\ndummy = "asdf:dummy"\n' "$selector" >mise.toml
+
+  # Missing plugins defer resolution for every selector, keeping the request
+  # discoverable so the installer can install the plugin and resolve again.
+  mise ls --current --json >listing.json 2>warning
+  assert "jq -r '.dummy[0].version' listing.json" "$selector"
+  assert "jq -r '.dummy[0].installed' listing.json" "false"
+  assert_not_contains "cat warning" "not in the lockfile"
+done
+
+mv "$HOME/dummy-plugin" "$MISE_DATA_DIR/plugins/dummy"
+
+for selector in 1 prefix:1 ref:main; do
+  printf '[tools]\ndummy = "%s"\n\n[tool_alias]\ndummy = "asdf:dummy"\n' "$selector" >mise.toml
+
+  # Once the plugin exists, the deferred request must pass the locked check
+  # before any unlocked resolution or installation can take place.
+  assert_fail_contains "mise which dummy --tool dummy@$selector 2>&1" "dummy@$selector is not in the lockfile"
+done
--- a/src/lockfile.rs
+++ b/src/lockfile.rs
@@ -3734,10 +3734,10 @@ pub(crate) fn get_locked_version(
     };
     if let Some(tools) = lockfile.tools.get(short) {
         if lockfile.uses_request_bindings() {
-            let matching =
+            let (matching, binding_error) =
                 matching_request_bindings(lockfile.as_ref(), short, specifier, |tool| {
                     Ok(binding_options(tool)?.is_some_and(|(options, _)| tool.options == options))
-                })?;
+                });
             match matching.as_slice() {
                 [] => {}
                 [found] => {
@@ -3751,13 +3751,14 @@ pub(crate) fn get_locked_version(
                 }
             }
 
-            let legacy = matching_request_bindings(lockfile.as_ref(), short, specifier, |tool| {
-                Ok(
-                    binding_options(tool)?.is_some_and(|(options, allow_fallback)| {
-                        allow_fallback && !options.is_empty() && tool.options.is_empty()
-                    }),
-                )
-            })?;
+            let (legacy, legacy_error) =
+                matching_request_bindings(lockfile.as_ref(), short, specifier, |tool| {
+                    Ok(
+                        binding_options(tool)?.is_some_and(|(options, allow_fallback)| {
+                            allow_fallback && !options.is_empty() && tool.options.is_empty()
+                        }),
+                    )
+                });
             match legacy.as_slice() {
                 [] => {}
                 [found] => {
@@ -3812,6 +3813,11 @@ pub(crate) fn get_locked_version(
                     )));
                 }
             }
+            // A stale backend's options must not hide a valid binding or legacy
+            // pin, but its error remains useful when none of those matches exist.
+            if let Some(err) = binding_error.or(legacy_error) {
+                return Err(err);
+            }
             return Ok(None);
         }
         let version_matches = |v: &LockfileTool| {
@@ -3865,14 +3871,22 @@ fn matching_request_bindings<'a>(
     short: &str,
     specifier: &str,
     mut matches_options: impl FnMut(&LockfileTool) -> Result<bool>,
-) -> Result<Vec<&'a LockfileTool>> {
+) -> (Vec<&'a LockfileTool>, Option<Report>) {
     let mut matching = Vec::new();
+    let mut first_error = None;
     for tool in lockfile.tools.get(short).into_iter().flatten() {
-        if tool.specifiers.contains(specifier) && matches_options(tool)? {
-            matching.push(tool);
+        if !tool.specifiers.contains(specifier) {
+            continue;
+        }
+        match matches_options(tool) {
+            Ok(true) => matching.push(tool),
+            Ok(false) => {}
+            Err(err) => {
+                first_error.get_or_insert(err);
+            }
         }
     }
-    Ok(matching)
+    (matching, first_error)
 }
 
 /// Newest-first ordering for the lockfile entries that all satisfy one
--- a/src/lockfile.rs
+++ b/src/lockfile.rs
@@ -1,17 +1,18 @@
 pub(crate) mod generate;
 
-use crate::backend::Backend;
 use crate::backend::backend_type::BackendType;
 use crate::backend::conda::CondaBackend;
 use crate::backend::pkgx::PkgxBackend;
 use crate::backend::platform_target::PlatformTarget;
+use crate::backend::{self, Backend};
+use crate::cli::args::BackendArg;
 use crate::config::{Config, Settings};
 use crate::env;
 use crate::file;
 use crate::file::display_path;
 use crate::path::PathExt;
 use crate::platform::Platform;
-use crate::toolset::{ToolSource, ToolVersion, ToolVersionOptions, Toolset};
+use crate::toolset::{ToolRequest, ToolSource, ToolVersion, ToolVersionOptions, Toolset};
 use eyre::{Report, Result, bail, eyre};
 use indexmap::IndexSet;
 use itertools::Itertools;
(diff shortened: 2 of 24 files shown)


jdx:
Thanks for tracking this down. The ambiguity error is a real bug, but I think runtime resolution should follow the effective tool configuration rather than search lockfiles independently.

If a project defines `hk`, it overrides the global definition, so the global lockfile’s `hk` entries shouldn’t participate. If the project has no matching pin, I’d expect normal unlocked resolution—or an error in locked mode. Global pins should remain available when the project doesn’t define the tool.

Could we preserve the effective config source for runtime requests and use its corresponding lockfile? Please also add a regression where the project overrides a globally configured tool but lacks a matching lock entry, ensuring it doesn’t fall back to the overridden global pin. Commands that intentionally bypass lockfiles should keep that behavior.

I don’t think different project/global pins need a warning; that’s a normal override.


For `mise which --tool`, I’d also like a targeted warning when the selector has no matching project pin but does match an overridden global pin. Explain that the global pin is being ignored because the project overrides the tool, and add regression coverage for that case. This should not warn when the project pin matches normally, and it should not cause the global pin to be used.

nettlesh:
@jdx no problem, will remove the fallback to the global lockfile. Quick questions:

1. Do you want "error in locked mode" to also extend to read-only runtime lookups, i.e `mise which --tool`
2. Should that targeted warning also apply when the project has no matching pin but an overridden parent-project or environment lockfile does, or should do you want to keep it to just global pins?

jdx:
1. Yes—`mise which --tool` should error in locked mode when the effective config source has no matching lock entry. Read-only lookups should follow the same locked resolution semantics. Commands that intentionally bypass lockfiles should remain unchanged.

2. Please apply the warning to any overridden lower-precedence configured source, including parent-project and environment-specific sources. It should be config-aware: only warn when that lower-precedence source actually defined the overridden tool and its corresponding lockfile matches the selector. Don’t warn merely because an unrelated lockfile contains a match.

nettlesh:
@jdx pushed a follow-up fixing stale/disabled backend candidates blocking valid pins, and moving the override warning after successful resolution. The focused tests and checks pass.

One remaining clarification from CodeRabbit: `MISE_<TOOL>_VERSION` overrides have no owning config file and still use merged lockfiles. I reproduced the ambiguity with `MISE_DUMMY_VERSION=latest mise which dummy` when project and global configs both define `dummy = "latest"` but pin different versions. That lookup path is unchanged from before this PR.

Should an environment-variable override bypass config-owned pins, or retain its selector while using the otherwise-effective tool config's lockfile, including the missing-pin/locked-mode rules? I've left the shared merge unchanged pending clarification.

src/toolset/builder.rs
```diff
@@ -149,6 +157,71 @@ impl ToolsetBuilder {
...
+                    let has_match = match request.lockfile_resolve(config) {
+                        Ok(pin) => pin.is_some(),
+                        Err(err)
+                            if err
+                                .downcast_ref::<crate::lockfile::AmbiguousRequestBinding>()
+                                .is_some() =>
+                        {
+                            true
+                        }
+                        Err(err) => return Err(err),
```
jdx: [P2] Keep overridden-lockfile diagnostics best-effort

This propagates errors from an overridden configuration after the effective toolset has already resolved successfully. For example, an overridden global Aqua definition with conflicting variable spellings (`foo` and `vars.foo`) causes `resolve_lockfile_options` to fail when this diagnostic inspects it. If the project defines the tool without a matching pin, `mise which --tool` then fails despite successful project resolution. Errors encountered solely while inspecting ignored definitions should remain diagnostic rather than aborting the command. Please add a regression test with an invalid overridden definition and a successfully resolved effective request.

nettlesh:
moving this to draft while I look into the CI regressions and see how this change impacts install/use/upgrade. Taking a step back to re-eval this architecturally for a sec, I want to keep the blast radius small. Will run the full test suite before another review

EDIT: alright, figured out what was bothering me about the design. it conflates where the request came from with which lockfile it should use. Planning to separate those two things

src/toolset/builder.rs
nettlesh: addressed in 54ed19501: failures inspecting an overridden source are logged at debug level and don’t interrupt resolution or inspection of subsequent sources

src/lockfile.rs
```diff
@@ -3830,6 +3860,21 @@ pub(crate) fn get_locked_version(
...
+) -> Result<Vec<&'a LockfileTool>> {
+    let mut matching = Vec::new();
+    for tool in lockfile.tools.get(short).into_iter().flatten() {
+        if tool.specifiers.contains(specifier) && matches_options(tool)? {
+            matching.push(tool);
+        }
+    }
+    Ok(matching)
+}
+
```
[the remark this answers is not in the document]
nettlesh: pretty sure this predates this PR and also relates to my clarifying question above. Leaving it alone for now but happy to address if you'd like

nettlesh:
reopened PR. Separated where a request came from from which lockfile it reads, which feels cleaner to me and better preserves existing behavior. replied to coderabbit about the existing env var case, lmk if you'd like any additional changes

src/lockfile.rs
```diff
@@ -3685,62 +3690,87 @@ pub(crate) fn get_locked_version(
...
+            let Some(backend) =
+                backend::arg_to_backend(BackendArg::new(full.clone(), Some(full.clone())))
+            else {
+                return Ok(None);
+            };
+            if backend::is_disabled_backend_type(&backend.get_type()) {
+                return Ok(None);
+            }
+            return Ok(Some((
+                backend.resolve_lockfile_options(request, &PlatformTarget::from_current())?,
```
jdx, line 3729: [P2] Keep non-matching backend option errors from aborting binding selection

This evaluates `resolve_lockfile_options` for every recorded backend and propagates any error before `matching_request_bindings` can return an otherwise valid match. A scoped lockfile may legitimately retain a stale entry for another recognized backend. For example, with a valid `asdf:dummy` pin using empty options plus a stale `aqua:jdx/hk` entry whose recorded options do not match, a config containing both `foo` and `vars.foo` resolves successfully on `main` to the ASDF pin, but this branch fails with `conflicting aqua var 'foo'` and then reports the tool inactive. I reproduced that control: the focused e2e passes on `origin/main` and fails on `b2f9416`.

Please isolate per-candidate option-resolution failures so a non-matching stale backend cannot invalidate a valid binding, while still surfacing an error when no valid candidate exists.

nettlesh: Addressed in ee293da36. Candidate option errors no longer block a valid pin; they still surface when no valid match exists, and ambiguity checks remain intact. Added regression coverage; focused e2e tests and scoped hk checks pass.

AI-assisted — Tool: Codex; model: OpenAI/unavailable; version: unavailable.

nettlesh:
@coderabbitai resume

e2e/lockfile/test_lockfile_missing_plugin
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+export MISE_LOCKFILE=1
+export MISE_LOCKED=1
+
+# Temporarily move the harness's fixture symlink so resolution sees a missing plugin.
+mv "$MISE_DATA_DIR/plugins/dummy" "$HOME/dummy-plugin"
```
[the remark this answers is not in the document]
nettlesh: Each e2e test gets its own temporary HOME and plugin directory, so this only moves that test’s fixture symlink and cannot affect subsequent tests. Failed environments are intentionally preserved for debugging, and the repository guidelines leave cleanup to the harness. An EXIT trap does not appear necessary here.

AI-assisted — Tool: Codex; model: OpenAI/unavailable; version: unavailable.

src/lockfile.rs
```diff
@@ -3685,62 +3690,88 @@ pub(crate) fn get_locked_version(
...
         if lockfile.uses_request_bindings() {
-            let matching = tools
-                .iter()
-                .filter(|tool| {
-                    tool.specifiers.contains(specifier) && &tool.options == request_options
-                })
-                .collect_vec();
+            let (matching, binding_error) =
+                matching_request_bindings(lockfile.as_ref(), short, specifier, |tool| {
+                    Ok(binding_options(tool)?.is_some_and(|(options, _)| tool.options == options))
```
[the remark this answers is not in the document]
nettlesh: @jdx CodeRabbit flagged another existing issue: explicit backend requests can match a lockfile binding for a different backend when the lookup key, selector, and options match. Would you prefer that I address this here with focused regression coverage, or leave it for a separate PR?

AI-assisted — Tool: Codex; model: OpenAI/unavailable; version: unavailable.

PR #13056, opened 2026-09-10
perf(oci): cache local tool layers across image builds

jdx:
Building several images with overlapping tools currently tars and gzips each tool for every image, even when the installations are unchanged. `oci build`, `oci run`, and `oci push` now share a local tool-layer cache, so separate output layouts and first-time pushes can reuse packaged tool bytes.

The cache hashes the actual packaged inputs: file contents, paths, executable permissions, symlink targets, ownership, and relocated contents. Hits still read and hash the installation but skip tar construction and gzip compression. Same-version edits and reinstalls are detected without trusting timestamps. Per-entry locks prevent concurrent builds from packaging identical inputs twice; atomic publication and blob verification allow recovery from interrupted or corrupt cache writes.

Local layouts remain complete. `oci build --no-cache` bypasses the cache, and push's existing `--no-cache` bypasses both local and remote tool caches. Entries live in each tool's normal mise cache and are removed by `mise cache clear TOOL`.

Stacked on #13055; together they address [discussion #13054](https://github.com/jdx/mise/discussions/13054).

Validation: all 84 OCI unit tests; local-cache, real-registry push, and remote-layer-reuse end-to-end tests; workspace/all-features/all-targets Clippy with `-D warnings`; `mise run lint-fix`. CLI documentation, usage, completions, and man page regenerated.

*AI-assisted — Tool: Codex; model: OpenAI/GPT-6; version: unavailable.*

---

> [!NOTE]
> **Medium Risk**
> Changes OCI image byte production and caching semantics (including push `--no-cache` behavior), but fingerprints and blob validation limit stale-layer risk; install locking reduces race failures during concurrent installs.
> 
> **Overview**
> Adds a **shared local cache** for OCI tool layers so `oci build`, `oci run`, and `oci push` can reuse tar/gzip work across different output layouts and registry destinations when installations are unchanged. Cache keys fingerprint packaged inputs (paths, file bytes including relocated shebangs, modes, symlinks, ownership, mount prefix)—not mtimes— with per-entry locking, atomic writes, and verification so concurrent builds and corrupt entries recover safely.
> 
> **CLI:** `mise oci build` gains **`--no-cache`**; **`oci push --no-cache`** now skips **both** remote registry reuse and the local cache (docs/help updated). **`mise cache clear`** clears entries under each tool’s `oci-layers` cache dir.
> 
> **Build pipeline:** Non-remote tool layers call `build_cached_tool_layer` unless bypassed; install-path checks use tool install locks so same-version reinstalls don’t race packaging. E2E covers reuse, `--no-cache`, content invalidation, cache clear, and push to a fresh repo using locally cached layers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bca6dbcb56a7c4135d967e9eac0d228de824e378. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added local OCI tool-layer caching for `oci build`, `oci run`, and `oci push`, including cache reuse, validation, invalidation, and concurrency support.
  - Added `--no-cache` to `mise oci build` to bypass local layer caching.
  - Updated `mise oci push --no-cache` to bypass both remote and local layer caches.

- **Documentation**
  - Documented cache behavior, invalidation, storage, concurrency, and cache-clearing commands.

- **Tests**
  - Added coverage for cache reuse, invalidation, corruption recovery, concurrency, and registry pushes.

### DIFF
--- a/e2e/oci/test_oci_local_layer_cache
+++ b/e2e/oci/test_oci_local_layer_cache
@@ -52,3 +52,11 @@ assert_contains "cat changed-again.log" "layer from the local cache"
 mise cache clear jq
 mise oci build --from scratch --no-mise -o cleared >cleared.log 2>&1
 assert_not_contains "cat cleared.log" "layer from the local cache"
+
+# Missing installs fail before any system-package work is attempted.
+mise uninstall jq
+cat >>mise.toml <<'EOF'
+[bootstrap.packages]
+"apt:curl" = "latest"
+EOF
+assert_fail "mise oci build --from scratch --no-mise -o missing" "install path does not exist"
--- a/e2e/oci/test_oci_local_layer_cache
+++ b/e2e/oci/test_oci_local_layer_cache
@@ -54,9 +54,11 @@ mise oci build --from scratch --no-mise -o cleared >cleared.log 2>&1
 assert_not_contains "cat cleared.log" "layer from the local cache"
 
 # Missing installs fail before any system-package work is attempted.
-mise uninstall jq
+# Control: with the tool present, package processing must reject scratch.
 cat >>mise.toml <<'EOF'
 [bootstrap.packages]
 "apt:curl" = "latest"
 EOF
+assert_fail "mise oci build --from scratch --no-mise -o packages-control" "requires an apt-based base image"
+mise uninstall jq
 assert_fail "mise oci build --from scratch --no-mise -o missing" "install path does not exist"
--- a/e2e/oci/test_oci_local_layer_cache
+++ b/e2e/oci/test_oci_local_layer_cache
@@ -1,0 +1,54 @@
+#!/usr/bin/env bash
+# Local tool layers are reusable across independent, complete OCI layouts.
+export MISE_EXPERIMENTAL=1
+export SOURCE_DATE_EPOCH=1700000000
+
+cat >mise.toml <<'EOF'
+[tools]
+jq = "1.8.1"
+EOF
+mise install >/dev/null 2>&1
+
+mise oci build --from scratch --no-mise -o first >first.log 2>&1
+assert_not_contains "cat first.log" "layer from the local cache"
+mise oci build --from scratch --no-mise -o second >second.log 2>&1
+assert_contains "cat second.log" "layer from the local cache"
+assert "jq -r '.manifests[0].digest' second/index.json" "$(jq -r '.manifests[0].digest' first/index.json)"
+
+# Independently verify every referenced blob and uncompressed layer digest.
+python3 - <<'PY'
+import gzip
+import hashlib
+import json
+from pathlib import Path
+
+root = Path("second/blobs/sha256")
+
+def read(desc):
+    data = (root / desc["digest"].removeprefix("sha256:")).read_bytes()
+    assert len(data) == desc["size"]
+    assert "sha256:" + hashlib.sha256(data).hexdigest() == desc["digest"]
+    return data
+
+index = json.loads(Path("second/index.json").read_text())
+manifest = json.loads(read(index["manifests"][0]))
+config = json.loads(read(manifest["config"]))
+for layer, diff_id in zip(manifest["layers"], config["rootfs"]["diff_ids"], strict=True):
+    assert "sha256:" + hashlib.sha256(gzip.decompress(read(layer))).hexdigest() == diff_id
+PY
+
+mise oci build --from scratch --no-mise --no-cache -o uncached >uncached.log 2>&1
+assert_not_contains "cat uncached.log" "layer from the local cache"
+assert "jq -r '.manifests[0].digest' uncached/index.json" "$(jq -r '.manifests[0].digest' first/index.json)"
+
+# Mutating an installation at the same version invalidates the local cache.
+printf 'added content\n' >"$(mise where jq)/extra-file"
+mise oci build --from scratch --no-mise -o changed >changed.log 2>&1
+assert_not_contains "cat changed.log" "layer from the local cache"
+assert_succeed "test '$(jq -r '.manifests[0].digest' changed/index.json)' != '$(jq -r '.manifests[0].digest' first/index.json)'"
+mise oci build --from scratch --no-mise -o changed-again >changed-again.log 2>&1
+assert_contains "cat changed-again.log" "layer from the local cache"
+
+mise cache clear jq
+mise oci build --from scratch --no-mise -o cleared >cleared.log 2>&1
+assert_not_contains "cat cleared.log" "layer from the local cache"
--- a/src/cli/oci/push.rs
+++ b/src/cli/oci/push.rs
@@ -62,7 +62,7 @@ pub(super) struct Push {
     #[usage(long)]
     mount_point: Option<String>,
 
-    /// Don't reuse tool layers from the previously pushed image
+    /// Rebuild tool layers without using the remote or local layer cache
     #[usage(long)]
     no_cache: bool,
 
@@ -126,6 +126,7 @@ impl Push {
                     copy: vec![],
                     reuse_from: self.fetch_layer_cache().await?,
                     push_destination: Some(self.reference.clone()),
+                    no_cache: self.no_cache,
                 };
                 let built = perform_build(opts, self.include_global).await?;
                 reused_layers = built.tool_layers.iter().filter(|l| l.reused).count();
--- a/src/cli/oci/push.rs
+++ b/src/cli/oci/push.rs
@@ -17,7 +17,7 @@ use crate::oci::{BuildOptions, LayerOwner, registry};
 /// Tool layers whose tool, version, mount point, and file owner match the
 /// previously pushed image (or `--cache-from`) are reused without being
 /// rebuilt — those tools don't even need to be installed locally. Pass
-/// `--no-cache` to force a full local rebuild.
+/// `--no-cache` to rebuild tool layers without using the remote or local layer cache.
 ///
 /// Credentials are read from the same places docker and podman use:
 /// `$REGISTRY_AUTH_FILE`, `$XDG_RUNTIME_DIR/containers/auth.json`,
--- a/src/oci/builder.rs
+++ b/src/oci/builder.rs
@@ -61,6 +61,8 @@ pub(crate) struct BuildOptions {
     /// Push destination; permits base blobs already in that repository to
     /// remain remote when no build step needs to unpack them.
     pub push_destination: Option<String>,
+    /// Bypass the local tool-layer cache (remote reuse is supplied separately).
+    pub no_cache: bool,
 }
 
 /// Cache key for tool-layer reuse. All four parts must match — a layer built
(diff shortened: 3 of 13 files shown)


src/oci/layer/cache.rs
```diff
@@ -0,0 +1,322 @@
...
+        Ok(None) => {}
+        Err(err) => debug!("ignoring invalid OCI tool layer cache: {err:#}"),
+    }
+
+    let blob = build_layer_from_entries(&entries, target_prefix, owner, Some(relocation))?;
+    // Do not publish under the old key if the installation changed while we
+    // were building. The next invocation will fingerprint its new contents.
+    let after = collect_sorted_entries(src_dir, false, owner, Some(relocation))?;
+    if fingerprint(&after, target_prefix, owner, relocation)? == key
+        && let Err(err) = write_cached_layer(&record_path, cache_dir, &blob)
```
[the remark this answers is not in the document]
jdx: Fixed in 6d14b4155: local layer building now acquires the same logical tool-version transaction lock as install/link/uninstall before checking the source directory, and holds it through hashing, packaging, and cache publication. Remote-only reuse remains installation-independent. All 86 OCI unit tests pass.

src/oci/layer/cache.rs
```diff
@@ -0,0 +1,322 @@
...
+    {
+        debug!("could not cache OCI tool layer: {err:#}");
+    }
```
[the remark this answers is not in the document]
jdx: Fixed in 6d14b4155: the post-build scan, fingerprint, and cache write now run inside a best-effort publication step. Errors log at debug level and retain the successfully packaged blob. The builder also holds the tool installation transaction lock during packaging/publication to coordinate concurrent installs and uninstalls.

src/oci/layer/cache.rs
```diff
@@ -0,0 +1,322 @@
...
+
+fn read_cached_layer(record_path: &Path, cache_dir: &Path) -> Result<Option<LayerBlob>> {
+    let record = match std::fs::read(record_path) {
+        Ok(bytes) => bytes,
+        Err(err) if err.kind() == std::io::ErrorKind::NotFound => return Ok(None),
+        Err(err) => return Err(err.into()),
+    };
+    let record: CachedLayer = serde_json::from_slice(&record)?;
+    crate::oci::layout::validate_sha256_digest(&record.digest)?;
+    crate::oci::layout::validate_sha256_digest(&record.diff_id)?;
```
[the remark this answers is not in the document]
jdx: Fixed in 6d14b4155: cache-hit validation streams the gzip contents through SHA-256 and compares the uncompressed digest with the recorded diff ID. The new regression changes only diff_id to another syntactically valid SHA-256 and verifies a rebuild followed by a valid hit. All 86 OCI unit tests and all three focused OCI e2e tests pass.

src/oci/layer/cache.rs
```diff
@@ -0,0 +1,322 @@
...
+        Err(err) => {
+            debug!("could not lock OCI tool layer cache: {err:#}");
+            return Ok((
+                build_layer_from_entries(&entries, target_prefix, owner, Some(relocation))?,
+                false,
+            ));
+        }
+    };
+    match read_cached_layer(&record_path, cache_dir) {
+        Ok(Some(blob)) => return Ok((blob, true)),
```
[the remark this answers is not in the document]
jdx: Addressed in 6d14b4155 at the caller: Builder now acquires the canonical tool-version installation transaction lock before inspecting the install directory and holds it through fingerprinting, cache lookup, packaging, and publication. Install/link/uninstall use that same lock, so a reinstall cannot race the cache-hit path. Rehashing alone would still leave a race after the second hash. All 86 OCI unit tests and all three focused OCI e2e tests pass.

src/oci/layer/cache.rs
```diff
@@ -0,0 +1,375 @@
...
+fn fingerprint(
+    entries: &[Entry],
+    prefix: &str,
+    owner: LayerOwner,
+    relocation: &ToolRelocation,
+) -> Result<String> {
+    let mut hash = Sha256::new();
+    // Bump the format when layer construction changes without a mise release.
+    field(&mut hash, b"mise-oci-tool-layer-v1");
+    field(&mut hash, env!("CARGO_PKG_VERSION").as_bytes());
```
jdx: Including the crate version means every mise release invalidates every local layer cache, even though the packaged bytes are byte-identical across releases unless the writer changes. The format tag on the line above already exists for the writer-changed case, so this makes it redundant and undercuts the "CI jobs can preserve them along with `MISE_CACHE_DIR`" story in the docs for anyone tracking latest mise. I'd drop the version and rely on bumping the tag when layer construction changes.

src/oci/layer/cache.rs
```diff
@@ -0,0 +1,375 @@
...
+    relocation: &ToolRelocation,
+    cache_dir: &Path,
+) -> Result<(LayerBlob, bool)> {
+    if !src_dir.is_dir() {
+        eyre::bail!("not a directory: {}", src_dir.display());
+    }
+    let entries = collect_sorted_entries(src_dir, false, owner, Some(relocation))?;
+    let key = fingerprint(&entries, target_prefix, owner, relocation)?;
+    let record_path = cache_dir.join(format!("{key}.json"));
+    let _lock = match crate::lock_file::LockFile::at(&cache_dir.join(format!("{key}.lock"))).lock()
```
jdx: `LockFile::at` documents that the file must remain in place while any process could hold its lock, but this one lives inside the tool cache dir that `mise cache clear TOOL` and the 30-day auto-prune delete. If it is unlinked under a holder, two builders lose exclusion. The consequence is benign here (duplicate work; the writes are atomic and identical), so this is a note rather than a bug. `LockFile::new` would sidestep it, and would also stop a `<key>.lock` file accumulating in `oci-layers/` per fingerprint.

jdx (commented):
Reviewed the final revision after the follow-up fixes. No blocking bugs found.

The fingerprint covers every input the tar writer uses: relative path, normalized mode, symlink target after relocation rebasing (the rebase happens in `collect_sorted_entries`, so relocation-driven rewrites are hashed), relocated shebang/pyvenv bytes as written, owner, prefix, OS/arch. Mtime is a constant zero and the gzip level is fixed, so `SOURCE_DATE_EPOCH` correctly stays out of the key. The builder locks the same logical `(short, tv_pathname)` key that install, link, and uninstall use, and there is no `.await` inside the locked region.

Five suggestions inline, none of which I consider blocking.

src/oci/builder.rs
```diff
@@ -381,6 +369,21 @@ impl Builder {
...
             } else {
+                // Coordinate with install, link, and uninstall before inspecting
+                // the source. Hold through fingerprinting, packaging, and cache
+                // publication so a same-version reinstall cannot poison a key.
+                let _install_lock = crate::toolset::install_state::lock_tool_version(
+                    &tv.ba().short,
+                    &tv.tv_pathname(),
+                )?;
+                let install_path = tv.install_path();
+                if !install_path.is_dir() {
```
jdx: This check used to run before the system-packages layer (step 3). With `[bootstrap.packages]`, a missing tool install is now only reported after the rootfs download and package work, which can be a long wait for a "run `mise install` first" error. A cheap unlocked `is_dir()` pre-flight over all versions would restore the early failure while keeping this locked check as the authoritative one.

src/oci/builder.rs
```diff
@@ -381,6 +369,21 @@ impl Builder {
                 );
                 ToolLayer::Reused(reused.clone())
             } else {
+                // Coordinate with install, link, and uninstall before inspecting
+                // the source. Hold through fingerprinting, packaging, and cache
+                // publication so a same-version reinstall cannot poison a key.
+                let _install_lock = crate::toolset::install_state::lock_tool_version(
```
jdx: If another process is installing this version, this blocks with only a debug log, so `oci build` looks hung. The install path uses `lock_tool_version_with_notice` for exactly this reason; an `info!` or progress message here would help.

src/cli/oci/push.rs
```diff
@@ -62,7 +62,7 @@ pub(super) struct Push {
     #[usage(long)]
     mount_point: Option<String>,
 
-    /// Don't reuse tool layers from the previously pushed image
+    /// Rebuild tool layers without using the remote or local layer cache
```
jdx, line 65: The flag help was updated, but the struct doc comment above (line 20) still says `--no-cache` forces "a full local rebuild", and that sentence renders at `docs/cli/oci/push.md`. Worth aligning it with this wording since the flag now bypasses both caches.

src/oci/layer/cache.rs
jdx: Removed the crate version from the fingerprint in 3d4e2e289. Cache compatibility now depends on the explicit layer-format tag, with a comment requiring it to be bumped whenever layer construction changes.

Validation: 86 OCI unit tests, both focused cache/reuse e2e tests, scoped hk checks, and workspace all-features/all-targets Clippy passed.

src/oci/layer/cache.rs
jdx: Switched to LockFile::new keyed by the metadata path in 3d4e2e289, keeping the lock outside the prunable tool cache. Added an assertion that packaging does not create .lock files inside the tool cache; concurrent-build coverage still passes.

Validation: 86 OCI unit tests, both focused cache/reuse e2e tests, scoped hk checks, and workspace all-features/all-targets Clippy passed.

src/oci/builder.rs
jdx: Restored the early unlocked preflight before system-package work in 3d4e2e289, retaining the authoritative check under the install lock. Added a passing e2e regression with a missing jq install and bootstrap apt packages. Remote reuse still skips the local-install requirement.

Validation: 86 OCI unit tests, both focused cache/reuse e2e tests, scoped hk checks, and workspace all-features/all-targets Clippy passed.

src/cli/oci/push.rs
jdx: Aligned the struct-level long help in 3d4e2e289 and regenerated mise.usage.kdl, docs/cli/oci/push.md, and the man page. All explicitly state that --no-cache bypasses both remote and local tool-layer caches.

Validation: 86 OCI unit tests, both focused cache/reuse e2e tests, scoped hk checks, and workspace all-features/all-targets Clippy passed.

src/oci/builder.rs
jdx: Switched to lock_tool_version_with_notice in 3d4e2e289. A contended lock now emits an info-level message identifying the tool whose install lock we are waiting for; uncontended builds stay quiet. OCI unit/e2e tests and Clippy passed.

src/oci/builder.rs
```diff
@@ -333,17 +335,11 @@ impl Builder {
...
-            if !install_path.is_dir() {
-                bail!(
-                    "{} install path does not exist: {}. Run `mise install` first.",
-                    tv.style(),
-                    install_path.display()
-                );
+            if tool_reuse[i].is_none() {
+                require_tool_install(tv)?;
             }
         }
```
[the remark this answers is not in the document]
jdx: Fixed in bf9c56c6c. The preflight retains the cheap positive is_dir check, but acquires the canonical install transaction lock and rechecks before reporting a missing directory. It therefore waits for an in-progress reinstall, while still failing before system-package work for a genuinely missing install. Packaging retains its own locked check. Added a concurrent-reinstall regression. All 87 OCI unit tests, both focused e2e tests, scoped hk checks, and Clippy passed.

e2e/oci/test_oci_local_layer_cache
```diff
@@ -0,0 +1,62 @@
...
+mise oci build --from scratch --no-mise -o cleared >cleared.log 2>&1
+assert_not_contains "cat cleared.log" "layer from the local cache"
+
+# Missing installs fail before any system-package work is attempted.
+mise uninstall jq
+cat >>mise.toml <<'EOF'
+[bootstrap.packages]
+"apt:curl" = "latest"
+EOF
+assert_fail "mise oci build --from scratch --no-mise -o missing" "install path does not exist"
```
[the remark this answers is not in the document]
jdx: The test uses --from scratch, so package processing must fail with requires an apt-based base image before apt can run. In bf9c56c6c I made that control explicit: first assert the apt/scratch error with jq installed, then uninstall jq and assert the missing-install error with the same package configuration. This proves the ordering without running apt. Both focused e2e tests pass.

PR #13062, opened 2026-09-10
feat(doctor): add project diagnostic checks

jdx:
Projects can pin tools but still fail because a native library is undiscoverable, a service is unavailable, or an environment requirement is unmet. Add `mise doctor project` to run explicitly declared `[doctor.checks.<name>]` commands and collect actionable text or JSON results.

Checks support descriptions, remediation hints, timeouts, working directories, task-style shell strings, and the shared OS/architecture selectors. Execution is bounded by mise's jobs setting, with results sorted by name. They use the project's environment and installed tools, continue after failures, and exit nonzero when any check fails or cannot complete. Named checks follow configuration precedence. Ordinary `mise doctor` does not execute them, and remedies are never run automatically.

Reuse the bounded command runner for output limits and process-tree cleanup, including Unix Ctrl-C, SIGTERM, and SIGHUP cancellation. Working directories follow task conventions: relative paths use the declaring project root (or the invocation directory for global/system checks), and absolute paths are used as given. Project checks require normal config trust, cannot execute commands in safe mode, and have no fixed command effect because they execute project code. Empty and platform-excluded check lists still succeed in safe mode. JSON separates configuration errors into a top-level `errors` array; tool-environment failures retain each affected check and platform skip. Include schema/editor support, documentation, Unix integration coverage, and Windows regression tests.

Validation:
- `cargo clippy --workspace --all-features --all-targets -- -D warnings`
- Focused doctor, command-effect, bounded-runner, and configuration-trust unit tests
- `mise run --skip-deps test:e2e e2e/cli/test_doctor_project e2e/config/test_schema_doctor` (using the verified current build; includes conventions and concurrent signal cleanup)
- Scoped hk safe fixes/checks, CLI documentation tests, and `mise run render` (including CLI and LLM indexes)

Windows regression tests are included for cmd.exe quoting, PowerShell, and timeouts; they were not run on this Linux host.

---

> [!NOTE]
> **Medium Risk**
> Runs arbitrary project-defined shell commands with full project env (not sandboxed), though bounded by timeouts/output limits and blocked in safe mode; ordinary `mise doctor` is unchanged.
> 
> **Overview**
> Adds **`mise doctor project`** so repos can declare **`[doctor.checks.<name>]`** probes in `mise.toml` and run them on demand (text or **`--json`**), separate from mise’s own **`mise doctor`** diagnostics.
> 
> Checks merge across the config hierarchy (whole named check replaced by higher precedence), honor **timeouts**, **dir**, **shell**, and **os** filters, run **concurrently** via **`jobs`**, use the project’s **offline** tool environment without installing tools, and exit **1** on fail/error while **not** replaying probe stdout/stderr (64 KiB cap). **Safe mode** blocks execution; trust rules treat doctor checks like other sensitive config. **`CmdLineRunner::output_isolated`** supports non-zero exits for this path, with Unix **SIGTERM/SIGHUP** cleanup aligned to bounded runners.
> 
> Ships JSON schema, CLI/docs/man/usage updates, and Unix/Windows e2e coverage for precedence, timeouts, signals, and schema validation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b5856b65c984ac7be5b8487eafd2a84219dadefe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added `mise doctor project` to run project-defined diagnostic checks.
  - Supports text or JSON reports, check timeouts, platform filtering, concurrent execution, and status reporting.
  - Checks run with the project’s configured tools and environment, with failures reported through the command’s exit status.

- **Configuration**
  - Added support for defining diagnostic checks in `mise.toml`, including commands, descriptions, hints, timeouts, directories, shells, and OS selectors.
  - Updated configuration schema and validation guidance.

- **Documentation**
  - Added CLI, configuration, command reference, sidebar, and man-page documentation for project diagnostics.


src/cli/doctor/project.rs
```diff
@@ -0,0 +1,216 @@
...
+            check
+                .dir
+                .as_ref()
+                .map(|dir| root.join(dir))
+                .unwrap_or_else(|| root.to_path_buf()),
+        )
+        .env_clear()
+        .envs(environment)
+        .with_timeout(timeout)
+        .output_isolated(64 * 1024)
```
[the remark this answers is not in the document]
jdx: Reproduced the SIGTERM leak and fixed it in 7d681e62f. Doctor now cancels its check future on SIGTERM and SIGHUP, allowing the owned process group to be closed; Ctrl-C already uses the CLI cancellation path. The new integration test verifies probe and grandchild cleanup for all three signals, both inside an actual mise task and as a session leader. Keeping a separate owned group is necessary for bounded timeout/output-limit cleanup without killing the surrounding task. SIGKILL prevents supervisor cleanup and that limitation is now documented.

schema/mise.json
```diff
@@ -19,6 +19,70 @@
         {
           "type": "object",
           "properties": {
+            "doctor": {
+              "description": "Project diagnostic checks, run explicitly with mise doctor project.",
+              "type": "object",
+              "additionalProperties": false,
```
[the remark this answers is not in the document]
jdx: Fixed in 7d681e62f: moved the block to $defs.doctor and referenced it from the root properties.doctor. Added actual document-validation tests for valid checks, a missing run command, misspelled options, invalid OS names, and doctor incorrectly placed in a task dependency object. All pass.

src/config/doctor.rs
```diff
@@ -0,0 +1,74 @@
...
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub(crate) struct DoctorCheck {
+    pub run: String,
+    pub description: Option<String>,
+    pub hint: Option<String>,
+    pub timeout: Option<String>,
+    /// Relative to the configuration file's project root.
+    pub dir: Option<PathBuf>,
```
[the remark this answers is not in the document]
jdx: Clarified the intended contract in 7d681e62f: like task directories, relative paths resolve from the declaring config root and absolute paths are accepted as given. Added integration coverage for both absolute and parent-relative paths. This preserves intentional checks of sibling checkouts and shared service directories; the config root is an anchor, not a confinement boundary.

src/cli/doctor/project.rs
```diff
@@ -0,0 +1,216 @@
...
+    };
+    CmdLineRunner::new(program)
+        .cmd_body_args(args, &check.run)
+        .current_dir(
+            check
+                .dir
+                .as_ref()
+                .map(|dir| root.join(dir))
+                .unwrap_or_else(|| root.to_path_buf()),
+        )
```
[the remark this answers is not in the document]
jdx: Clarified the intended dir contract in 7d681e62f and added regression coverage: relative paths are anchored at the declaring config root, while absolute paths are accepted as given. Parent-relative paths are also intentional, matching task working-directory behavior and allowing checks of sibling checkouts or shared services. The config root is not a confinement boundary, so this is a documentation correction rather than a restriction on valid paths.

schema/mise.json
```diff
@@ -19,6 +19,70 @@
         {
           "type": "object",
           "properties": {
+            "doctor": {
```
jdx: **Schema block is in the wrong object.** This `doctor` property sits under `$defs.task_dependency_item.oneOf[2].properties` (next to `task`, `args`, `env`, `optional`), not the root `properties` map:

```sh
jq '.properties.doctor != null, (.["$defs"].task_dependency_item.oneOf[2].properties | keys)' schema/mise.json
# false
# ["args","doctor","env","optional","task"]
```

Effect: `[doctor.checks.<name>]` gets no completion or validation (the root has no `additionalProperties: false`, so a `timeuot` typo is silently accepted by the editor and then fails at runtime for every command), and `depends = [{ task = "x", doctor = {...} }]` now validates. Move the block next to `bootstrap`/`dotfiles`/`oci` in the root `properties`.

*AI-assisted — Tool: Claude Code; model: Anthropic/claude-fable-5-1; version: unavailable.*

src/cli/doctor/project.rs
```diff
@@ -0,0 +1,216 @@
...
+        let mut report = Report::default();
+        if checks.is_empty() {
+            return Ok(report);
+        }
+        let toolset = ToolsetBuilder::new()
+            .with_resolve_options(ResolveOptions {
+                offline: true,
+                ..Default::default()
+            })
+            .build(&config)
```
jdx: **`build(&config).await?` collapses the whole report.** Any toolset-build error propagates out of `check()`, and `run()` replaces the report with a single synthetic `configuration` entry, dropping every per-check and `skipped` result. Only `full_env` failures get the per-check handling the docs describe (project-diagnostics.md: "If configuration loads but preparing the tool environment fails, affected checks each report an error and platform-excluded checks remain skipped").

Realistic triggers with `offline: true`: `MISE_DUMMY_VERSION="-1"` or `[tools] dummy = "../x"` fail `ToolRequest::new` inside `load_runtime_env` / `to_tool_request_set` (src/toolset/builder.rs:121-138). Config itself loaded fine, so this contradicts the doc sentence. Suggest storing the build result like `environment` and reporting it per non-skipped check, or amending the docs.

*AI-assisted — Tool: Claude Code; model: Anthropic/claude-fable-5-1; version: unavailable.*

src/config/doctor.rs
```diff
@@ -0,0 +1,74 @@
...
+    /// Shell executable and arguments, including the command flag.
+    pub shell: Option<Vec<String>>,
+    /// Operating systems on which this check applies.
+    pub os: Option<Vec<OperatingSystem>>,
```
jdx: **`os` re-implements the shared selector.** `[tools].os`, `[bootstrap.packages].os`, and the dotfiles variants all model this as `Vec<String>` (scalar-or-list) matched through `crate::cli::version::os_selector_matches` (src/cli/version.rs:168), which normalizes `darwin`/`win`, accepts `os/arch` pairs, and is backed by the shared `$defs/os_filter` schema.

With this private enum, `os = ["darwin"]`, `os = "macos"`, or `os = ["linux/arm64"]` are serde errors, and because `DoctorConfig` is a field of `MiseToml` that error fails parsing the entire file for every command. It also adds a second hand-maintained OS list in schema/mise.json. Suggest `Vec<String>` + `os_selector_matches` and `$ref: #/$defs/os_filter`.

*AI-assisted — Tool: Claude Code; model: Anthropic/claude-fable-5-1; version: unavailable.*

src/config/doctor.rs
```diff
@@ -0,0 +1,74 @@
...
+#[serde(deny_unknown_fields)]
+pub(crate) struct DoctorCheck {
+    pub run: String,
+    pub description: Option<String>,
+    pub hint: Option<String>,
+    pub timeout: Option<String>,
+    /// Relative to the configuration file's project root.
+    pub dir: Option<PathBuf>,
+    /// Shell executable and arguments, including the command flag.
+    pub shell: Option<Vec<String>>,
```
jdx: **`shell` diverges from tasks/hooks.** Tasks (`Task::shell()`, src/task/mod.rs:2517-2532), run hooks (src/hooks.rs:667-672), and watch_files store `shell: Option<String>`, split it with `crate::path::split_shell_command`, then apply `Settings::get().maybe_no_profile`.

Consequences: (a) `shell = "bash -c"` copied from a task is a type error that fails the whole mise.toml; (b) an explicit doctor shell never gets `-NoProfile` injected, so `windows_powershell_no_profile` is silently ignored here, which is why the new docs tell users to hand-write `["pwsh", "-NoProfile", "-Command"]` while task docs say it is automatic; (c) the empty-shell bail duplicates `Task::shell()` validation. Reusing the task resolution path fixes all three.

*AI-assisted — Tool: Claude Code; model: Anthropic/claude-fable-5-1; version: unavailable.*

src/cli/doctor/project.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    async fn check(&self) -> Result<Report> {
+        if Settings::safe_mode() {
+            bail!("Project diagnostic commands cannot run in safe mode");
+        }
+        let config = Config::get().await?;
+        let mut checks = BTreeMap::new();
+        // Config files are ordered from highest to lowest precedence. Replace
+        // whole named checks so a local command cannot inherit a stale remedy.
+        for (path, cf) in config.config_files.iter().rev() {
```
jdx, line 124: **Global/system config checks run with cwd `$HOME`.** This loop includes `~/.config/mise/config.toml` and `/etc/mise/config.toml`, and `cf.config_root()` for those returns `MISE_GLOBAL_CONFIG_ROOT` (default `$HOME`, config_root.rs:63-66). Tasks use `project_root()`, which is `None` for those files and falls back to the invocation cwd.

So a global `[doctor.checks.env_file] run = "test -f .env"` runs in `$HOME` for every project and reports a misleading PASS/FAIL, while the docs say `dir` is relative to "the declaring configuration's project root" and the PR description says working directories follow task conventions. Either exclude non-project configs from `doctor project` or switch to `project_root()` with the task fallback, and document the choice.

*AI-assisted — Tool: Claude Code; model: Anthropic/claude-fable-5-1; version: unavailable.*

src/config/doctor.rs
```diff
@@ -0,0 +1,74 @@
+use std::collections::BTreeMap;
+use std::path::PathBuf;
+
+use serde::Deserialize;
+
+#[derive(Debug, Default, Clone, Deserialize)]
+#[serde(deny_unknown_fields)]
```
jdx: **Container-level `deny_unknown_fields` fails every command.** Because `DoctorConfig` is a plain `#[serde(default)]` field of `MiseToml`, an unknown key anywhere under `[doctor]` goes through the `serde_ignored` path at mise_toml.rs:589-597 as a hard error, so `mise install`, `mise env`, and `hook-env` activation all fail, not just `mise doctor project`. The unit test and e2e only assert the doctor-side rendering.

Precedent is mixed: `[oci]`, `[tool_config]`, and `[tasks.*]` are strict at the container, but the newer `[bootstrap]`, `[history]`, and `[deps]` are lenient at the container and strict only at leaves. For an opt-in diagnostics table, a future `[doctor.settings]` or per-check `env` key breaking older clients for all commands is a steep cost. Suggest dropping this attribute on the container and deciding deliberately about the leaf.

*AI-assisted — Tool: Claude Code; model: Anthropic/claude-fable-5-1; version: unavailable.*

src/cli/doctor/project.rs
```diff
@@ -0,0 +1,216 @@
...
+                description: check.description.clone(),
+                source: Some(source),
+                status: Status::Pass,
+                message: None,
+                hint: check.hint.clone(),
+            };
+            if check
+                .os
+                .as_ref()
+                .is_some_and(|os| !os.iter().any(|os| os.as_ref() == std::env::consts::OS))
```
jdx: **`os = []` silently skips everywhere.** For an empty list, `is_some_and(|os| !os.iter().any(..))` is `true`, so the check is reported `skipped` on every platform and the command exits 0. Nothing rejects `[]` at deserialize time, and the doctor schema lacks the `minItems: 1` that both `$defs/os_filter` and `bootstrap.packages.*.os` have. Bootstrap packages treat an empty selector as run-everywhere; `[tools]` forbids it in the schema. Pick one and enforce it.

*AI-assisted — Tool: Claude Code; model: Anthropic/claude-fable-5-1; version: unavailable.*

src/cli/doctor/project.rs
```diff
@@ -0,0 +1,216 @@
...
+            .iter()
+            .any(|check| matches!(check.status, Status::Fail | Status::Error))
+        {
+            return Err(crate::request_exit(1));
+        }
+        Ok(())
+    }
+
+    async fn check(&self) -> Result<Report> {
+        if Settings::safe_mode() {
```
jdx: **Safe mode exits 1 with no checks configured.** This bail runs before `Config::get()` and before the `checks.is_empty()` early return, so `MISE_SAFE=1 mise doctor project` in a project with no `[doctor]` table exits 1 with a `configuration` error, although nothing would have run and the docs say "An empty check list or only skipped checks exits successfully." Safe mode still loads config (config_file/mod.rs:504-509), so the bail could move after the empty-list return; otherwise qualify the docs. Minor: the message also lacks the `(MISE_SAFE=1)` suffix `ensure_not_safe` emits.

*AI-assisted — Tool: Claude Code; model: Anthropic/claude-fable-5-1; version: unavailable.*

src/cli/doctor/project.rs
```diff
@@ -0,0 +1,216 @@
...
+        let toolset = ToolsetBuilder::new()
+            .with_resolve_options(ResolveOptions {
+                offline: true,
+                ..Default::default()
+            })
+            .build(&config)
+            .await?;
+        // No tool installation, task dependencies, or task hooks are run here.
+        let environment = toolset.full_env(&config).await;
+        for (name, (source, root, check)) in checks {
```
jdx: **Sequential execution multiplies the deadline.** Each check is awaited in turn with a 10s default timeout, so five unreachable probes cost 50s. Nothing serializes concurrent `output_isolated` calls: `RAW_LOCK` is only read-locked there and output is captured, not printed. `join_all` (or a `Settings::jobs`-capped semaphore like src/parallel.rs:14) over futures built in BTreeMap order keeps the name-ordered results the e2e tests assert on. If sequential is a deliberate contract, the docs sentence should say why; otherwise this is an easy win.

*AI-assisted — Tool: Claude Code; model: Anthropic/claude-fable-5-1; version: unavailable.*

src/cli/doctor/project.rs
```diff
@@ -0,0 +1,216 @@
...
+    checks: Vec<CheckResult>,
+}
+
+impl Project {
+    pub(crate) async fn run(self, parent_json: bool) -> Result<()> {
+        let report = match self.check().await {
+            Ok(report) => report,
+            Err(err) => Report {
+                checks: vec![CheckResult {
+                    name: "configuration".into(),
```
jdx: **Config failures as a fake check named `configuration`.** Parent `mise doctor --json` reports this class in top-level `errors`/`warnings` arrays (doctor/mod.rs:302-309); here it is a `checks[]` entry with `source: null`. Nothing reserves the name, so a user `[doctor.checks.configuration]` produces a same-named entry, and a consumer cannot tell "no checks ran" from "one of N failed" except by the magic name plus `source == null`. A top-level `error` field on `Report` (or rejecting the reserved name at parse time) would be cleaner.

*AI-assisted — Tool: Claude Code; model: Anthropic/claude-fable-5-1; version: unavailable.*

jdx (commented):
Reviewed the full diff plus the enclosing modules. Ten findings below as inline comments, most severe first. Summary:

1. **schema/mise.json** — the `doctor` block landed inside `$defs.task_dependency_item` instead of the root `properties`, so `[doctor.checks.*]` gets no schema validation.
2. **project.rs:125** — a `ToolsetBuilder::build` error collapses the whole report into one `configuration` entry, contrary to the docs' per-check error promise.
3. **config/doctor.rs:25** — `os` re-implements the shared `os_selector_matches` selector; `darwin`, scalar `os`, and `os/arch` are rejected here but accepted by `[tools]`/`[bootstrap.packages]`/dotfiles.
4. **config/doctor.rs:23** — `shell` diverges from the task/hook `Option<String>` model and skips `maybe_no_profile`.
5. **project.rs:111** — global/system config checks run with cwd `$HOME`.
6. **config/doctor.rs:7** — `deny_unknown_fields` on the `[doctor]` container fails every mise command on a typo or a future key.
7. **project.rs:141** — `os = []` silently skips the check everywhere.
8. **project.rs:104** — safe mode exits 1 even with no checks configured, contrary to the docs.
9. **project.rs:129** — checks run strictly sequentially; a `jobs`-capped concurrency would preserve ordering.
10. **project.rs:57** — config failures are a fake check named `configuration` rather than a top-level error field.

Verified along the way and fine: config precedence via `config_files.iter().rev()`, `full_env` carrying the pristine env on Windows, `read_isolated` behaviour unchanged, trust gating via `is_safe_config_body`, `UNCLASSIFIED` effects placement, and the generated usage/man/docs files.

*AI-assisted — Tool: Claude Code; model: Anthropic/claude-fable-5-1; version: unavailable.*

schema/mise.json
jdx: Fixed in 7d681e62f: the doctor definition is referenced from root properties and removed from task dependencies. The schema e2e validates a real root-level doctor block and rejects missing run, unknown check fields, and doctor nested under a task dependency.

src/cli/doctor/project.rs
jdx: Addressed in 075abcac3. Toolset construction and full_env now share the retained environment Result. A build error produces an error for each applicable check while excluded checks remain skipped. The regression uses MISE_DUMMY_VERSION=-1 and asserts the full report.

src/config/doctor.rs
jdx: Addressed in 075abcac3. Switched to scalar-or-list string selectors, the shared os_selector_matches matcher, and the shared os_filter schema. Coverage includes darwin/win aliases and OS/architecture pairs. Explicit empty or malformed selectors are rejected; omitted selectors run everywhere.

src/config/doctor.rs
jdx: Addressed in 075abcac3. Shell overrides now use strings, split_shell_command, and maybe_no_profile, like hooks and tasks. Updated cmd.exe/PowerShell fixtures and docs. Explicit empty or malformed shells remain check errors rather than falling back to another interpreter; tests verify neither executes the probe.

src/cli/doctor/project.rs
jdx: Addressed in 075abcac3. Checks now use project_root() with invocation-cwd fallback for global/system configs. Regression coverage invokes doctor from a nested project directory and verifies project, global, and relative global directories independently. Documented the fallback.

src/config/doctor.rs
jdx: Addressed in 075abcac3. Removed deny_unknown_fields from the doctor container. Unknown container options no longer prevent mise env or project diagnostics from loading (both covered). Kept strict check leaves deliberately to catch misspelled execution controls such as timeuot, following the newer bootstrap container/leaf convention; documented that choice.

src/cli/doctor/project.rs
jdx: Addressed in 075abcac3. Explicit os=[] is now rejected during deserialization and by the shared os_filter schema. Omission means run everywhere. Runtime and schema regressions cover the empty-list case.

src/cli/doctor/project.rs
jdx: Addressed in 075abcac3. The safe-mode guard now runs only when preparing an environment for applicable probes, using ensure_not_safe for the standard message. Empty and entirely excluded projects succeed; applicable checks report errors without executing. Tests cover empty, excluded, and mixed projects.

src/cli/doctor/project.rs
jdx: Addressed in 075abcac3. Checks now execute concurrently through a jobs-bounded stream, preserving sorted report order. Regression tests require two probes to overlap with MISE_JOBS=2 and verify serialized execution with jobs=1 or 0. The signal cleanup test also runs two probes and verifies both process trees are gone.

src/cli/doctor/project.rs
jdx: Addressed in 075abcac3. Added top-level errors: configuration-load failures return errors with checks empty. Successful configuration loading leaves errors empty and retains real check results, including a user check named configuration. Both cases have regression coverage and the JSON contract is documented.

src/cli/doctor/project.rs
```diff
@@ -0,0 +1,253 @@
...
+        let mut checks = BTreeMap::new();
+        // Config files are ordered from highest to lowest precedence. Replace
+        // whole named checks so a local command cannot inherit a stale remedy.
+        for (path, cf) in config.config_files.iter().rev() {
+            for (name, check) in cf.doctor_config().checks {
+                checks.insert(
+                    name,
+                    (
+                        path.clone(),
+                        cf.project_root().unwrap_or(std::env::current_dir()?),
```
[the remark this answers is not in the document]
jdx: Fixed in b5856b65c. The invocation directory is now looked up only in the None branch of project_root(), so project-scoped checks do not depend on getcwd(). The build, existing directory/configuration regression tests, scoped hk checks, and workspace Clippy all pass.

PR #13083, opened 2026-09-11
feat(brew): implement `mise bootstrap packages where` for keg-only packages

himkt:
Installing bootstrap packages leaves shell PATH configuration to the user. Add `mise bootstrap packages where brew:<formula>` to print an installed formula's stable, absolute `opt` root, so scripts can locate its executables without hardcoding the Homebrew prefix or a Cellar version.

### Use case: prefer the installed `brew:unzip`

After installing `brew:unzip`, a user wants their shell to use that formula's executable. They can locate its root and prepend its `bin` directory to PATH:

```sh
mise bootstrap packages apply brew:unzip
if package_root="$(mise bootstrap packages where brew:unzip)"; then
  export PATH="$package_root/bin:$PATH"
fi
```

The lookup returns a path such as `/opt/homebrew/opt/unzip` on macOS arm64. The guarded assignment changes PATH only when lookup succeeds, and the stable `opt` path follows upgrades that repoint the link. The same approach works for keg-only formulae whose executables are absent from the shared prefix's `bin` directory, and for formulae installed by either mise or Homebrew.

Lookup reads local records and reports missing or invalid installations with a nonzero exit status and empty stdout. It uses environment variables and global CLI options for settings, leaving configuration templates, credential helpers, package downloads, and startup housekeeping outside the query path. Includes documentation, generated CLI usage, and unit/process regressions for path validation and startup isolation.

### Validation

- 356 brew unit tests and 8 query unit tests passed.
- After review fixes, all 16 focused package-root tests and 15 process cases passed; one unsupported-platform case was skipped.
- Generated usage/completions and three CLI reference tests passed.
- Repository lint passed; temporary verification artifacts containing deliberately invalid TOML were excluded from the final lint rerun.
- Reviewer approved after two rounds. Local validation used macOS arm64; Linux and Windows remain unverified locally.

*AI-assisted — Tool: Codex; model: OpenAI/gpt-6-astra; version: unavailable.*

## Summary by CodeRabbit

* **New Features**
  * Added `mise bootstrap packages where <PACKAGE>` to locate an installed Homebrew formula’s stable root path.
  * Supports macOS and Linux, including keg-only and library-only formulae.
  * Added the `mise daemons` command family with start, stop, restart, listing, logs, status, and TUI operations.
  * Missing or invalid package installations return empty output with a nonzero status.

* **Documentation**
  * Added CLI help, manual, and usage documentation.

* **Tests**
  * Added coverage for lookups, invalid inputs, platform restrictions, symlinks, and configuration isolation.


himkt:
@coderabbitai resume

himkt:
@coderabbitai resume

himkt:
Not confident but https://github.com/jdx/mise/pull/13083#pullrequestreview-5184198532 is not related to this PR... PTAL @jdx.

PR #13094, opened 2026-09-11
fix(task): explain MISE_CONFIG_DIR boundary in monorepo root error

i-api:
Monorepos rooted under `$MISE_CONFIG_DIR` can never establish a project root, so `//` task paths fail even though a root `mise.toml` sets `monorepo_root = true`. The old error only said a root configuration was required, which sent reporters trace-diving (see #13076).

This keeps that message but appends the boundary when a declaring config was found and rejected for lack of a project root:

- lists each `monorepo_root = true` path with no project root
- when under `$MISE_CONFIG_DIR`, names it and explains configs there are treated as global configuration
- advises moving the monorepo outside `$MISE_CONFIG_DIR` or pointing `MISE_CONFIG_DIR` elsewhere

Generic no-root output is unchanged (no boundary hint).

Verification:

- `cargo fmt --check` clean
- `cargo check --bin mise --locked` exit 0
- `mise run test:e2e e2e/tasks/test_task_monorepo_config_dir_error e2e/tasks/test_task_monorepo_validation` passes (new test 6s, validation 1s)
- `mise run lint`: `cargo-fmt`, `shellcheck`, `shfmt`, `cargo-check` pass; only pre-existing missing-tool failures for `markdownlint`/`prettier`/`ajv` in this environment

*AI-assisted — Tool: opencode; model: opencode/muse-spark-1.3-contributor-free; version: unavailable.*

## Summary by CodeRabbit

- **New Features**
  - Added the `meli` terminal mail client to the package registry, including Linux support and version verification.
  - Improved monorepo configuration errors with clearer guidance when configuration files declare a monorepo root without a project root, including relevant configuration locations and boundary details.

- **Tests**
  - Added end-to-end coverage for monorepo configuration errors involving `MISE_CONFIG_DIR`.

### DIFF
--- a/e2e/tasks/test_task_monorepo_config_dir_error
+++ b/e2e/tasks/test_task_monorepo_config_dir_error
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+# Monorepo under $MISE_CONFIG_DIR is treated as global config and never
+# receives a project root, so // task paths cannot resolve there.
+# The error should name that boundary instead of only saying no root was found.
+# See https://github.com/jdx/mise/discussions/13076
+
+export MISE_TRUSTED_CONFIG_PATHS="$MISE_TRUSTED_CONFIG_PATHS:$MISE_CONFIG_DIR"
+
+MONOREPO="$MISE_CONFIG_DIR/my_monorepo"
+mkdir -p "$MONOREPO/workspace2"
+cat <<'TOML' >"$MONOREPO/mise.toml"
+monorepo_root = true
+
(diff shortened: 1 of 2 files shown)


jdx:
sorry, looking at this again i think this is really edge-casey and I don't think we actually need it. i know this contradicts what i said earlier

i-api:
> sorry, looking at this again i think this is really edge-casey and I don't think we actually need it. i know this contradicts what i said earlier

ok no problem 👍

PR #13138, opened 2026-09-13
fix(brew): support appdir DSL in tap cask metadata evaluation

Guria:
## What

Define `appdir` on `CaskMetadata` and `CaskFlightSteps` in the tap-cask metadata evaluator, answering Homebrew's default (`/Applications`).

## Why

`brew-cask:` packages from third-party taps hard-fail the packages phase when the cask interpolates `appdir` (e.g. `binary "#{appdir}/..."`): `method_missing` raises `unsupported cask metadata DSL`, and since third-party taps have no published API metadata to fall back on, bootstrap fails entirely. The install-time shim already defines `appdir` (`cask_shim.rb`); it was only missing from `tap_cask_metadata.rb`.

The metadata path never knows the real install location and the installer re-evaluates the cask via the shim, so a default-backed placeholder is sufficient here.

## Test

- Extended `extracts_cask_metadata_without_homebrew` to interpolate `appdir` in both `binary` and a flight-step `run` and assert the resolved path.
- Verified against the real-world failing cask (`umputun/apps/agterm`): baseline raises, this branch evaluates and emits the expected binary artifact.

Fixes #13136

*AI-assisted — Tool: pi; model: unavailable.*

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected macOS cask metadata to use relocatable application directory references.
  - Improved handling of application executable paths so they resolve consistently relative to the configured application directory.
  - Updated command metadata to preserve the correct relative path while retaining the appropriate installation base.


Guria:
Addressed all review findings in c99663522 (and it also picked up the branch update merging main):

- `appdir` now interpolates the relocatable `` marker instead of a literal `/Applications` path, so `MISE_BREW_CASK_OPT_APPDIR` and prefix-local appdirs keep working.
- `CaskFlightSteps#run` records `/`-prefixed commands relative to `base: :appdir` (defaulting the base when omitted), which is what `parse_run_command` accepts.

The unit-macos failures are nextest 1s timeouts on tests unrelated to this change (`cli::mcp`, `command_effects`, `aube_node_gyp`, `usage`; 4004/4015 passed) — looks like an overloaded runner; retriggering.

*AI-assisted — Tool: pi; model: unavailable.*

Guria:
The remaining `unit-macos` / `test-ci` failures are not from this PR: the run times out at the ~1s per-test cap on exactly the same test set seen on #13147 and #13165 (`cli::mcp::*`, `cli::command_effects::*`, `aube_node_gyp_bootstrap_trampoline`, `aqua_registry_wrapper`) — 4004/4015 pass, all failures are `TIMEOUT [~1.01s]` on unrelated spawn-heavy tests, reproducible across three independent branches. Looks like the macOS runners got slow enough to trip the per-test timeout repo-wide. The brew/tap tests touched here pass locally (13/13).

*AI-assisted — Tool: pi; model: unavailable.*

PR #13141, opened 2026-09-13
feat(lock): store Python and npm dependency graphs in native sidecars

jdx:
Python CLI tools can lock and replay their dependencies with uv, and both Python and npm dependency graphs now live in native sidecar files instead of expanding `mise.lock` into thousands of nested TOML lines.

```toml
[tools]
"pypi:black" = "24.10.0"
"npm:prettier" = "3.3.3"
```

Run `mise lock` (or `mise lock --upgrade` for an older lockfile), then commit `mise.lock` and the generated sidecars:

```text
.mise/locks/pipx-black/24.10.0/{pyproject.toml,uv.lock}
.mise/locks/npm-prettier/3.3.3/{package.json,aube-lock.yaml}
```

`mise.lock` records each directory's relative path and the SHA-256 of its native graph file. Projects using `.config/mise/` keep their sidecars there; local/environment lockfiles have separate cleanup scopes. Paths remain stable, option variants coexist, and removing an entry prunes its sidecars. Revision stays **2**, which has not shipped; revisions 0–1 retain their format until explicit upgrade. Unreleased inline revision-2 graphs remain readable and migrate on save.

`pypi:` is the preferred backend name; `pipx:` remains supported without warnings or filesystem migration, including settings and environment compatibility. Existing explicit spellings and registry aliases keep their identities.

Python graph locking requires an installed **uv >= 0.12.10** and published hashed wheels. Resolution retains universal Python/platform markers, extras, index authentication, and release-age policy without executing source builds or storing credentials. Installation validates the root and interpreter and replays with `uv sync --frozen` and source builds disabled. Standalone pipx and Git sources retain their legacy behavior outside graph locking; unsupported free-form installer arguments are rejected for graph locking.

Ordinary locks reuse graphs; `mise lock --bump pypi:black` refreshes dependencies even when Black's version stays unchanged. Valid external graph edits are accepted by ordinary `mise install` or `mise lock`; the actual graph bytes determine the new installation identity and recorded digest. `mise install --locked` rejects digest mismatches. Formatting edits also change identity. Environment resolution and installed-tool lookup use the recorded digest without reading sidecars.

Renovate 44.79.2's local pep621 extractor detected an actual generated sidecar project, its adjacent `uv.lock`, and the pinned fixture dependency. Native aube YAML is preserved byte-for-byte, but scanners that only understand npm lockfiles do not gain transitive aube visibility.

Validation on macOS: 177 lockfile unit tests, 38 Python unit tests, Python/npm fixture e2e tests, lockfile revision and pruning suites, task dry-run and outdated checks, full workspace Clippy, lint fixes, and documentation build passed. The Python/npm fixture tests, revision/rollback tests, 177 lockfile unit tests, 38 Python unit tests, and full Clippy check were repeated successfully after incorporating current main (aube 2.2.16). The 27 CLI lock tests also pass after fixing graphless variant pruning. Windows console-script and path assertions are included for CI.

Stacked on #13146, the small collection refactor required to keep lazy graph caches out of hash keys without Clippy exclusions.

---

> [!NOTE]
> **Medium Risk**
> Changes core lockfile persistence and Python/npm install replay paths; compatibility aliases reduce breakage, but digest/sidecar mismatches and graph-only installs can still surprise existing projects until they re-lock.
> 
> **Overview**
> Introduces **`pypi:`** as the preferred Python CLI backend name while keeping **`pipx:`** as a full alias (same installs, lock entries, and on-disk `pipx-*` layout). Settings move to **`pypi.uvx`** / **`pypi.registry_url`** with **`pipx.*`** compatibility, conflict detection, and targeted repair when aliases disagree in one file.
> 
> **Lockfile revision 2** no longer embeds huge npm/Python graphs in `mise.lock`. Entries now store a **sidecar path plus SHA-256 digest** (e.g. `.mise/locks/pipx-black/1.0.0/` with `pyproject.toml` + `uv.lock`, or npm `aube-lock.yaml`). `mise lock` generates/refreshes sidecars; **`mise install --locked`** replays frozen graphs and rejects digest mismatches; ordinary install/lock can accept edited sidecars and update identity.
> 
> **Python tools** with uv ≥ 0.12.10 get portable **wheel-only** graphs: `uv lock` during lock, **`uv sync --frozen`** during install, interpreter binding, extras support, and install dirs keyed by graph digest. Version-only pipx installs and unsupported `uvx_args`/`pipx_args` stay on the legacy path. **npm embedded-aube** graphs follow the same sidecar model with validation on load.
> 
> Docs, schema, registry build mapping, and e2e (including Windows) cover locking, bump, alias equivalence, and npm sidecar migration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 44c5d1c6f31cd01233854aaca4353b4da331f5d1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/backend/pipx/lock.rs
```diff
@@ -0,0 +1,529 @@
...
+        config: &Arc<Config>,
+        tv: &ToolVersion,
+        uv: &Path,
+        project: &Path,
+    ) -> Result<CmdLineRunner<'static>> {
+        let registry = self.get_registry_url(config).await?;
+        let base = registry.split("{}").next().unwrap_or(&registry);
+        let base = base
+            .trim_end_matches('/')
+            .trim_end_matches("/pypi")
```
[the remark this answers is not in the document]
jdx: Fixed in 4014cabbf. Private registry prefixes are preserved; only the public pypi.org endpoint maps to /simple/. The end-to-end fixture now uses /pypi/{}/json and verifies lock generation and frozen installation through /pypi/simple/.

src/backend/pipx.rs
```diff
@@ -853,15 +947,15 @@ impl PIPXBackend {
                 .await
         else {
             warn!(
-                "minimum_release_age is set for pipx:{} but could not determine uv version required to verify --exclude-newer support. Release-age filtering for transitive dependencies may not work as expected. See https://mise.jdx.dev/dev-tools/backends/pipx.html",
+                "minimum_release_age is set for pypi:{} but could not determine uv version required to verify --exclude-newer support. Release-age filtering for transitive dependencies may not work as expected. See https://mise.jdx.dev/dev-tools/backends/pipx.html",
```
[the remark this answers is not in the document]
jdx: Updated both diagnostics to link to the preferred pypi.html page in 4014cabbf. The old URL remains supported.

docs/dev-tools/backends/pypi.md
```diff
@@ -0,0 +1,234 @@
...
+Other syntax may work but is unsupported and untested.
+
+## Settings
+
+Set these with `mise settings set [VARIABLE]=[VALUE]` or by setting the environment variable listed.
+
+<script setup>
+import Settings from '/components/settings.vue';
+</script>
+<Settings child="pipx" :level="3" />
```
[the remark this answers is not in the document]
jdx: Changed the settings reference to child="pypi" in 4014cabbf and verified the documentation build.

src/backend/pipx/lock.rs
```diff
@@ -0,0 +1,529 @@
...
+            .trim_end_matches("/simple");
+        let index = format!("{base}/simple/");
```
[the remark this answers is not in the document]
jdx: Fixed in 4014cabbf. Private registry prefixes are retained when deriving the simple index. The end-to-end test now exercises the documented /pypi/{}/json layout for resolution and frozen replay.

src/backend/pipx/lock.rs
```diff
@@ -0,0 +1,529 @@
...
+                    .as_str()
+                    .split(['?', '#'])
+                    .next()
+                    .unwrap_or("")
+                    .rsplit('/')
+                    .next()
+                    .unwrap_or("");
+                if Self::version_from_distribution_filename(&self.tool_name(), filename).as_deref()
+                    == Some(&tv.version)
+                    && filename.ends_with(".whl")
```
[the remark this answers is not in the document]
jdx: Fixed in 4014cabbf by sharing URL-decoded filename extraction between version listing and graph metadata matching. Added a regression for a wheel containing an encoded +local version; all 31 focused Python unit tests pass.

src/backend/pipx/lock.rs
```diff
@@ -0,0 +1,529 @@
...
+            let url = url::Url::parse(s)?;
+            if !matches!(url.scheme(), "https" | "http")
+                || !url.username().is_empty()
+                || url.password().is_some()
+                || url.query().is_some()
+            {
+                bail!(
+                    "Python locks require credential-free HTTP artifact URLs; configure authentication outside the lockfile"
+                );
+            }
```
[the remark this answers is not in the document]
jdx: Retaining HTTP support intentionally. mise already supports explicitly configured HTTP package indexes, including local/private indexes; making only Python graph replay HTTPS-only would break those configurations and the controlled local index fixture. Locked wheels must have SHA256 hashes and uv verifies them during replay, and this validator rejects credential-bearing URLs. A broader transport-policy change is separate from this dependency-locking PR.

e2e-win/pypi_lock.Tests.ps1
```diff
@@ -0,0 +1,47 @@
...
+        New-Item -ItemType Directory -Path $project | Out-Null
+        Push-Location $project
+        $server = $null
+        try {
+            @'
+[tools]
+python = "3.12.11"
+uv = "0.12.10"
+'@ | Set-Content mise.toml
+            mise trust
```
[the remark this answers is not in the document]
jdx: Added an immediate exit-status assertion after mise trust in 741c93f88, before mise install can overwrite it.

src/backend/pipx/lock.rs
```diff
@@ -0,0 +1,529 @@
...
+        let names: Vec<String> = serde_json::from_str(names.trim())?;
+        if names.is_empty() {
+            bail!("pypi:{} exposes no executable scripts", self.tool_name());
+        }
+        let bin = tv.install_path().join("bin");
+        crate::file::create_dir_all(&bin)?;
+        for name in names {
+            if name.is_empty() || name.contains(['/', '\\']) || name == "." || name == ".." {
+                bail!("invalid Python entry point name");
+            }
```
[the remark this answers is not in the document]
jdx: Switched to the existing platform-aware is_plain_file_name validator in 741c93f88, preserving separator/dot checks and rejecting Windows drive prefixes before either join.

src/config/settings.rs
```diff
@@ -1194,6 +1195,36 @@ impl Settings {
...
+        fn merge<T: PartialEq>(
+            preferred: &mut Option<T>,
+            legacy: &mut Option<T>,
+            name: &str,
+        ) -> Result<()> {
+            if let (Some(a), Some(b)) = (preferred.as_ref(), legacy.as_ref())
+                && a != b
+            {
+                eyre::bail!("conflicting pypi.{name} and pipx.{name} settings");
+            }
```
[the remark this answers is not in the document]
jdx: Added a narrowly scoped pre-load path for pypi/pipx settings set/unset in 741c93f88. Normal settings consumers still reject conflicts. The end-to-end test verifies both repair commands against conflicting file aliases with --cd, and passed.

src/backend/pipx/lock.rs
```diff
@@ -0,0 +1,556 @@
...
+    let base = registry.split("{}").next().unwrap_or(registry);
+    let mut url = url::Url::parse(base)?;
+    if url.host_str() == Some("pypi.org") {
+        url.set_path("/simple/");
+    } else {
+        let path = url.path().trim_end_matches('/').trim_end_matches("/simple");
+        url.set_path(&format!("{path}/simple/"));
+    }
+    Ok(url.into())
+}
```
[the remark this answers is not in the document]
jdx: Extended the public-index mapping to pypi.org subdomains in 741c93f88. Regression cases cover Test PyPI and verify that an unrelated domain ending in notpypi.org retains its private path. All 31 focused Python unit tests passed.

jdx:
Addressed the outside-diff CodeRabbit finding in 565c427ed: simple-index tags are now matched without truncating quoted angle brackets. A regression checks raw >= and escaped &gt;= requirements in both single- and double-quoted attributes, including an encoded local-version wheel filename. All 32 focused Python unit tests and required lint fixes passed.

jdx:
Python and npm graphs now share lockfile revision 2 (7515e2909). Revision 2 has not shipped: the latest release, [v2026.9.6](https://github.com/jdx/mise/blob/v2026.9.6/src/lockfile.rs), still uses revision 1; #13131 introduced revision 2 afterward. This preserves the released-client rejection boundary without another unreleased format bump. Revision 0–1 lockfiles retain their format until explicit upgrade.

Validation passed: 171 lockfile unit tests; Python locking, npm/aube locking, format-upgrade, and pruning end-to-end suites; lint fixes; documentation build.

jdx:
Addressed the outside-diff graph-merge finding in 3e2a56311. A missing primary uv graph is inherited only when both backend identities are present and equal; an existing primary graph is preserved. The regression test covers matching, conflicting, and missing identities and retained request specifiers. All 172 lockfile unit tests and `mise run lint-fix` passed.

src/toolset/tool_version.rs
```diff
@@ -103,7 +107,17 @@ impl ToolVersion {
         opts: &ResolveOptions,
     ) -> Result<Self> {
         let install_env = request.options().core.install_env;
-        crate::env::with_install_env(install_env, Self::resolve_(config, request, opts)).await
+        let mut tv =
+            crate::env::with_install_env(install_env, Self::resolve_(config, request, opts))
+                .await?;
+        if tv.uv_lock.is_some() {
```
jdx, line 113: **Blocker: interpreter probe on every resolve.**

`resolve` now calls `bind_uv_python` whenever `uv_lock` is set. That builds the dependency toolset and spawns `python -I -c …` for every graph-locked pypi tool on every toolset resolution: `mise env`, `hook-env` on a cache miss, `ls`, `which`, `x`, shims. Each is a fork/exec plus interpreter startup, multiplied by the number of locked pypi tools. And because the probe uses `?`, a broken interpreter (or a stale PATH entry) fails the whole resolve and takes down `mise env` for unrelated tools.

Suggest deriving the identity without a subprocess: from the python `ToolVersion` in the dependency toolset (backend short + resolved version + install path), or persist an identity file in the install dir and discover installs by the `version~uv~*` prefix the way `aube_install_path_version` already does. At minimum make the probe lazy (install / satisfaction only) and non-fatal for resolution.

src/cli/args/backend_arg.rs
```diff
@@ -276,7 +276,10 @@ impl BackendArg {
         opts: Option<ToolVersionOptions>,
         resolution: BackendResolution,
     ) -> Self {
-        let pathname = short.to_kebab_case();
+        let short = unalias_backend(&short).into_owned();
```
jdx: **Blocker: user spelling is rewritten.**

Unaliasing `short` here means a user who wrote `"pipx:black"` sees `pypi:black` everywhere: `mise ls`, `mise.lock` keys, install-state manifests, `mise use` output. If `pipx:` is what's in their config we should keep using `pipx:`, including under the new lockfile revision.

Suggest keeping `short` as written and comparing via `canonical_backend_full` only where identity matters (`BackendArg` Eq/Hash, `tool_directory_name`, lockfile lookup), so both spellings resolve to the same install and lock entry without rewriting the user's text.

src/lockfile.rs
```diff
@@ -1165,7 +1185,26 @@ impl Lockfile {
                     "invalid lockfile format for tool {short}: expected array ([[tools.{short}]])"
                 ),
             };
-            lockfile.tools.insert(short, versions);
+            let short = crate::backend::unalias_backend(&short).into_owned();
```
jdx: **Blocker (same root cause as the `short` rewrite).**

`Lockfile::read` canonicalizes `[[tools."pipx:black"]]` to `pypi:black` for revision 0–1 files too, so the next save rewrites the key while the revision stays at 1. An older mise client sharing that `mise.lock` still has `pipx:black` in config, finds no matching entry, and silently loses the pin; it doesn't reject the file because the revision didn't change.

Lookup should accept either spelling, and the written key should stay whatever the file / config used.

src/config/settings.rs
```diff
@@ -1194,6 +1195,36 @@ impl Settings {
...
+    fn normalize_pypi_aliases(&mut self) -> Result<()> {
+        fn merge<T: PartialEq>(
+            preferred: &mut Option<T>,
+            legacy: &mut Option<T>,
+            name: &str,
+        ) -> Result<()> {
+            if let (Some(a), Some(b)) = (preferred.as_ref(), legacy.as_ref())
+                && a != b
+            {
+                eyre::bail!("conflicting pypi.{name} and pipx.{name} settings");
```
jdx, line 1240: **Blocker: conflict across layers should follow precedence, not error.**

This bails whenever `pypi.*` and `pipx.*` differ after all layers are merged. `MISE_PIPX_UVX=false` baked into a CI image plus `pypi.uvx = true` in the repo `mise.toml` now fails every command, `mise env` included, where env normally wins over file.

The per-file migration in `parse_settings_file` already gives file-vs-file precedence; env-vs-file should work the same way. Rejecting only same-source conflicts (both spellings in one file, or both env vars set to different values, which `test_pypi_lock` already covers) keeps the loud failure for the cases that are actually ambiguous.

src/backend/pipx/lock.rs
```diff
@@ -0,0 +1,586 @@
...
+        &self,
+        config: &Arc<Config>,
+        tv: &mut ToolVersion,
+    ) -> Result<()> {
+        let Some(python) = self.spawnable_dependency(config, None, "python").await else {
+            // Resolve on a clean machine before the install graph installs Python.
+            tv.uv_python = None;
+            return Ok(());
+        };
+        let identity = CmdLineRunner::new(&python).args(["-I", "-c", "import sys, sysconfig; print((sys.implementation.name, sys.version_info[:3], sysconfig.get_config_var('SOABI'), sysconfig.get_platform(), sys.executable))"]).read().await?;
```
jdx: **Concern: identity is stricter than what selects wheels.**

Identity = `(impl, version_info[:3], SOABI, platform, sys.executable)`. Two consequences:

1. When no python is in `[tools]`, `spawnable_dependency` falls through to `which_non_pristine_spawnable("python")`, so the identity is the machine's system python. A Homebrew/apt patch bump orphans every pypi install and `mise ls` reports them missing until reinstall, even though the venv would keep working.
2. `is_install_satisfied` returns false whenever `uv_python` is `None`, so installed tools show as missing any time python isn't discoverable.

Hashing `impl + major.minor + platform` (what actually drives wheel and marker selection), or the mise python tool version string, and dropping `sys.executable` would make this much less brittle.

src/backend/pipx/lock.rs
```diff
@@ -0,0 +1,586 @@
...
+        Ok(uv)
+    }
+
+    pub(crate) async fn bind_uv_python(
+        &self,
+        config: &Arc<Config>,
+        tv: &mut ToolVersion,
+    ) -> Result<()> {
+        let Some(python) = self.spawnable_dependency(config, None, "python").await else {
+            // Resolve on a clean machine before the install graph installs Python.
```
jdx: Nit: I believe `uv lock` still discovers an interpreter (with `--no-python-downloads` it errors if none is found); it just doesn't need the *configured* one. Worth rewording so nobody relies on "clean machine" literally.

src/backend/pipx/lock.rs
```diff
@@ -0,0 +1,586 @@
...
+        crate::file::write(project.join("uv.lock"), toml::to_string(&lock.graph)?)?;
+        ctx.pr
+            .set_message("installing frozen Python dependencies".to_owned());
+        self.uv_lock_command(&ctx.config, tv, &uv, &project)
+            .await?
+            .args([
+                "sync",
+                "--frozen",
+                "--no-build",
+                "--no-cache",
```
jdx: Question: `--no-cache` means every frozen install re-downloads every wheel (and again on each identity change). Hashes are verified from the lock regardless of cache, so is this intentional, or left over from debugging?

src/lockfile/generate.rs
```diff
@@ -633,6 +649,63 @@ pub(crate) async fn populate_aube_locks(
...
+            .and_then(|entry| entry.uv.as_ref())
+            && !force
+        {
+            backend.validate_uv_lock(tv, lock)?;
+            continue;
+        }
+        if backend
+            .spawnable_dependency(&config, None, "uv")
+            .await
+            .is_none()
```
jdx, line 725: **Concern: silent skip creates a loop.**

When uv isn't spawnable this `continue`s silently, `mise lock` writes a revision-2 file with no graph, and `mise install --locked` then fails with "has no uv dependency graph; run `mise lock`". A warning here (tool skipped because uv >= 0.12.10 wasn't found) would tell the user what to actually do.

src/backend/pipx.rs
```diff
@@ -300,7 +302,89 @@ impl Backend for PIPXBackend {
...
     }
 
+    async fn prepare_install_version(
+        &self,
+        ctx: &InstallContext,
+        mut tv: ToolVersion,
+    ) -> Result<ToolVersion> {
+        if let Some(lock) = &tv.uv_lock {
+            self.validate_uv_lock(&tv, lock)?;
+        } else if self.uv_lock_allowed(&tv)
```
jdx: **Concern: `--locked` falls through when uv is missing.**

If the lockfile has no graph *and* uv isn't spawnable, this `else if` is skipped and a `--locked` install proceeds down the version-only pipx path with no error. That contradicts "Missing graphs in revision-2 locked uv installs are errors" in the docs. Probably should bail under `ctx.locked` regardless of uv availability.

src/lockfile.rs
```diff
@@ -96,6 +96,24 @@ impl AubeLock {
     }
 }
 
+/// Lossless native uv lock plus the synthetic project used to resolve it.
+#[derive(Debug, Clone, Default, Serialize, Deserialize, PartialEq)]
+#[serde(deny_unknown_fields)]
+pub(crate) struct UvLock {
```
jdx, line 120: Question: the full universal `uv.lock` (every package, every wheel for every platform) is embedded per tool version. For something like `ansible` or `aws-cli` that's thousands of lines per entry, and `mise lock --bump` rewrites all of it. Aube graphs set the precedent, but Python wheel lists are much larger. Would pruning wheels to the platforms recorded in `mise.lock` be acceptable, or is full portability the explicit goal here?

jdx (commented):
Solid direction, and the e2e fixture is good (real wheels, publish/corrupt toggles, marker and extras coverage). I'm not repeating the bot findings that are already on the thread.

Three items I'd treat as blocking before merge, then a handful of concerns and questions; details inline.

**Blocking**
1. `ToolVersion::resolve` now spawns `python` for every graph-locked pypi tool on every toolset resolution (`mise env`, `hook-env`, `ls`, `which`, `x`, shims). Startup cost plus a new way for `mise env` to fail outright.
2. The `pipx:` spelling is silently rewritten to `pypi:` in `BackendArg::short`, `mise.lock` keys, install-state manifests, and `mise ls`. If the user wrote `pipx:` we should keep using `pipx:`, including in the new lockfile revision. Older clients sharing a rev-1 lockfile lose their pin when the key is rewritten.
3. `pypi.*` vs `pipx.*` conflicts across layers are a hard error instead of normal precedence, so an env var in CI plus a repo setting fails every command.

**Concerns / questions**
- Interpreter identity includes `version_info[:3]` and `sys.executable` and falls back to system python; a patch upgrade orphans every install.
- `mise lock` silently skips graphs when uv isn't available, then `mise install --locked` errors telling you to run `mise lock`.
- `--locked` install falls through to a version-only install when uv is missing and the lockfile has no graph.
- `--no-cache` on `uv sync`, the "clean machine" comment, and lockfile size from embedding full universal graphs.

src/toolset/tool_version.rs
jdx: Fixed in 93d2e8f7c. Ordinary resolution no longer probes Python. Configured Python identity comes from its resolved mise backend/version/install path; system-Python environments are rediscovered from saved installation metadata. The only system interpreter probe is during installation, and missing/broken Python cannot fail this resolution step. A regression test discovers an existing environment with a nonexistent interpreter and rejects a different graph.

src/cli/args/backend_arg.rs
jdx: Fixed in 93d2e8f7c. BackendArg retains explicit pipx spelling while equality, ordering, hashing, and filesystem identity treat the names as aliases. Output, install-state discovery, and lockfile keys retain that spelling. Cross-alias locking/bumping and installation lookup pass the real-wheel e2e fixture.

src/lockfile.rs
jdx: Fixed in 93d2e8f7c. Reading retains the original short key; lookup accepts the alternate spelling. Saving legacy revisions emits the compatible pipx backend name as well. The revision-1 round-trip regression checks the key/backend remain pipx and the revision is unchanged, while pypi lookup still finds the pin.

src/config/settings.rs
jdx: Fixed in 93d2e8f7c. Environment aliases are normalized as their own layer before file settings are merged, so environment values override either file spelling. Same-file and same-environment contradictions still fail. The e2e fixture covers both env/file spelling directions plus conflict repair.

src/backend/pipx/lock.rs
jdx: Fixed in 93d2e8f7c. Configured Python uses its resolved mise identity. System Python records implementation, major/minor, ABI, and platform at install time, without sys.executable or patch version in the compatibility identity. Later resolution restores that saved environment identity without requiring Python on PATH; uv is no longer required merely to recognize an installed graph environment.

src/backend/pipx/lock.rs
jdx: Updated in 93d2e8f7c. The misleading clean-machine comment is gone, and the docs explicitly say uv needs a discoverable interpreter to generate locks, but it need not be the configured tool interpreter.

src/backend/pipx/lock.rs
jdx: Removed --no-cache in 93d2e8f7c. Ordinary frozen installs now reuse uv’s artifact cache. The checksum regression explicitly uses an empty UV_CACHE_DIR to exercise the modified download and still verifies Hash mismatch.

src/lockfile/generate.rs
jdx: Added a warning in 93d2e8f7c identifying the skipped tool, uv >= 0.12.10 requirement, and the need to install uv and run mise lock again. Existing standalone-pipx/version-only locking remains available.

src/backend/pipx.rs
jdx: Fixed in 93d2e8f7c. Missing-graph validation no longer depends on uv being spawnable, including installation-satisfaction checks. The e2e regression removes uv and verifies --locked fails with the missing-graph diagnostic rather than falling through.

src/lockfile.rs
jdx: Full portability is an explicit goal of this first version, so I retained the universal native uv graph. Pruning to recorded platforms would discard targets that the committed graph can currently replay without resolving again. The docs and PR description now call out the potentially substantial lockfile-size cost for dependency-heavy tools; this can be revisited as an explicit coverage option.

jdx:
@coderabbitai resume

The latest review fixes are ready for another pass on 93d2e8f7c.

jdx:
bugbot run

Retrying the current-head review because the previous check reports it failed to post one finding. Please publish that finding so it can be addressed.

src/backend/pipx.rs
jdx: Verified 93d2e8f7c: `--locked` now fails correctly without uv. But dropping the uv gate from `is_install_satisfied` as well introduces a reinstall loop.

With `lockfile = true` and **no uv installed**, the `else if self.uv_lock_allowed(tv)` branch now always returns `Ok(false)` for a version-only PyPI install: `prepare_install_version` returns early without a graph, the pipx path installs the tool, and the next `mise install` reinstalls it again because satisfaction never becomes true. `mise ls` reports the tool as missing between runs. That contradicts "standalone pipx retains version-only behavior" in the description, and the new e2e only checks that `--locked` fails without uv, not that a plain install settles.

Suggested fix: keep the `ctx.locked` bail independent of uv, but in `is_install_satisfied` fall through to `is_version_installed` when uv is not spawnable (or more generally whenever `prepare_install_version` would return without a graph). An e2e that installs twice without uv and asserts the second run is a no-op would pin it.

Two minor notes, not blocking:
- `configured_python_identity` still builds and resolves the dependency toolset for every locked pypi tool on every resolution. In-process and far cheaper than the fork, but repeated work that could be memoized per backend.
- On save, a revision-1 lockfile keyed `pypi:black` gets its `backend` rewritten to `pipx:black` while the key stays `pypi:`. Harmless for old clients (they can't use that key either way), just inconsistent.

The other nine threads look addressed as described; resolving those.

src/lockfile/generate.rs
```diff
@@ -633,6 +648,66 @@ pub(crate) async fn populate_aube_locks(
...
+        {
+            warn!(
+                "{} dependency graph skipped: uv >= 0.12.10 is unavailable; install uv and run `mise lock` again",
+                ba.short
+            );
+            continue;
+        }
+        let graph = backend.resolve_uv_lock(tv).await?;
+        lockfile.set_uv_lock(&ba.short, &tv.version, &backend_name, &options, graph)?;
```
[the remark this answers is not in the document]
jdx: Keeping the explicit version error here intentionally. The requested feature requires uv >= 0.12.10 for graph creation and replay; when uv is selected but too old, silently producing a version-only lock would hide that the dependency graph was not locked. The diagnostic tells the user the minimum version. The no-uv case retains standalone-pipx compatibility with an explicit skipped-graph warning; users who want that behavior with uv installed can select `uvx = false`. Existing revision-0/1 and explicit standalone-pipx paths remain version-only.

jdx:
## Plan: move dependency graphs out of `mise.lock` into native sidecar lockfiles

For PR jdx/mise#13141 (`codex/pypi-uv-lock`). Applies to **both** the uv graphs this PR adds and the aube graphs from #13131. Lockfile revision stays at **2** (revision 2 has not shipped; v2026.9.6 is still revision 1), so this is a change to what revision 2 means, not a new revision.

### Why

Embedding the full universal `uv.lock` (and the aube graph) as nested TOML inside `mise.lock` has two costs:

1. **Visibility.** Dependabot, Renovate, GitHub's dependency graph, and advisory scanners only parse files literally named `uv.lock` / `pyproject.toml` (and the npm equivalents). Wrapped inside `mise.lock` the transitive dependencies are invisible to all of them.
2. **Size and diff noise.** A mid-size CLI produces thousands of lines under `[[tools."pypi:x".uv.graph.package]]`, and `mise lock --bump` rewrites all of it inside the one file everyone reviews.

Storing each graph as the native lockfile it already is, in its own directory, fixes both. `mise.lock` keeps the top-level pin and a pointer.

### Target layout

```
mise.lock
.mise/locks/pipx-black/24.10.0/pyproject.toml
.mise/locks/pipx-black/24.10.0/uv.lock
.mise/locks/npm-prettier/3.3.3/package.json   # or .config/mise/locks/… in that layout
.mise/locks/npm-prettier/3.3.3/aube-lock.yaml
```

- **Follow the project's existing config layout.** `lockfile_path_for_config` already places the lockfile next to the config (`.config/mise/config.toml` → `.config/mise/mise.lock`, `.mise/config.toml` → `.mise/mise.lock`, bare `mise.toml` → `./mise.lock`, `conf.d` → its parent, monorepo → root). Derive the sidecar root from `<lockfile_dir>` with one helper, `sidecar_root(lockfile_path) -> PathBuf`:
  - lockfile is inside a `.config/mise/` directory → `<lockfile_dir>/locks/` (i.e. `.config/mise/locks/`)
  - lockfile is inside a `.mise/` directory → `<lockfile_dir>/locks/` (i.e. `.mise/locks/`)
  - lockfile is inside `.config/` (the `.config/mise.toml` single-file layout) → `<lockfile_dir>/mise/locks/` (i.e. `.config/mise/locks/`)
  - otherwise (lockfile at the project or monorepo root next to `mise.toml`) → `<lockfile_dir>/.mise/locks/`

  Never create `.mise/` in a project that already uses `.config/mise/`. Unit-test the helper for all four cases.
- For any lockfile name other than `mise.lock` (`mise.local.lock`, `mise.<env>.lock`, `mise.<env>.local.lock`) append `<stem>/` to the root (e.g. `.config/mise/locks/mise.local/`) so garbage collection can be scoped per lockfile.

  Layout examples:

  ```
  # bare mise.toml project            # .config/mise layout
  mise.toml                           .config/mise/config.toml
  mise.lock                           .config/mise/mise.lock
  .mise/locks/pipx-black/24.10.0/     .config/mise/locks/pipx-black/24.10.0/
  ```
- **One directory per lockfile entry, and the recorded `path` is sticky.** Sidecars map 1:1 to entries in `tools.<short>`, never to packages. `black` (alias or registry shorthand) and `pypi:black` are two entries and two directories today for installs, so they are two sidecars too, even if their `uv.lock` contents match. The only merge is `pypi:` vs `pipx:`, which `tool_directory_name` already collapses to `pipx-<name>`, matching the alias-aware `tools_for` lookup.
- **Default name, then disambiguate, then persist.** On first write the default is `<tool_directory_name(short)>/<version>`. Append `~<first 8 hex of sha256(canonical TOML of {backend, options})>` when the entry's `options` map is non-empty, or when another entry with the same short and version already holds the plain name. Every option that survives `resolve_lockfile_options` can change the graph (`extras` changes the root requirement, `registry_url` changes index URLs in `uv.lock`, `uvx = false` means no graph), and one `short` can lock two backends at the same version (registry fallbacks resolved differently on two machines), so both go into the hash. Reuse `hash_canonical_toml` from `src/lockfile.rs`.
- Once written, the `path` in the entry is authoritative: later saves reuse it and never rename, so adding a variant later does not move existing directories. `generate` already carries `uv`/`aube` forward from the previous lockfile by matching version, backend, and options; carry `path` and `digest` in that same match. GC works from recorded paths.
- Paths recorded in `mise.lock` always use forward slashes, relative to the lockfile's directory.

### `mise.lock` schema (revision 2)

```toml
[[tools."pypi:black"]]
version = "24.10.0"
backend = "pypi:black"
uv = { path = ".mise/locks/pipx-black/24.10.0", digest = "sha256:…" }

[[tools."npm:prettier"]]
version = "3.3.3"
backend = "npm:prettier"
aube = { path = ".mise/locks/npm-prettier/3.3.3", digest = "sha256:…" }
```

`path` names the directory. `digest` is the SHA-256 of the graph file's bytes (`uv.lock` or `aube-lock.yaml`), stored so install identity can be computed without opening the sidecar (see "Lazy loading"). The files inside are fixed by kind:

| key    | files                                   | source of truth today                                                        |
|--------|-----------------------------------------|------------------------------------------------------------------------------|
| `uv`   | `pyproject.toml`, `uv.lock`             | `UvLock { project, graph }` in `src/lockfile.rs`; written by `install_uv_lock` in `src/backend/pipx/lock.rs` |
| `aube` | `package.json`, `aube-lock.yaml`        | `AubeLock { graph }` (native YAML, currently converted to TOML via `from_yaml`/`to_yaml`); `write_aube_root_dependency` + `resolve_aube_lock` in `src/backend/npm.rs` |

Store the native formats verbatim: `uv.lock` as the TOML uv wrote, `aube-lock.yaml` as the YAML aube wrote. Do not round-trip through `toml::Table` on disk. Only persist files that frozen replay needs; the aube embed project config that `write_aube_embed_project` derives from tool options and `before_date` is still written fresh at install time, not persisted.

Inline `uv = { project = …, graph = … }` and `aube = { graph = … }` tables: keep the existing read path so a developer with a main-built revision-2 lockfile is not broken, but never write them. Add a debug-level note on read that the inline form is being migrated; the next `mise lock` or auto-lock save emits sidecars. Remove the inline read path before revision 2 ships if a single format is preferred; that is a one-line decision later.

### Code changes

#### `src/lockfile.rs`

- Replace `LockfileTool.uv: Option<UvLock>` / `.aube: Option<AubeLock>` with a lazy handle, one generic type for both kinds:

  ```rust
  pub(crate) enum GraphRef<T> {
      /// Freshly resolved this run; not yet written to disk.
      Inline(T),
      /// Referenced from mise.lock; body read on first `load()`.
      Sidecar { dir: PathBuf /* absolute */, digest: String, cell: OnceLock<Result<T>> },
  }
  impl<T> GraphRef<T> {
      fn identity(&self) -> String;          // digest; for Inline, computed once from the serialized bytes
      fn load(&self) -> Result<&T>;          // read + parse + verify digest, memoized
  }
  ```

  `ToolVersion.uv_lock` / `.aube_lock` hold the same `GraphRef`. `uv_install_identity` / `aube_install_identity` call `identity()` only, so `mise env`, `hook-env`, `ls`, `which`, `x`, and shims never open a sidecar. `identity()` replaces the current `hash_canonical_toml` identity for both kinds; existing main-built installs get a new identity suffix once, which is acceptable because revision 2 is unreleased.
- `LockfileTool` TOML conversion (`TryFrom<toml::Value>` around line 4355 and `into_toml_value` around line 4418): accept `{ path = "…" }` for both keys; emit only the path form. The path is not part of `lockfile_entry_key`; it is derived from short/version/options at write time, so re-deriving must produce the same path (deterministic).
- `Lockfile::read(path)`: after parsing, resolve each `path` relative to `path.parent()` into a `GraphRef::Sidecar` **without touching the filesystem**. `read` must stay as cheap as it is today: `read_all_lockfiles` runs on every backend-identity lookup.
- `GraphRef::load()` is where the body is read: `UvLock.project` from `pyproject.toml`, `.graph` from `uv.lock`, `AubeLock` via `AubeLock::from_yaml`. It verifies the graph file's bytes against `digest` and fails with a message naming the directory and suggesting `mise lock` on a missing directory, unparseable file, or digest mismatch. Callers that need the body: `validate_uv_lock` at install time, `install_uv_lock`, the npm frozen install, and `populate_*_locks` when deciding whether an existing graph is reusable. `is_install_satisfied` must **not** load: an install directory whose `~uv~`/`-aube-` suffix matches the identity was built from that graph, so the identity comparison already covers it. Move the `validate_uv_lock` call out of `is_install_satisfied` and keep it in `prepare_install_version`.
- Under `--locked`, a `load()` failure is fatal (same class as the existing "has no uv dependency graph" error). Outside `--locked`, `mise lock` treats an unreadable sidecar as missing and regenerates it; a digest mismatch with a readable file means someone edited the sidecar (Renovate, a human), and `mise lock` re-reads it, re-validates the root requirement and hashes, and rewrites the `digest` rather than regenerating the graph. That keeps external edits usable while `mise install --locked` still refuses a sidecar that disagrees with `mise.lock`.
- Ordinary `mise install` (not `--locked`) on a digest mismatch: `prepare_install_version` calls `load()`, which reports the mismatch; the backend validates the edited graph exactly as it would a fresh one (`validate_uv_lock` / aube strict frozen mode), and **the install identity must be computed from the actual file bytes, not the stale recorded digest**, so the install directory name matches what was installed. The auto-lock pass at the end of that same run then writes the new digest to `mise.lock`. Never install an edited graph into a directory named after the old digest. Error text for the `--locked` case: name the sidecar path and say `run \`mise lock\` to accept the edited graph`.
- `save` / `prepare_write` (line ~1206): `GraphRef::Inline` entries are serialized to their sidecar files and become `Sidecar { path, digest }` in the written `mise.lock`; `Sidecar` entries whose directory exists are left alone (no read, no write). Write sidecars **before** `mise.lock` so a crash never leaves a pointer to nothing. Each sidecar file is written atomically (temp + rename) and skipped when content is byte-identical, matching how `prepare_write` avoids rewriting an unchanged lockfile. Whatever `PreparedWrite` is used for (dry-run diff / unchanged detection) must include sidecar changes.
- Garbage collection on save: after writing, remove any directory under this lockfile's sidecar root that is not referenced by any entry being written. Never touch a different lockfile's root (that is what the per-stem subdir is for). Remove empty parent dirs.
- `merge_lockfile_for_lookup` / `merge_lockfile_preserving_root` / `read_all_lockfiles`: no change in logic; they operate on the already-loaded in-memory graphs. Monorepo sub-lockfiles get their own `.mise/locks/` next to them.
- Keep the existing revision-2 boundary check (`lockfile_version < 2 && uv.is_some()` bail) as-is.

#### `src/lockfile/generate.rs`

- `populate_uv_locks` / `populate_aube_locks` / `set_uv_lock` / `set_aube_lock`: unchanged; they set the in-memory graph. Path derivation happens in `save`.
- The carry-forward loops (`entry.uv = previous.tools_for(...)…`) keep working because `read` loaded the previous sidecars.

#### `src/backend/pipx/lock.rs`, `src/backend/npm.rs`

- `resolve_uv_lock` / `resolve_aube_lock` return `GraphRef::Inline`. `install_uv_lock` still writes `pyproject.toml` and `uv.lock` into `<install>/.mise-uv` from the in-memory `UvLock`; npm still writes `aube-lock.yaml` into the install path. Optionally have `install_uv_lock` copy the sidecar files verbatim when they exist, but this is not required for correctness.
- `UvLock` currently holds `graph: toml::Table`. Keep the raw `uv.lock` text alongside the parsed table (`graph_text: String`, populated by `resolve_uv_lock` and `load()`), and write that text verbatim to the sidecar so the on-disk bytes are exactly what uv produced. **Identity is the byte digest of that file, full stop.** Do not keep a second, formatting-insensitive identity: it would require parsing the sidecar, which defeats lazy loading, or a second stored digest that has to stay in sync. uv and aube emit deterministic output for the same graph, so formatting churn only appears when something outside mise rewrote the file, and that is exactly when a reinstall is wanted. Delete the existing `hash_canonical_toml`-based `identity()` methods on `UvLock` / `AubeLock` once `GraphRef::identity()` replaces them.

#### `src/toolset/tool_version.rs`

- `uv_install_identity` / `aube_install_identity` use `GraphRef::identity()` (the digest). `from_lockfile` copies the `GraphRef`. `legacy_aube_install_path_version` keeps reading `aube-lock.yaml` from the install dir for the pre-revision-2 `-aube-` suffix; leave it.

### Tests

Unit (`src/lockfile.rs`):
- Round-trip: a `Lockfile` with one uv entry and one aube entry saved to a tempdir produces `mise.lock` containing only `{ path = … }` for both keys, plus the four sidecar files; `read` yields equal `UvLock` / `AubeLock`.
- Options variant: two entries, same version, different `options`, get distinct sidecar dirs and both survive a round-trip.
- GC: save entry A and B, then save only A; B's directory is gone, A's is byte-identical (mtime unchanged).
- Lazy: `Lockfile::read` on a lockfile with sidecar pointers performs no reads under the sidecar root (assert with a nonexistent `path`; `read` succeeds, `identity()` returns the digest, `load()` errors).
- Digest mismatch: edit `uv.lock`, `load()` errors naming the directory; `mise lock` then rewrites `digest` without regenerating the graph.
- Sticky path: write entry A, add variant B at the same version, save; A keeps its plain path and B gets the `~hash` suffix; drop B, save; A's path unchanged and B's directory GC'd.
- Non-default lockfile name uses the `<stem>` subdir and its GC does not touch `.mise/locks/` siblings.
- Inline form: an existing inline `uv = { project, graph }` table still reads (until removed).
- Path is forward-slash on Windows (`cfg(windows)` unit test or string assertion on the generated path).

E2E:
- `e2e/backend/test_pypi_lock`: replace the `assert_contains "cat mise.lock" 'uv.graph'` / `'mise-lock-marker'` assertions with: `mise.lock` contains `uv = { path = ".mise/locks/pipx-mise-lock-cli/1.0.0" }`; `.mise/locks/pipx-mise-lock-cli/1.0.0/uv.lock` exists and contains `mise-lock-marker`; the `extras = "feature"` step produces a `1.0.0~<hash>` directory; after `cp original.toml mise.toml` + `mise lock` the variant directory is GC'd. Existing `cmp first.lock mise.lock` checks stay valid (the pointer does not change on a no-op relock; the sidecar does not either).
- `e2e/backend/test_npm_aube`: lines 75/85 currently assert `tools."npm:cowsay".aube.graph.packages` in `mise.lock`; change to assert the `aube = { path = … }` pointer plus `.mise/locks/npm-cowsay/<version>/aube-lock.yaml` containing `packages`. Line 92 (`assert_not_contains '.aube.graph.'` after a package-manager change) becomes: no `aube` key **and** the sidecar directory was removed.
- `e2e/lockfile/test_lockfile_version` and the pruning suite: confirm `mise lock` pruning a tool also removes its sidecar dir.
- `e2e-win/pypi_lock.Tests.ps1`: assert the sidecar path in `mise.lock` uses `/`.

### Docs

- `docs/dev-tools/mise-lock.md`: the aube section (~lines 49, 141–146, 171) and the new Python section describe revision 2 as storing graphs **in native lockfiles under `.mise/locks/`**, referenced by `path`. Mention: commit `.mise/locks/` alongside `mise.lock`; if `mise.local.lock` is gitignored, also ignore the matching `locks/mise.local/` directory; Dependabot/Renovate see `uv.lock` and can bump it, after which `mise install` treats the edited graph as a new install identity and `mise lock` is not required to accept it (validation still checks the root requirement and wheel hashes).
- `docs/dev-tools/backends/pypi.md` and `npm.md`: one paragraph each pointing at the layout, and the `uv tree` / `uv lock --upgrade-package` debugging tip for the pypi dir.
- PR title/description: rewrite to describe the final result (sidecar layout, both backends, revision unchanged), per the repo guide's release-note rule.

### Non-goals

- No dedup of identical graphs across aliases. Two entries, two directories.

- No revision bump. No lockfile compatibility window beyond tolerating the unreleased inline form on read.
- No pruning of the universal graph to recorded platforms; portability stays as decided on the PR.

### Verification

```
mise run build
cargo test --bin mise lockfile
cargo test --bin mise pipx
mise run test:e2e e2e/backend/test_pypi_lock e2e/backend/test_npm_aube e2e/lockfile/test_lockfile_version
mise run lint-fix
mise run docs:build
```

Also do one manual check that a repo with `mise.lock` + `.mise/locks/` shows the uv dependencies in GitHub's dependency graph (or run Renovate/Dependabot's file detection locally) to confirm the visibility goal actually holds with this directory layout.

### Optional aube-side work (a separate jdx/aube PR is acceptable)

Nothing in this plan requires an aube change: mise stores the `aube-lock.yaml` that `aube::embed::install_with_overrides` already writes, and frozen replay reads it from the install dir as today. Two things would make the npm side match the uv side more fully, and belong in aube rather than mise:

1. **Scanner-readable output.** `aube-lock.yaml` is not a format Dependabot, Renovate, or GitHub's dependency graph recognize. If aube can emit (or mise can ask it for) a `package-lock.json`-compatible file alongside `aube-lock.yaml`, write that into the same sidecar directory purely for scanners; mise keeps replaying from `aube-lock.yaml`. Only do this if the extra file is a faithful projection of the same graph, otherwise it misleads the tools it is meant to inform.
2. **Stable serialization.** If aube's YAML output ordering or formatting is not deterministic for the same graph, sidecar diffs will churn on no-op relocks. Verify with two consecutive `mise lock` runs producing a byte-identical `aube-lock.yaml`; fix ordering in aube if not.

If an aube PR is opened, bump `aube` and `aube-registry` together in this PR per the repo guide's "Updating embedded aube" checklist, and run the focused aube checks listed there.

_Revised: sidecars are lazy-loaded via a digest in the pointer, paths are sticky per entry, the hash covers backend plus options, the byte digest is the only identity, and ordinary installs accept validated external edits while `--locked` rejects them._

src/backend/pipx.rs
```diff
@@ -300,7 +302,79 @@ impl Backend for PIPXBackend {
...
+        config: &Arc<Config>,
+        tv: &ToolVersion,
+        check_symlink: bool,
+    ) -> Result<bool> {
+        if let Some(lock) = &tv.uv_lock {
+            self.validate_uv_lock(tv, lock)?;
+            if tv.uv_python.is_none() {
+                return Ok(false);
+            }
+        } else if self.uv_lock_allowed(tv) {
```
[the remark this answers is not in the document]
jdx: Fixed in bdd599aa8: graphless non-locked installations fall back to ordinary satisfaction when uv is unavailable. Locked requests still fail closed, and available uv still triggers graph creation.

src/backend/pipx/lock.rs
```diff
@@ -0,0 +1,677 @@
...
+        if let Some(identity) = self.configured_python_identity(config).await {
+            tv.uv_python = Some((PathBuf::new(), identity));
+            return;
+        }
+        Self::restore_system_uv_python(tv);
+    }
+
+    fn restore_system_uv_python(tv: &mut ToolVersion) {
+        // Installed environments remain usable without rediscovering a system Python.
+        let Ok(entries) = std::fs::read_dir(&tv.ba().installs_path) else {
```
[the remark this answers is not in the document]
jdx: Fixed in bdd599aa8: system Python metadata discovery scans the primary directory followed by shared install roots, using the same tool directory name and final install-path resolution as normal installation lookup.

src/cli/lock.rs
```diff
@@ -1682,10 +1688,13 @@ impl Lock {
...
             // Build map of tool args with explicit versions
-            let specified_versions: std::collections::HashMap<String, Option<ToolRequest>> = self
+            let specified_versions: std::collections::HashMap<
+                Arc<crate::cli::args::BackendArg>,
+                Option<ToolRequest>,
+            > = self
                 .tool
                 .iter()
-                .map(|t| (t.ba.short.clone(), t.tvr.clone()))
+                .map(|t| (t.ba.clone(), t.tvr.clone()))
```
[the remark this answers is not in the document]
jdx: Fixed in bdd599aa8: specified version requests are keyed, filtered, and retrieved by full backend identity.

src/toolset/install_state.rs
```diff
@@ -338,7 +337,7 @@ fn scan_tool_dir(
 
     let tool = InstallStateTool {
         short,
```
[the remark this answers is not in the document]
jdx: Fixed in bdd599aa8: the single-tool manifest lookup compares canonical shorts, retaining legacy pipx manifests and directories while accepting pypi requests.

src/backend/pipx/lock.rs
```diff
@@ -0,0 +1,699 @@
...
+                    .flatten()
+                    .filter_map(toml::Value::as_str)
+                    .map(str::to_owned)
+                    .collect::<std::collections::BTreeSet<_>>();
+                if requirements.len() != 1
+                    || requirement.get("name").and_then(toml::Value::as_str)
+                        != Some(Self::normalize_package_name(&self.tool_name()).as_str())
+                    || requirement.get("specifier").and_then(toml::Value::as_str)
+                        != Some(format!("=={}", tv.version).as_str())
+                    || extras != locked_extras
```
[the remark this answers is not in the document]
jdx: Fixed in 44c5d1c6f. Requested extras are trimmed before normalization, and recorded extras are normalized before comparison. The new regression test validates `postgres, S_3` against uv’s normalized names. All 38 focused Python tests pass.

src/lockfile/graph.rs
```diff
@@ -0,0 +1,656 @@
...
+            } else {
+                plain
+            }
+        };
+        self.referenced.insert(dir.clone());
+        if matches!(graph, GraphRef::Sidecar { dir: source, .. } if *source == dir && source.is_dir())
+        {
+            return Ok(graph.clone());
+        }
+        for (name, contents) in graph.load()?.files()? {
```
jdx, line 347: **Medium: a missing or unreadable sidecar makes every save of this lockfile fail.**

For a `Sidecar` whose directory is gone (or whose file no longer matches the digest) and that nothing replaced with an `Inline`, `prepare` falls through to `graph.load()?`, which errors with "run `mise lock` to repair it". That propagates out of `prepare_write`, so the *whole* lockfile save fails, including saves that have nothing to do with this tool.

Cases where nothing replaces it before save:
- `populate_uv_locks` hits the `uv >= 0.12.10 is unavailable` branch and `continue`s with the broken pointer still in place; the save then fails, so the warning's advice ("install uv and run `mise lock` again") is preceded by an error on the very same run, and every `mise install` auto-lock fails too.
- Entries not in the current toolset: monorepo `preserve_absent_tool_entries`, versions retained for another platform, or a variant the current host doesn't resolve. `populate_*_locks` only iterates `tools`, so those never get refreshed or regenerated.
- Someone cloned a repo that gitignored `.mise/locks/` by mistake and has no uv: `mise install` is now unusable until they hand-edit `mise.lock`.

Suggest: when `graph` is a `Sidecar` under `self.root` whose body cannot be loaded, keep the pointer as-is (it is still valid data), skip file writes for it, and `warn!` naming the directory. The install path already fails clearly on `load()`, and `mise lock` regenerates when it can. Only an `Inline` graph should ever be able to fail `prepare`. A unit test for "save succeeds with a dangling pointer and preserves it byte-for-byte" would pin this.

docs/dev-tools/mise-lock.md
```diff
@@ -677,3 +678,66 @@ the tool's installer/compiler.
...
+these native Python projects. Configure their repository scope to include the
+sidecar directories. `aube-lock.yaml` is aube's native format; it is not an npm
+`package-lock.json` and scanners may not recognize its transitive dependencies.
+
+After an external tool or a person edits a sidecar, ordinary `mise install`
+validates the graph and installs it under the identity of its actual bytes, then
+updates the recorded digest through auto-locking. `mise lock` also accepts valid
+edits without resolving dependencies again. `mise install --locked` instead rejects
+an inconsistent digest and directs you to run `mise lock` to accept the edit.
+Formatting-only edits also change identity. Missing or unreadable sidecars are
```
jdx, line 738: Nit: "Missing or unreadable sidecars are regenerated by `mise lock`" is only true when the tool is in the current toolset and uv (or embedded aube) is available. With the `prepare` behavior above, today it is actually an error. Once the save path tolerates dangling pointers, qualify this as "regenerated by `mise lock` when the tool is configured and its installer is available; otherwise the entry is kept and installation of that tool fails until it is regenerated".

src/toolset/tool_request.rs
```diff
@@ -470,6 +470,10 @@ impl ToolRequest {
     pub(crate) async fn is_install_satisfied(&self, config: &Arc<Config>) -> bool {
         if let Some(backend) = backend::get(self.ba()) {
             match self.resolve(config, &Default::default()).await {
+                // Explicit installs must validate sidecars even when the recorded digest
+                // already has an installation. Ordinary resolution uses the backend's
+                // filesystem-only satisfaction check directly.
+                Ok(tv) if tv.uv_lock.is_some() || tv.aube_lock.is_some() => false,
```
jdx, line 476: Minor / confirm intent: returning `false` for every graph-bearing tool means explicit `mise install` (including the no-argument form) always enters `install_version` for those tools, which loads and validates the sidecar and, for system-python setups, runs the `bind_uv_python` interpreter probe on every `mise install`. The pipeline still skips the actual install via `is_install_satisfied_or_false` after `prepare_install_version`, so this is a cost and a UX change (tool is queued, then reports "already installed") rather than a correctness issue. If that is the intended "explicit installs validate" behavior, fine; just flagging that it applies to `mise install` with no arguments too.

src/lockfile/graph.rs
```diff
@@ -0,0 +1,656 @@
...
+impl<T: NativeGraph> GraphRef<T> {
+    pub(crate) fn identity(&self) -> String {
+        match self {
+            Self::Sidecar { digest, .. } => digest.clone(),
+            Self::Inline { graph, digest, .. } => digest
+                .get_or_init(|| {
+                    digest_bytes(
+                        graph
+                            .graph_text()
+                            .expect("native graph serialization")
```
jdx, line 61: Nit: `identity()` is called from `install_path()`, so an `expect` here turns a serialization failure into a panic during ordinary resolution. `graph_text()` is `Result`-returning already; consider having `identity()` return `Result<String>` or fall back to hashing the canonical table on error rather than panicking.

jdx (commented):
Re-reviewed at 44c5d1c6f (sidecar implementation plus the earlier follow-ups).

**Confirmed fixed / working as intended**
- The `is_install_satisfied` reinstall loop for pipx-only users is gone: the "no graph yet" branch is now gated on `--locked` or uv being spawnable, and the e2e ends with the no-uv `--locked` failure case.
- `Lockfile::read` does no filesystem work for sidecars; identity comes from the recorded digest and the unit test asserts `load()` fails after the directory is removed while `identity()` still matches.
- Variant directories are sticky, the `~hash` covers backend plus options, GC is scoped per lockfile stem, and the `.config/mise` / `.mise` / bare layouts all resolve as agreed.
- Rollback ordering in `mise lock` is right: sidecar files join the snapshot set before rollbacks are prepared, and GC deletions are deferred until after the commit succeeds.
- Edited-sidecar flow is covered end to end: `--locked` rejects the digest mismatch, ordinary install accepts it under the new identity, and auto-lock rewrites the digest.
- `pipx:` spelling survives in `BackendArg::short`, lockfile keys, install-state, and revision-1 saves.

**One item I'd fix before merge**, plus a doc nit and two minor notes, inline.

jdx:
## Follow-up decisions after the 44c5d1c6f review

These are jdx's calls on the open questions from the sidecar review; treat them as the plan for the next revision.

### 1. Hot `mise install` is cheap

Revert the `ToolRequest::is_install_satisfied` special case that returns `false` for every graph-bearing tool. For a graph-locked tool whose install directory for the recorded identity already exists, `mise install` must not load the sidecar, must not run `validate_uv_lock` / `validate_aube_lock`, and must not run the `bind_uv_python` interpreter probe. Do one `stat` on the sidecar's graph file (`uv.lock` / `aube-lock.yaml`) and `warn!` if it is missing, so a gitignored `.mise/locks/` surfaces on the maintainer's machine rather than on a fresh clone. Full validation stays in `--locked` installs and `mise lock`.

This only holds if the generate-mode auto-lock also stays out of the sidecars on a no-op run: `populate_uv_locks(force = false)` calls `load()` on every existing graph, so `is_current()` must short-circuit before `generate()` when nothing changed. Add a comment at that call site and a test: generate mode, uv not installed, unchanged toolset, `mise install` performs no reads under the sidecar root.

### 2. Sidecar garbage collection: `mise lock` and generate-mode saves only

Unreferenced sidecar directories are deleted by explicit `mise lock` (already under transaction locks) and by generate-mode saves from `mise install`, `mise x`, `mise use`, and task tool installs. Merge-mode auto-lock writes sidecars but never deletes. Any path that deletes must hold the lockfile lock across read → prepare → publish → prune; without that, process A's GC (computed from a stale read) can `remove_dir_all` a directory process B just wrote, leaving B's pointer dangling, which unlike the pre-existing last-writer-wins on `mise.lock` is not recoverable by rerunning. Confirm the generate-mode install path holds that lock or make it do so, and add a two-process test if practical.

### 3. Directory names follow the spelling in `mise.toml`, for installs and sidecars

Reversal of the earlier "both spellings share `pipx-<name>`" design. `pypi:black` installs to `installs/pypi-black/…` and locks to `.mise/locks/pypi-black/…`; `pipx:black` keeps `pipx-black/…` for both. Nothing on disk should say `pipx` unless the user wrote `pipx:`.

Consequence, which is the simplest consistent model: `pipx:black` and `pypi:black` become **distinct tool identities on the same backend type**, not aliases of one tool. Concretely:

- Remove the `pipx:` special case from `tool_directory_name` (plain kebab-case of `short`), the `unalias_backend` calls in `install_state`, the alias lookup in `Lockfile::tool_key` / `tools_for`, and the canonical-form `PartialEq` / `Ord` / `Hash` on `BackendArg` (compare `short` as written, as before this PR).
- Keep: `BackendType::guess` accepting both prefixes, the registry emitting `pypi:`, `pypi.*` / `pipx.*` settings aliasing, and backend-level settings such as `disable_backends` and `minimum_release_age_excludes` treating both names as the same backend.
- Switching a config from `pipx:black` to `pypi:black` is a fresh install under the new directory and a new lockfile entry; the old entry is pruned as stale by `mise lock` and the old install directory by `mise prune`. Revision-1 lockfiles need no special handling: whatever spelling is in the file is the key.
- Update the assertions that currently encode sharing: `mise where pypi:x == mise where pipx:x` in `test_pypi_lock` and `pypi_lock.Tests.ps1`, the `pypi_and_pipx_share_backend_and_install_paths` unit test, `pypi_discovers_legacy_pipx_manifest_without_moving_it`, and the lockfile alias round-trip test. Remove the "same existing installation directories" claim from the PR description and docs.

### 4. Monorepo

- Legacy-lockfile migration into a root lockfile copies sidecars under the root's `.mise/locks/` but never removes `apps/<sub>/.mise/locks/`. Delete the legacy sidecar root after a successful migration so scanners do not keep seeing a stale `uv.lock`.
- Document that sidecars follow the *root* lockfile's layout; a subproject's `.config/mise/` layout does not influence where its graphs land.

### 5. Dangling pointers must not fail the save (filed inline on `graph.rs`)

`SidecarWrites::prepare` must keep a `Sidecar` pointer whose body cannot be loaded, warn, and skip its file writes. The `preserve_absent_tool_entries` monorepo path makes this a daily case, not an edge: `mise install` in `apps/a` writes `apps/b`'s entries untouched, and `populate_*_locks` never refreshes them.

### 6. Docs

Qualify "missing or unreadable sidecars are regenerated by `mise lock`" (only when the tool is configured and its installer is available), describe the per-spelling directory naming, and note the GC policy per lockfile mode.

src/lockfile/graph.rs
jdx: Fixed in 1fd1a3f05 on the active PR #13146: unavailable Sidecar bodies preserve their pointers and skip writes. A regression saves a dangling pointer twice and verifies byte-identical lockfile output. Missing uv no longer causes that pointer to be cleared.

docs/dev-tools/mise-lock.md
jdx: Updated in 1fd1a3f05 on #13146: regeneration is qualified by configured-tool and available-installer requirements. The docs also describe hot-install behavior, per-spelling directories, root monorepo layout, and mode-specific garbage collection.

src/toolset/tool_request.rs
jdx: Implemented your follow-up decision in 1fd1a3f05 on #13146. Hot installs skip validation/probing for an existing recorded identity, including explicit tool arguments. Full validation remains for locked/forced installs; generate-mode no-op detection now handles universal graphs lazily.

src/lockfile/graph.rs
jdx: Keeping the byte-digest identity contract for this revision rather than introducing a formatting-insensitive fallback. Sidecar identity only returns the recorded string and cannot serialize or panic. Inline graphs are generated in-process from native serializable graph values; a future fallible inline constructor would be preferable to adding a second identity scheme. This does not affect the sidecar read/validation fixes in 1fd1a3f05.

PR #13179, opened 2026-09-14
fix(upgrade): persist explicit selectors when bumping versions

zeitlinger:
## Summary

`mise upgrade --bump` now saves explicit version selectors to the tool's existing configuration instead of leaving the configured version stale after a successful upgrade.

For example, with Flint pinned in `mise.toml`:

```bash
mise upgrade --bump aqua:grafana/flint@latest
```

upgrades Flint and writes `"latest"` to its existing entry. Like `mise use`, explicit version selectors are preserved: `node@22` writes `"22"`, and `node@latest` writes `"latest"`. With `settings.pin = true`, version requests are saved as their resolved concrete version instead.

To retain the existing version precision while bumping it, omit the selector:

```bash
mise upgrade --bump aqua:grafana/flint
```

Once an entry says `"latest"`, later bumps leave that selector unchanged; an enabled lockfile still records the concrete version. Configuration updates also work when the requested version is already installed, and dry runs show the pending configuration change. Failed requests do not overwrite working configuration.

## Validation

Focused upgrade e2e coverage and all 10 upgrade unit tests pass. Formatting, ShellCheck, shfmt, and Cargo checks pass. Local compilation required exposing the installed CMake and supplying temporary libclang/header paths; no project configuration changes were needed.

Additional change: this PR also includes the unrelated e2e/config/test_required_env_vars_tools_filter pipefail-safe grep helper fix in commit fc676ed8f.


jdx:
I don't think we should solve this only for the literal `@latest` case. The underlying problem is that an explicit tool request becomes `ToolSource::Argument`, so mise loses the configuration source it needs to persist the requested change.

This should also cover explicit targets such as:

- `mise up --bump dummy@3` → resolve the latest version matching `3` and write `dummy = "3"`
- `dummy@3.1` and `dummy@3.1.2` → preserve that requested precision
- channels such as `dummy@stable` or `dummy@nightly`
- typed requests such as `dummy@prefix:3`

A bare tool or `@latest` should mean latest overall while preserving the existing config precision. An explicit non-`latest` request should remain the target constraint and become the new configured request. Simply clearing every explicit request would not work, since that would discard `@3`; conversely, keeping it command-line-sourced still loses the config path.

I think the implementation should represent the CLI target separately from the configuration source being updated, rather than deleting `@latest` as a special case. For non-release requests such as `path:`, `ref:`, `branch:`, `rev:`, and `system`, we should either persist them deliberately or reject the combination clearly, not silently reinterpret them as latest.

Please add coverage for fuzzy, exact, channel, and typed requests as well as `latest`. This broader model should also use identity-aware matching so equivalent shorthand and explicit backend spellings update the same configured tool, as noted in the Greptile review.

src/cli/upgrade.rs
```diff
@@ -250,6 +296,13 @@ impl Upgrade {
             }
         }
         if outdated.is_empty() {
+            if !explicit_config_bumps.is_empty() {
+                if self.is_dry_run() {
+                    print_explicit_config_bumps(&explicit_config_bumps)?;
+                } else {
+                    apply_explicit_config_bumps(&explicit_config_bumps).await?;
```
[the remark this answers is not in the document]
zeitlinger: Fixed in commit `0d4ec525f`. Explicit bumps are now persisted only when the exact requested selector is resolved and installed; already-installed valid selectors remain supported in mixed upgrades. Added regression coverage for unresolved-only, unresolved-plus-successful, and already-installed-plus-outdated cases. Verified with focused e2e, 9 upgrade unit tests, cargo check, formatting, and full lint.

zeitlinger:
The fix is pushed at `0d4ec525fbc5fb0717a68f855e979aad3855089b`; local focused e2e, upgrade unit tests, cargo-check, formatting, and full lint pass. The e2e job [104444405306](https://github.com/jdx/mise/actions/runs/34986354030/job/104444405306) failed because the GitHub installation hit its core API rate limit (0/5000; reset 2026-09-15 15:56:41 UTC), across unrelated GitHub-backed tests. Our token cannot rerun jobs (GitHub returned 403 Must have admin rights). Please rerun the failed e2e job after the reset, and the cancelled title job [104439319808](https://github.com/jdx/mise/actions/runs/34986350746/job/104439319808) if needed. The macOS unit job is still queued.

e2e/cli/test_upgrade_explicit_selector_bump
```diff
@@ -0,0 +1,79 @@
...
+tiny-local = "1.0.0"
+EOF
+mise install tiny-local@1.0.0
+mise up --bump dummy@path:/does-not-exist tiny-local --prune >/dev/null 2>&1 || :
+assert_contains "cat mise.toml" 'dummy = "1.0.0"'
+assert_contains "cat mise.toml" 'tiny-local = "3.1.0"'
+
+mise up --bump dummy@3 --prune
+assert_contains "cat mise.toml" 'dummy = "3"'
+
```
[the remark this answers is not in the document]
zeitlinger: Addressed in `ed4e646c1`: added coverage for concrete `3.1.0` and successful typed `prefix:3` selectors, while retaining unresolved-selector coverage. The focused e2e suite passes.

src/cli/upgrade.rs
```diff
@@ -544,6 +605,18 @@ impl Upgrade {
                     ));
                 }
             }
+            let explicit_config_bumps = explicit_config_bumps
+                .iter()
+                .filter(|bump| {
+                    successful_versions.iter().any(|version| {
+                        backend_args_match(version.ba(), bump.request.ba())
```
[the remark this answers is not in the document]
zeitlinger: Addressed in `ed4e646c1`: explicit bumps now correlate the original selector to its `OutdatedInfo.tool_request`, then require that bumped selector among successful installations. A same-backend failed-plus-successful regression fails on the prior backend-only implementation and passes now; a direct unit test covers the matching helper.

jdx:
Thanks for tracking this down — the underlying bug is real.

I'd like to drop the `latest` special case though. The rule the rest of this PR implements — an explicit selector on the command line is persisted verbatim — is already how `mise use` behaves: `mise use node@latest` writes `node = "latest"`, and only resolves to a concrete version when `--pin` or `settings.pin` is set. I'd rather `mise up --bump node@latest` match that and write `latest`, rather than being a silent synonym for plain `mise up --bump node` (right now `tool.tvr = None` discards the `@latest` entirely, so there's no way to express "widen my pin" through `upgrade` at all).

That removes the carve-out and lets the `ExplicitConfigBump` path handle every selector uniformly. Two follow-ons if you take it:

- `settings.pin` should still resolve to the concrete version — roughly the behavior you have today, just conditional.
- Once a config says `latest`, future `--bump` runs won't rewrite it (`check_semver_bump` has no numeric precision to work from). That's intended, and the lockfile still holds the concrete version, but worth a test asserting it.

*AI-assisted — Tool: Claude Code; model: Anthropic/claude-opus-5; version: 2.1.270.*

zeitlinger:
Implemented in 565d596a5. Removed the `latest` special case: explicit selectors now persist verbatim, so `mise upgrade --bump aqua:grafana/flint@latest` writes `"latest"`. With `settings.pin` enabled, version requests instead persist their resolved concrete version.

Added regression coverage for both behaviors, fully qualified backend requests, already-installed versions, and subsequent bumps preserving `"latest"` while the lockfile retains a concrete version. The focused e2e test, all 10 upgrade unit tests, and scoped lint checks pass. Updated the PR title and description to reflect the final behavior.

src/cli/upgrade.rs
```diff
@@ -397,6 +477,7 @@ impl Upgrade {
                     display_path(cf.get_path())
                 );
             }
+            print_explicit_config_bumps(explicit_config_bumps)?;
```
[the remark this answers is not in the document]
zeitlinger: Fixed in eddf7a8d2. Dry-run now validates each pending install request before reporting its configuration bump, and requires path targets to exist. Matching includes the original selector as well as backend identity, so a valid sibling selector cannot make a failed one eligible. Already-installed selectors and valid pending installs remain visible.

Added a mixed-tool regression that fails on the previous code and passes with this fix. The focused e2e test, lint-fix, and scoped Cargo/formatting/ShellCheck/shfmt checks pass.

src/cli/upgrade.rs
```diff
@@ -1033,6 +1163,94 @@ fn backend_matches(backends: &HashSet<String>, ba: &BackendArg) -> bool {
...
+    let (ba, versions) = toolset
+        .versions
+        .iter()
+        .find(|(ba, _)| backend_args_match(ba, requested))?;
+    let mut supported = versions
+        .requests
+        .iter()
+        .filter(|request| request.is_os_supported());
+    let request = supported.next()?;
+    if supported.next().is_some() || request.source().path().is_none() {
```
jdx: This accepts any source that has a path, including `ToolSource::MiseTomlDaemon`. Daemon tool requests are merged into the toolset by `ToolsetBuilder::load_config_files`, so for a tool declared only under `[daemons.x] tool = "dummy@1.0.0"`, `mise up --bump dummy@latest` writes a brand-new `[tools] dummy = "latest"` entry while the daemon's own spec stays stale. The existing bump path guards exactly this (`!o.source.is_mise_toml_daemon()`, ~line 316) — same guard is needed here.

Separately, the `supported.next().is_some()` bail (a tool configured as `node = ["20", "22"]`) returns silently, where the old path warns `upgrading multiple versions with --bump is not yet supported`.

src/cli/upgrade.rs
```diff
@@ -546,6 +662,20 @@ impl Upgrade {
...
+                        bump,
+                        &outdated,
+                        &successful_versions,
+                        &ts,
+                        config,
+                    )
+                })
+                .cloned()
+                .collect::<Vec<_>>();
+            apply_explicit_config_bumps(&explicit_config_bumps).await?;
```
jdx: The `?` here short-circuits before the `config_file_errors` reporting immediately below, so a failure applying an explicit bump discards every error already collected for the other tools in this run. Push into `config_file_errors` instead of returning, so this matches the accumulate-and-report style it's sitting in.

src/cli/upgrade.rs
```diff
@@ -191,6 +199,41 @@ impl Upgrade {
...
+                .build(&config)
+                .await?;
+            for tool in &self.tool {
+                let Some(request) = tool.tvr.as_ref() else {
+                    continue;
+                };
+                let Some((configured_ba, configured_request)) =
+                    effective_persistable_request(&effective, &tool.ba)
+                else {
+                    continue;
```
jdx, line 248: Every skip in this loop is silent — version comes from `MISE_<TOOL>_VERSION`, tool isn't in any config file, or the backend has multiple requests. The command exits 0, prints nothing unusual, and the config is unchanged, so there's no way for a user to find out why their bump didn't stick.

This is the "reject the combination clearly" part of my earlier comment. One `warn!` naming the tool and the source it actually came from covers all three cases.

src/cli/upgrade.rs
```diff
@@ -191,6 +199,41 @@ impl Upgrade {
             unimplemented!("mise upgrade --monorepo is not implemented yet");
         }
         let mut config = Config::get().await?;
+        let mut explicit_config_bumps = Vec::new();
+        if self.bump && !self.tool.is_empty() {
+            let scope = self.scope();
+            let effective = ToolsetBuilder::new()
```
jdx, line 209: `build()` resolves the entire toolset, but `effective_persistable_request` only reads `versions.requests`, `is_os_supported()` and `source()` — all of which `load_config_files` populates before resolution runs. That's a full extra resolve pass on every `mise up --bump <tool>` for data that's already available unresolved.

src/cli/upgrade.rs
```diff
@@ -397,6 +477,42 @@ impl Upgrade {
...
+                let pending = outdated.iter().find(|outdated| {
+                    backend_args_match(outdated.tool_version.ba(), bump.request.ba())
+                        && outdated.tool_version.request.version() == bump.request.version()
+                });
+                let eligible = if let Some(pending) = pending {
+                    // Outdated entries can include requests that failed initial resolution.
+                    // Validate the actual install request before advertising its config update.
+                    pending
+                        .tool_request
+                        .resolve(config, &resolve_options)
```
jdx: This re-resolves each pending request a second time, and the eligibility logic here duplicates `explicit_bump_is_successful` in a subtly different form — that one matches against `successful_versions`, this one re-resolves and probes install state. Two implementations of "would this bump actually apply?" will drift; worth one shared predicate so dry-run can't disagree with the real path.

jdx (commented):
Thanks — this is close. I checked each bot finding against the current code rather than the replies, and they're all genuinely fixed; both rounds of my earlier feedback are in, including the `latest`/pin behavior and the test asserting a config that says `latest` stays `latest`.

A few things left. Most are inline; one has no line to hang off:

**Docs.** The diff touches three files and none of them is documentation. The `--bump` help text still says "It keeps the same precision as what was there before" — that was unconditionally true before this PR, because an explicit selector became `ToolSource::Argument` and never reached the config. It's now only true for the bare form: `node@latest` writes `"latest"`, `node@3` writes `"3"`, `node@prefix:3` writes `"prefix:3"`. So the one sentence a user reads to understand `--bump` now describes a special case without saying it's one.

It needs a paragraph covering the selector-is-persisted rule, that `settings.pin` writes the resolved concrete version instead, and that nothing is persisted when the version comes from a non-writable source. Worth an example line too — `mise upgrade node@latest` is the new capability here and there's no way to discover it from `--help` today. That means `mise run render` to regenerate `docs/cli/upgrade.md`.

Minor:
- `test_upgrade_explicit_latest_bump` is misnamed now that the `latest` carve-out is gone.
- No coverage for `ref:`/`branch:`/`rev:`/`system` from my original list — only `path:` and `prefix:` are exercised.
- `fc676ed8f` (the pipefail fix in `e2e/config/test_required_env_vars_tools_filter`) is unrelated to this PR and isn't mentioned in the description.

*AI-assisted — Tool: Claude Code; model: Anthropic/claude-opus-5; version: 2.1.270.*

src/cli/upgrade.rs
zeitlinger: Fixed in c688ac941: explicit config discovery now excludes daemon sources while retaining the multiple-version warning. Added deterministic daemon-source regression coverage. Local full lint and both focused e2e suites pass.

src/cli/upgrade.rs
zeitlinger: Fixed in c688ac941: explicit config-application failures are accumulated with other config errors instead of short-circuiting. Full lint and focused tests pass.

src/cli/upgrade.rs
zeitlinger: Fixed in c688ac941: skipped explicit bumps now warn with the effective source, including environment, daemon/non-writable, multiple-request, and absent-config cases. Regression e2e coverage passes.

src/cli/upgrade.rs
zeitlinger: Fixed in c688ac941: request discovery uses ToolsetBuilder::build_unresolved, avoiding the redundant full resolution pass. Cargo check and full lint pass.

src/cli/upgrade.rs
zeitlinger: Fixed in c688ac941: dry-run and execution now share explicit-bump eligibility and selector-correlation logic, using resolved candidates or successful installations as evidence. Both focused e2e suites and 10 upgrade unit tests pass.

zeitlinger:
Should be good now

src/cli/upgrade.rs
```diff
@@ -547,6 +672,17 @@ impl Upgrade {
...
             }
+            let mut eligible_explicit_config_bumps = Vec::new();
+            for bump in explicit_config_bumps {
+                let pending = find_explicit_bump_outdated(&outdated, bump);
+                if explicit_bump_is_eligible(pending, bump, &successful_versions, &ts, config)
+                {
+                    eligible_explicit_config_bumps.push(bump.clone());
+                }
+            }
+            if let Err(e) = apply_explicit_config_bumps(&eligible_explicit_config_bumps).await {
```
jdx, line 764: This writes the explicit selector (`"2"`) even when the install was the rewritten request (`"3"` → 3.1.0), leaving the configured version uninstalled. See the review summary.

src/cli/upgrade.rs
```diff
@@ -398,6 +509,20 @@ impl Upgrade {
                     display_path(cf.get_path())
                 );
             }
+            let candidates = ts
```
jdx: These candidates are just the requested versions, so the check either passes trivially or, when `check_semver_bump` rewrote the request, never matches. `mise up --bump tiny-local@2 --dry-run` prints no `Would bump` line, but the real run writes `"2"`.

src/cli/upgrade.rs
```diff
@@ -1034,6 +1170,107 @@ fn backend_matches(backends: &HashSet<String>, ba: &BackendArg) -> bool {
...
+    candidates: &[ToolVersion],
+    toolset: &Toolset,
+    config: &Arc<Config>,
+) -> bool {
+    if let Some(pending) = pending {
+        let candidate = explicit_bump_matches_successful(pending, candidates);
+        if !candidate {
+            return false;
+        }
+        return !matches!(&pending.tool_request, ToolRequest::Path { path, .. } if !path.exists());
```
jdx: This `path:` special case only exists because dry-run has no real evidence of resolution (see above). If explicit selectors resolved against themselves, it could go.

src/cli/upgrade.rs
```diff
@@ -252,6 +349,18 @@ impl Upgrade {
...
+            let installed_explicit_config_bumps = explicit_config_bumps
+                .iter()
+                .filter(|bump| explicit_bump_is_installed(&ts, &config, bump))
+                .cloned()
+                .collect::<Vec<_>>();
+            if !installed_explicit_config_bumps.is_empty() {
+                if self.is_dry_run() {
+                    print_explicit_config_bumps(&installed_explicit_config_bumps)?;
+                } else {
+                    apply_explicit_config_bumps(&installed_explicit_config_bumps).await?;
```
jdx, line 425: Minor: this path changes the config but still prints "All tools are up to date" and leaves `mise.lock` as it was. With `tiny-local = "1.0.0"` locked and 3.1.0 installed, `--bump tiny-local@3` writes `"3"` while the lock stays at 1.0.0.

e2e/cli/test_upgrade_explicit_selector_bump
```diff
@@ -0,0 +1,179 @@
...
+assert_contains "cat mise.toml" 'dummy = "2.0.0"'
+
+# Explicit selectors constrain resolution and are persisted verbatim.
+cat <<'EOF' >mise.toml
+[tools]
+dummy = "1.0.0"
+EOF
+mise install dummy@3.1.0
+
+output=$(mise up --bump dummy@3 --dry-run)
```
jdx, line 38: `dummy` only lists up to 2.0.0, so this never exercises a selector below the backend's latest. Worth adding a `tiny-local@2` case (latest 3.1.0).

jdx (commented):
Thanks. I checked each item from my last round against the code, and they're all fixed.

One thing is still missing from my first comment: with `--bump`, an explicit selector is persisted but doesn't constrain what gets installed. `OutdatedInfo::resolve` uses the backend's overall latest when `bump` is set, and `check_semver_bump` then rewrites the request to that. So with the `tiny` plugin:

```
# mise.toml: tiny-local = "1.0.0"
mise up --bump tiny-local@2
# installs 3.1.0, writes tiny-local = "2"; mise ls shows 2.1.0 (missing)
```

The tests don't catch it because `dummy`'s latest is 2.0.0 and every `dummy@3` case installs 3.1.0 first, so nothing is ever outdated past the selector. A `tiny-local@2` case would show it.

I think requests with an explicit selector should resolve without the bump behavior (latest matching the selector, request kept as written), leaving backend-latest + precision bump for bare tools. That should also fix the dry-run issue inline.

*AI-assisted — Tool: Claude Code; model: Anthropic/claude-opus-5; version: 2.1.270.*

src/cli/upgrade.rs
zeitlinger: I fixed this in [`4be8a233`](https://github.com/zeitlinger/mise/commit/4be8a2336d070df18a1e389d93ff03c26d5d1546). Explicit CLI selectors are now resolved with `bump=false`, and the config is updated only when that selector is among the successful installations. The focused e2e covers `tiny-local@2` resolving to 2.1.0, so the persisted selector remains `"2"` for the installed 2.1.0 version. Full current-head CI is green.

src/cli/upgrade.rs
zeitlinger: I fixed this in [`4be8a233`](https://github.com/zeitlinger/mise/commit/4be8a2336d070df18a1e389d93ff03c26d5d1546). Dry-run eligibility now uses the resolved `OutdatedInfo` entry instead of `ts.list_current_versions()`, so requested selectors are no longer treated as installation evidence. The focused e2e now asserts that `mise up --bump tiny-local@2 --dry-run` reports `Would install tiny-local@2.1.0` and `Would bump tiny-local@2`.

src/cli/upgrade.rs
zeitlinger: I fixed this in [`4be8a233`](https://github.com/zeitlinger/mise/commit/4be8a2336d070df18a1e389d93ff03c26d5d1546). I removed the path-specific eligibility exception: nonexistent explicit `path:` selectors are rejected before a persistable bump is created, while dry-run uses the resolved request as its evidence. The regression asserts that an invalid path produces neither a path bump nor a spurious `dummy@1.0.0` configuration bump.

src/cli/upgrade.rs
zeitlinger: I fixed this in [`4be8a233`](https://github.com/zeitlinger/mise/commit/4be8a2336d070df18a1e389d93ff03c26d5d1546). After a config-only explicit bump, mise reloads the config, re-resolves without the stale lockfile pin, and rebuilds the lockfile; it also suppresses the misleading `All tools are up to date` message. The focused e2e covers a locked `tiny-local@1.0.0` upgraded with `--bump tiny-local@3`, asserting the config is `"3"` and the lockfile is 3.1.0.

e2e/cli/test_upgrade_explicit_selector_bump
zeitlinger: I added the requested `tiny-local@2` coverage in [`4be8a233`](https://github.com/zeitlinger/mise/commit/4be8a2336d070df18a1e389d93ff03c26d5d1546). The local backend latest is 3.1.0; the test now checks both dry-run and real execution, including persistence of `"2"` and installation of 2.1.0. The focused e2e passes.

src/cli/upgrade.rs
```diff
@@ -252,6 +406,35 @@ impl Upgrade {
             }
         }
         if outdated.is_empty() {
+            let installed_explicit_config_bumps = explicit_config_bumps
+                .iter()
+                .filter(|bump| explicit_bump_is_installed(&ts, &config, bump))
```
jdx, line 417: **[P2] Honor exclusions when persisting explicit bumps**

`--exclude` only filters the outdated list; this separate persistence path still accepts an excluded request when its version is already installed. With `dummy = "1.0.0"` configured and `dummy@3.1.0` installed, `mise up --bump dummy@3 --exclude dummy` reaches this branch and writes `dummy = "3"` despite the exclusion. The same installed-request fallback exists in `explicit_bump_is_eligible` when another tool is upgraded. Filter excluded tools out of `explicit_config_bumps` before either persistence path, and add coverage for both config-only and mixed-tool runs.

jdx (commented):
Two remaining findings from code inspection: explicit configuration bumps bypass `--exclude`, and successful config-only bumps print the out-of-range upgrade warning. I have not run local tests.

src/cli/upgrade.rs
```diff
@@ -267,7 +450,7 @@ impl Upgrade {
                 .filter(|o| o.bump.is_some())
                 .collect::<Vec<_>>()
             };
-            if bump_outdated.is_empty() {
+            if bump_outdated.is_empty() && !has_explicit_config_bumps {
```
jdx, line 459: **[P3] Avoid the out-of-range warning after a config-only bump**

When the requested version is already installed, a successful explicit bump sets `has_explicit_config_bumps = true`, while `bump_outdated` is necessarily empty because `self.bump` is true. This condition therefore takes the `else` branch and prints “Newer versions are available but do not match the configured version ranges” with an empty list, followed by advice to run `mise upgrade --bump` again. Gate that warning on a nonempty `bump_outdated` list so a successful config-only bump does not report nonexistent pending upgrades; the current e2e assertion only excludes the old “All tools are up to date” message.

src/cli/upgrade.rs
zeitlinger: I fixed this in `4283f1dcb`: explicit configuration bumps are now filtered against `--exclude` using backend identity before either the config-only or normal persistence path. The e2e regression covers both an excluded config-only bump and an excluded tool alongside a sibling upgrade.

src/cli/upgrade.rs
zeitlinger: I fixed this in `4283f1dcb`: the out-of-range warning is now emitted only when `bump_outdated` is non-empty. The config-only e2e assertion now verifies that neither the up-to-date message nor the out-of-range warning is printed.

zeitlinger:
addressed

PR #13191, opened 2026-09-14
registry: add nubr

colinhacks:
Adds `nubr`, the TypeScript runner from the Nub project, published on npm as [`@nubjs/runner`](https://www.npmjs.com/package/@nubjs/runner). It runs a file, a `package.json` script, or an installed bin on plain Node, without the `nub` binary; `nub` itself is already in the registry (#10544).

**Popularity:** [nubjs/nub](https://github.com/nubjs/nub) has 4,296 stars and 60 forks. The runner ships from that repository and shares its release cadence.

**Backend:** `npm:` is the only backend that exists for this tool. `@nubjs/runner` is a Node package, a `--import` preload plus a `nubr` script that re-execs the user's Node, with a native addon in per-platform optional dependencies. There is no standalone binary, no GitHub release asset, no aqua entry, and no packslip to point a higher tier at; a Node on PATH is what the tool runs on, not an incidental dependency. This is the same situation as the existing `nub` entry, which was accepted with the npm backend.

Version listing and install verified with the npm backend (mise 2026.9.7). The first real release, 0.9.2, is under 24 hours old at the time of writing, so the default `minimum_release_age` still hides it from a plain `ls-remote`; `0.0.0` is the name reservation that preceded it.

```sh
MISE_MINIMUM_RELEASE_AGE=0 mise ls-remote npm:@nubjs/runner
# 0.0.0 0.9.2 
mise exec npm:@nubjs/runner@0.9.2 -- nubr --version
# 0.9.2
```

## Summary by CodeRabbit

* **Chores**
  * Updated the `nubr` registry configuration to use the revised backend format.
  * Enabled support for packages with lower download counts in this registry entry.


jdx:
Going to take this despite the low download numbers, but two things need to happen first.

**1. It needs `allow_low_downloads = true`.**

As written the entry can't install — aube's reputation gate refuses the package:

```sh
printf '[tools]\n"npm:@nubjs/runner" = "0.9.2"\n' > mise.toml && mise install
# mise ERROR Failed to install npm:@nubjs/runner@0.9.2: aube install failed: user aborted `mise add @nubjs/runner`
```

`@nubjs/runner` is at 0 weekly downloads and is a newly registered name, so it trips two of the three
reputation checks ([docs](https://mise.jdx.dev/dev-tools/backends/npm.html#allow-low-downloads)).
With the option added it installs and `nubr --version` prints `0.9.2`.

This is a different gate than the `allow_builds` on the `nub` entry — `@nubjs/runner` and its
platform packages have no lifecycle scripts, so `allow_builds` is correctly omitted here.

Heads up that the green `test-tool` check doesn't cover this: it runs a dev build, and the install
fails with a released binary.

**2. We need to wait out `minimum_release_age`.**

0.9.2 published 2026-09-13 22:22Z and the default window is 24h, so at the moment:

```sh
mise latest npm:@nubjs/runner
# 0.0.0
```

`0.0.0` is the name-reservation placeholder and has no `bin` at all, so merging before the window
closes would ship an entry where `mise use nubr` installs a package containing no `nubr`. Resolves
itself later today — just not mergeable until then.

---

**Separately, not this PR:** `nub` itself should move from `npm:@nubjs/nub` to `github:nubjs/nub`.
The releases ship real standalone binaries, and `bin/nub` out of `nub-linux-x64.tar.gz` runs on its
own with no Node involved. Today the npm path installs `#!/usr/bin/env node` launchers that
self-heal into an sh trampoline via a `postinstall`, and on Windows that Node launcher runs on every
single call with no fast path. Going through GitHub drops the trampoline, drops the `allow_builds`
exemption, and stops requiring Node on PATH for a Rust CLI.

`nubx` still works there — it's argv0 dispatch on the same binary, so a symlink named `nubx` next to
`nub` is enough:

```sh
ln -s nub nubx && ./nubx --help
# Run a tool from `node_modules/.bin`, fetching it on a local miss (`npx`/`pnpm dlx`).
```

Release assets cover darwin/linux/win32 on x64 and arm64, plus linux musl variants.

colinhacks:
Added `allow_low_downloads = true` in `[backends.options]`, matching the other curated npm entries.

Reproduced on a clean data dir with mise 2026.9.7: without the option `mise install` fails on the reputation gate; with it, `npm:@nubjs/runner@0.9.2` installs and `nubr --version` prints `0.9.2`. The release-age window on 0.9.2 closes at 22:22Z today.

PR #13252, opened 2026-09-15
fix(pypi): lock tools whose injected dependencies need a newer Python

jdx:
Reported in [#13247](https://github.com/jdx/mise/discussions/13247).

`mise lock` failed for any PyPI tool whose `with` or `expose` list pinned a requirement needing a newer Python than the tool itself:

```toml
[tools]
"pypi:mkdocs" = { version = "1.6.1", with = ["mkdocstrings==1.0.6"] }
uv = "latest"
```

mkdocs 1.6.1 declares `requires-python = ">=3.8"`; mkdocstrings 1.0.6 declares `>=3.10`. mise built the sidecar's `requires-python` from the tool's own release metadata alone, so the generated project advertised `>=3.8` while depending on a release that cannot be installed below 3.10. uv forked the universal resolution at `>=3.8, <3.10`, found no mkdocstrings release for that fork, and gave up:

```console
$ mise lock
  × No solution found when resolving dependencies for split (markers:
  │ python_full_version >= '3.8' and python_full_version < '3.10'):
  ╰─▶ Because the requested Python version (>=3.8) does not satisfy
      Python>=3.10 and mkdocstrings==1.0.6 depends on Python>=3.10, we can
      conclude that mkdocstrings==1.0.6 cannot be used.
mise ERROR uv exited with non-zero status: exit code 1
```

The sidecar's Python range is now intersected across every requirement mise writes into it, so the same config locks and installs, with the floor raised to `>=3.10`:

```console
$ mise lock
Resolved 21 packages in 83ms
✓ Lockfile written to /tmp/test-mise/mise.lock

$ cat .mise/locks/pypi-mkdocs/1.6.1~6716a419/pyproject.toml
[project]
name = "mise-pypi-tool-environment"
version = "0.0.0"
requires-python = ">=3.10"
dependencies = ["mkdocs==1.6.1", "mkdocstrings==1.0.6"]

$ mise x --locked -- mkdocs --version
mkdocs, version 1.6.1 from …/.mise-uv/.venv/lib/python3.12/site-packages/mkdocs (Python 3.12)
```

## What narrows the range, and what does not

Only a requirement pinned to one exact release narrows it, because that is the case uv cannot spread across the range on its own:

- **Unpinned requirements are untouched.** `with = ["mkdocstrings"]` still locks 0.26.1 below 3.9, 0.30.1 on 3.9, and 1.0.6 from 3.10 — uv's universal forking already handles that correctly, and narrowing there would needlessly cut off older interpreters.
- **A pin behind an interpreter marker is skipped.** `with = ["mkdocstrings==1.0.6; python_version >= '3.10'"]` keeps `requires-python = ">=3.8"`, because the requirement is simply absent from the versions the marker excludes.
- **Any other marker still narrows.** `; sys_platform == "linux"` leaves the requirement active across the whole Python range, so uv fails without narrowing; the floor is raised for every platform, which is the only way such a config locks at all.
- **A compound specifier containing an `==` clause counts as a pin.** `mkdocstrings==1.0.6,!=1.0.5` selects one release just as a bare `==` does.
- **A release whose metadata cannot be read leaves the range alone.** uv reports an unusable requirement more precisely than a partial lookup could, so a failed lookup degrades to today's behavior rather than failing the lock.

The `>=3.8` floor mise applies for portability is unchanged; it is now one constraint among several rather than the only one alongside the tool's.

## Compatibility

Existing lockfiles are unaffected and nothing needs regenerating. Sidecar validation compares a stored `requires-python` against that sidecar's own graph, not against a fresh calculation, so a lockfile written before this change stays valid until it is bumped. A tool that locks a higher floor may need a newer interpreter than it declares by itself — that is inherent to installing the pinned requirement at all, and the docs now say so under [`with`](https://mise.jdx.dev/dev-tools/backends/pypi.html#with).

## Validation

- `mise lock` + `mise install` + `mise x --locked` against real PyPI for the reported mkdocs/mkdocstrings config, before and after.
- `e2e/backend/test_pypi_lock` gains a case pinning a `with` requirement that needs `>=3.12` against a root at `>=3.10`. The e2e index fixture previously served one hardcoded `Requires-Python` for every package, so it could not express this; it now takes a per-package value. Confirmed the new case fails against the old single-floor sidecar and passes with the fix.
- Each of the four cases above confirmed end to end against real PyPI: bare pin and compound pin lock at `>=3.10`, the platform-marked pin locks at `>=3.10`, and the interpreter-marked pin keeps `>=3.8`.
- Unit coverage for which requirement spellings count as an exact pin (extras, whitespace, compound clauses, `===`, wildcards, interpreter vs. other markers, direct URLs).
- `mise run lint` and `cargo clippy --workspace --all-features --all-targets -- -D warnings` clean.

*AI-assisted — Tool: Claude Code; model: anthropic/claude-opus-5; version: 2.1.271.*

## Summary by CodeRabbit

* **New Features**
  * Locked Python package environments now account for the Python versions supported by all locked requirements.
  * Exact-version dependencies added through tool configuration can narrow the supported Python version range.
  * Interpreter-specific dependency markers can preserve broader Python version support when appropriate.
  * Locked environments continue to include published wheel targets and reuse cached artifacts for frozen installs.

* **Documentation**
  * Updated PyPI backend guidance explains Python version range handling, pinned dependencies, interpreter markers, and potentially larger sidecar environments.


jdx:
Both review findings were real. Fixed in d59eddfd8.

I reproduced each against the e2e index fixture (root at `>=3.10`, injected pin at `>=3.12`) to check what uv actually does before changing anything:

| injected requirement | uv with the old rule |
| --- | --- |
| `mise-lock-strict==1.0.0; sys_platform == "linux"` | **fails** — fork `python_full_version >= '3.10' and python_full_version < '3.12'` |
| `mise-lock-strict==1.0.0,!=2.0.0` | **fails** — same fork |
| `mise-lock-strict==1.0.0; python_version >= "3.12"` | resolves, 5 packages |

So skipping every marker was too broad, but skipping none would have been wrong too. uv does not fork on `sys_platform` here — it requires the whole Python range to resolve regardless of platform — so a platform-marked pin has to narrow, while an interpreter-marked one must not, since that requirement is simply absent from the versions the marker excludes.

The rule is now: skip only markers mentioning `python_version`, `python_full_version`, or `implementation_version`. For the compound case, the parser reads an `==` clause out of the clause list rather than rejecting anything containing a comma.

Confirmed end to end against real PyPI with mkdocs 1.6.1 (`>=3.8`) plus mkdocstrings 1.0.6 (`>=3.10`):

```console
# with = ["mkdocstrings==1.0.6,!=1.0.5"]          → requires-python = ">=3.10"  (was: lock failed)
# with = ["mkdocstrings==1.0.6; sys_platform == 'linux'"]  → requires-python = ">=3.10"  (was: lock failed)
# with = ["mkdocstrings==1.0.6; python_version >= '3.10'"] → requires-python = ">=3.8"   (unchanged, 36 packages)
```

One consequence worth naming for review: a platform-marked pin now raises the floor on every platform, including ones where the requirement never installs. That is over-narrowing, but the table above shows the alternative is a lock that cannot be produced at all, so it seemed the better trade. Users who want the wider range can guard the pin with an interpreter marker instead, which the `with` docs now mention.

*AI-assisted — Tool: Claude Code; model: anthropic/claude-opus-5; version: 2.1.271.*

PR #13198, opened 2026-09-14
fix(bootstrap): ignore mise shims when selecting tap Ruby

jacobbednarz:
Evaluating a third-party Homebrew tap could select a mise Ruby shim. The metadata sandbox then prevented that shim from loading configuration, causing package bootstrap to fail even when Ruby was installed as described in #13197.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved system Ruby detection when a mise shim appears earlier in the system `PATH`.
  * Compatible system Ruby installations are now selected directly, improving reliability for Ruby-based workflows.
* **Tests**
  * Added coverage for system Ruby selection on Linux and macOS.
  * Expanded formula and cask metadata validation across supported platforms and architectures.

PR #13265, opened 2026-09-16
fix(bootstrap): honour adopt for explicitly named brew-cask packages

waynehoover:
Closes the bug reported in #13260.

## The problem

`packages_from_specs_with_config` resolved with an empty option map:

```rust
resolve_managers(by_mgr, IndexMap::new(), true)
```

So naming a package on the command line silently discarded its adopt
configuration. `mise bootstrap packages apply brew-cask:foo` replaced the app
bundle even with `adopt = true` set, in either form:

```toml
[bootstrap.brew]
adopt = true

"brew-cask:ukelele" = { version = "latest", adopt = true }
```

The config-driven path (`mise bootstrap packages apply` with no arguments) was
always correct, which made this confusing to diagnose: adopt appeared broken
outright, when only the explicit-spec path was affected.

Replacing the bundle is the one outcome `adopt` exists to avoid. macOS
re-evaluates an app's Privacy & Security grants when its identity at that path
changes, and mise's own warning recommends setting exactly the option that was
being ignored:

```
mise WARN  brew-cask: replacing /Applications/Ukelele.app - macOS may revoke
Privacy & Security permissions for this app (Accessibility, Screen Recording,
Full Disk Access, etc.) ... To take over an existing app without replacing it,
set adopt = true or [bootstrap.brew] adopt = true
```

This matters most when migrating an existing machine's casks to mise ownership,
which is naturally done one package at a time. On a machine with ~165 casks, a
large share are apps whose function depends on a TCC grant.

## The change

Naming a package scopes the run; it does not opt out of that package's
configuration. The spec path now builds the same options the config path does,
with per-package `adopt` overriding `[bootstrap.brew] adopt`.

The logic moves into `packages_from_specs_with_config_files`, taking a
`ConfigMap` so it can be tested the same way `packages_from_config_files`
already is.

A CLI argument is usually written exactly as the config key, but may omit a tap
qualifier the config carries, so lookup falls back to matching on package name.

## Tests

Two added beside the existing `test_brew_cask_adopt_option_reaches_package_request`:

- `test_brew_cask_adopt_option_reaches_explicit_specs` covers all three cases:
  a declared package inheriting the global default, a declared package whose
  per-package `adopt = false` must win over it, and an undeclared package
  falling back to the global.
- `test_explicit_specs_without_adopt_config_carry_no_options` pins the default,
  so anyone not opting in keeps the existing replace behaviour.

## Verification

Confirmed against a real machine before and after. With ownership held by
Homebrew, mise reports:

```
mise brew-cask:wispr-flow: installed and managed by Homebrew; leaving unchanged
```

After transferring ownership it reports `already up to date`, and the marker is
visible on disk: mise writes `.mise-cask.toml` into the Caskroom version
directory where brew writes `.metadata/`. Adoption via the config path leaves
the bundle untouched, verified by inode rather than content hash, since a
replacement with identical bytes still changes the inode.

## Note on the test suite

Four tests in `system::packages::brew::tap::tests` fail on my machine both with
and without this change:

```
extracts_cask_homebrew_prefix_as_relocatable_metadata
extracts_cask_metadata_without_homebrew
extracts_formula_metadata_without_homebrew
metadata_sandbox_blocks_tap_processes_and_temp_writes
```

They appear to be the local symptom of #13197 (tap Ruby evaluation selecting a
mise shim inside a deny-read sandbox), which #13198 addresses. Verified
pre-existing by stashing this change and re-running on a clean tree.

## Summary by CodeRabbit

- **Bug Fixes**
  - Explicitly named Homebrew cask packages now resolve correctly against tap-qualified configuration entries.
  - Unqualified cask names match a configured entry when there is exactly one match.
  - Ambiguous matches now defer to the global Homebrew adoption setting instead of selecting unpredictably.
  - Package specifications continue to honor relevant settings from loaded configuration files.

- **Documentation**
  - Added documentation for package resolution with configuration settings.


waynehoover:
Closing: superseded by #13262, which landed the same fix an hour before I noticed.

For anyone reading later, #13262 is the better version. It handles a case mine
did not: `homebrew/cask/<token>` and a bare `<token>` resolve to the same cask
from the same API URL, so an option written against either should apply to both,
while two *other* taps shipping the same token must stay distinct. Mine simply
returned `None` when a token was ambiguous and fell back to the global default,
which is safe but loses the per-package setting in a case that can be resolved
correctly. #13262 also carries `test_brew_cask_adopt_does_not_leak_between_taps_sharing_a_token`
and `test_official_cask_alias_covers_only_the_official_tap`, which cover it
properly.

I have verified the fix on my own machine: `mise bootstrap packages apply
brew-cask:<name>` now honours `adopt`, and adoption leaves the bundle untouched
(same inode), which is what matters, since replacing an app bundle is what costs
it its Privacy and Security grants.

No hard feelings about the duplicate. Discussion #13260 was filed at 00:47 and
#13262 opened at 01:06, so the report did its job even if this PR did not.
Thanks for the quick turnaround.

PR #13280, opened 2026-09-16
feat(install): invoke sudo for system binaries and reject sudo mise

mxcl:
`mise install --system uv` currently needs elevated filesystem permissions, but running the entire command through sudo can leave root-owned caches, downloads, and project lockfiles in a user's directories.

This change runs downloads, verification, staging, and project bookkeeping as the invoking user. On Unix, mise invokes sudo only to publish the prepared installation, installation manifests, runtime links, and system shims. User-invoked `sudo mise ...` fails at the start of `main`, before configuration, logging, migrations, or the async runtime. Direct root execution without sudo provenance remains available for containers.

```sh
mise install --system uv
```

The initial scope is relocatable installations from Aqua, GitHub, GitLab, Forgejo, HTTP, and S3 backends. Other backends and tools with postinstall hooks receive an error when protected system publication would be required; user-writable installations continue to work. This intentionally does not relocate Python virtualenvs or compiled environments with embedded installation prefixes.

The privileged helper consumes an archive streamed by the unprivileged process, rather than opening user-controlled source paths. It checks destination ownership/modes, rejects archive traversal and external symlinks, strips special permission bits, serializes publication, and stages replacements before replacing an existing version. It does not load mise configuration or execute backend code. Elevation uses the existing sudo policy, including `system_packages.sudo = false` and noninteractive passwordless-sudo checks.

### Validation

- Focused regression test passes: sudo provenance, archive round-trip, rebased symlinks, empty directories, stripped setuid/setgid bits, manifest publication, duplicate installs, reserved metadata names, rejected malicious replacements preserving the old installation, link publication, and unsafe directory modes.
- `cargo clippy --locked --bin mise -- -D warnings` passes.
- Formatting and `git diff --check` pass.
- The repository's `mise run lint-fix` task could not run its checks because `hk` is not installed.

Draft because an actual privileged installation/password prompt has not been exercised: passwordless sudo is unavailable on the development host. Publication tests exercise the same filesystem code using the test user's ownership. This also needs maintainer review of the intentional compatibility change rejecting `sudo mise` globally.

## Summary by CodeRabbit

- **New Features**
  - Added support for installing tools into protected Unix system directories without running `mise` as root.
  - Supported binary-download installations now use `sudo` only when needed to publish files and system shims.
  - System installations safely stage updates before replacing existing versions, with rollback protection.

- **Documentation**
  - Expanded guidance for `mise install --system`, including supported backends, permissions, restrictions, and passwordless sudo requirements.
  - Updated command help to recommend running `mise` as the normal user rather than with `sudo`.


mxcl:
Human here!

This is one option. The other is more invasive but still valid, ie:

- prevent sudo instantiated use of `mise` from creating or writing files to the user’s directory
- or pre-emptively abort if it *would* do that providing the user a command that ensures all those files are pre-created.

`sudo` has code smells and leaving root owned files in the user’s directory is one. Even though I don't understand all the details of it IME it's just best to abort and handle the sudo yourself.

Consequence: some people hate it when tools call sudo for them. But I think I spotted mise already doing this somewhere else so figured this would be the preferred route.

Either way: happy to do it differently, or if you consider this not a bug, just close it.

mxcl:
Only not fixing as waiting on an opinion from @jdx

PR #13286, opened 2026-09-16
fix(release): unbreak the alpine aports bump and honor dry_run=false

jdx:
The `release-alpine` workflow — which bumps `community/mise` in Alpine's aports and opens the MR — has been broken for over two weeks, and it could not be run by hand to work around that. This fixes both halves.

## The clone was blocked

The job died ~20 seconds in, at the very first network call:

```
+ git clone https://gitlab.alpinelinux.org/alpine/aports.git/ /home/packager/aports
Cloning into '/home/packager/aports'...
fatal: unable to access 'https://gitlab.alpinelinux.org/alpine/aports.git/': The requested URL returned error: 418
```

`gitlab.alpinelinux.org` answers git requests from CI IP ranges with an HTTP 418 anti-bot challenge. The workflow only runs for tags ending in `0`, so this was easy to miss — but both releases that actually triggered it failed identically:

- v2026.9.0 — [run 33502453327](https://github.com/jdx/mise/actions/runs/33502453327) (2026-09-01)
- v2026.9.10 — [run 35128087151](https://github.com/jdx/mise/actions/runs/35128087151) (2026-09-16)

So the Alpine package has been stuck since 2026.8.x.

**Fix:** clone `github.com/alpinelinux/aports`, the official mirror under the `alpinelinux` org. It carries the same history — `master` resolves to the same commit on both hosts:

```
gitlab: fbc4c0f028a07f6ee8064974b55c325de646d924  refs/heads/master
github: fbc4c0f028a07f6ee8064974b55c325de646d924  refs/heads/master
```

Only the clone moves. The branch is still pushed to the `jdxcode/aports` fork on GitLab and the MR is still opened against `alpine/aports` there, so the resulting MR is unchanged.

Authenticating the clone was the obvious first attempt, since the script already has `ALPINE_GITLAB_TOKEN` for the push remote. It does not work — the challenge is served before credentials are considered, and a dry run with the token in the clone URL failed with the identical 418 ([run 35145179593](https://github.com/jdx/mise/actions/runs/35145179593)).

## `dry_run=false` did nothing

`DRY_RUN` read `github.event.inputs.dry_run`. Values in `github.event.inputs` come from the raw webhook payload and are always strings, and the string `"false"` is truthy in a GitHub expression — so unchecking "dry run" on a manual dispatch still produced `DRY_RUN=1` and skipped both the push and the MR. There was no way to trigger a real bump by hand, which matters here because it is the natural way to recover a release whose automatic run failed.

**Fix:** read the `inputs` context, which honors the declared `type: boolean`.

```yaml
DRY_RUN: ${{ (github.event_name == 'release' || (github.event_name == 'workflow_dispatch' && !inputs.dry_run)) && '0' || '1' }}
```

| event | `dry_run` | before | after |
|---|---|---|---|
| `release` | — | `0` | `0` |
| `workflow_dispatch` | `true` (default) | `1` | `1` |
| `workflow_dispatch` | `false` | `1` ❌ | `0` |

Release-triggered runs are unaffected, and manual runs still default to dry-run.

## Validation

Verified by running the workflow itself via `workflow_dispatch` with `dry_run=true` on this branch ([run 35145339243](https://github.com/jdx/mise/actions/runs/35145339243)). It **succeeded end to end**, where every previous run died after 14 seconds. Dry-run mode guards `git push` and `glab mr create`, so nothing was published to Alpine.

```console
+ git clone https://github.com/alpinelinux/aports.git /home/packager/aports
Cloning into '/home/packager/aports'...
+ sed -i 's/pkgver=.*/pkgver=2026.9.10/' APKBUILD
...
-pkgver=2026.8.15
+pkgver=2026.9.10
+ git commit -m 'community/mise: upgrade to 2026.9.10'
+ '[' 1 == 0 ']'                                   # DRY_RUN=1, push skipped
++ glab mr list -R alpine/aports --author=@me
+ [[ No open merge requests match your search in alpine/aports. != Showing* ]]
```

That confirms the clone, the `abuild -r` build of 2026.9.10 from source, and the APKBUILD bump. It also shows aports is still on **2026.8.15**, one release behind where it should be.

Usefully, `glab mr list -R alpine/aports` succeeded against `gitlab.alpinelinux.org` from the same runner. So the host is reachable from CI for authenticated API traffic; it is specifically the anonymous git clone that gets challenged. That is good evidence the later push and MR creation will work too, though `git push jdxcode` is the one step a dry run cannot exercise — it is only reached on a real release.

The `dry_run=false` path is deliberately **not** exercised: a real manual run would push to Alpine's aports and open an MR. It rests on the expression semantics above and on `actionlint`, not on a live run.


jdx:
> Push `mise` with an explicit refspec. ... With Git's default `push.default=simple`, `git push jdxcode -f` can fail with `current branch mise has no upstream branch`.

Checked this one and I don't think it applies. `push.default=simple` only requires an upstream when you push to the remote you normally pull from. From `git config`:

> `simple` — ... When pushing to a remote that is different from the remote you normally pull from, work as `current`.

`jdxcode` is a different remote from `origin`, so the push falls back to `current`, which pushes the current branch to a same-named branch and creates it. Reproduced both paths on git 2.53 with no upstream configured:

```console
$ git rev-parse --abbrev-ref mise@{upstream}
fatal: no upstream configured for branch 'mise'

$ git push jdxcode -f      # different remote -> works as `current`
To ../fork.git
 * [new branch]      mise -> mise

$ git push origin -f       # the fetch remote -> the error you predicted
fatal: The current branch mise has no upstream branch.
```

The predicted failure is real, but only for the `origin` case, which the script never does. Same result with `push.default` set explicitly rather than left at its default.

This is also unaffected by the PR: `jdxcode` was a different remote from `origin` both before and after the clone moved to the GitHub mirror. Leaving it as written.

PR #13302, opened 2026-09-17
fix(use): fail when explicit postinstall hooks are skipped

zeitlinger:
## Summary

An explicitly requested postinstall command should not silently succeed without running. Currently, if the selected tool version is already installed, this exits successfully without running setup:

```sh
mise use --global --postinstall "mbx setup --yes" mr-boxington
```

This change returns an error naming the skipped hook's tool and explaining that the user can run the command manually or retry with `--force`. The requested configuration changes are not saved on failure; other tools in a mixed invocation may already have been installed.

Hooks merely stored in configuration keep their existing install-only behavior. Fresh installations, forced reinstalls, and dry runs retain their behavior. The check uses the actual installation outcome, including when another process installed the tool while this invocation waited for its lock, and keeps explicit hooks scoped to their original tool/version requests.

## Validation

- Regression test fails against unmodified mise and passes with the fix.
- Five focused e2e tests pass: postinstall, tool options, dry run, retained options, and postinstall environment.
- Covers global/versionless/latest requests, canonicalized relative paths, mixed installs, multiple versions, `--force`, and unchanged configured hooks.
- `mise run lint-fix` and scoped hk checks pass; CLI docs, usage spec, and man page regenerated (including 3 passing documentation tests).

## Summary by CodeRabbit

- **Bug Fixes**
  - `mise use --postinstall` now reports an error when the requested tool is already installed and the hook is skipped, instead of silently succeeding.
  - Use `--force` to reinstall the tool and run the postinstall command, or run the command manually.
  - Skipped postinstall requests no longer save the requested configuration changes. This behavior is consistent across versioned, global, path, and multi-tool installations.

- **Documentation**
  - Updated CLI help, the man page, and developer documentation to explain postinstall behavior, including dry-run handling.


jdx:
i don't think i agree with the premise

zeitlinger:
> i don't think i agree with the premise

you mean it should ignore it silenly?

zeitlinger:
@jdx you mean it should ignore it silenly?

jdx:
if it isn't installing it shouldn't run the postinstall hook

PR #13367, opened 2026-09-18
fix(install): detect stale installs when a locked checksum changes

rabadin:
An install directory can predate a `mise.lock` checksum change: `mise lock` re-resolves the same version to a different asset (a musl build replaced by a glibc-dependent one, for example) without anything removing the existing install. `is_install_satisfied` only checks that the install path exists, so `--locked` installs silently keep serving that stale artifact instead of refreshing it.

This records the checksum an install was verified against (reusing the `.mise.checksum` marker already used for rolling-release tracking) and, in `--locked` mode only, treats a mismatch against the current lockfile entry as "not installed" so the tool gets reinstalled. A missing marker — a pre-fix install, or a backend whose `is_install_satisfied` override never wrote one — counts as drifted too, rather than being trusted by default, matching how rolling-version checksum tracking already treats an absent stored checksum. Ordinary (non-locked) installs are unaffected.

The github/gitlab/forgejo backend has its own `is_install_satisfied` override and now consults the same shared check, including for its additional-asset installs: each additional asset's recorded checksum is compared against the current lockfile entry under `--locked`, so a pattern-preserving asset swap isn't mistaken for an unchanged install.

`e2e/backend/test_http_lock_stale_install` reproduces the primary scenario in three phases: install one asset, re-lock the same version against a different one and confirm a plain `mise install --locked` (no `--force`) refreshes it, then delete the checksum marker and confirm a missing marker is also treated as drift. Unit tests cover the additional-asset checksum comparison directly.

## Summary by CodeRabbit

- **Bug Fixes**
  - Locked installations now detect when an installed tool or additional asset is missing its checksum record or no longer matches the lockfile.
  - When checksum drift is detected, the affected artifact is reinstalled instead of being incorrectly treated as up to date.
  - Locked installs now correctly handle asset replacements that retain the same patterns but use different content.
  - Installed checksums are now recorded to support reliable verification during future locked installs.


rabadin:
Closing to reopen as a non-draft PR, ready for review.

PR #13454, opened 2026-09-21
feat(self-update): let a distribution disable self-update machine-wide

jdx:
https://entire.io/gh/jdx/mise/trails/85

## The problem

A packager can turn off `mise self-update` with a `.disable-self-update` marker or an instructions file, but both are found *relative to the running binary* — two levels up from the canonicalized path. That covers the install the packager laid down and nothing else.

On a distribution that manages mise for its users, that is not enough. Omarchy ships mise as the `mise-bin` pacman package, which installs `/usr/bin/mise` plus `/usr/lib/mise/.disable-self-update`, and updates it through `omarchy update`. A second mise installed alongside it — the `/usr/local/bin/mise` the install script leaves behind, or `~/.local/bin/mise` — finds no marker, self-updates happily, and then shadows the packaged copy on `PATH`, because `/usr/local/bin` comes first. The machine ends up running a mise that nothing updates, and the marker that was supposed to prevent exactly this never gets consulted.

## Now

Both files are also read from the system config directory (`/etc/mise`, or `MISE_SYSTEM_CONFIG_DIR`). That path is not relative to any binary, so one file covers every mise on the machine:

```toml
# /etc/mise/mise-self-update-instructions.toml
message = "mise is managed by pacman on Omarchy. To update it, run:\n\n  omarchy update\n"
```

```
$ /usr/local/bin/mise self-update
mise WARN  mise is managed by pacman on Omarchy. To update it, run:

  omarchy update

mise ERROR mise is installed via a package manager, cannot update
```

The same message replaces the otherwise dead-end `mise 2026.9.12 available` warning, so a user on an unmanaged copy is told what to run rather than being pointed at a self-update that should not happen.

`/etc/mise/.disable-self-update` works the same way without a message of its own. For a machine-wide marker the instructions file is usually the better of the two, since it is read by binaries the packager never installed, whose users need to be told what to run instead.

Unchanged from the per-binary markers:

- Per-binary markers keep their meaning and are checked first. They remain the right choice for a package installed on systems it does not own, such as Homebrew or a `.deb`.
- `MISE_SELF_UPDATE_AVAILABLE=true` still overrides, for a user deliberately running a standalone mise alongside the packaged one.
- `mise self-update --force` still bypasses the availability check.
- `mise doctor` reports `self_update_available: no` and prints the instructions, as it does for the existing markers.

Nothing changes on a machine with no `/etc/mise` marker, which is every machine today.

## Validation

- New e2e test `e2e/cli/test_self_update_system_marker`, covering both files, the version-available warning, and the `MISE_SELF_UPDATE_AVAILABLE=true` override. The e2e harness already isolates the system config directory, so the test writes to its own `/etc/mise`.
- Packager documentation in `docs/contributing.md` gains the two paths and a section on when a machine-wide marker is the right one.
- `mise run lint-fix`, `cargo clippy --workspace --all-features --all-targets -- -D warnings`.

Companion change on the distribution side (not in this PR): Omarchy would ship `/etc/mise/mise-self-update-instructions.toml` pointing at `omarchy update`.

---

> [!NOTE]
> **Medium Risk**
> Changes how self-update availability is resolved for all mise binaries when `/etc/mise` markers exist; misconfiguration could block updates or show wrong packager messages, but behavior is opt-in and overridable.
> 
> **Overview**
> Distributions can now **disable or redirect `mise self-update` for every mise binary on the host**, not only the install next to a packager marker.
> 
> **Runtime:** After the existing prefix-relative search (`lib/`, `lib/mise/`, etc.), mise also looks in the system config directory (`/etc/mise` or `MISE_SYSTEM_CONFIG_DIR`) for `.disable-self-update` and `mise-self-update-instructions.toml`. Per-binary markers are unchanged and still win when present; `MISE_SELF_UPDATE_AVAILABLE=true` and `mise self-update --force` behave as before.
> 
> **Docs:** `docs/contributing.md` documents the two `/etc/mise` paths and when a machine-wide marker fits (e.g. distro-managed mise vs Homebrew).
> 
> **Tests:** New E2E `e2e/cli/test_self_update_system_marker` covers instructions and bare markers, the “newer version available” message, and the env override.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a1b9065c209bda11cc8fd82cbe84321c670f2124. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added support for machine-wide self-update markers and instructions.
  * Users can now see package-manager-specific guidance when self-updates are unavailable.
  * System-wide markers apply to all mise installations on the machine, including package-manager-installed binaries.

* **Documentation**
  * Clarified that self-update marker and instruction paths are relative to the install prefix unless explicitly identified as machine-wide.
  * Documented the machine-wide configuration locations and behavior.


jdx:
Closing: the machine-wide lookup is the wrong shape. A marker that is not relative to a binary cannot say anything about which install it covers, so it necessarily catches a mise a user deliberately installed outside the package manager — and that install should stay updatable.

The prefix-relative lookup already handles the case this came from. `/usr/local/bin/mise` canonicalizes to prefix `/usr/local`, so `/usr/local/lib/mise/mise-self-update-instructions.toml` covers it, while `~/.local/bin/mise` (prefix `~/.local`) is untouched and self-updates normally. Verified on a debug build:

```
$ ./prefix/usr-local/bin/mise self-update
mise WARN  mise in /usr/local is managed by pacman on Omarchy. To update it, run:

  omarchy update

mise ERROR mise is installed via a package manager, cannot update

$ ./prefix/home-local/bin/mise self-update
mise is already up to date
```

No mise change needed; this is a packaging decision about whether a distribution claims the `/usr/local` prefix. #13453 (a clear error instead of `Permission denied … .mise.__temp__XKV5Oz`) stands on its own and stays open.

PR #13455, opened 2026-09-21
test: don't reset the shared fixture tree in --list invocations

betaboon:
## The problem

`cargo nextest run` can abort during its list phase, before a single test has
run:

```
thread '<unnamed>' panicked at src/test.rs:93:28:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }

thread '<unnamed>' panicked at .../core/src/panicking.rs:225:5:
panic in a function that cannot unwind
...
13: mise::test::init::_::__ctor_private
command `.../deps/mise-<hash> --list --format terse --ignored` aborted with signal 6 (SIGABRT)
```

nextest lists each test binary by executing it twice — `--list --format
terse`, with and without `--ignored` — and those two processes run
concurrently. Both run the test ctor, which resets the shared fixture tree:

```rust
file::remove_all(&*env::HOME.join("cwd")).unwrap();
file::create_dir_all(&*env::HOME.join("cwd").join(".mise").join("tasks")).unwrap();
env::set_current_dir(env::HOME.join("cwd")).unwrap();
// ... a dozen fixture writes, one via env::current_dir()
```

`$HOME` in tests is the compile-time `CARGO_MANIFEST_DIR/test`, shared by
every process of the same build. One process's `remove_all()` unlinks the
directory the other just `set_current_dir()`'d into; that process is left on
a dead inode and aborts at the ctor's next `current_dir()` call. The window
is only the dozen writes between the chdir and that call, so CI usually wins
the timing — a loaded machine doesn't. A nix-sandbox build on aarch64-darwin
hit it on the first attempt.

## Now

Invocations that only list tests skip the ctor. Listing enumerates
compiled-in test names; it needs neither the fixture tree nor a cwd.

The run phase is unchanged: nextest executes tests in per-test processes with
no `--list` argument, and `.config/nextest.toml`'s `test-threads = 1` already
serializes those against the shared fixture path — the listing pair was the
one concurrent execution left.

## Validation

- With the guard, a from-scratch nix-sandbox build on aarch64-darwin running
  `cargo nextest run --all-features --no-fail-fast` (profile `ci-shared`)
  completes: `Summary [ 154.579s] 4163 tests run: 4163 passed (8 slow), 348
  skipped`. The same build without the guard aborts in the list phase as
  above.
- Context: found while moving nix packaging (nixpkgs and downstream) onto
  cargo-nextest, the runner this suite is written for — under plain `cargo
  test`'s single shared process, a dropped-tempdir cwd strands ~23 unrelated
  `current_dir()`-calling tests, which process-per-test resolves.

*AI-assisted — Tool: Claude Code; model: anthropic/claude-fable-5; version: 2.1.245.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed list-only test runs so they no longer reset shared test fixtures or abort due to concurrent removal.
  * Improved end-to-end output checks without changing their expected behavior.


betaboon:
I dug into the e2e failure (tranches 2 and 6, `config/test_required_env_vars_tools_filter`) — it's unrelated to this diff (which is `#[cfg(test)]`-only and never reaches the e2e binary), and a re-run may not clear it. Summary so you can decide what to do with it:

The assert helper's failure branch dumps the content it searched (`echo "Expected to find '$*' in:"; echo "${output}"`), and in the job log that dump *contains the pattern it reports as missing*:

```
Expected to find 'NON_TOOL_VAR' in:
Error:
   0: Required environment variable 'NON_TOOL_VAR' is not defined. ...
      Help: Set NON_TOOL_VAR for development
```

So grep matched, yet the pipeline reported failure — the exit status is lying, not the search.

The cause is the test's local assert helpers (it shadows `e2e/assert.sh` with its own copies), which run under the script's `set -o pipefail`:

```bash
if ! echo "${output}" | grep -q "$@"; then
```

`grep -q` exits on the first match; once `${output}` outgrows the pipe buffer, the still-writing `echo` takes SIGPIPE and pipefail turns the pipeline into exit 141 — so a **matched** assertion reports failure. The boundary is exact:

```bash
set -o pipefail
out="NON_TOOL_VAR
$(head -c $N /dev/zero | tr '\0' x)"
echo "$out" | grep -q NON_TOOL_VAR   # N=64000: exit 0; N=70000: exit 141
```

In CI the captured stderr includes the color-eyre backtrace (`RUST_BACKTRACE: 1` in the workflow), which puts `error.log` right at that threshold — I measure ~64 KB from the job log, which undercounts since the log truncates long lines. That threshold-sitting explains why it failed both retry attempts here while main stays green: the file straddles the buffer size, and small build-layout or runner differences push it either way.

The test itself passes locally through `e2e/run_test` (all 4 sub-tests) when the error output is small.

A two-line fix would be feeding grep from a herestring in both local helpers — no pipe, no SIGPIPE, argument passthrough unchanged:

```bash
if ! grep -q "$@" <<<"${output}"; then
```

(the shared `e2e/assert.sh` helpers already use plain substring checks and are unaffected; this test's local copies are the only occurrence of the pattern in `e2e/`). Pushed that as a second commit here, following the shape of #13421's in-PR `test(windows)` fixes — say the word if you'd rather have it split out into its own PR.

